PR #31, opened 2013-11-13
Rename handler to path in coarsemux label set

grobie:
The term "handler" is more or less specific to golang while the commonly
used term in HTTP is "path". In order to not leak golang terminology
into other clients and also to help people to understand the meaning of
the labels, this change renames handler to path.

@matttproud @bernerdschaefer @discordianfish @juliusv

There are no rules or alerts using that label name, so let's rename as long as it's easy.


beorn7:
This is obsolete.

PR #38, opened 2014-12-15
Make Registry public

tsenart:
This change set exports a Registry interface implemented by the local
registry type. I added TODO comments along the way with improvements
that could be done to reduce the surface area of a Registry's behaviour
and to make the code more orthogonal overall. I'd be glad to contribute
these changes over time, if you like the direction I propose.

This has been done on top of the `standard-exports` branch.

All tests pass except for the Linux specifc `/proc` related tests. I
suggest that we add linux specific build tags to those files.


beorn7:
This PR intentionally left open as a starting point for a future public Registry interface.

DBarney:
Has there been any movement on this? I just realized today that I need to create two endpoints for prometheus to scrape...

cagedmantis:
@beorn7 Has there been any further development along these lines? Do you need somebody to pick up this pr or use this pr as a base to attempt this work again?

beorn7:
Work is in progress. I'm sorry for the lack of speed but higher priority issues kept interfering.

beorn7:
Superseeded by #214.

PR #41, opened 2015-01-06
Re-add some simple example binaries.

juliusv:
I intentionally left out the Makefile infrastructure that we had
previously for this, as it's not strictly needed and adds
complexity/maintenance burden.


juliusv:
@beorn7

beorn7:
:+1:

PR #44, opened 2015-01-09
Remove a Go 1.4ism to allow compilation under Go 1.3.

jmyounker:
One line syntactic change with no semantic effects.


juliusv:
:+1:

PR #45, opened 2015-01-12
Allow error reporting during metrics collection and simplify Register().

beorn7:
@juliusv @grobie @xla 

Both are interface changes I want to get in before public
announcement. They only break rare usage cases, and are always easy to
fix, but still we want to avoid breaking changes after a wider
announcement of the project.

The change of Register() simply removes the return of the Collector,
which nobody was using in practice. It was just bloating the call
syntax. Note that this is different from RegisterOrGet(), which is
used at various occasions where you want to register something that
might or might not be registered already, but if it is, you want the
previously registered Collector back (because that's the relevant
one).

WRT error reporting: I first tried the obvious way of letting the
Collector methods Describe() and Collect() return error. However, I
had to conclude that that bloated _many_ calls and their handling in
very obnoxious ways. On the other hand, the case where you actually
want to report errors during registration or collection is very
rare. Hence, this approach has the wrong trade-off. The approach taken
here might at first appear clunky but is in practice quite handy,
mostly because there is almost no change for the "normal" case of "no
special error handling", but also because it plays well with the way
descriptors and metrics are handled (via channels).

Explaining the approach in more detail:

- During registration / describe: Error handling was actually already
  in place (for invalid descriptors, which carry an error anyway). I
  only added a convenience function to create an invalid descriptor
  with a given error on purpose.

- Metrics are now treated in a similar way. The Write method returns
  an error now (the only change in interface). An "invalid metric" is
  provided that can be sent via the channel to signal that that metric
  could not be collected. It alse transports an error.

NON-GOALS OF THIS COMMIT:

This is NOT yet the major improvement of the whole registry part,
where we want a public Registry interface and plenty of modular
configurations (for error handling, various auto-metrics, http
instrumentation, testing, ...). However, we can do that whole thing
without breaking existing interfaces. For now (which is a significant
issue) any error during collection will either cause a 500 HTTP
response or a panic (depending on registry config). Later, we
definitely want to have a possibility to skip (and only report
somehow) non-collectible metrics instead of aborting the whole scrape.

### DIFF
--- a/prometheus/collector.go
+++ b/prometheus/collector.go
@@ -33,8 +33,9 @@ type Collector interface {
 	// duplicate descriptors. Those duplicates are simply ignored. However,
 	// two different Collectors must not send duplicate descriptors.) This
 	// method idempotently sends the same descriptors throughout the
-	// lifetime of the Collector. A Collector unable to describe itself must
-	// send an invalid descriptor (created with NewInvalidDesc).
+	// lifetime of the Collector. If a Collector encounters an error while
+	// executing this method, it must send an invalid descriptor (created
+	// with NewInvalidDesc) to signal the error to the registry.
 	Describe(chan<- *Desc)
 	// Collect is called by Prometheus when collecting metrics. The
 	// implementation sends each collected metric via the provided channel
--- a/prometheus/collector.go
+++ b/prometheus/collector.go
@@ -33,7 +33,8 @@ type Collector interface {
 	// duplicate descriptors. Those duplicates are simply ignored. However,
 	// two different Collectors must not send duplicate descriptors.) This
 	// method idempotently sends the same descriptors throughout the
-	// lifetime of the Collector.
+	// lifetime of the Collector. A Collector unable to describe itself must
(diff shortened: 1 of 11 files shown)


prometheus/collector.go
```diff
@@ -33,7 +33,8 @@ type Collector interface {
 	// duplicate descriptors. Those duplicates are simply ignored. However,
 	// two different Collectors must not send duplicate descriptors.) This
 	// method idempotently sends the same descriptors throughout the
-	// lifetime of the Collector.
+	// lifetime of the Collector. A Collector unable to describe itself must
```
juliusv: "A Collector unable to describe itself" could be interpreted by a newbie as a Collector which generally doesn't support describing itself. Since such a thing is not supported, maybe write it like "If a Collector encounters an error while describing itself, it must send an invalid descriptor..."

juliusv:
:+1: otherwise

prometheus/collector.go
beorn7: Makes sense.

PR #51, opened 2015-01-25
Better sample value string formatting.

juliusv:
This forces a float format without exponent that uses the minimum number
of digits necessary to display a given sample value.


brian-brazil:
:+1:

beorn7:
:+1:

PR #52, opened 2015-01-26
Make examples buildable from make

brian-brazil:


beorn7:
:+1: just for the record...

PR #53, opened 2015-01-27
Add benchmarks to compare text and protobuf parsing.

beorn7:
@brian-brazil @juliusv 

I guess that means for high sample ingestion rates, we should aim for (optionally) supporting protobuf in the new Java client, too...

On the other hand: As the parsing is happening in parallel with many scrapes, the bottleneck for sample ingestion on spinning disk machines is persisting the chunks to disk. (With more efficient compression for chunks, that might change, though.) On SSD, the bottleneck (with protobuf ingestion) appeared to be the fingerprint calculation, which happens serially at the moment. Should anybody ever want more than 50k samples/sec ingestion, we would improve the fingerprint calculation first.

Still, especially for situations where CPU is tighter (and the server is also serving expensive queries and such), a 4x speed-up on parsing seems worth the deal...


text/bench_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"io"
+	"io/ioutil"
+	"testing"
+
+	dto "github.com/prometheus/client_model/go"
+
+	"github.com/matttproud/golang_protobuf_extensions/ext"
+)
+
+// Benchmarks to show how much penalty text format parsing accually inflicts.
```
juliusv: accually -> actually

text/bench_test.go
```diff
@@ -0,0 +1,165 @@
...
+		if err != nil {
+			b.Fatal(err)
+		}
+		if _, err := parser.TextToMetricFamilies(in); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
+// BenchmarkParseProto benchmarks the parsinge of a protobuf-format scrape into
```
juliusv: parsinge -> parsing

text/bench_test.go
```diff
@@ -0,0 +1,165 @@
...
+// metric family DTOs. Note that this does not build a map of matric families
```
juliusv: matric -> metric

juliusv:
:+1: otherwise

brian-brazil:
That'd indicate ~5us per sample in text format. With a 60s interval and 500k timeseries (rough m1.large limit), that's only .04 of a core on a 2 core box. I don't think we need to worry about performance in the general case, the rules/graphs/storage will dominate.

beorn7:
@juliusv Oh my... I hope my typos are not a representation for my mental state...

beorn7:
@brian-brazil Wait for fish running Prometheus on a Raspberry Pi... :)

But you are right, not our most pressing concern right now. The gzipped text format is actually not much larger than the gzipped protobuf, which gets rid of the bandwidth concern.

CPU-wise: Even the 50k samples/s scenario will only need 0.25 cores. GC-pressure might hurt more, here... but who knows.

Issue #61, opened 2015-01-31
Bug: data race in Register() vs. ServeHTTP()


stapelberg:
I’m using this simple test program:
```go
package main

import (
    "fmt"
    "net/http"
    "time"

    "github.com/prometheus/client_golang/prometheus"
)

func main() {
    http.Handle("/metrics", prometheus.Handler())
    go http.ListenAndServe(":8089", nil)

    i := 0
    for {
        prometheus.Register(prometheus.NewCounter(prometheus.CounterOpts{
            Subsystem: "promrace",
            Name:      fmt.Sprintf("counter_%d", i),
            Help:      "counter",
        }))
        i++
        time.Sleep(1 * time.Second)
    }
}
```
According to the example in https://godoc.org/github.com/prometheus/client_golang/prometheus#Register, registering new counters should work just fine after the HTTP server is already up and running.

However, when running the test program with `go run -race` and accessing /metrics repeatedly, e.g. in a browser, or with a benchmark program, I get warnings about data races:

```text
==================
WARNING: DATA RACE
Read by goroutine 17:
  github.com/prometheus/client_golang/prometheus.(*registry).writePB()
      /home/michael/gocode/src/github.com/prometheus/client_golang/prometheus/registry.go:403 +0x85
  github.com/prometheus/client_golang/prometheus.(*registry).ServeHTTP()
      /home/michael/gocode/src/github.com/prometheus/client_golang/prometheus/registry.go:383 +0x15b
  net/http.Handler.ServeHTTP·fm()
      /home/michael/gocode/src/github.com/prometheus/client_golang/prometheus/http.go:58 +0x64
  github.com/prometheus/client_golang/prometheus.func·007()
      /home/michael/gocode/src/github.com/prometheus/client_golang/prometheus/http.go:144 +0x241
  net/http.HandlerFunc.ServeHTTP()
      /home/michael/go/src/net/http/server.go:1265 +0x4e
  net/http.(*ServeMux).ServeHTTP()
      /home/michael/go/src/net/http/server.go:1541 +0x20c
  net/http.serverHandler.ServeHTTP()
      /home/michael/go/src/net/http/server.go:1703 +0x1f6
  net/http.(*conn).serve()
      /home/michael/go/src/net/http/server.go:1204 +0x1087

Previous write by main goroutine:
  runtime.mapassign1()
      /home/michael/go/src/runtime/hashmap.go:376 +0x0
  github.com/prometheus/client_golang/prometheus.(*registry).Register()
[...]

PR #62, opened 2015-01-31
Fix race condition in writePB().

juliusv:
The RLock already needs to be acquired when reading r.dimHashesByName.

This fixes https://github.com/prometheus/client_golang/issues/61

@beorn7 @stapelberg

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -400,7 +400,6 @@ func (r *registry) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 }
 
 func (r *registry) writePB(w io.Writer, writeEncoded encoder) (int, error) {
-	metricFamiliesByName := make(map[string]*dto.MetricFamily, len(r.dimHashesByName))
 	var metricHashes map[uint64]struct{}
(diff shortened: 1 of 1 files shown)


grobie:
What about adding the referenced example to the test suite to ensure concurrent/later registration of metrics work and race conditions can be checked for with the test suite?

juliusv:
@grobie I could add a test case, but don't know a way to test the race directly. I'd have to add a separate Makefile step to run that specific test with the race detector enabled. Is that worth it? I think if we have races again, it would be much more likely a *different* race that that particular test wouldn't catch anyways, and it's not possible to run the whole test suite with `-race` enabled because it uses too much RAM.

beorn7:
:+1:

beorn7:
@grobie @juliusv In principle, think it makes sense to add a test for it, and then mention in the doc string that it can be run with `-race` to test for races. We have a similar case in `prometheus/prometheus/storage/local`. The test would be able to detect some other races, too, especially if it's a bigger fuzz test of some kind.

We can do that later, though.

PR #63, opened 2015-02-02
Introduce versioning and clean up the headers.

beorn7:
@juliusv 

Don't be afraid, most of the changes is just the header cleanup (I should have done before)...

This fixes https://github.com/prometheus/client_golang/issues/54


juliusv:
Phew, the fear subsides.

:+1:

PR #65, opened 2015-02-02
Remove dependency on procfs/cgo in non-procfs systems

grobie:
The procfs package has a cgo dependency (necessary to calculate system
times). As procfs is not available under windows, darwin and supposely
all newer BSD systems, this change removes the procfs dependency on these
systems.

This will enable cross compilation for linux->darwin again.

@juliusv @beorn7 @discordianfish

### DIFF
--- a/prometheus/process_collector.go
+++ b/prometheus/process_collector.go
@@ -1,6 +1,17 @@
-package prometheus
+// Copyright 2015 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


prometheus/process_collector.go
```diff
@@ -64,9 +75,9 @@ func NewProcessCollectorPIDFn(
 		}),
 	}
 
-	// Use procfs to export metrics if available.
-	if _, err := procfs.NewStat(); err == nil {
-		c.collectFn = c.procfsCollect
+	// Setup process metric collection if supported by the runtime.
```
juliusv: s/Setup/Set up/ for the verb, not the noun.

juliusv:
Interesting approach :) :+1: besides the nitpick

PR #66, opened 2015-02-02
Use clear error interface for process pidFn

grobie:
If a given pidFn for a process collector can't determine the pid, this
had to be signaled with an invalid pid so far (pid <= 0). In order to
make the error interface clearer for users, this change introduces an
explicit error parameter.

As requested by @juliusv.

@brian-brazil 

---

I'll handle any possible merge conflict errors depending whether #65 or this change will be merged first.


juliusv:
Yay, thanks! :+1: :lollipop: :candy: :100:

PR #70, opened 2015-02-10
Handle process collector errors gracefully

grobie:
As it is expected that the process collector can fail under certain
conditions (proc information for a process only readable by root or
other user for example) and as there is currently no option to configure
the error behavior of the client, this change reverts the error
reporting introduced in 159e96f. This effectively means that errors are
simply ignored and there won't be any samples for the process_* metrics
in case of an error.

Once a user can control how to behave in case of errors returned by
collectors, this change should probably be reverted again.

@beorn7 @juliusv @brian-brazil

### DIFF
--- a/prometheus/process_collector_procfs.go
+++ b/prometheus/process_collector_procfs.go
@@ -27,23 +27,15 @@ func processCollectSupported() bool {
 func (c *processCollector) processCollect(ch chan<- Metric) {
 	pid, err := c.pidFn()
 	if err != nil {
-		c.reportCollectErrors(ch, err)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


brian-brazil:
:+1:

juliusv:
It feels a bit weird to do... nothing (no logs, no error counter) when an error occurs, but since errors are going to be the norm in certain environments (chroot), I guess that's the way to go.

:+1:

beorn7:
And let's revisit once we have more granular error handling.

beorn7:
:+1: But perhaps add a TODO with this commit as a reference so that you can put it back once error handling is more configurable?

PR #71, opened 2015-02-11
Remove obsolete comment and add Earliest and Latest timestamp.

beorn7:
The latter is to eradicate the various MaxInt64 and MinInt64 in the
codebase.

@juliusv


juliusv:
:+1:

PR #72, opened 2015-02-11
Update README.md

wiremine:
Seems like indexed and size are being initialized, so they would require := instead of =

### DIFF
--- a/prometheus/README.md
+++ b/prometheus/README.md
@@ -19,13 +19,13 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


grobie:
This is part of a surrounding `var ( ... )`. `:=` is not valid in that context.

grobie:
Variable declaration in go: https://golang.org/ref/spec#Short_variable_declarations

Thanks though!

wiremine:
Duh. Sorry for the spam. :-)

Issue #74, opened 2015-02-12
Fingerprinting function misses a separator


beorn7:
x{a="bb",b="c} and x{a="b",bb="c"} have the same fingerprint with the current fingerprinting method.
Reason is that we have a separator between label name and label value, but not between whole label pairs.

This is easy to fix, but invalidates all storage. :-(((

PR #75, opened 2015-02-12
Reduce allocations during fingerprinting.

beorn7:
Also, add a test to expose
https://github.com/prometheus/client_golang/issues/74 .

benchmark           old ns/op     new ns/op     delta
BenchmarkMetric     7034          6272          -10.83%

benchmark           old allocs     new allocs     delta
BenchmarkMetric     52             32             -38.46%

benchmark           old bytes     new bytes     delta
BenchmarkMetric     1976          1800          -8.91%

@juliusv

### DIFF
--- a/model/metric.go
+++ b/model/metric.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 )
 
+var separator = []byte{0}
+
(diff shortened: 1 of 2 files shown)


juliusv:
:+1:

PR #76, opened 2015-02-12
Add support for histograms to parsers, extraction and creation.

brian-brazil:
This does not include a histogram metric usable from the go client.

See https://docs.google.com/document/d/1uSenXRDjDaJLV3qnSD09GqgPdEEDPjER0mVsnGaCYF0/edit#

Make format did a few changes too.

@beorn7 @juliusv

### DIFF
--- a/text/create.go
+++ b/text/create.go
@@ -139,6 +139,39 @@ func MetricFamilyToText(out io.Writer, in *dto.MetricFamily) (int, error) {
 				float64(metric.Summary.GetSampleCount()),
 				out,
 			)
+		case dto.MetricType_HISTOGRAM:
+			if metric.Histogram == nil {
+				return written, fmt.Errorf(
+					"expected summary in metric %s", metric,
+				)
+			}
+			for _, q := range metric.Histogram.Bucket {
+				n, err = writeSample(
+					name+"_bucket", metric,
+					"le", fmt.Sprint(q.GetUpperBound()),
+					float64(q.GetCumulativeCount()),
+					out,
+				)
+				written += n
+				if err != nil {
+					return written, err
+				}
+                        // TODO: Add +inf bucket if it's missing.
+			}
+			n, err = writeSample(
+				name+"_sum", metric, "", "",
+				metric.Histogram.GetSampleSum(),
+				out,
+			)
+			if err != nil {
+				return written, err
+			}
+			written += n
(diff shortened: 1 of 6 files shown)


prometheus/registry.go
```diff
@@ -426,7 +426,7 @@ func (r *registry) writePB(w io.Writer, writeEncoded encoder) (int, error) {
 
 	// Drain metricChan in case of premature return.
 	defer func() {
-		for _ = range metricChan {
+		for range metricChan {
```
juliusv: We said a while ago that we don't want to require users of the lib to be on Go >=1.4, so this kind of syntactic change around the `range` statement (automatically done by gofmt) would need to be reverted for older Go versions to work with this library.

beorn7:
Will have a look ASAP. (Which might not be today, sadly...)

prometheus/registry.go
brian-brazil: I think I've gotten them all.

text/parse.go
```diff
@@ -376,6 +406,25 @@ func (p *Parser) readingValue() stateFn {
 				},
 			)
 		}
+	case dto.MetricType_HISTOGRAM:
+		// *sigh*
```
beorn7: After sanitizing all my other comments, you have left this in place? ;-)

brian-brazil: I de-Beornified, it's different from sanitising ;-)

text/create.go
```diff
@@ -139,6 +139,38 @@ func MetricFamilyToText(out io.Writer, in *dto.MetricFamily) (int, error) {
...
+				n, err = writeSample(
+					name+"_bucket", metric,
+					"le", fmt.Sprint(q.GetUpperBound()),
+					float64(q.GetCumulativeCount()),
+					out,
+				)
+				written += n
+				if err != nil {
+					return written, err
+				}
```
beorn7: Here we have to add the +Inf bucket, too, if it is not in the protobuf. (Just add an TODO, as above, if you wish.)

brian-brazil: Done.

text/parse.go
beorn7: De-Beornification? That's clearly cruelty to animals or something...

beorn7:
:+1:

PR #77, opened 2015-02-13
Update protobuf library package name.

juliusv:
The Golang protocol buffer library has now moved to GitHub:

https://github.com/golang/protobuf

Although "go get"-ing the old package name still works, moving
everything to the new one will make vendoring cleaner.

See also https://github.com/matttproud/golang_protobuf_extensions/pull/7


juliusv:
@beorn7 for TBR, but I will just merge for now because my Prometheus server vendoring cleanups depend on it.

PR #78, opened 2015-02-19
Histograms for the exposition library.

beorn7:
@juliusv See commit descriptions for details.

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -1,0 +1,329 @@
+// Copyright 2015 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package prometheus
+
+import (
+	"fmt"
+	"hash/fnv"
+	"math"
+	"sync/atomic"
+
+	"github.com/golang/protobuf/proto"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+// A Histogram counts individual observations from an event or sample stream in
+// configurable buckets. Similar to a summary, it also provides a sum of
+// observations and an observation count.
+//
+// On the Prometheus server, quantiles can be calculated from a Histogram using
+// the histogram_quantile function in the query language.
+//
+// Note that Histograms, in contrast to Summaries, can be aggregated with the
+// Prometheus query language (see the documentation for detailed
+// procedures). However, Histograms requires the user to pre-define suitable
+// buckets, and they are in general less accurate. The Observe method of a
+// Histogram has a very low performance overhead in comparison with the Observe
+// method of a Summary.
+//
+// To create Histogram instances, use NewHistogram.
+type Histogram interface {
+	Metric
+	Collector
+
+	// Observe adds a single observation to the histogram.
+	Observe(float64)
+}
+
+// DefBuckets are the default Histogram buckets. Most likely, you want to define
+// buckets customized to your use case.
+var (
+	DefBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}
+)
+
+// LinearBuckets creates 'count' buckets, each 'width' wide, where the lowest
+// bucket has an upper bound of 'start'. The final +Inf bucket is counted
+// towards the total 'count', but not included in the returned slice. The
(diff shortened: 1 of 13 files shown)


prometheus/histogram.go
```diff
@@ -0,0 +1,331 @@
...
+// A Histogram counts individual observations from an event or sample stream in
+// configurable buckets. Similar to a summary, it also provides a sum of
+// observations and an observation count.
+//
+// On the Prometheus server, quantiles can be calculated from a Histogram using
+// the histogram_quantile function in the query language.
+//
+// Note that Histograms, in contrast to Summaries, can be aggregated with the
+// Prometheus query language (see the documentation for detailed
+// procedures). However, Histograms requires the user to pre-define suitable
```
juliusv: s/requires/require/

prometheus/histogram.go
```diff
@@ -0,0 +1,331 @@
...
+	Metric
+	Collector
+
+	// Observe adds a single observation to the histogram.
+	Observe(float64)
+}
+
+// DefBuckets are the default Histogram buckets. The default buckets are
+// tailored to broadly measure response time in seconds for a typical online
+// serving system. Most likely, however, you will be required to define buckets
```
juliusv: s/online serving/online-serving/

(http://www.gingersoftware.com/content/grammar-rules/adjectives/compound-adjectives/ - only being pedantic here because it's user-visible docs :))

prometheus/summary.go
```diff
@@ -158,6 +163,17 @@ func newSummary(desc *Desc, opts SummaryOpts, labelValues ...string) Summary {
 		panic(errInconsistentCardinality)
 	}
 
+	for _, n := range desc.variableLabels {
+		if n == "quantile" {
```
juliusv: I think "quantile" and "le" should be constants in https://github.com/prometheus/client_golang/blob/master/model/labelname.go#L20-L36

prometheus/histogram.go
beorn7: Done.

prometheus/histogram.go
beorn7: Just that an online serving system is a serving system that happens to be online, and not a system that is "online-serving". @brian-brazil As the word came from you, what did you actually mean by it?

brian-brazil: We say online-serving elsewhere. http://prometheus.io/docs/practices/instrumentation/#online-serving-systems

prometheus/histogram.go
```diff
@@ -0,0 +1,331 @@
...
+	his.Bucket = buckets
+	out.Histogram = his
+	out.Label = h.labelPairs
+	return nil
+}
+
+// HistogramVec is a Collector that bundles a set of Histograms that all share the
+// same Desc, but have different values for their variable labels. This is used
+// if you want to count the same thing partitioned by various dimensions
+// (e.g. http request latencies, partitioned by status code and method). Create
```
juliusv: nit: s/http/HTTP/ here and for the other type(s) that have this comment.

prometheus/summary.go
```diff
@@ -35,6 +35,12 @@ import (
 // Summary provides the median, the 90th and the 99th percentile of the latency
 // as rank estimations.
 //
+// Note that the rank estimations cannot be aggregated in a meaningful way with
+// the Prometheus query language (i.e. you cannot average or add them). If you
+// need aggregatable quantiles (e.g. you want the 99th percentile latency of all
+// queries served across all instances of a service), check out the Histogram
```
juliusv: s/check out/see/

prometheus/histogram.go
juliusv: Yeah, as its about the serving happening online, vs. the system being online.

juliusv:
:+1: besides my comments.

prometheus/summary.go
beorn7: Good point. Done everywhere in `client_golang`.

prometheus/histogram.go
beorn7: Done everywhere in `client_golang`.

prometheus/summary.go
beorn7: Made it 'consider'.

prometheus/histogram.go
beorn7: Fun fact:

The term "online serving system" is used only 24 times in the whole web. 23 times it is spelled "online serving system", only once "online-serving system". And that one time is here: http://prometheus.io/docs/practices/instrumentation/

So I'd either vote for coming up with a more common term altogether, or drop the hyphen as I think we are talking about a serving system that has been put online, and I don't even know what "online-serving" should be (opposite of "offline-serving"? What is that? )

juliusv: Ok, my impression was always that it was about the serving happening online... but I agree that the term is a bit fuzzy. We do at least have our own definition here: http://prometheus.io/docs/practices/instrumentation/#online-serving-systems - so maybe someone can think of an even better name for that concept, but I guess it's not *that* important to spend a lot of time on.

PR #80, opened 2015-02-20
Fix for Histogram in registry.writePB

peterbourgon:
Should be self-explanatory.


beorn7:
Thanks. That should be it...

PR #83, opened 2015-02-25
Use godep with import rewriting for vendoring.

juliusv:
The new vendoring was produced by running:

    godep save -r ./examples/... ./prometheus/... ./text/... ./model/... ./extraction/...

Two things to note:

- "extraction/processor0_0_{1,2}_test.go" imported a package from
  "github.com/prometheus/prometheus", all for just one tiny testing
  function. To not have to deal with a circular vendoring dependency, I
  simply replaced the usage of the function by some in-line logic.

- godep grouped the rewritten imports slightly differently for some
  reason, but at least the standard library imports are still in a
  separate section. Not sure if it's worth manually keeping our old
  import grouping scheme or if we should simply use that godep-generated
  one.


juliusv:
@beorn7

beorn7:
Thanks for doing this.

Newly vendored parts need to be added to the NOTICE file.

juliusv:
@beorn7 Done.

beorn7:
:+1:

PR #87, opened 2015-03-03
Unify fingerprinting aka signature calculation.

beorn7:
This fixes https://github.com/prometheus/client_golang/issues/74 .

IT CHANGES THE FINGERPRINTING AND THEREFORE INVALIDATES EACH AND EVERY
PERSISTED FINGERPRINT (I.E. YOU HAVE TO WIPE THE STORAGE TO USE THIS).

This commit removes the LabelValuesToSignature function as it is used
nowhere (and broken in the way it is implemented right now).

Also, remove one more golint warning.

@juliusv

### DIFF
--- a/model/metric.go
+++ b/model/metric.go
@@ -14,10 +14,8 @@
 package model
 
 import (
-	"encoding/binary"
 	"encoding/json"
 	"fmt"
-	"hash/fnv"
 	"sort"
 	"strings"
 )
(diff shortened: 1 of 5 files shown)


beorn7:
Old benchmark:
BenchmarkMetric   200000              6262 ns/op            1800 B/op         32 allocs/op

New benchmark:
BenchmarkMetric   300000              5117 ns/op            1393 B/op          9 allocs/op

Note that the 9 remaining allocations are all done by the test setup. The benchmark is not particularly well written. This PR adds a better benchmark:

BenchmarkMetricLabelsToFingerprintScalar        200000000                6.34 ns/op            0 B/op          0 allocs/op
BenchmarkMetricLabelsToFingerprintSingle         5000000               354 ns/op               0 B/op          0 allocs/op
BenchmarkMetricLabelsToFingerprintDouble         3000000               454 ns/op               0 B/op          0 allocs/op
BenchmarkMetricLabelsToFingerprintTriple         3000000               562 ns/op               0 B/op          0 allocs/op

juliusv:
:+1:

PR #97, opened 2015-03-15
Allow the metric family injection hook to merge with existing metric fam...

beorn7:
...ilies.

If a metric family returned by the injection hook already exists (with
the same name), then its metrics are simply merged into that metric
family. With enabled collect-time checks, even uniqueness is checked,
but in general, things stay the same that the caller is responsible to
ensure metric consistency.

This fixes https://github.com/prometheus/pushgateway/issues/27 .

@juliusv @brian-brazil 

Note that we still want a way to switch off the automatic process metrics (which is being worked on, it just takes some time to do it right).
With this PR, at least everything works, but it will create inconsistent metrics (pushed process metrics end up with an additional "exporter_job" label, for example). But that problem is a bigger one, too. (I believe we discussed the problem with adding the exporter labels elsewhere...)


juliusv:
:+1:

PR #103, opened 2015-03-31
Always build the procfs collector.

juliusv:
The cgo dependency in the "procfs" package is being removed in:

  https://github.com/prometheus/procfs/pull/4

So now it should be fine to always import the "procfs" package and have
it determine dynamically at runtime whether the proc filesystem is
actually there.

This fixes a problem for users which were vendoring client_golang on
MacOS X, but building the result on Linux (the procfs package was
missing in the vendored packages).


juliusv:
@grobie @joshdevins @beorn7 @brian-brazil

brian-brazil:
Should we merge the two process_collector files now?

juliusv:
@brian-brazil I don't feel strongly. I added a second commit to merge the files, let's see how you like it :)

grobie:
:+1:

beorn7:
:+1:

brian-brazil:
:+1:

prometheus/process_collector.go
```diff
@@ -100,3 +102,48 @@ func (c *processCollector) Describe(ch chan<- *Desc) {
 func (c *processCollector) Collect(ch chan<- Metric) {
 	c.collectFn(ch)
 }
+
+func processCollectSupported() bool {
+	if _, err := procfs.NewStat(); err == nil {
+		return true
+	}
+	return false
+}
```
grobie: This indirection isn't really necessary anymore. You can replace the `if processCollectSupported` with the condition above.

PR #110, opened 2015-04-21
Add constants for instance/address labels and hidden prefix.

fabxc:
Some constants being used in the retrieval changes. As these are only additions we don't have to sync merging.


beorn7:
:+1:

PR #111, opened 2015-04-21
Replace hidden label prefix with meta prefix.

fabxc:


model/labelname.go
```diff
@@ -34,10 +34,8 @@ const (
 	// label names.
 	ReservedLabelPrefix = "__"
 
-	// HiddenLabelPrefix is a prefix which is legal in user-supplied label names
-	// but will not appear in the eventual metrics.
-	// Reserved labels may be excepted from that rule.
-	HiddenLabelPrefix = "_"
+	// MetaLabelPrefix is a prefix for labels that provide meta information.
```
juliusv: I would add something like "...that provide meta-information useful during target relabeling, but will not be attached to time series themselves."

fabxc: Certain yes to the final part. `Meta` is completely generic though and we might use it outside of SD in the future.

juliusv: :+1:

fabxc:
Changed.

juliusv:
:+1:

PR #116, opened 2015-05-05
Fingerprinting fixes in client_golang

beorn7:
@juliusv These are the fixes we need in client_golang. prometheus/prometheus code that uses these fixes needs to be changed to not change the fingerprint used in storage.

(Separate PR imminent.)

### DIFF
--- a/model/metric.go
+++ b/model/metric.go
@@ -28,11 +28,13 @@ type Metric map[LabelName]LabelValue
 
 // Equal compares the fingerprints of both metrics.
 func (m Metric) Equal(o Metric) bool {
+	// TODO do an actual map comparison
 	return m.Fingerprint().Equal(o.Fingerprint())
 }
 
 // Before compares the fingerprints of both metrics.
 func (m Metric) Before(o Metric) bool {
+	// TODO do an actual map comparison
 	return m.Fingerprint().Less(o.Fingerprint())
 }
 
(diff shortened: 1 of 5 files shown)


model/metric.go
```diff
@@ -26,14 +26,67 @@ var separator = []byte{0}
...
+	}
+	return true
 }
 
-// Before compares the fingerprints of both metrics.
+// Before compares the metrics, using the following criteria:
+//
+// If m has fewer labels than o, it is before o. If it has more, it is not.
+//
+// If the number of labels is the same, the subset of all label values is sorted
```
juliusv: s/label values/labels names/?

model/signature.go
```diff
@@ -46,30 +47,37 @@ func getHashAndBuf() *hashAndBuf {
...
 
 func putHashAndBuf(hb *hashAndBuf) {
+	hb.h.Reset()
+	hb.b.Reset()
 	hashAndBufPool.Put(hb)
 }
 
-// LabelsToSignature returns a unique signature (i.e., fingerprint) for a given
-// label set.
+// LabelsToSignature returns an quasi-unique signature (i.e., fingerprint) for a
```
juliusv: s/an/a/

juliusv:
:+1: besides nits

model/metric.go
beorn7: I screwed up here anyway, should be superset, not subset... will reword...

model/signature.go
beorn7: Done.

juliusv:
:+1:

PR #129, opened 2015-06-01
gofmt

robx:
Keeps showing up when I gofmt our whole tree including godeps.

### DIFF
--- a/text/bench_test.go
+++ b/text/bench_test.go
@@ -19,9 +19,10 @@ import (
 	"io"
 	"io/ioutil"
 	"testing"
-	dto "github.com/prometheus/client_model/go"
 
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
+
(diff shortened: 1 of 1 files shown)


fabxc:
Thanks. Could you move it back down there but insert a blank line above?

robx:
Sure. There you go.

text/bench_test.go
```diff
@@ -19,6 +19,7 @@ import (
 	"io"
 	"io/ioutil"
 	"testing"
+
 	dto "github.com/prometheus/client_model/go"
```
beorn7: To fix this properly (according to the Prometheus style guide) this has to go to the bottom most group.

(The original placing was a glitch, and older gofmt liked it there.)

robx: Hah, I wanted to do this, but then I saw it was different in other places.

fabxc:
Thanks!

PR #131, opened 2015-06-04
Solve "The Proxy Problem" of http.InstrumentHandler

telendt:
`InstrumentHandler` provides proxy object over given `http.ResponseWriter` - `responseWriterDelegator` that only implements the bare minimum required of an `http.ResponseWriter` and doesn’t implement any interface upgrades (`http.Flusher`, `http.Hijacker`, etc.). One example of negative side-effect of this is decreased performance of `http.FileServer` handler after instrumentation.
This PR fixes it by providing `fancyResponseWriterDelegator` with all the fancy bells and whistles if standard library `http.ResponseWriter` is detected. More info [here](https://avtok.com/2014/11/05/interface-upgrades.html#the-proxy-problem).

I also switched `responseWriterDelegator.written` to int64 (to allow responses greater than 4GB) in different commit. If you prefer two different PRs or one PR with squashed commits just let me know.

### DIFF
--- a/prometheus/http.go
+++ b/prometheus/http.go
@@ -14,6 +14,9 @@
 package prometheus
 
 import (
+	"bufio"
+	"io"
+	"net"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


fabxc:
Looks good to me, thanks!
AFAIK @beorn7 had plans to do some work on the `client_golang` and on this section in particular. So he can tell whether this should be merged now or rather be integrated with the larger changes he planned.

telendt:
@fabxc Yes, I've heard that @beorn7 as well. I just wanted to add that this change should be backward compatible :)

beorn7:
:+1: Yeah, at least this incrementally improve things. I still have to think about how to solve generic HTTP instrumentation properly...

Thanks.

PR #141, opened 2015-06-16
Remove www subdomain from Prometheus link.

ttacon:
The prometheus link in the README is broken - there is no page at www.prometheus.io (it's at promethius.io - I guess they didn't set up a redirect). @beorn7


grobie:
Thanks!

beorn7:
OK, got it. The original comment was in @ttacon 's fork, for which the notifications are sent to my private email... so forget all my noise, please.

@juliusv what still matters: Please create a redirect www.prometheus.io -> prometheus.io.

juliusv:
@beorn7 Added the redirect, should be active in a while.

PR #142, opened 2015-06-23
Rename ExporterLabelPrefix to ExportedLabelPrefix.

fabxc:
This is related to prometheus/prometheus#786

### DIFF
--- a/model/labelname.go
+++ b/model/labelname.go
@@ -21,9 +21,9 @@ import (
 )
 
 const (
-	// ExporterLabelPrefix is the label name prefix to prepend if a
-	// synthetic label is already present in the exported metrics.
-	ExporterLabelPrefix LabelName = "exporter_"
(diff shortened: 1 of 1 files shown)


model/labelname.go
```diff
@@ -21,9 +21,9 @@ import (
 )
 
 const (
-	// ExporterLabelPrefix is the label name prefix to prepend if a
+	// ExportedLabelPrefix is the label name prefix to prepend if a
 	// synthetic label is already present in the exported metrics.
```
beorn7: The doc comment should be updated to reflect the priority.

Suggestion:

ExportedLabelPrefix is the prefix to prepend to the label names present in exported metrics if a label of the same name is added by the server.

fabxc: Done.

beorn7:
:+1:

PR #152, opened 2015-08-03
More explicit documentation of SummaryOpts.Objectives.

robx:
I ended up following the source to the quantile package https://godoc.org/github.com/prometheus/client_golang/Godeps/_workspace/src/github.com/beorn7/perks/quantile#NewTargeted to understand the meaning of `Objectives`, so some kind of documentation change would be appreciated. Maybe this works?

### DIFF
--- a/prometheus/summary.go
+++ b/prometheus/summary.go
@@ -112,7 +112,9 @@ type SummaryOpts struct {
 	ConstLabels Labels
 
 	// Objectives defines the quantile rank estimates with their respective
-	// absolute error. The default value is DefObjectives.
+	// absolute error. If Objectives[q] = e, then the reported q-quantile
+	// value will be the r-quantile value with r between q-e and q+e.
+	// The default value is DefObjectives.
 	Objectives map[float64]float64
 
 	// MaxAge defines the duration for which an observation stays relevant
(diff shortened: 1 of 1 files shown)


prometheus/summary.go
```diff
@@ -112,7 +112,9 @@ type SummaryOpts struct {
 	ConstLabels Labels
 
 	// Objectives defines the quantile rank estimates with their respective
-	// absolute error. The default value is DefObjectives.
+	// absolute error. If Objectives[q] = e, then the reported q-quantile
+	// value will be the r-quantile value with r between q-e and q+e.
+	// The default value is DefObjectives.
```
beorn7: The naming of quantiles is very confusing in the literature. An _r-quantile_ would suggest something like '4-quantile', which would refer to quartiles. The most fitting name for the fractional quantile I could find is _φ-quantile_. My co-developers lack the humanist education I have enjoyed and regularly chastise me for using Greek letters, but in this case we concluded it is least confusing after all. Also to be consistent with the documentation on prometheus.io, I'd suggest the following:

```
If Objectives[q] = e, then the value reported for q
will be the φ-quantile value with φ between q-e and q+e.
The default value is DefObjectives.
```

robx:
Ack, updated (with small variant).

beorn7:
Nice. Thanks. Merging...

PR #154, opened 2015-08-03
Use spaces instead of tabs to align equals (=)

ashishgandhi:
Using tabs to align things has the downside that people who have set the
tabstop to be of a different lenght than the orignial author's run the
risk of not having the alignment stay. In the current Makefile before
this change people with tabstop of 4 do not see the equals line up.

set tabstop=4
set tabstop=8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,17 +32,17 @@ endif
 # Never honor GOBIN, should it be set at all.
 unexport GOBIN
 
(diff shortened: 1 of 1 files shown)


beorn7:
Sure, why not... :)

Thanks for the contribution.

But note that the Makefile is really only used for the examples and the tests. We are currently evaluating if we can remove it altogether.

ashishgandhi:
That's fair. We aren't using it either. We vendor our own dependencies (like old school) with two `gopath`s.

PR #156, opened 2015-08-06
Add scheme as label to allow relabelling in discovery

jimmidyson:
As discussed in https://github.com/prometheus/prometheus/issues/910#issuecomment-128305407

### DIFF
--- a/model/labelname.go
+++ b/model/labelname.go
@@ -29,6 +29,10 @@ const (
 	// timeseries.
 	MetricNameLabel LabelName = "__name__"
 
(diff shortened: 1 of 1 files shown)


fabxc:
:+1: 

Assuming you want to use this in Prometheus :)
The usual procedure is now to do the same modification in Prometheus's vendoring directory along with your change because we know for certain that this will get in with the next vendoring upgrade. This avoids a hassle of timing and constant vendoring upgrades.

jimmidyson:
@fabxc That was my next question - when will I be able to use this? Thanks for being psychic ;)

PR #160, opened 2015-08-12
Update histogram_test.go

ardan-bkennedy:
These tests are always timing out in our Jenkins CI environment. We have moved our timeout to 2 minutes and it still timed out. I have studies the code and can't find any reason for this. There does not appear to be a deadlock at all. The stack traces for the time out change a bit with just the slightest changes in the test code. I have added a check for the short flag so other can skip these tests as we are until this can be identified.

### DIFF
--- a/prometheus/histogram_test.go
+++ b/prometheus/histogram_test.go
@@ -124,6 +124,10 @@ func BenchmarkHistogramWrite8(b *testing.B) {
 var testBuckets = []float64{-2, -1, -0.5, 0, 0.5, 1, 2, math.Inf(+1)}
 
 func TestHistogramConcurrency(t *testing.T) {
+	if testing.Short() {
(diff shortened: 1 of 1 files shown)


fabxc:
I'd prefer to fix the actual issue before the next release. It's arguably a hack that masks a problem which will slow down finding a solution if anything.

ardan-bkennedy:
No worries. But these seem to be long running tests when they do work. I thought it was ok to use the flag on these.

fabxc:
Ah I see. Then it probably makes sense.
@beorn7 might want to have a look at this too after his vacation.

beorn7:
Yeah, these concurrency tests are long-running by design. Makes sense to mark them as such.

Note that by default, they are still run. You have to explicitly request to skip long-running tests.

Issue #159, opened 2015-08-12
TestEmptyLabelSignature


ardan-bkennedy:
This test is always failing in our linux environment using go 1.4.2

input := []map[string]string{nil, {}}

var ms runtime.MemStats
runtime.ReadMemStats(&ms)

alloc := ms.Alloc

for _, labels := range input {
	LabelsToSignature(labels)
}

runtime.ReadMemStats(&ms)

if got := ms.Alloc; alloc != got {
	t.Fatal("expected LabelsToSignature with empty labels not to perform allocations")
}

fabxc:
Can you provide some details about your linux environment? Anything special that's explicitly different from regular distributions such as Ubuntu?

Regardless of the cause in your environment, if succeeding of this test is dependent on anything but the implementation it should probably be removed.

ardan-bkennedy:
I did not allocate the box, it comes from our open stack environment. I will try to find out more for you.

This is how we run the tests and we are using go 1.4.2

cd src/github.comcast.com/ventris/kober
GORACE="log_path=stdout" go test -short -bench=Bench* -test.benchmem -timeout 2m -cpu 4 -race -v ./... | tee $WORKSPACE/test.output 
cat $WORKSPACE/test.output | "$WORKSPACE/bin/go2xunit" -fail-on-race=true -output $WORKSPACE/tests.xml
cat $WORKSPACE/test.output | "$WORKSPACE/bin/gobench2plot" > $WORKSPACE/benchmarks.xml

ardan-bkennedy:
I found out information about the machine/VM

CentOS 6.4 and a openstack VM
4 X CPU, 8GB memory

fabxc:
This is no longer a test. It was moved to a benchmark so the allocation efficiency can be verified by developers but doesn't make tests fail based on environment circumstances.

PR #161, opened 2015-08-13
Remove runtime-dependent test.

fabxc:
Success of this test was runtime dependent and did not validate
the implementation logic itself.

@brian-brazil

### DIFF
--- a/model/signature_test.go
+++ b/model/signature_test.go
@@ -14,7 +14,6 @@
 package model
 
 import (
-	"runtime"
 	"sync"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


brian-brazil:
I'm afraid my knowledge of go runtimes isn't sufficient to know if this is correct. Are there docs or something on the behaviour of the different runtimes?

fabxc:
#159 describes a case where it is failing. Reason is not clear but I think unit test should validate implementation semantics.

There are just too many variables to do such runtime checks. Changes to the compiler's escape analysis etc. may break such tests in the future. Optimizing for allocations is good but should not be validated through unit tests.

brian-brazil:
:+1: 

I'm betting it was the race condition checked in that case.

fabxc:
The issue says it's _always_ failing, though.

brian-brazil:
Presuambly they were always testing with the race condition checker turned on, and that's doing some allocations.

fabxc:
Ah, I see what you mean. One reason more why this should not be a test, I guess :)

grobie:
Hmmm. I think it's valid to check for allocations in tests. Do we know why this is failing exactly? If it's because of the race condition checker, maybe extract this into another file and add some flags to not run it in this case?

fabxc:
Horrible timing.

It's a super fragile test. There are no documented guarantees by the compiler based on which we can make such assumptions in a test. Plus the implementation is stable and simple - easy to validate that the code logically strives for minimum allocations.

Of course we can revert this should there be strong opposing opinions.

grobie:
I'd vote to keep it in some way. For example in a different file with some tag, so that developers have at least a chance to run it optionally.

grobie:
Or a benchmark.

fabxc:
As agreed on other channels, I'll turn it into a benchmark.

PR #162, opened 2015-08-13
Turn allocation test into a benchmark.

fabxc:
It even stays at logging an error. But now it is not in the testing path which should succeed without exception if the code compiles.

@grobie

### DIFF
--- a/model/signature_test.go
+++ b/model/signature_test.go
@@ -14,6 +14,7 @@
 package model
 
 import (
+	"runtime"
 	"sync"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


grobie:
:+1:

grobie:
So if one runs the benchmark with trace, will we see the same problem? Should we consider to detect incompatible flags and use [B.Skipf](https://godoc.org/testing#B.Skipf) in that case?

grobie:
I can't get the benchmark to fail. It's probably good to go.

fabxc:
Yes, I expect developers running benchmarks to be aware of conditions that might take effect.

fabxc:
@grobie @brian-brazil 
Just a little follow-up. In the port of the model to common/model I still had this test. From one test run to another I got this test failing. There were no changes at all to my environment.

Just as an indicator how fragile the whole thing was.

PR #167, opened 2015-08-23
Remove client_golang/model als direct dependency

fabxc:
This removes client_golang/model as a direct dependency from the instrumentation.
It's still a dependency via the text parser, however.

It was only used for constants, which are well-specified enough that replicating them is okay.

@brian-brazil @juliusv


brian-brazil:
:+1:

PR #168, opened 2015-08-24
Remove client_golang/model dependency from text package

fabxc:
@brian-brazil @juliusv @beorn7 

As for the instrumentation package the constants are well defined and stable, so replicating is fine. The label set hashing is internal to the package, so it doesn't have to be identical with anything else. Thus, replicating the hashing code is worth dropping the dependency.

With that the instrumentation package no longer depends on client_golang/model at all. Neither does the Prometheus server.

Is there any other place using it? Otherwise we should delete it soon.

### DIFF
--- a/text/create.go
+++ b/text/create.go
@@ -27,7 +27,6 @@ import (
 	"math"
 	"strings"
 
-	"github.com/prometheus/client_golang/model"
 	dto "github.com/prometheus/client_model/go"
 )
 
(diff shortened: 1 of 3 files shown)


brian-brazil:
:+1:

juliusv:
:+1: but `migrate`, `pushgateway`, and `statsd_bridge` still have this dependency and need to be updated first.

juliusv:
(at least the latter two, which don't use vendoring yet)

beorn7:
@fabxc Is that update of other repos still ongoing?

fabxc:
Theorey yes, practically, not gotten anywhere close to it. `migrate` is wasted effort, though and has it vendored.

beorn7:
We'll actually move everything over from `client_golang/text` to `common/expfmt`. Then the whole text package can be deleted.

In other words: This is obsolete.

PR #173, opened 2015-09-17
Remove the text, model, and extraction packages.

beorn7:
They have all been moved to common/expfmt. All uses of it have been
migrated.

This is the final PR to pull the plug. Only to be merged once everything else has been merged.


beorn7:
@fabxc

juliusv:
:+1: RIP

eparis:
kubernetes PRs keep popping up here. We have a test that verifies that godeps are reproducable. Kubernetes is using github.com/prometheus/client_golang/extraction and model in two places.

https://github.com/kubernetes/kubernetes/blob/master/test/e2e/kubelet_stats.go
https://github.com/kubernetes/kubernetes/blob/master/test/e2e/util.go

Since the directories were removed `godep restore` fails. Even though the directories exist in the version we vendored, since they don't exist in your HEAD, godep dies. I'm about to go fill a bug in on godep.

But I guess I should also try to figure out how to port to a newer version. I know nothing about prometheus.  Maybe you guys have a second to let me know they new way to do whatever we are doing with extraction and model?

beorn7:
@eparis Sorry for that. This code was dead for quite some time. We blundered in letting it lie around here for so long and not documenting the migration path. The same functionality is now in https://github.com/prometheus/common/tree/master/expfmt - and of course, the documentation there is outdated, too... We had a bit of urge to straight all this out today. I'm working on the documentation. Should all be in place by tomorrow.

PR #176, opened 2015-09-23
allow creating seperate registries

mattkanwisher:
While not as full featured as @tsenart 's PR, it does enough for me to have a separate registry inside my app which is good enough.

### DIFF
--- a/prometheus/process_collector_test.go
+++ b/prometheus/process_collector_test.go
@@ -16,7 +16,7 @@ func TestProcessCollector(t *testing.T) {
 		t.Skipf("skipping TestProcessCollector, procfs not available: %s", err)
 	}
 
-	registry := newRegistry()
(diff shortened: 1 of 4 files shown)


tsenart:
Hi Matt! Making the constructor public isn't enough. You should make the returned type public too. Returning private types from public methods isn't a good practice.

While you're at it, document the new public types and functions. :-)

mattkanwisher:
@tsenart hah thanks *fixed* just had the same conversation with @bliles

beorn7:
@mattkanwisher I'm actively working on exactly this problem. It's unfortunately much more complicated than it appears to the uninitiated... It would be utterly confusing to merge an ad-hoc solution now.

Feel free to use this for your own needs, but we really cannot do anything like that here in the public repo.

Sorry...

PR #177, opened 2015-10-01
[WIP] Add initial alertmanager client package

fabxc:
Currently only covers `/api/v1/silences` and `/api/v1/silence/:id`. More to follow, including tests.

Usage mostly by AM acceptance test suite and later on Prometheus itself.

### DIFF
--- a/api/alertmanager/api.go
+++ b/api/alertmanager/api.go
@@ -1,0 +1,351 @@
+// Copyright 2015 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package alertmanager
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/url"
+	"path"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/prometheus/common/model"
+	"golang.org/x/net/context"
+	"golang.org/x/net/context/ctxhttp"
+)
+
+const (
+	statusAPIError = 422
+	apiPrefix      = "/api/v1"
+
+	epSilence  = "/silence/:id"
+	epSilences = "/silences"
+	epAlerts   = "/alerts"
+)
+
+type ErrorType string
+
+const (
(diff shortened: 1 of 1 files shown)


api/alertmanager/api.go
```diff
@@ -0,0 +1,351 @@
...
+
+	if code == statusAPIError && result.Status == "error" {
+		err = &Error{
+			Type: result.ErrorType,
+			Msg:  result.Error,
+		}
+	}
+
+	return resp, []byte(result.Data), err
+}
```
juliusv: Hmm, up until here it's mostly a 1:1 copy of the code in https://github.com/prometheus/client_golang/blob/master/api/prometheus/api.go. Given it's 200 lines, wouldn't it make sense to share more of this common code?

fabxc: Absolutely. I'm mostly interested to have something working right now but that should be done eventually.

api/alertmanager/api.go
```diff
@@ -0,0 +1,351 @@
...
+		return err
+	}
+
+	req, _ := http.NewRequest("POST", u.String(), &buf)
+
+	_, _, err := h.client.do(ctx, req)
+	return err
+}
+
+// SilenceAPI provides bindings the Alertmanager's silence API.
```
juliusv: s/bindings/bindings for/

Add similar docstring to `AlertAPI`

api/alertmanager/api.go
```diff
@@ -0,0 +1,351 @@
...
+type httpSilenceAPI struct {
+	client Client
+}
+
+func (h *httpSilenceAPI) Get(ctx context.Context, id uint64) (*model.Silence, error) {
+	u := h.client.url(epSilence, map[string]string{
+		"id": strconv.FormatUint(id, 10),
+	})
+
+	req, _ := http.NewRequest("GET", u.String(), nil)
```
juliusv: don't want to bail early here (and below) if there is an error?

api/alertmanager/api.go
```diff
@@ -0,0 +1,351 @@
...
+		u      *url.URL
+		method string
+	)
+
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(sil); err != nil {
+		return 0, err
+	}
+
+	// Talk to different endpoints depending on whether its a new silence or not.
```
juliusv: its -> it's

api/alertmanager/api.go
juliusv: Hmm, AFAICS it's even mandatory to act on the error in this situation because otherwise `req` will be `nil` and the subsequent `do` call will internally get `nil` pointer derefs if there's an error? Like in https://github.com/golang/net/blob/master/context/ctxhttp/cancelreq.go#L13

api/alertmanager/api.go
juliusv: Cool.

juliusv:
Looks good generally.

fabxc:
Those should rather live within the respective repositories. Too much synching issues even with theoretically stable APIs.

PR #180, opened 2015-11-09
Make metric sort stable and sort by timestamp

dnesting:
Presently metricSorter is unstable because Less returns "true" instead of "false" when two metrics compare the same.  This means when Collectors emit timestamped metrics in order, this sort will perturb the ordering and result in metrics delivered out-of-order to Prometheus.  This PR fixes the return value for Less.

Since we're going through the trouble of reordering metrics here anyway, explicitly sort by TimestampMs as well.

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -32,7 +32,6 @@ import (
 	"sort"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/prometheus/common/expfmt"
@@ -723,26 +722,22 @@ func (s metricSorter) Less(i, j int) bool {
 			return vi < vj
 		}
 	}
-	if s[i].GetTimestampMs() != s[j].GetTimestampMs() {
-		// While this is not strictly supported, until we have
-		// an import API some people are having some success
-		// getting data imported by supplying the same metric
-		// with different timestamps.  At the very least we
-		// shouldn't make things more difficult, so ensure that
-		// metrics get ordered by timestamp to make it possible
-		// for Prometheus to import them.
-		ti, tj := s[i].GetTimestampMs(), s[j].GetTimestampMs()
-		if ti == 0 {
-			ti = nowMillis()
-		} else if tj == 0 {
-			tj = nowMillis()
-		}
-		return ti < tj
-	}
-	return false
-}
 
-func nowMillis() int64 {
-	now := time.Now()
-	return now.Unix()*1000 + int64(now.Nanosecond())/1000000
+	// While this is not strictly supported, until we have an
+	// import API some people are having some success getting
+	// data imported by supplying the same metric with different
+	// timestamps.  At the very least we shouldn't make things
+	// more difficult, so ensure that metrics get ordered by
+	// timestamp so that Prometheus doesn't complain about
+	// out-of-order metrics. A missing timestamp (implies "now")
+	// will be ordered last.
+	if s[i].TimestampMs == nil && s[j].TimestampMs != nil {
+		return false
+	} else if s[i].TimestampMs != nil && s[j].TimestampMs == nil {
+		return true
+	} else if s[i].GetTimestampMs() != s[j].GetTimestampMs() {
+		return s[i].GetTimestampMs() < s[j].GetTimestampMs()
+	}
+
+	return false
 }
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -723,21 +723,17 @@ func (s metricSorter) Less(i, j int) bool {
 		}
 	}
 
-	// While this is not strictly supported, until we have an
-	// import API some people are having some success getting
-	// data imported by supplying the same metric with different
-	// timestamps.  At the very least we shouldn't make things
-	// more difficult, so ensure that metrics get ordered by
-	// timestamp so that Prometheus doesn't complain about
-	// out-of-order metrics. A missing timestamp (implies "now")
-	// will be ordered last.
-	if s[i].TimestampMs == nil && s[j].TimestampMs != nil {
+	// We should never arrive here. Multiple metrics with the same
+	// label set in the same scrape will lead to undefined ingestion
+	// behavior. However, as above, we have to provide stable sorting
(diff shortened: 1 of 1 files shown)


dnesting:
I don't know if it's appropriate to treat a zero timestamp as equivalent to "now" or not, but it seemed right, thinking about how Prometheus might interpret it.

@beorn7

prometheus/registry.go
```diff
@@ -722,5 +723,23 @@ func (s metricSorter) Less(i, j int) bool {
...
 		}
 	}
-	return true
+	if s[i].GetTimestampMs() != s[j].GetTimestampMs() {
+		// Ensure metrics are ordered by timestamp.  Treat
+		// a zero timestamp as "now" in case we have
+		// untimestamped metrics mixed with timestamped
+		// metrics.
+		ti, tj := s[i].GetTimestampMs(), s[j].GetTimestampMs()
+		if ti == 0 {
```
beorn7: Two things here:

(1) problem of an unset timestamp

It's a general issue how to distinguish an unset timestamp and a timestamp == 0, especially as the latter might appear as the former after a serialization round-trip. We are considering using MinInt64 as default value in the protobuf definition.

To not break in that case, you should check for `s[i].TimestampMs == nil` to determine if the timestamp is unset.

(2) problem of "now"

With calling `time.Now()`, the sorting depends on the time of the call (and it might even change during sorting). That's dangerous. I'd suggest to sort an unset timestamp always at the beginning or always at the end.

beorn7:
Thanks for your contribution. The current behavior breaks the `Less` contract, so it's indeed a bug. However, it shouldn't matter because the metrics must never be equal. (You could create equal metrics with a custom `Collector` and not setting `EnableCollectChecks(true)`. But it would be against the contract for metric exposition. The ingestion behavior of the server is undefined in that case. (It might ingest fine, it might only ingest one of the equal metrics, it might ingest one of the equal metrics with a timestamp of another, or it might ingest none.))

Still I think we should guarantee consistent sorting. Could you acknowledge the circumstances why we are doing it in a comment? Similar to https://github.com/dnesting/client_golang/blob/dnesting-sort-metrics/prometheus/registry.go#L711-L716 .

dnesting:
The contract precludes multiple metrics at different timestamps?  I am specifically writing an importer, and this largely works save this one issue.  I'll add a comment once I'm clear on this question.  Thanks!

beorn7:
We should definitely document the contract more clearly. So far, we essentially tell people to not use the `timestamp_ms` field unless they know what they are doing.

If you have the same metric multiple times with different time-stamps in the same scrape, then it only works by accident. (I can imagine it works with protobuf exposition, but I'm pretty sure it wouldn't work with the text format.)

Bulk import has to be implemented properly by the server, see https://github.com/prometheus/prometheus/issues/535 .

The only way I see it work is to distribute the metrics over different scrapes, but that clearly shows it's not meant that way…

juliusv:
As I mention in https://github.com/prometheus/prometheus/issues/10#issuecomment-150024410, using the existing format could be an option for the import API (albeit not an efficient one if one wants to import many points for one series), but then it would indeed be great to clarify what happens if the same metric / metric family is encountered multiple times with the same timestamp.

dnesting:
I've updated the comment.  I would also be OK dropping this block of code and just keeping the change to the return value of Less, leaving the ordering burden on the Collector implementation.  This should work equally well and wouldn't add complexity for an unsupported use case.  Thoughts?

dnesting:
Sorry I just now noticed your line comments and will revise accordingly.

prometheus/registry.go
dnesting: Sounds good.  I've updated to order a nil timestamp at the end (most likely that is the best "now" location) and eliminated the dependency on time.  PTAL?

prometheus/registry.go
```diff
@@ -722,5 +722,22 @@ func (s metricSorter) Less(i, j int) bool {
...
+	// will be ordered last.
+	if s[i].TimestampMs == nil && s[j].TimestampMs != nil {
+		return false
+	} else if s[i].TimestampMs != nil && s[j].TimestampMs == nil {
+		return true
+	} else if s[i].GetTimestampMs() != s[j].GetTimestampMs() {
+		return s[i].GetTimestampMs() < s[j].GetTimestampMs()
+	}
+
+	return false
```
beorn7: I think the whole section (from line 734) can be written simpler as

```
if s[i].TimestampMs == nil {
    return false
}
if s[j].TimestampMs == nil {
    return true
}
return s[i].GetTimestampMs() < s[j].GetTimestampMs()
```

beorn7:
So let's definitely sort by timestamp (at the very least, we might use the metric sort for something else that exposition one day). Error handling happens elsewhere (e.g. if `EnableCollectChecks(true)` has been called). But let's not even mention that you do bulk imports in that way. That would set a really bad precedence.

prometheus/registry.go
```diff
@@ -722,5 +722,22 @@ func (s metricSorter) Less(i, j int) bool {
...
-	return true
+
+	// While this is not strictly supported, until we have an
+	// import API some people are having some success getting
+	// data imported by supplying the same metric with different
+	// timestamps.  At the very least we shouldn't make things
+	// more difficult, so ensure that metrics get ordered by
+	// timestamp so that Prometheus doesn't complain about
+	// out-of-order metrics. A missing timestamp (implies "now")
+	// will be ordered last.
```
beorn7: Sorry for being picky, but I really don't want to suggest in any way here that you could do bulk import in that way. (Even your current success may cease to work with any new commit we do.) Suggestion for the comment:

"We should never arrive here. Multiple metrics with the same label set in the same scrape will lead to undefined ingestion behavior. However, as above, we have to provide stable sorting here, even for inconsistent metrics. So sort equal metrics by their timestamp, with missing timestamps (implying "now") coming last."

dnesting:
I agree, that's simpler and I've updated the comment.  Do you prefer squashed commits?

beorn7:
It depends. I tend to only squash if it makes a big difference. Not in this case...

Thanks.

PR #195, opened 2016-03-17
Remove anti-pattern of having target labels in instrumentation example.

brian-brazil:


beorn7:
:+1: 

Obviously, we need better explanation for how to use const labels, but I'll take care of that during the rewrite...

beorn7:
Failed test is the flaky one...

brian-brazil:
There's an example further down the file that's okay.

Issue #46, opened 2015-01-12
Create a public, configurable Registry.


beorn7:
For testing purposes and for advanced configuration (and possibly other uses), we want a public Registry interface with the possibility of plugging together implementations that behave according to the user's requirements (error handling, which standard metrics to export, http instrumentation, exposition - which might or might not be via HTTP, ...).

The Registry interface should be lean, possibly with parts of the functionality delegated to other interfaces or even modules (HTTP serving and/or instrumentation, Pushgateway handling, ...).

Currently, there is only one global registry, and it is not exposed to the user. It is used only via global convenience methods. We want to avoid a further explosion of the number of those functions. The good news is that for as long we support the existing functions, we can change everything behind it at will. Basic idea is to keep the existing convenience functions (with the option of deprecating those that are not used a lot), while providing advanced functionality via a more elegant but also more complex to use API (where you create a Registry implementation first that is configured to your needs and then register metrics there and decide about HTTP exposition and such). For inspiration, see https://github.com/prometheus/client_golang/pull/38 (code and discussion).

The convenience functions to keep are:
- `EnableCollectChecks`
- `UninstrumentedHandler`
- `Handler`
- `InstrumentHandler`
- `InstrumentHandlerFunc`
- `InstrumentHandlerWithOpts`
- `InstrumentHandlerFuncWithOpts`
- `PanicOnCollectError`
- `Push`
- `PushAdd`
- `SetMetricFamilyInjectionHook`
- `Unregister`
- `Registe`
- `RegisterOrGet`
- `MustRegister`
- `MustRegisterOrGet`

beorn7:
@tsenart Lest we forget...

tsenart:
:-)

abustany:
That would be really convenient to eg. set ConstLabels on all metrics that are implicitly exposed, for example the "number of goroutines" and "gc stats" ones. Right now I don't see any way to alter those (the inject hook can apparently only be used to inject new metrics).

brian-brazil:
> That would be really convenient to eg. set ConstLabels on all metrics that are implicitly exposed

You're looking for relabelling on the prometheus side, see http://prometheus.io/docs/operating/configuration/#target-relabeling-relabel_config

abustany:
[...]

PR #198, opened 2016-04-19
add a Registry interface and NewRegistry

alfred-landrum:
For testing purposes, it would be convenient to create a new
registry that is unrelated to the default registry. For example,
this allows running tests in parallel that update and verify metrics.

@beorn7

### DIFF
--- a/prometheus/process_collector_test.go
+++ b/prometheus/process_collector_test.go
@@ -16,7 +16,7 @@ func TestProcessCollector(t *testing.T) {
 		t.Skipf("skipping TestProcessCollector, procfs not available: %s", err)
 	}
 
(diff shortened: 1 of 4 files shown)


beorn7:
Yes, this is a well know problem. See #46 .

I'm already working on this. Unfortunately, it's not as easy as it looks. Feel free to play with it on your own, but merging any half-bred solution right now would be counter-productive.

alfred-landrum:
ok, will study #46, thanks for the quick response.

beorn7:
Essentially, the (as of now internal) interface of the registry is flawed, and I don't want to make those flaws exported. The new registry will follow better design practices and solve a number of other problems, too.

PR #199, opened 2016-05-12
Fix max_fds process collector test

mattharden:
This test fails when max_fds is a large value; say 4.5e+06, for example. Change it to match virtual_memory_bytes, which also has to handle large values.


brian-brazil:
Thanks!

Issue #200, opened 2016-05-15
Race with httputil.ReverseProxy


raliste:
We are using `prometheus.InstrumentHandler("test", handler)` where handler is an `httputil.ReverseProxy`. 

```
==================
WARNING: DATA RACE
Read by goroutine 36:
  github.com/prometheus/client_golang/prometheus.computeApproximateRequestSize()
      /Volumes/PreySecure/code/src/github.com/prometheus/client_golang/prometheus/http.go:176 +0x1eb

Previous write by goroutine 42:
  net/textproto.MIMEHeader.Set()
      /usr/local/go/src/net/textproto/header.go:22 +0x94
  net/http.Header.Set()
      /usr/local/go/src/net/http/header.go:31 +0x56
  net/http/httputil.(*ReverseProxy).ServeHTTP()
      /usr/local/go/src/net/http/httputil/reverseproxy.go:207 +0xb19
  net/http.(Handler).ServeHTTP-fm()
      /usr/local/go/src/net/http/h2_bundle.go:3847 +0x5d
  github.com/prometheus/client_golang/prometheus.InstrumentHandlerFuncWithOpts.func1()
      /Volumes/PreySecure/code/src/github.com/prometheus/client_golang/prometheus/http.go:158 +0x3c1
  net/http.HandlerFunc.ServeHTTP()
      /usr/local/go/src/net/http/server.go:1618 +0x47
  main.(*Server).ServeHTTP()
      /Users/raliste/Work/PreySecure/code/src/github.preyhq.com/prey/net/upload/preyfrontend.go:204 +0x47e
  net/http.(*ServeMux).ServeHTTP()
      /usr/local/go/src/net/http/server.go:1910 +0x212
  github.preyhq.com/prey/net/frontend/go/server.(*Server).ServeHTTP()
      /Volumes/PreySecure/code/src/github.preyhq.com/prey/net/frontend/go/server/server.go:36 +0x398
  net/http.serverHandler.ServeHTTP()
      /usr/local/go/src/net/http/server.go:2081 +0x206
  net/http.(*conn).serve()
      /usr/local/go/src/net/http/server.go:1472 +0x1565

Goroutine 36 (running) created at:
  github.com/prometheus/client_golang/prometheus.InstrumentHandlerFuncWithOpts.func1()
      /Volumes/PreySecure/code/src/github.com/prometheus/client_golang/prometheus/http.go:146 +0x1ff
  net/http.HandlerFunc.ServeHTTP()
[...]

PR #201, opened 2016-05-15
Document the issues of InstrumentHandler

beorn7:
Obvious next step: Fix those issues. (Which I'm already working on.)

@raliste This at least documents what you have run into in #200 .

### DIFF
--- a/prometheus/http.go
+++ b/prometheus/http.go
@@ -57,12 +57,31 @@ func nowSeries(t ...time.Time) nower {
 // has a constant label named "handler" with the provided handlerName as
 // value. http_requests_total is a metric vector partitioned by HTTP method
 // (label name "method") and HTTP status code (label name "code").
+//
+// Note that InstrumentHandler has several issues:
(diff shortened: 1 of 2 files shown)


raliste:
Thank you @beorn7 

I'll use direct instrumentation with request timing in the mean time.

beorn7:
We need to fix the CI tests... unrelated to this (doc-comment only) change, though. Merging...

F21:
@beorn7 Sorry for going a bit off topic, but is there any progress on fixing the InstrumentHandler? I am happy to help if needed :smile:

beorn7:
I recommend explicit instrumentation for now. There is no easy "fix" for the issues of `InstrumentHandler`. The problem has to be approached from a very different angle, which I'm working on. Explaining the context for that would take longer than implementing it myself.

F21:
Cool :smile: Looking forward to seeing your implementation. If you need any help, let me know!

PR #202, opened 2016-05-31
prometheus/docs: Added a link to the metric types page

mattyw:
Sorry to driveby like this. Thought it might be useful to link to the metric types doc somewhere


beorn7:
:+1: Thanks.

PR #209, opened 2016-06-27
Fix GoDoc indentation

mattbostock:
The line after the first in each bullet point was appearing as
preformatted blocks when viewed as HTML in GoDoc.

E.g.: https://godoc.org/github.com/prometheus/client_golang/prometheus#InstrumentHandler


brian-brazil:
Thanks!

PR #210, opened 2016-07-06
Missing alignment issue for 32bit machines.

jaredeh:
I found this issue as I audited an older version included in coreos/etcd.  The background here is that if you don't take care with the alignment of 64bit fields in structs you can't safely use sync/atomic primitives on them on 32bit architectures.  The comments in the struct look like at some point there care was taken to address this alignment.  Perhaps counts[] was added at a later time?

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -216,11 +216,12 @@ func newHistogram(desc *Desc, opts HistogramOpts, labelValues ...string) Histogr
 
 type histogram struct {
 	// sumBits contains the bits of the float64 representing the sum of all
-	// observations. sumBits and count have to go first in the struct to
-	// guarantee alignment for atomic operations.
+	// observations. sumBits, count, and counts have to go first in the struct
+	// to guarantee alignment for atomic operations.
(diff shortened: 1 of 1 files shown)


beorn7:
Thanks for raising this point.

My assumption always has been that the int64s in the slice are aligned within the slice, which is not related to the alignment of `count []uint64` within the struct. The latter, after all, just keeps a pointer to the int64 slice.

DId you see any errors showing up? I cannot create any errors with the current code on the 32bit 386 arch.

jaredeh:
etcd crashed because of related issue so I audited everything.  But I didn't get a chance to see a crash directly from this issue.

I too was unsure of the the slice alignment.  So I wrote a test for it, but as I read your question I realized the test was flawed.  I was testing the behavior of `counts [10]uint64` instead of `count []uint64; h.counts = make([]uint64, 10) `oops.  I just ran a corrected tests.

My conclusion is that the make() will properly align counts[] as you expected.  Sorry for the inconvenience.  I'll close this out.

beorn7:
Never mind. Thanks for checking.

PR #211, opened 2016-07-08
Add *RegisterAll functions with variadic args

andrewstuart:
This enables a little bit more flexibility and less boilerplate for registering many Collectors.

- One-liner `MustRegisterAll(coll1, coll2, coll3, coll4)`
- Spread a slice `MustRegisterAll(collectors...)`

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -115,11 +115,13 @@ func Register(m Collector) error {
 }
 
 // MustRegister works like Register but panics where Register would have
-// returned an error.
-func MustRegister(m Collector) {
-	err := Register(m)
-	if err != nil {
-		panic(err)
+// returned an error. MustRegister is also Variadic, where Register only
+// accepts a single Collector to register.
+func MustRegister(m ...Collector) {
+	for i := range m {
+		if err := Register(m[i]); err != nil {
+			panic(err)
+		}
 	}
 }
 


prometheus/registry.go
```diff
@@ -114,6 +114,16 @@ func Register(m Collector) error {
 	return err
 }
 
+// RegisterAll works like Register but takes a variadic list of Collectors.
+func RegisterAll(m ...Collector) error {
+	for i := range m {
+		if err := Register(m[i]); err != nil {
+			return err
```
beorn7: The problem here is that you don't know at which Collector the failure has happen if an error is returned.

prometheus/registry.go
```diff
@@ -123,6 +133,16 @@ func MustRegister(m Collector) {
 	}
 }
 
+// MustRegisterAll works like MustRegister but takes a variadic list of
+// Collectors.
+func MustRegisterAll(m ...Collector) {
```
beorn7: How about making the normal `MustRegister` variadic instead? It doesn't break the current call syntax of `MustRegister` and avoids introducing yet another function.

beorn7:
Thanks for your contribution.

Ironically, I'm currently working on making the registry interface leaner (to finally expose a public registry interface).

Those variadic functions only save one line of code, essentially. I'd claim it's more idiomatic Go to write an additional line of code instead of bloating the number of functions. In addition, as commented before, in the case of `RegisterAll`, the behavior if an error is returned somewhere in the middle is problematic.

I suggest to not add a `RegisterAll` because of the error return problem, and to make `MustRegister` variadic so that we can even avoid adding `MustRegisterAll`. Does that make sense?

prometheus/registry.go
andrewstuart: What would you think about returning a descriptive error containing the exact argument that failed? Then the usage between the two functions gets really clean, since it's essentially the same as before, but variadic for both.

prometheus/registry.go
andrewstuart: Sounds good to me. :-)

andrewstuart:
So I've gone ahead and updated MustRegister, though I'd like to propose merging the original RegisterAll and Register, while returning a more-descriptive error about where in the arg list the registration failed.

I think this interface gives the best of both worlds: still only two functions, flexible input, and helpful, descriptive errors. Let me know what you think. If you'd still rather keep the same arity for Register(), I can refactor again; just let me know. :-)

prometheus/registry.go
beorn7: I think that is way to complex to save the user a single for loop. Also, it makes the error handling way more convoluted.

beorn7:
I would say, leave `Register` as it is and just have `MustRegister` variadic (and adjust the doc comment accordingly).

andrewstuart:
Alright, let me know how this looks. I also squashed to keep a cleaner history.

beorn7:
:+1:  I like that. Merging...

PR #212, opened 2016-07-11
Added support for relatetive path when Prom API is behind nginx

roman-vynar:
If Prometheus is behind Nginx at some path, e.g. `/prometheus`, this change will allow to pass the address along with relative URL and will not be ignored.

```
config := prometheus.Config{Address: "http://my-nginx/prometheus")}
```

Currently, `/prometheus` is ignored in this example and the path is hardcoded to `apiPrefix      = "/api/v1"`

### DIFF
--- a/api/prometheus/api.go
+++ b/api/prometheus/api.go
@@ -107,7 +107,7 @@ func New(cfg Config) (Client, error) {
 	if err != nil {
 		return nil, err
 	}
-	u.Path = apiPrefix
+	u.Path = u.Path + apiPrefix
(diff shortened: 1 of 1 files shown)


roman-vynar:
Nevermind, the conflict says
```
<<<<<<< HEAD
        u.Path = u.Path + apiPrefix
=======
        u.Path = strings.TrimRight(u.Path, "/") + apiPrefix
>>>>>>> upstream/master
```
So I am late and closing this :D

PR #213, opened 2016-07-31
api: document goroutine safeness

stapelberg:
From what I can see, both `Client` and `QueryAPI` are safe for use from multiple goroutines as they don’t modify any state. Let’s document this so that others don’t need to review the source as well :).


beorn7:
Totally for documenting goroutine safety. But to make the call here, we need @fabxc .

fabxc:
Thanks 👍

Issue #205, opened 2016-06-02
prometheus.Unregister documentation is incomplete/misleading


haspeterd:
Current documentation for promethus.Unregister() states:
```golang
// Unregister unregisters the Collector that equals the Collector passed in as
// an argument. (Two Collectors are considered equal if their Describe method
// yields the same set of descriptors.) The function returns whether a Collector
// was unregistered.
```

If you dig into the code of registry.Unregister, this comment is in the implementation:
```golang
	// dimHashesByName is left untouched as those must be consistent
	// throughout the lifetime of a program.
```

The fact that dimHashesByName is untouched means that if you naively call prometheus.Unregister() expecting that it will remove all traces of the collector, you can still run into issues later if you try to re-register the same collector name with, e.g., a different help string.

This behavior is a significant part of this function's contract, and it should be called out on the public-facing documentation for prometheus.Unregister.

beorn7:
@haspeterd Now that we have discussed the context so meticulously, how about you create a PR that amends the doc comment? ;-)

Issue #197, opened 2016-04-14
Provide graphite bridge


grobie:
In order to convince people to integrate the prometheus client directly in their project, it would help to have support for a graphite bridge. The Python client already [supports that](https://github.com/prometheus/client_python#graphite).

beorn7:
This depends on https://github.com/prometheus/client_golang/issues/46 , which will separate the registry from exposition, so instead of exposing via HTTP, you can also regularly sent to Graphite.

Issue #170, opened 2015-09-03
Include body in error upon pushing to pushgateway


beorn7:
Currently, only the status code is included. But the body contains often a message that is helpful for understanding the problem (like a parse error in the pushed metrics).

Issue #100, opened 2015-03-17
Change signature for functions pushing to pushgateway.


beorn7:
In anticipation of https://github.com/prometheus/pushgateway/issues/18, we want the functions pushing to pushgateway to look like

    Push(job string, grouping map[string]string, url string)

`grouping` can be empty or nil, in which case, only the job is used. Initially, the only supported grouping other than that will be `{"instance": "foo"}`. Once the pushgateway supports arbitrary grouping, the grouping can contain anything except `job`. To avoid creating inconsistent metrics on the pushgateway (which will lead to an error on the pushgateway itself), one of the label names used in the grouping, and neither `job`, must be used by any of the pushed metrics directly. (The functions will check for that.)

There will be a convenience function to push with the own hostname as `instance`:

    prometheus.Push("my_job", prometheus.GroupByHostname(), "pushgateway:9091")

PR #214, opened 2016-08-01
Create a public registry interface and separate out HTTP exposition

beorn7:
@grobie who might be most keen on this.
@juliusv whom I gave a short preview in person.
But then probably everybody else is interested, @fabxc , @brian-brazil , …

General context and approch
===========================

This is the first part of the long awaited wider refurbishment of
`client_golang/prometheus/...`. After a lot of struggling, I decided
to not go for one breaking big-bang, but cut things into smaller steps
after all, mostly to keep the changes manageable and easy to
review. I'm aiming for having the invasive breaking changes
concentrated in as few steps as possible (ideally one). Some steps
will not be breaking at all, but typically there will be breaking
changes that only affect quite special cases so that 95+% of users
will not be affected. This first step is an example for that, see
details below.

What's happening in this commit?
================================

This step is about finally creating an exported registry
interface. This could not be done by simply export the existing
internal implementation because the interface would be _way_ too
fat. This commit introduces a qutie lean `Registry` interface
(compared to the previous interval implementation). The functions that
act on the default registry are retained (with very few exceptions) so
that most use cases won't see a change. However, several of those are
deprecated now to clean up the namespace in the future.

The default registry is kept in the public variable
`DefaultRegistry`. This follows the example of the http package in the
standard library (cf. `http.DefaultServeMux`, `http.DefaultClient`)
with the same implications. (This pattern is somewhat disputed within
the Go community but I chose to go with the devil you know instead of
creating something more complex or even disallowing any changes to the
default registry. The current approach gives everybody the freedom to
not touch DefaultRegistry or to do everything with a custom registry
to play save.)

Another important part in making the registry lean is the extraction
of the HTTP exposition, which also allows for customization of the
HTTP exposition. Note that the separation of metric collection and
exposition has the side effect that managing the MetricFamily and
Metric protobuf objects in a free-list or pool isn't really feasible
anymore. By now (with better GC in more recent Go versions), the
returns were anyway dimisishing. To be effective at all, scrapes had
to happen more often than GC cycles, and even then most elements of
the protobufs (everything excetp the MetricFamily and Metric structs
themselves) would still cause allocation churn. In a future breaking
change, the signature of the Write method in the Metric interface will
be adjusted accordingly. In this commit, avoiding breakage is more
important.

The following issues are fixed by this commit (some solved "on the
fly" now that I was touching the code anyway and it would have been
stupid to port the bugs):

#46
#100
#170
#205

Documentation including examples have been amended as required.

What future changes does this commit enable?
============================================

The following items are not yet implemented, but this commit opens the
possibility of implementing these independently.

- The separation of the HTTP exposition allows the implementation of
  other exposition methods based on the Registry interface, as known
  from other Prometheus client libraries, e.g. sending the metrics to
  Graphite.
  Cf. #197

- The public `Registry` interface allows the implementation of
  convenience tools for testing metrics collection. Those tools can
  inspect the collected MetricFamily protobufs and compare them to
  expectation. Also, tests can use their own testing instance of a
  registry.
  Cf. #58

Notable non-goals of this commit
================================

Non-goals that will be tackled later
------------------------------------

The following two issues are quite closely connected to the changes in
this commit but the line has been drawn deliberately to address them
in later steps of the refurbishment:

- `InstrumentHandler` has many known problems. The plan is to create a
  saner way to conveniently intrument HTTP handlers and remove the old
  `InstrumentHandler` altogether. To keep breakage low for now, even
  the default handler to expose metrics is still using the old
  `InstrumentHandler`. This leads to weird naming inconsistencies but
  I have deemed it better to not break the world right now but do it
  in the change that provides better ways of instrumenting HTTP
  handlers.
  Cf. #200

- There is work underway to make the whole handling of metric
  descriptors (`Desc`) more intuitive and transparent for the user
  (including an ability for less strict checking,
  cf. #47). That's
  quite invasive from the perspective of the internal code, namely the
  registry. I deliberately kept those changes out of this commit.

- While this commit adds new external dependency, the effort to vendor
  anything within the library that is not visible in any exported
  types will have to be done later.

Non-goals that _might_ be tackled later
---------------------------------------

There is a strong and understandable urge to divide the `prometheus`
package into a number of sub-packages (like `registry`, `collectors`,
`http`, `metrics`, …). However, to not run into a multitude of
circular import chains, this would need to break every single existing
usage of the library. (As just one example, if the ubiquitious
`prometheus.MustRegister` (with more than 2,000 uses on GitHub alone)
is kept in the `prometheus` package, but the other registry concerns
go into a new `registry` package, then the `prometheus` package would
import the `registry` package (to call the actual register method),
while at the same time the `registry` package needs to import the
`prometheus` package to access `Collector`, `Metric`, `Desc` and
more. If we moved `MustRegister` into the `registry` package,
thousands of code lines would have to be fixed (which would be easy if
the world was a mono repo, but it is not). If we moved everything else
the proposed registry package needs into packages of their own, we
would break thousands of other code lines.)

The main problem is really the top-level functions like
`MustRegister`, `Handler`, …, which effectively pull everything into
one package. Those functions are however very convenient for the easy
and very frequent use-cases.

This problem has to be revisited later.

For now, I'm trying to keep the amount of exported names in the
package as low as possible (e.g. I unexported expvarCollector in this
commit because the NewExpvarCollector constructor is enough to export,
and it is now consistent with other collectors, like the goCollector).

Non-goals that won't be tackled anytime soon
--------------------------------------------

Something that I have played with a lot is "streaming collection",
i.e. allow an implementation of the `Registry` interface that collects
metrics incrementally and serves them while doing so. As it has turned
out, this has many many issues and makes the `Registry` interface very
clunky. Eventually, I made the call that it is unlikely we will really
implement streaming collection; and making the interface more clunky
for something that might not even happen is really a big no-no. Note
that the `Registry` interface only creates the in-memory
representation of the metric family protobufs in one go. The
serializaton onto the wire can still be handled in a streaming fashion
(which hasn't been done so far, without causing any trouble, but might
be done in the future without breaking any interfaces).

What are the breaking changes?
==============================

- Signatures of functions pushing to Pushgateway have changed to allow
  arbitrary grouping (which was planned for a long time anyway, and
  now that I had to work on the Push code anyway for the registry
  refurbishment, I finally did it,
  cf. #100).
  With the gained insight that pushing to the default registry is almost
  never the right thing, and now that we are breaking the Push call
  anyway, all the Push functions were moved to their own package,
  which cleans up the namespace and is more idiomatic (pushing
  Collectors is now literally done by `push.Collectors(...)`).

- The registry is doing more consistency checks by default now. Past
  creators of inconsistent metrics could have masked the problem by
  not setting `EnableCollectChecks`. Those inconsistencies will now be
  detected. (But note that a "best effort" metrics collection is now
  possible with `HandlerOpts.ErrorHandling = ContinueOnError`.)

- `EnableCollectChecks` is gone. The registry is now performing some
  of those checks anyway (see previous item), and a registry with all
  of those checks can now be created with `NewPedanticRegistry` (only
  used for testing).

- `PanicOnCollectError` is gone. This behavior can now be configured
  when creating a custom HTTP handler.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/client_golang/214)


beorn7:
Oops... will fix test soon.

prometheus/registry.go
```diff
@@ -11,232 +11,253 @@
...
+	// possible. Hence, if a non-nil error is returned, the returned
+	// MetricFamily slice could be nil (in case of a fatal error that
+	// prevented any meaningful metric collection) or contain a number of
+	// MetricFamily protobufs, some of which might be incomplete, and some
+	// might be missing altogether. The returned error (which might be a
+	// multierror.Error) explains the details. In any case, the MetricFamily
+	// protobufs are consistent and valid for Prometheus to ingest (e.g. no
+	// duplicate metrics, no invalid identifiers). In scenarios where
+	// complete collection is critical, the returned MetricFamily protobufs
+	// should be disregareded if the returned error is non-nil.
```
brian-brazil: disregarded

brian-brazil:
It's great to finally see this.

> There is a strong and understandable urge to divide the prometheus package into a number of sub-packages (like registry, collectors, http, metrics, …). However, to not run into a multitude of

It boils down mostly to a question of dependencies. Like Python, Go comes with HTTP support out of the box so having it all in one package is practical. By contrast that's not the case with Java where we need to pull in external dependencies to work. 

Where we run into issues here is protobuf, as that shouldn't be a dependency of the instrumentation (registry+collector+metrics+http instrumentation) part of the library.

> As just one example, if the ubiquitious prometheus.MustRegister (with more than 2,000 uses on GitHub alone)

To be honest, we should be replacing most of those uses in favour of the metrics auto-registering by default.

What I've done elsewhere is to have the register method on the registry, but also on the metric for convenience.

prometheus/doc.go
```diff
@@ -61,51 +62,135 @@
...
+//
+// So far, everything we did operated on the so-called default registry, as it
+// can be found in the global DefaultRegistry variable. With NewRegistry, you
+// can create a custom registry, or you can even implement the Registry
+// interface yourself. The methods Register and Unregister work in the same way
+// on a custom registry as the global functions Register and Unregister on the
+// default registry. The other global functions are not methods in the Registry
+// interface but there are utility functions that implement the same
+// functionality on top of the interface and take a Registry instance as a
+// parameter. Those are: MustRegister → MustRegisterWith, RegisterOrGet →
```
brian-brazil: This seems a rather big interface, and not following a OOP model by having them as standalone functions rather than methods.

Can we remove the Get methods? The only time I've seen them used is in cases where a ConstMetric should have been used instead.

prometheus/http.go
```diff
@@ -15,14 +15,169 @@ package prometheus
...
+// Handler returns an HTTP handler for the DefaultRegistry. It is
+// already instrumented with InstrumentHandler (using "prometheus" as handler
+// name).
+//
+// Please note the issues described in the doc comment of InstrumentHandler. You
+// might want to consider using UninstrumentedHandler instead.
+//
+// The returned Handler is using the same HandlerOpts as the Handler returned by
+// UninstrumentedHandler. See there for details.
+func Handler() http.Handler {
```
brian-brazil: There's three different constructors here, can this be simplified?

prometheus/push.go
```diff
@@ -19,47 +19,153 @@
...
 //
-// Note that all previously pushed metrics with the same job and instance will
-// be replaced with the metrics pushed by this call. (It uses HTTP method 'PUT'
-// to push to the Pushgateway.)
-func Push(job, instance, url string) error {
-	return defRegistry.Push(job, instance, url, "PUT")
+// Note that all previously pushed metrics with the same job and other grouping
+// labels will be replaced with the metrics pushed by this call. (It uses HTTP
+// method 'PUT' to push to the Pushgateway.)
+func PushFrom(r Registry, job string, grouping map[string]string, url string) error {
```
brian-brazil: I'd have this as the primary variant, with no default registry convenience function. When using the pushgateway as intended you'll be using a custom registry.

beorn7:
> Where we run into issues here is protobuf, as that shouldn't be a dependency of the instrumentation (registry+collector+metrics+http instrumentation) part of the library.

In the view of both the server and the Go client, the protobuf is _the_ central data model of Prometheus.

> > As just one example, if the ubiquitious prometheus.MustRegister (with more than 2,000 uses on GitHub alone)
> To be honest, we should be replacing most of those uses in favour of the metrics auto-registering by default.

I'm planning something like that for the next iteration. However, as Go has no exceptions, and panicking in a library if not explicitly called out (via `Must…`) is frowned upon in Go. So we cannot really auto-register in the same way as in Python because there is no way to report the error.

> What I've done elsewhere is to have the register method on the registry, but also on the metric for convenience.

My plan is essentially to have a convenience `.MustRegister()` method on the metrics that return the registered metric so that you can write 

```Go
var (
	cpuTemp = prometheus.NewGauge(prometheus.GaugeOpts{
		Name: "cpu_temperature_celsius",
		Help: "Current temperature of the CPU.",
	}).MustRegister()
)
```

But yeah, not really part of this PR.

prometheus/push.go
```diff
@@ -19,47 +19,153 @@
...
-	return r.Push(job, instance, url, method)
+	return push(r, job, grouping, url, method)
+}
+
+// GroupByHostname returns a label map with the only entry
+// {instance="<hostname>"}. This can be conveniently used as the grouping
+// parameter if metrics should be pushed with the hostname as label. The
+// returned map is created upon each call so that the caller is free to add more
+// labels to the map.
+func GroupByHostname() map[string]string {
```
brian-brazil: This is called `instance_ip_grouping_key` and `instanceIPGroupingKey` in other clients, we should be consistent.

prometheus/registry.go
```diff
@@ -11,232 +11,253 @@
...
-	"net/http"
-	"net/url"
 	"os"
 	"sort"
-	"strings"
 	"sync"
 
 	"github.com/golang/protobuf/proto"
-	"github.com/prometheus/common/expfmt"
+	"github.com/hashicorp/go-multierror"
```
brian-brazil: We should be working to remove dependencies from instrumentation, not add more of them - particularly for trivial things like this.

prometheus/registry.go
```diff
@@ -11,232 +11,253 @@
...
+// convenience functions. It has a ProcessCollector and a GoCollector
+// pre-registered.
+var DefaultRegistry = NewRegistry()
 
-	acceptEncodingHeader = "Accept-Encoding"
-	acceptHeader         = "Accept"
-)
+func init() {
+	MustRegister(NewProcessCollector(os.Getpid(), ""))
+	MustRegister(NewGoCollector())
```
brian-brazil: For Python I kept around a public variable with a copy these, so the user could unregister them if they wanted.

prometheus/http.go
beorn7: Possibly, but for now, I want to not break any common use case.
In a later iteration, where we have more central breaking changes, at least one constructor will go (`UninstrumentedHandler`).

brian-brazil: Can we get a deprecation notice on all of these?

With so many top-level functions, this API is going to be very hard for a user to figure out what is and isn't important.

prometheus/doc.go
beorn7: Yeah, this is a problem with Go's desire for lean interfaces and at the same time not being a "proper" OOP language. This is essentially a side effect of the absence of implementation inheritance. The standalone functions contain the code that would otherwise be repeated in every implementation of the `Registry` interface. While this pattern might appear weird, it's very common in the standard library, e.g. `ioutiles.ReadAll(io.Reader)` instead of having a `ReadAll` method in the Reader interface.

WRT removing the Get methods: We can definitely consider that. I had a use case, but if it's rare, it can still be implemented by the few users in the same way as it is done here.

However, I don't want to have any avoidable breaking changes in this PR. So that people can have the public registry without breaking changes. Future iterations will be more aggressive WRT breaking existing usage, and then we can consider removal.

prometheus/registry.go
beorn7: fixed

brian-brazil:
>> Where we run into issues here is protobuf, as that shouldn't be a dependency of the instrumentation (registry+collector+metrics+http instrumentation) part of the library.
> In the view of both the server and the Go client, the protobuf is the central data model of Prometheus.

Basically if I'm exporting to Graphite via a Prometheus client library, why should I be pulling in protobuf?

This is a really important use case for me, I don't want any friction to users using our instrumentation library.

>> To be honest, we should be replacing most of those uses in favour of the metrics auto-registering by default.
> I'm planning something like that for the next iteration. However, as Go has no exceptions, and panicking in a library if not explicitly called out (via Must…) is frowned upon in Go. So we cannot really auto-register in the same way as in Python because there is no way to report the error.

I don't think it'd be inappropriate to fatal when this happens. We should design the library with the presumption that this is the way things are meant to be used.

> My plan is essentially to have a convenience .MustRegister() method on the metrics that return the registered metric so that you can write

That's how Java works, but that's because of how we have to structure the options in Java. 

In Go we don't have that limitation, so I expect that there'd be no such a method. Rather you'd pass a registry in with the rest of the options, and provide a special no-registry value in case that's needed.

prometheus/push.go
beorn7: That's a good point. As we are breaking the Push signature anyway, we can do that.

(It's in fact outright stupid to ever push the default registry as it contains all the process metrics.)

So done. (With the next push... no pun intended.)

prometheus/doc.go
brian-brazil: How many registry implementations are you expecting? We've not seen a need in the other languages, and of all the other instrumentation libraries out there I can think of only one which is even vaguely this general.

I think we can simplify be having Registry as a type rather than an interface.

> However, I don't want to have any avoidable breaking changes in this PR. So that people can have the public registry without breaking changes. Future iterations will be more aggressive WRT breaking existing usage, and then we can consider removal.

I'm expecting us to have to leave a lot of cruft lying around for a while for the purposes of transition.

prometheus/push.go
brian-brazil: The other question is if this should all be inside one class, and the url is provided to the constructor. That's what's done in Java/Python. If we're going to have one big mega-package that sort of structure will be beneficial.

An additional challenge that has not been resolved yet in the other client libraries is how to handle basic auth, client ca certs etc. for pushing so we should think about that.

prometheus/push.go
brian-brazil: This is also the IP in other clients, as this is meant to be replacing the old pgw behaviour of adding in the client IP.

prometheus/registry.go
```diff
@@ -11,232 +11,253 @@
...
+//
+// This is a way to directly inject MetricFamily protobufs managed and owned by
+// the caller. The caller has full responsibility. As no registration of the
+// injected metrics has happened, there was no check at registration-time. If
+// the injection results in inconsistent metrics, the Collect call will return
+// an error. Some problems may even go undetected, like invalid label names in
+// the injected protobufs.
+//
+// The hook function must be callable at any time and concurrently.
+func NewRegistryWithInjectionHook(hook func() []*dto.MetricFamily) Registry {
```
brian-brazil: I'm not sure about this. This would require a custom registry for the node exporter, which doesn't feel right.

prometheus/push.go
beorn7: Yeah, definitely.

Question: Did you intentionally pick the IP number rather than the hostname? That needed to be changed here then, too.

prometheus/registry.go
```diff
@@ -505,22 +470,30 @@ func (r *registry) writePB(encoder expfmt.Encoder) error {
 		sort.Sort(metricSorter(mf.Metric))
 	}
 
-	// Write out MetricFamilies sorted by their name.
+	// Write out MetricFamilies sorted by their name, skipping those without
+	// metrics.
```
brian-brazil: These are exposed in other clients. I'd see it as useful for spotting where you've not exposed any labels for a vector yet.

prometheus/registry.go
beorn7: The problem is that something like multierror should be part of the standard library. Now that it is not, gazillions of implementations are flying around. This ones appeared to me to be the closest to a "canonical one". We could of course copy the implementation (bot note that it is not completely trivial and even pulls in yet another package).

This package would also be a candidate to library-internal vendoring.
(Although that would break people who want to type-convert the error.)

Internal vendoring should be considered for client_colang (but not in this PR).

prometheus/registry.go
beorn7: `Unregister(NewGoCollector())` would work (as the collectors don't have to be the same, equality is enough).

brian-brazil: I wouldn't like to depend on that in general.

This is a very rare use case though.

prometheus/doc.go
```diff
@@ -61,51 +62,135 @@
...
+// Custom Collectors and constant Metrics
+//
+// While you could create your own implementations of Metric, most likely you
+// will only ever implement the Collector interface on your own. At a first
+// glance, a custom Collector seems handy to bundle Metrics for common
+// registration (with the prime example of the different metric vectors above,
+// which bundle all the metrics of the same name but with different labels).
+//
+// There is a more involved use-case, too: If you already have metrics
+// availably, created outside of the Prometheus context, you don't need the
```
brian-brazil: available

beorn7:
> Basically if I'm exporting to Graphite via a Prometheus client library, why should I be pulling in protobuf?

Because the canonical Prometheus data model is destribed in `metrics.proto`.

> This is a really important use case for me, I don't want any friction to users using our instrumentation library.

Writing a consumer of the new `Registry` interface that calls `Collect` and sends the metrics to Graphite will be trivial now. The only friction is that the Go representation of protobuf is (arguably) difficult to work with, but it would only affect the one programmer who write the code exporting to Graphite (i.e. probably myself), and not our users.

The discussion if `metrics.proto` should really be the canonical representation of the Prometheus data model is a legitimate one, but way way out of scope of this PR.

beorn7:
> I don't think it'd be inappropriate to fatal when this happens. We should design the library with the presumption that this is the way things are meant to be used.

We are creating a library that aspires to be used in a massive amount of Go programs. I don't think we can work against fundamental conventions in the Go community while doing so.

> In Go we don't have that limitation, so I expect that there'd be no such a method. Rather you'd pass a registry in with the rest of the options, and provide a special no-registry value in case that's needed.

Yes, that was my initial thought, too, but I couldn't find a way to do this in agreement with Go conventions.

But this is another discussion I deliberately didn't want to have now. Let's have it later when I going for the metrics themselves.

prometheus/doc.go
beorn7: Fixed.

prometheus/registry.go
beorn7: How do you expose a MetricFamily without any metrics?
The expfmt package returns an encoding error in that case. (which is how I noticed the case in the first place.)

prometheus/registry.go
beorn7: That's a valid concern.

But I really don't want the super-special use-case of the injection hook in the interface.

On the other hand, we probably want error logging in the node exporter anyway, with all the possible things that can go wrong during collection. In that case, you need a custom registry anyway to configure the logger.

prometheus/registry.go
brian-brazil:
```
# HELP blah some help
# TYPE blah gauge
```

This is something I think we should be supporting.

prometheus/registry.go
brian-brazil: We could adjust the node exporter not to use it, and then remove the feature.

prometheus/push.go
beorn7: The reason for attaching the IP number was that we had no other information on the Pushgateway.

Picking the IP-number on the side of the client is a bit hard. It's not trivial through with IP-number the push will go out. And IP-numbers might get translated, or they might be long IPv6 numbers...

I'd actually favor the hostname here. (It's a convenience function anyway, and might even not be used a lot.)

brian-brazil:
> The discussion if metrics.proto should really be the canonical representation of the Prometheus data model is a legitimate one, but way way out of scope of this PR.

The question here is what dependencies the instrumentation library should have, which is a different question to how we mentally think about Prometheus data.

We need to accept that some users have aversions to protobuf (in the same way you or I would about XML), and aversions to seemingly unnecessary dependencies as that's a sign of a bloated library. As you say "We are creating a library that aspires to be used in a massive amount of Go programs."

prometheus/doc.go
```diff
@@ -61,51 +62,135 @@
...
+// interface of the various Metric types. You essentially want to mirror the
+// existing numbers into Prometheus Metrics during collection. An own
+// implementation of the Collector interface is perfect for that. You can create
+// Metric instances “on the fly” using NewConstMetric, NewConstHistogram, and
+// NewConstSummary (and their respective Must… versions). That will happen in
+// the Collect method, while the Describe method has to return separate Desc
+// instances, the very same you later use to create the constant “throw-away”
```
brian-brazil: This is a bit hard to parse

prometheus/doc.go
```diff
@@ -61,51 +62,135 @@
...
+// collect time. The former will usually be detected at start-up time of a
+// program, while the latter will only happen at scrape time, possibly not even
+// on the first scrape if the inconsistency only becomes relevant later. That is
+// the main reason why a Collector and a Metric have to describe themselves to
+// the registry.
+//
+// A particular error case to handle is if a Collector has already be registered
+// (or a different Collector that collects the same metrics). Registration fails
+// in that case, but you can detect from the kind of error what has happened. A
+// common pattern is that you want to register the new Collector if an equal
```
brian-brazil: I don't think that's a common pattern

prometheus/doc.go
beorn7: Having an interface is the "fail safe" approach. Because this library is used quite widely, I assume a lot of requirements. I have observed the recent tendency to go for interfaces in doubt, and may people complaining about std library types that were not done as interfaces.

The normal use case is anyway to use the default registry. So only power user are even concerned with the registry as an own type. So I expect a higher percentage of "weird" use cases there, which we cannot anticipate right now. In any case, we don't have to optimize the registry itself for the naive user, as the latter will use the top-level functions.

prometheus/http.go
beorn7: I doubt that `Handler` will ever go away. That's just too convenient. A deprecation notice about the handler instrumentation is already in place. I just cannot put a DEPRECATED marker on any of Handler/HandlerFor/UninstrumentedHandler right now, because people _should_ use UnistrumentedHandler right now, but then switch to Handler once it is not instrumented anymore (or instrumented in a saner fashion).

Let's solve that later.

I think the number of top-level identifiers can only be reduced by re-packaging. We can consider that later, see commit description.

prometheus/push.go
brian-brazil: I'd expect IPs to be more unique than hostnames, as it's not a FQDN.

I think we should aim for consistency here.

prometheus/push.go
beorn7: Let's solve the auth and cert question later. Once we know how that's done, we can consider wrapping things into a class. For now, I see little value in that.

prometheus/registry.go
beorn7: Yes, agreed that it's rare. I would tend to tell people to use a custom registry in that case.

prometheus/registry.go
```diff
@@ -11,232 +11,253 @@
...
-func Register(m Collector) error {
-	_, err := defRegistry.Register(m)
-	return err
+// Usually, a Registry will be happy as long as the union of all collected
+// Metrics is consistent and valid even if some metrics are not consistent with
+// their own Desc or with one of the Descs provided by their
+// Collector. Well-behaved Collectors and Metrics will only provide consistent
+// Descs. This Registry is useful to test the implementation of Collectors and
+// Metrics.
+func NewPedanticRegistry() Registry {
```
brian-brazil: If we're adding this can we remove the check that a collector provides at least one Desc?

beorn7:
To be clear: The question if protobuf should disappear from client_golang is not on the table right now. That discussion has to happen outside of this PR (and in fact outside of the refurbishment I'm planning).

prometheus/registry.go
beorn7: This is not really adding, it's the replacement for the boolean swich for "collect checks".

The handling and requirements of Descs will be dealt with in future iterations.

prometheus/doc.go
brian-brazil: My presumption would be that anything custom could be done without needing an interface. You could accept collectors and then be a standard collector yourself, and go out via a normal Registry to the exposition formats.

I don't at all like all these top-level functions, it feels very disjointed. A user should never have to touch the top-level functions, only instrumentation.

prometheus/doc.go
beorn7: "Common" might indeed be exaggerated... softened that a bit.

prometheus/push.go
brian-brazil: I see value in that we'd be reducing the number of top-level functions, and being consistent with other client libraries.

prometheus/doc.go
beorn7: Tried to reword.

(Note that this part will definitely be changed in later iterations. No need to touch Desc directly for const metrics.)

prometheus/registry.go
beorn7: I see. Right now, that would create an error in encoding (and possibly decoding - not sure how the text parser deals with that). This needs a more basic approach. First the expfmr package has to be changed, and then it has to propagate into all Go binaries (server, PGW) before we can allow it in the client.

prometheus/registry.go
beorn7: Pushgateway uses it, too.

Conclusion for now: NodeExporter will need to use a custom registry anyway (for logging).
So I'll leave it here as it is. Fine?

prometheus/registry.go
brian-brazil: The text parser deals with it correctly, if not we'd certainly have heard by now.

prometheus/push.go
beorn7: How have you picked the IP number in the other clients? The first one that doesn't look like localhost?

prometheus/registry.go
brian-brazil: We should also remove it from the pgw then, and use a collector instead. Given how the node exporter already works it doesn't require a custom registry, just a custom collector.

Needing `2^n` factory functions depending on options doesn't seem like a good idea. We've already two options.

prometheus/push.go
brian-brazil: Java has methods for it. For Python we bind to a socket and see what the IP is (which turns out not to be terribly portable).

prometheus/http.go
```diff
@@ -15,14 +15,169 @@ package prometheus
...
+	// Panic upon the first error encountered (useful for "crash only" apps).
+	PanicOnError
 )
 
+// HandlerOpts specifies options how to serve metrics via an http.Handler. The
+// zero value of HandlerOpts is a reasonable default.
+type HandlerOpts struct {
+	// ErrorLog specifies an optional logger for errors collecting and
+	// serving metrics. If nil, errors are not logged at all.
+	ErrorLog *log.Logger
```
fabxc: This should take an interface.

beorn7: Yes, it should.

Sadly, the standard library only offers the `log.Logger` type, which is a `struct`. I'd love to use an interface like https://github.com/go-kit/kit/blob/master/log/log.go#L10 but I don't think we can coerce users into using a particular logging library. By using the standard library `log.Logger`, we allow everybody to use it with any library that provides an adapter, e.g. https://github.com/go-kit/kit/blob/master/log/stdlib.go#L54 .

Full example:

```Go
package main

import (
	stdlog "log"
	"os"

	"github.com/go-kit/kit/log"
)

func main() {
	logger := log.NewLogfmtLogger(os.Stdout)

	handler := prometheus.HandlerFor(prometheus.DefaultRegistry, prometheus.HandlerOpts{
		ErrorLog: stdlog.New(log.NewStdlibAdapter(logger), "", 0),
	})
}
```

beorn7: This is also in line with https://golang.org/pkg/net/http/#Server , i.e. it's the devil people know.

fabxc: Known problem that stdlib should've made it an interface. I think so far a good practice was just adding your own interface with the minimal methods being used, i.e. `type Logger interface { Println(...interface{}) }`.

Allows passing in almost all loggers without relying on go-kit (but can still be used).

prometheus/doc.go
beorn7: I found ~30 uses on GitHub. I didn't evaluate every single case for how legitimate they are. My own legitimate use case is in `InstrumentHandler` (which currently allows to instrument different handlers with the same metric).

In the spirit of reducing top-level identifiers, I'll remove the `...OrGet` for custom registries, see other comment thread.

prometheus/registry.go
beorn7: OK, did the research.

The text parser protects against empty metric families here: https://github.com/prometheus/common/blob/master/expfmt/text_parse.go#L105-L110

So this is an issue beyond `client_golang`. I have filed https://github.com/prometheus/common/issues/50 for it.

prometheus/registry.go
beorn7: In this particular case, there is no 2^n problem, as (incidentally) the two special cases here are effectively mutually exclusive (the pedantic checks are useless if you set an injection hook, which has no descriptor and cannot be checked).

Having said that, I thought a while about this. Conclusions:

- Both, the PGW and the Node Exporter are legitimate uses of the injection hook. In both cases, you have `MetricFamily` protobuf objects already (for very different reasons). `Collector` collects `Metric` protobuf, so to emulate the injection hook in terms of a `Collector`, you needed to break the `MetricFamily` up, reverse-engineer `Desc` for them (which is where the registry gets the metric name and help from). Avoiding this tedious procedure was the reason (back then) for creating the injection hook in the first place. Thus, entirely removing the injection hook is not an option, even if the use case is rare.
- My argument above about "we need a custom registry for the Node Exporter" is actually invalid. Wisely, I have moved the logging concern into the handler, so all we need is a custom handler. Which makes the requirement to set an injection hook in the default registry a valid one again.
- One of my incentives for removing the setter was that it was never meant to be called after the first scrape. So the form was not in line with the semantics. But it is easy to bring the semantics in line with the form by allowing later setting of the injection hook (one mutex acquisition should do the trick).

So this is what I will do now:
- Reinstate ` SetMetricFamilyInjectionHook`, but in a concurrency-safe form, callable at any time. Big plus: One fewer breaking change.
- Make it a first-class citizen in the `Registry` interface and remove the special constructor. This makes the interface fatter, but the conclusions above show that it's essentially another type of registration, which is a core functionality of the registry.

prometheus/push.go
beorn7: Hmm, so the Go stdlibrary allows to list all network interfaces. Picking one seems really adventurous to me. I dislike the IP number idea more and more. The hostname may be pretty arbitrary, too, but at least it's well defined. I like that more than the IP number. But I'm tempted to remove this convenience function completely (it is rarely useful anyway).

Does anybody else have an opinion on this?

prometheus/push.go
beorn7: In a not truly OO language like Go, moving top level functions into "classes" is not really helping. It might sugar-coat the problem a bit, but that's arguably worse than leaving it in the open.

prometheus/http.go
beorn7: I'll explain the situation with the handlers in the doc comment a bit more.

prometheus/doc.go
beorn7: I don't like the number of top-level identifiers, either.
However, bundling them up in classes in a language that doesn't even have proper classes is not really improving anything.

A real solution would be a meaningful re-packaging and of course avoiding many top-level identifiers in the first place. As re-packaging is only an option later (at the price of breaking the world), let's apply extra scrutiny to reducing the number of identifiers.

In that spirit, I'll declare the old `...OrGet` top level function as deprecated, and will not have a utility function for custom registries anymore (will be done in the next push).

Making the exported Registry a struct rather than an interface is a different question. I'll think about it... Other opinions welcome.

prometheus/http.go
beorn7: And BTW, @brian-brazil , this is an example, where historically, a `struct` was chosen over an `interface`, and now everybody is sad about it.

prometheus/push.go
brian-brazil: It helps if we can encapsulate and organise related behaviour. As it stands all these top-level functions are going to make the API very unwieldy.

prometheus/registry.go
brian-brazil: Two levels of dependencies just to get a list of errors? That seems a bit much.

(And this is why dependencies are a problem, particularly in what is intended to be a very core library).

prometheus/registry.go
brian-brazil: > Avoiding this tedious procedure was the reason (back then) for creating the injection hook in the first place. Thus, entirely removing the injection hook is not an option, even if the use case is rare.

Could we not write a collector that'll take the protos and convert it into const metrics?

> This makes the interface fatter, but the conclusions above show that it's essentially another type of registration, which is a core functionality of the registry.

I don't see why it's fundamentally another type of registration, it's just a different type of object we're passing around. We should try and consolidate this and keep the overall API small.

prometheus/http.go
beorn7: @fabxc Since we are only using `Println`, that's indeed doable. I'll rework in the next push. (There is the potential for a future problem should we decide to use more methods of the logger. But that's unlikely.)

prometheus/registry.go
beorn7: Since we have a number of dependencies anyway, I see the cost of adding two more as the lesser evil compared to implementing the 47th variant of a multi error. (Plus, this is one that could be vendored internally, once we go for it.)

But I'm just repeating myself, and so are you.

Do others have additional arguments here or an opinion?

prometheus/registry.go
beorn7: > Could we not write a collector that'll take the protos and convert it into const metrics?

Yes, we could. But that would require exactly the tedious procedure I have described above. And that exact procedure was the reason for creating the injection hook, as said already. There was an intense discussion back then with many stakeholders, and I don't see any changed circumstances that would require us to revisit the decision.

> I don't see why it's fundamentally another type of registration, it's just a different type of object we're passing around. We should try and consolidate this and keep the overall API small.

OK, then call it "a different type of object" instead of "fundamentally another type of registration". It still requires a different method. I tried to keep it out of the default path, but as you have brought up yourself, it would require a custom registry for the node exporter.

I don't see a way to consolidate metric family injection with metric collection without causing exactly the type of hassle that we removed by introducing the injection hook.

brian-brazil: > I don't see any changed circumstances that would require us to revisit the decision.

We're creating a new API, it's the time to look any any ways we can simplify that interface. Two  use cases that could be implemented in another fashion doesn't justify a public API to me.

> I tried to keep it out of the default path, but as you have brought up yourself, it would require a custom registry for the node exporter.

I never stated it must require a custom registry, merely that the current proposed implementation forces it. I want to continue using use the default registry in the node exporter.

prometheus/push.go
beorn7: > It helps if we can encapsulate and organise related behaviour. As it stands all these top-level functions are going to make the API very unwieldy.

You said that before. I have understood you. As said before, I completely agree that many top-level identifiers are unwieldy. There is no point in repeating that again and again. However, as I have also said before, I also think that a `struct` in Go is not meant to organize related behavior, and doing so doesn't solve the problem. Just repeating what you said before won't change my opinion here. It merely triggers me to repeat what I said, too.

To add something of new value, hopefully:

In most cases, users of a PGW push exactly once (at the end of their batch job). With the current approach, that's one call. With a struct wrapping, that becomes a two step process: Create an instance of the struct, then call a method. In the code, we would replace four top-level identifiers by one top-level identifier (the struct) with four methods, which results in five entries in the Godoc. Even before, all four functions were next to each other in the Godoc and quite obviously related by naming. I'd argue the situation is now worse. One more thing to mentally parse, and one more step in pushing.

Other suggestion: Since I have removed the global push function for the default registry, the push stuff is, for a change, a part we can move into its own package right now. How about doing that? There is a bit of friction, because it would be the first package, but it's a first step, and then pushing will stay a one-step procedure, and it will be really idiomatic:

```Go
import ".../prometheus/push"

push.Registry(reg, "myjob", nil, url)
push.Collectors("myjob", nil, url, c1, c2)
```

Perhaps move the registry argument to the end to mirror the collector location? (The collectors have to be at the end to make them variadic.)

prometheus/registry.go
beorn7: On 2 August 2016 at 14:28, Brian Brazil <notifications@github.com> wrote:
>
> We're creating a new API, it's the time to look any any ways we can simplify that interface. Two use cases that could be implemented in another fashion doesn't justify a public API to me.

Yes, that mirrors an argument we had in the discussion back in early
2013. As said, I thought about changes in the circumstances, but I
couldn't detect any relevant changes, so I don't see a different
outcome of the discussion.

> > I tried to keep it out of the default path, but as you have brought up yourself, it would require a custom registry for the node exporter.
>
> I never stated it must require a custom registry, merely that the current proposed implementation forces it. I want to continue using use the default registry in the node exporter.

Yes, that's what I meant. By using the default registry in the node
exporter, we need a way to inject metric families with the default
registry. Since the conclusion after a discussion with many
stakeholders was that we need an entry point for that separate from
regular registration and collection, a first-class support in the
interface has to be the way until somebody comes up with a better idea
that changes the odds.

prometheus/http.go
brian-brazil: > I think the number of top-level identifiers can only be reduced by re-packaging.

We can reduce the number of top-level identifiers by putting things in classes.

The reason to split things out to other packages is to avoid dependencies in the core instrumentation, not to reduce top-level identifiers.

I'd overall expect that the Go client library would have roughly the same user-facing API as the Java client. The only top-level functions I'd expect in Go would be factories (Java has only one top-level function, for reasons that don't apply to Go).

beorn7: On 2 August 2016 at 14:50, Brian Brazil <notifications@github.com> wrote:
>
> We can reduce the number of top-level identifiers by putting things in classes.

That's a purely formal argument. Sticking functions to structs is
sugarcoating and not reducing the mental load. But I said that again.

> The reason to split things out to other packages is to avoid dependencies in the core instrumentation, not to reduce top-level identifiers.

Packages are the Go way of grouping related functions.

> I'd overall expect that the Go client library would have roughly the same user-facing API as the Java client. The only top-level functions I'd expect in Go would be factories (Java has only one top-level function, for reasons that don't apply to Go).

I don't think it's a good idea to coerce the Go API into that of a
very different language.

-- 
Björn Rabenstein, Engineer
http://soundcloud.com/brabenstein

SoundCloud Ltd. | Rheinsberger Str. 76/77, 10115 Berlin, Germany
Managing Director: Alexander Ljung | Incorporated in England & Wales
with Company No. 6343600 | Local Branch Office | AG Charlottenburg  |
HRB 110657B

prometheus/registry.go
brian-brazil: That discussion was 3 years ago, lots has happened since then.

The Collect interface is already defined in terms of protos via Metric, converting doesn't seem like it should be a big deal.

prometheus/push.go
brian-brazil: > However, as I have also said before, I also think that a struct in Go is not meant to organize related behavior, and doing so doesn't solve the problem.

I believe we should have these operations on particular on objects as methods, rather than as top-level functions.

> With the current approach, that's one call. With a struct wrapping, that becomes a two step process: Create an instance of the struct, then call a method.

With custom registries it becomes two calls anyway, as you need to create the new registry whereas previously it used the default.

> How about doing that? 

We could do that. What would the plan be for handling auth?

prometheus/registry.go
beorn7: On 2 August 2016 at 14:56, Brian Brazil <notifications@github.com> wrote:
>
> That discussion was 3 years ago, lots has happened since then.
>
> The Collect interface is already defined in terms of protos via Metric, converting doesn't seem like it should be a big deal.

To have a meaningful discussion, you have to trust me. I'm not trying
to deceive you.

I considered exactly what you said. While the code back then (as
sketched out by Bernerd) is not line-by-line exactly the same as it
would be today, it would be very similar in essence. We would still
convert a good part of a protobuf object into a different
representation, stuff it into (nowadays) a ConstMetric of the
appropriate type, just to convert everything back then.

-- 
Björn Rabenstein, Engineer
http://soundcloud.com/brabenstein

SoundCloud Ltd. | Rheinsberger Str. 76/77, 10115 Berlin, Germany
Managing Director: Alexander Ljung | Incorporated in England & Wales
with Company No. 6343600 | Local Branch Office | AG Charlottenburg  |
HRB 110657B

prometheus/push.go
beorn7: On 2 August 2016 at 15:04, Brian Brazil <notifications@github.com> wrote:
>
> I believe we should have these operations on particular on objects as methods, rather than as top-level functions.

Yes. I got that. Really!

I still disagree, though.

> > With the current approach, that's one call. With a struct wrapping, that becomes a two step process: Create an instance of the struct, then call a method.
>
> With custom registries it becomes two calls anyway, as you need to create the new registry whereas previously it used the default.

Most people will use the `PushCollectors` form. That would be two
instead of one calls.

With a custom registry, it would be three instead of two calls. The
instantiation of a "class" (which doesn't even exist in Go) would be
an additional step in any case.

> We could do that. What would the plan be for handling auth?

Handling auth is out of scope of this PR. (I'm not saying it's not
important. And breaking the usage only once is great, but I'm afraid
this will drag on forever if we load more changes into it.) In any
case, moving push related stuff into an own packages is at least
containing future complexity in that package.

-- 
Björn Rabenstein, Engineer
http://soundcloud.com/brabenstein

SoundCloud Ltd. | Rheinsberger Str. 76/77, 10115 Berlin, Germany
Managing Director: Alexander Ljung | Incorporated in England & Wales
with Company No. 6343600 | Local Branch Office | AG Charlottenburg  |
HRB 110657B

prometheus/http.go
brian-brazil: > That's a purely formal argument. Sticking functions to structs is
sugarcoating and not reducing the mental load. But I said that again.

It is reducing the mental load, as when e.g. I'm looking for functions related to the Registry they're right there as methods rather than spread around the package.

> Packages are the Go way of grouping related functions.

Classes are a Go way to group functions related to a specific type.

> I don't think it's a good idea to coerce the Go API into that of a very different language.

Considering the languages we have clients in, it's actually pretty similar (you could make all the above arguments about Java for example). The only real major difference API-wise is that functions can only have one signature. I'd also expect Ruby and Python to look roughly the same as each other.

prometheus/registry.go
brian-brazil: > We would still convert a good part of a protobuf object into a different
representation, stuff it into (nowadays) a ConstMetric of the
appropriate type, just to convert everything back then.

You can guess what my opinion on how we resolve that is :)

Given the current code, it's not that bad. You need to create a Desc from the MetricFamily, and can then send each Metric down the channel.

prometheus/push.go
brian-brazil: > Most people will use the `PushCollectors` form.

I'm sure on that. I half wonder sometimes if we should be keeping it around, as it makes PushAdd usage where you want to change different metrics based on success/fail more difficult as you've to enumerate all collectors twice rather than toggling the one or two who are special.

prometheus/http.go
beorn7: > It is reducing the mental load, as when e.g. I'm looking for functions related to the Registry they're right there as methods rather than spread around the package.

Yes, everything else put aside, grouping can be helpful in that way. But it's a secondary concern. I'd prefer packages over that, where a package has a small enough surface to not run into the issue in the first place. I pack things in structs primarily not for grouping but for other reasons, because it helps the code using the functions, or - as a specific Go concern - because I want to implement interfaces.

To not devolve that into a fundamental discussion about software design: We are discussing two concrete cases here, where the above matters: The push related functions and the registry related functions. My point for the push related functions is that having a struct complicates pushing by 33 to 50%. That has priority over grouping the functions. Also, we can use a package for that, in that particular case. For the registry, the grouping matters more because of the many legacy functions we have. I hope we'll be able to cut this down, currently working on it with scrutiny. The core question with the registry is if we want a registry `interface` or a registry `struct`. This is an open question (and yes, I know you believe it should be a `struct`. Still I have to think about it. Opinions of others are welcome, BTW, to speed up my thought process). If it's an interface, there is no way we would load it up with methods. If it is a struct, our options have to be weighted against each other. But let's first get the decision about the interface done. Otherwise, our endless discussion is merely academic (not a bad thing per se, but we have to get forward with this PR). 

> Classes are a Go way to group functions related to a specific type.

Go doesn't even have classes. (OK, it's a nit, but it speaks volumes about the Go mindset.)

Also, as a data point, note how https://golang.org/pkg/strings/ is a package, while the equivalent are all string methods in Java.

In different news, I see it as the perfect way to unidiomatic Go code to treat Go as similar to Java. But yeah, let's not have a fundamental programming paradigm discussion here. Let's just decide about the issues at hand. See above.

prometheus/registry.go
beorn7: On 2 August 2016 at 15:29, Brian Brazil <notifications@github.com> wrote:

> Given the current code, it's not that bad. You need to create a Desc from
> the MetricFamily, and can then send each Metric down the channel.


Yes, I'm a decent enough programmer to know how I would implement that. It
was considered in my decision.

-- 
Björn Rabenstein, Engineer
http://soundcloud.com/brabenstein

SoundCloud Ltd. | Rheinsberger Str. 76/77, 10115 Berlin, Germany
Managing Director: Alexander Ljung | Incorporated in England & Wales with
Company No. 6343600 | Local Branch Office | AG Charlottenburg  | HRB 110657B

prometheus/push.go
brian-brazil: Consider the case where you've a last_success and duration metrics. Depending on whether the batch job succeeds you'll either want to report one or two of those metrics.

As you add more metrics, that gets messier.

beorn7: OK, but people can still use a custom registry if they have many metrics,
but the frequent case where you push one or two Collectors is still covered
by `PushCollectors`, no? I don't see the win of removing the latter. It
would make things more complicated for most cases.

beorn7:
Sooo, several things updated:

- Setting the injection hook is now a 1st class method again. The discussion convinced me that it is required.
- The `...OrGet` type of functions are now deprecated, and no versions for a custom registry are provided.
- I added a bit of notes about the messy deprecation situation with `Handler` vs. `UninstrumentedHandler`.
- I implemented @fabxc suggestion about the logger interface.
- I moved the push related stuff into its own package (which I really like).

That leaves us with really only one function that could be a method on the Registry interface (or struct, see below): `MustRegister`. In the struct situation, I would totally add it as a method, no controversy about that. Now that it's only one method left, we might even add it to the interface to avoid the one oddball. I'm still reluctant to make the interface a struct as this rules out things like creating test mocks or really plain implementation of your own registry, even if we don't see that as likely. Still thinking about that one. If more people than just Brian want to chime in, please do so.

A number of other top-level identifiers will disappear with future iterations (like `NewDesc`, `BuildFQName`, ...). If you also mentally remove the now deprecated functions and then look at the Godoc, it doesn't look that bad anymore.

Which leaves us with three remaining open questions (if I haven't missed anything):
1. Registry interface vs. struct, see above
2. The hostname (or IP) convenience grouping key for pushing (I renamed it `HostnameGroupingKey` for now, but we have to decide if we (a) leave it like it is now (b) remove it entirely or (c) try to guess a meaningful IP number like the other client libraries. (BTW: This is my order of preference, where (a) and (b) are not much different, but I'd really hate (c) as it seems very fragile.)
3. The multi-error question: Should we (a) keep it as now (pulling in two hashicorp packages and potentially internally vendor them later) (b) copying over the code (which is poor-man's vendoring) or (c) implement our own solution. (This is again my preference, but here I tend quite strongly towards (a) and would really not like (b) and (c).

Since Brian and I have apparently exchanged all our arguments already, it would be good if others could chime in.

Note that I have changed the commit description according to all of this. I'll update the PR description manually soon. The breaking changes are really quite minimal by now.

beorn7:
So this would fix #216, too. :)

beorn7:
Another idea about interface vs. struct, which might resolve all issues around it:

The important part about the registry interface is not that we only expose the interface, the important part is that we only use the interface when we _use_ the registry. In that way, people can still inject own implementations, mock it for testing etc. Then we can totally expose the implementation of the registry as a struct.

For that, it's actually advantageous to have a separate interface for the registration side (`Register`, `Unregister`) and for the collection site (`Collect`) because you will usually only want one of these.

I'll add a separate commit implementing my thoughts.

beorn7:
Done. Please have a look.
This would leave us with open questions number 2 and 3, which are hopefully easy to resolve.

prometheus/push.go
grobie: Yes, and I strongly prefer hostname over IP.

With mutliple NICs, bonding, docker bridges, etc. there is an endless possibility to pick the wrong IP. On the other side, the hostname has a clear definition and retrieval interface. I've had to write code to retrieve the internal network IP of all our nodes before and it required many exceptions and knowledge about our network to not pick a wrong one. I've seen the same issues with other software trying to auto-pick the right IP (e.g. Consul).

prometheus/collector.go
```diff
@@ -37,39 +37,39 @@ type Collector interface {
...
+	// in their variable label values. This method may be called
+	// concurrently and must therefore be implemented in a concurrency safe
+	// way. Blocking occurs at the expense of total performance of rendering
+	// all registered metrics. Ideally, Collector implementations support
+	// concurrent readers.
 	Collect(chan<- Metric)
 }
 
-// SelfCollector implements Collector for a single Metric so that that the
+// selfCollector implements Collector for a single Metric so that that the
```
grobie: s/that that/that/

prometheus/collector.go
```diff
@@ -37,39 +37,39 @@ type Collector interface {
...
 // Metric collects itself. Add it as an anonymous field to a struct that
 // implements Metric, and call Init with the Metric itself as an argument.
```
grobie: Not sure how relevant that last sentence still is, but s/Init/init/.

prometheus/counter.go
```diff
@@ -35,6 +35,9 @@ type Counter interface {
 	// Prometheus metric. Do not use it for regular handling of a
 	// Prometheus counter (as it can be used to break the contract of
 	// monotonically increasing values).
+	// This method is DEPRECATED. Use NewConstMetric to create a counter for
```
grobie: > To signal that an identifier should not be used, add a paragraph to its doc comment that begins with "Deprecated:" followed by some information about the deprecation.

https://blog.golang.org/godoc-documenting-go-code

grobie: Some background on that convention https://groups.google.com/forum/#!topic/golang-dev/ya5vE0G62Z4

prometheus/example_clustermanager_test.go
```diff
@@ -55,76 +50,69 @@ func (c *ClusterManager) ReallyExpensiveAssessmentOfTheSystemState() (
...
 		Zone: zone,
-		OOMCount: prometheus.NewCounterVec(
-			prometheus.CounterOpts{
-				Subsystem:   "clustermanager",
-				Name:        "oom_count",
-				Help:        "number of OOM crashes",
-				ConstLabels: prometheus.Labels{"zone": zone},
-			},
+		OOMCountDesc: prometheus.NewDesc(
+			"clustermanager_oom_count",
```
grobie: By our convention, this should use _total instead of _count.

prometheus/example_clustermanager_test.go
```diff
@@ -55,76 +50,69 @@ func (c *ClusterManager) ReallyExpensiveAssessmentOfTheSystemState() (
...
+		RAMUsageDesc: prometheus.NewDesc(
+			"clustermanager_ram_usage_bytes",
+			"RAM usage as reported to the cluster manager.",
 			[]string{"host"},
+			prometheus.Labels{"zone": zone},
 		),
 	}
 }
 
 func ExampleCollector_clustermanager() {
```
grobie: While you're on it, any reason to use underscores in function names?

prometheus/doc.go
```diff
@@ -61,51 +62,115 @@
...
+// Metrics
+//
+// The number of exported identifiers in this package might appear a bit
+// overwhelming. Hovever, in addition to the basic plumbing shown in the example
+// above, you only need to understand the different metric types and their
+// vector versions for basic usage.
+//
+// Above, you have already touched the Counter and the Gauge. There are two more
+// advanced metric types: the Summary and Histogram. A more thorough description
+// of those four metric types can be found in the prometheus docs:
```
juliusv: s/prometheus/Prometheus/

prometheus/doc.go
```diff
@@ -61,51 +62,115 @@
...
-// NewConstMetric and MustNewConstMetric to create "throw away" Metrics at
-// collection time, MetricVec to bundle custom Metrics into a metric vector
-// Collector, SelfCollector to make a custom Metric collect itself.
-//
-// A good example for a custom Collector is the ExpVarCollector included in this
-// package, which exports variables exported via the "expvar" package as
-// Prometheus metrics.
+// HistogramVec, and UntypedVec.
+//
+// While only the fudamental metric types implement the Metric interface, both
```
juliusv: fudamental -> fundamental

prometheus/doc.go
```diff
@@ -61,51 +62,115 @@
...
+// registration (with the prime example of the different metric vectors above,
+// which bundle all the metrics of the same name but with different labels).
+//
+// There is a more involved use-case, too: If you already have metrics
+// available, created outside of the Prometheus context, you don't need the
+// interface of the various Metric types. You essentially want to mirror the
+// existing numbers into Prometheus Metrics during collection. An own
+// implementation of the Collector interface is perfect for that. You can create
+// Metric instances “on the fly” using NewConstMetric, NewConstHistogram, and
+// NewConstSummary (and their respective Must… versions). That will happen in
```
juliusv: Here and above, you use "&" as a wildcard... is that a common thing? Wouldn't "*" make more sense?

prometheus/doc.go
```diff
@@ -61,51 +62,115 @@
...
+// the Collect method. The Describe method has to return separate Desc
+// instances, representative of the “throw-away” metrics to be created
+// later. NewDesc comes in handy to create those Desc instances.
+//
+// The Collector example illustrates the use-case. You can also look at the
+// source code of the processCollector (mirroring process metrics), the
+// goCollector (mirroring Go metrics), or the exvarCollector (mirroring expvar
```
juliusv: exvar->expvar

prometheus/doc.go
```diff
@@ -61,51 +62,115 @@
...
+// Also note that the DefaultRegistry comes registered with a Collector for Go
+// runtime metrics (via NewGoCollector) and a Collector for process metrics (via
+// NewProcessCollector). With a custom registry, you are in control and decide
+// yourself about the Collectors to register.
+//
+// HTTP Exposition
+//
+// The Handler function used so far to get an http.Handler for serving the
+// metrics is also acting on the DefaultRegistry. With HandlerFor, you can
+// create a handler for a custom registry or anything the implements the
```
juliusv: anything the -> anything that

prometheus/doc.go
```diff
@@ -61,51 +62,115 @@
...
+// Deliverer interface. It also allows to create handler that act differently on
```
juliusv: handler -> handlers

prometheus/example_clustermanager_test.go
```diff
@@ -55,76 +50,69 @@ func (c *ClusterManager) ReallyExpensiveAssessmentOfTheSystemState() (
...
-// NewClusterManager creates the two metric vectors OOMCount and RAMUsage. Note
+// NewClusterManager creates the two Descs OOMCountDesc and RAMUsageDesc. Note
 // that the zone is set as a ConstLabel. (It's different in each instance of the
-// ClusterManager, but constant over the lifetime of an instance.) The reported
-// values are partitioned by host, which is therefore a variable label.
+// ClusterManager, but constant over the lifetime of an instance.) Then there is
+// a variable label "host", since we want to partition the collected metrics by
+// host. Since all Descs created in this way are consistent across instances,
+// with a guaranteed distinction by the "zone" label, we can register different
+// ClusterManager with the same registry.
```
juliusv: "ClusterManager" -> "ClusterManagers" | "ClusterManager" instances?

prometheus/push/examples_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/push"
+)
+
+func ExampleCollectors() {
+	completionTime := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "db_backup_last_completion_time",
```
grobie: _seconds

grobie:
Cool! Thanks a lot @beorn7. This looks great and I like the direction of splitting the client up into multiple sub-packages. I found name of the `push.Registry()` function almost a bit too clever, I guess it can be confusing in godocs at first. It should be cool in code were it actually matters though. Looking forward to a test registry (or do you want others to contribute that?).

To answer your questions:
1. The interface is cool.
2. I prefer a) over b), I don't like c) as this will cause too many issues.
3. I don't have a strong opinion here. I wouldn't have seen an issue with a reasonable dependency from a trustworthy source, but I also remember PRs were people objected to including a library because of too many dependencies.

prometheus/push/examples_test.go
beorn7: Fixed.

prometheus/example_clustermanager_test.go
beorn7: Done.

prometheus/doc.go
beorn7: Done.

prometheus/doc.go
beorn7: Done.

prometheus/doc.go
beorn7: Done.

prometheus/doc.go
beorn7: Display issue? As the typography nerd I am, I use the …-character, http://www.fileformat.info/info/unicode/char/2026/index.htm
Display nicely in my editor and in the browser displaying godoc and here in GH.

prometheus/doc.go
beorn7: Done.

prometheus/doc.go
beorn7: Done.

prometheus/example_clustermanager_test.go
beorn7: https://godoc.org/github.com/fluhus/godoc-tricks#Examples

beorn7: But actually, now that I have removed the other collector example, I don't need the suffix anymore.
Removed.

prometheus/example_clustermanager_test.go
beorn7: Totally. very outdated example...

prometheus/counter.go
beorn7: Thanks for the pointer. I'll fix it where it is "real" deprecation. Situation with Handler and friends is special, but I'll have another trick up my sleeves there in the next commit. Stay tuned…

prometheus/collector.go
beorn7: Done.

prometheus/collector.go
beorn7: Done.

beorn7:
Addressed the comments.

Also some push tweaks.

Two more commits in the pipeline. Stay tuned.

beorn7:
@grobie I have added an example of `push.Registry` to make the cleverness of the naming more obvious.

This PR opens the way for others to implement testing tooling (I suggest a `promtest` sub-package for it). A test registry (or test Registerer) could be part of it, but I would more expect people to simply call `Deliver()` on the normal registry and then have some tooling for verifying the content (e.g. use the text parser to compare it to a text representation of your expected scrape).

beorn7:
OK, the latest commit removes more items from the prometheus top-level namespace. And paves the road for removing more. (Many functions now deprecated that will be removed in a breaking change later.)

I think this settles the concerns about namespace pollution as far as it can be done in this PR. Interface vs. struct is also settled (by having the best of both worlds). So far, the votes in terms of the push helper are in favor of `HostnameGroupingKey`. Leaves us with the question about multi-error.

For the latter, I'd still argue that replicating the code is bad, and that avoiding the dependency only has value if it were the only dependency we had. Since we have a number of external dependencies already, I doubt we will lose a single user by going the current way. I filed #217 for the internal vendoring discussed a number of times.

prometheus/example_clustermanager_test.go
grobie: Ooh, TIL. Thanks.

beorn7:
Talked to St. Peter (@peterbourgon) to calm my mind. The enlightened version of my utterings about "tendency towards interfaces" is: "Return concrete objects, accept interfaces." And that's exactly what I did in yesterday's commit. Good!

Also, I learned that the resistance against global state is even stronger than I imagined. So really good that we finally allow to work without global state for those who desire so.

beorn7:
Minor doc improvement in the latest commit.

prometheus/doc.go
juliusv: Strange:

char

It's a `&` in the code, but looks like an ellipsis in your comment - I thought, hey, maybe different fonts, but if I copy&paste it here, it's still a `&`. Copy-and-pasting the ellipsis from your comment gives me an ellipsis though.

beorn7: Eeeenteresting... displays perfectly fine for me in GH. Encoding setting in your browser?

beorn7:
In person discussion with @fabxc @juliusv @brancz resulted in:
- let's have a trivial own multi-error implementation
- `Deliver` is not a good name. It's kind of 2nd order collecting and merging. Brainstorm resulted in `Gather` to imply the collect-and-then-merge-together aspect.

I'll change that in two follow-up commits.

beorn7:
OK, done. This should settle the fundamentals.
Please review at your earliest convenience as this fixes the most pressing issues of users.
I'd like to release quickly so that I can then work on those parts that will create more breakage (while the users can still benefit from the many low hanging fruit harvested here).

crosbymichael:
This looks good to me.  I was just coming to this repo to ask for these types to be exported so we don't always have this global registry.

prometheus/registry.go
```diff
@@ -613,90 +605,26 @@ func (r *registry) checkConsistency(metricFamily *dto.MetricFamily, dtoMetric *d
...
+//
+// This is a way to directly inject MetricFamily protobufs managed and owned by
+// the caller. The caller has full responsibility. As no registration of the
+// injected metrics has happened, there was no check at registration time. If
+// the injection results in inconsistent metrics, the Collect call will return
+// an error. Some problems may even go undetected, like invalid label names in
+// the injected protobufs.
+//
+// The hook function must be callable at any time and concurrently.
+func (r *Registry) SetInjectionHook(hook func() []*dto.MetricFamily) {
```
grobie: Have you thought about just using a custom gatherer instead of adding another function to the interface?

beorn7: Nit: This is not part of any of the interfaces. It's a use case so special that we don't need to offer that level ob abstraction.

For the real question: What do you mean by "custom gatherer"? A gatherer is doing the job of gathering and merging metric families (and return the result). An injection hook is the source of more metric families to merge.

grobie: Excuse the ambiguous wording, I was referring to the package interface of client library.

I've never used the injection hook so far, so excuse if I'm missing something, but from looking at the signatures of the current functions, I assumed one could just write a MergeGatherer or so which would call Gather() on two provided gatherer objects (or one gatherer object, one function) and do the merging.

beorn7: Got it. Yeah, that's actually a neat idea. The injection hook is essentially a gatherer. This sounds so neat that I'm afraid people will abuse it... ;) But it will make the semantics quite clear. I'll play with it. Thanks for the inspiration.

prometheus/doc.go
grobie: To add to the confusion, I'm able to see both versions :)

Here on the discussion page I only see the ampersand.

screenshot from 2016-08-04 19-03-17

But on the diff page everything is fine:

screenshot from 2016-08-04 19-02-47

grobie:
Reviewed 13 of 28 files at r1, 2 of 15 files at r2, 4 of 6 files at r4, 2 of 4 files at r5, 1 of 7 files at r6, 9 of 9 files at r8.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/prometheus/client_golang/214)*

beorn7:
I implemented @grobie 's idea about the "merge gatherer". That turned out really nice. On the way, I refactored and then improved the consistency checking.

beorn7:
I'm on vacation with only intermittent internet access. I'd still appreciate feedback about the latest version so that I can work on it when I find time and in particular wrap this up soon after being back. Thanks everybody.

prometheus/push/push.go
```diff
@@ -60,18 +60,18 @@ const contentTypeHeader = "Content-Type"
...
+func FromGatherer(job string, grouping map[string]string, url string, g prometheus.Gatherer) error {
+	return push(job, grouping, url, g, "PUT")
 }
 
-// RegistryAdd works like Registry, but only previously pushed metrics with the
-// same name (and the same job and other grouping labels) will be replaced. (It
-// uses HTTP method 'POST' to push to the Pushgateway.)
-func RegistryAdd(job string, grouping map[string]string, url string, reg prometheus.Gatherer) error {
-	return push(job, grouping, url, reg, "POST")
+// AddFromGatherer works like Gatherer, but only previously pushed metrics with
```
grobie: FromGatherer

prometheus/push/push.go
```diff
@@ -60,18 +60,18 @@ const contentTypeHeader = "Content-Type"
 // Note that all previously pushed metrics with the same job and other grouping
 // labels will be replaced with the metrics pushed by this call. (It uses HTTP
 // method 'PUT' to push to the Pushgateway.)
-func Registry(job string, grouping map[string]string, url string, reg prometheus.Gatherer) error {
-	return push(job, grouping, url, reg, "PUT")
+func FromGatherer(job string, grouping map[string]string, url string, g prometheus.Gatherer) error {
```
grobie: I'm not sold on these function names.  I don't find it intuitive, also as none of the other client libraries has the concept of a gatherer. The usage of the client library became very abstract with the introduction of the gatherer interface. I'll need to think about a better name though ...

prometheus/registry.go
```diff
@@ -426,105 +457,267 @@ func (r *Registry) Gather() ([]*dto.MetricFamily, error) {
...
+// Gather calles are all returned in a flattened MultiError. Duplicate and
+// inconsistent Metrics are skipped (first occurrence in slice order wins) and
+// reported in the returned error.
+//
+// Gatherers can be used to merge the Gather results from multiple
+// Registries. It also provides a way to directly inject existing MetricFamily
+// protobufs into the gathering by creating a custom Gatherer with a Gather
+// method that simply returns the existing MetricFamily protobufs. Note that no
+// registration is involved (as with Collector registration), so obviously no
+// registration-time checks cannot happen. Any inconsistencies between the
```
grobie: double negation

grobie:
Cool! The only thing I don't really like are the function names of the push package, but maybe others have some opinions or suggestions here. Otherwise LGTM.

prometheus/push/push.go
beorn7: Done.

prometheus/push/push.go
beorn7: Not entirely happy with the name, either. The concept is sound, IMHO, and in a certain way Go-specific (have interfaces for parts of the functionality of a concrete type).
All up for more intuitive naming, we just need ideas.

prometheus/registry.go
beorn7: Fixed.

beorn7:
Thanks, @grobie . I'll wait for ideas and objections from others over the weekend.

beorn7:
Apparently, we have found the least bad name. ;) Merging now.

PR #221, opened 2016-08-12
metricvec: handle hash collision for labeled metrics

### LINKED ISSUE #220: MetricVec hashing does not deal with collisions
If we look at https://github.com/prometheus/client_golang/blob/master/prometheus/vec.go#L167, we see there is no code to deal with collisions. For example, if two value sets come through that compute the same hash `h`, the incorrect metric will be returned.

Let me know if you want me to send a PR to fix this.

stevvooe:
While hash collisions are quite rare, the current state of the client
library carries a risk of merging two separate label values into a
single metric bucket. The effects are near impossible to detect and the
result will be missing or incorrect counters.

This changeset handles hash collisions by falling back to collision
resolution if multiple label values hash to the same value. This works
similar to separate chaining using a slice. Extra storage is minimized
to only the value key slice to that metrics can be differentiated
within a bucket.

In general, the cost of handling collisions is completely minimized
under normal operation. Performance does show slight increases in
certain areas, but these are more likely statistically anomalies. More
importantly, zero allocation behavior for metrics is preserved on the
fast path. Minimal allocations may be made during collision handling but
this has minimal effect.

Benchmark comparisons with and without collision resolution follow.

```
benchmark                                                         old ns/op     new ns/op     delta
BenchmarkCounterWithLabelValues-4                                 99.0          107           +8.08%
BenchmarkCounterWithLabelValuesConcurrent-4                       79.6          91.0          +14.32%
BenchmarkCounterWithMappedLabels-4                                518           542           +4.63%
BenchmarkCounterWithPreparedMappedLabels-4                        127           137           +7.87%
BenchmarkCounterNoLabels-4                                        19.5          19.1          -2.05%
BenchmarkGaugeWithLabelValues-4                                   97.4          110           +12.94%
BenchmarkGaugeNoLabels-4                                          12.4          10.3          -16.94%
BenchmarkSummaryWithLabelValues-4                                 1204          915           -24.00%
BenchmarkSummaryNoLabels-4                                        936           847           -9.51%
BenchmarkHistogramWithLabelValues-4                               147           147           +0.00%
BenchmarkHistogramNoLabels-4                                      50.6          49.3          -2.57%
BenchmarkHistogramObserve1-4                                      37.9          37.5          -1.06%
BenchmarkHistogramObserve2-4                                      122           137           +12.30%
BenchmarkHistogramObserve4-4                                      310           352           +13.55%
BenchmarkHistogramObserve8-4                                      691           729           +5.50%
BenchmarkHistogramWrite1-4                                        3374          3097          -8.21%
BenchmarkHistogramWrite2-4                                        5310          5051          -4.88%
BenchmarkHistogramWrite4-4                                        12094         10690         -11.61%
BenchmarkHistogramWrite8-4                                        19416         17755         -8.55%
BenchmarkHandler-4                                                11934304      13765894      +15.35%
BenchmarkSummaryObserve1-4                                        1119          1105          -1.25%
BenchmarkSummaryObserve2-4                                        3679          3430          -6.77%
BenchmarkSummaryObserve4-4                                        10678         7982          -25.25%
BenchmarkSummaryObserve8-4                                        22974         16689         -27.36%
BenchmarkSummaryWrite1-4                                          25962         14680         -43.46%
BenchmarkSummaryWrite2-4                                          38019         35073         -7.75%
BenchmarkSummaryWrite4-4                                          78027         56816         -27.18%
BenchmarkSummaryWrite8-4                                          117220        132248        +12.82%
BenchmarkMetricVecWithLabelValuesBasic-4                          138           133           -3.62%
BenchmarkMetricVecWithLabelValues2Keys10ValueCardinality-4        150           144           -4.00%
BenchmarkMetricVecWithLabelValues4Keys10ValueCardinality-4        263           256           -2.66%
BenchmarkMetricVecWithLabelValues2Keys100ValueCardinality-4       145           155           +6.90%
BenchmarkMetricVecWithLabelValues10Keys100ValueCardinality-4      606           634           +4.62%
BenchmarkMetricVecWithLabelValues10Keys1000ValueCardinality-4     746           641           -14.08%

benchmark                                                         old allocs     new allocs     delta
BenchmarkCounterWithLabelValues-4                                 0              0              +0.00%
BenchmarkCounterWithLabelValuesConcurrent-4                       0              0              +0.00%
BenchmarkCounterWithMappedLabels-4                                2              2              +0.00%
BenchmarkCounterWithPreparedMappedLabels-4                        0              0              +0.00%
BenchmarkCounterNoLabels-4                                        0              0              +0.00%
BenchmarkGaugeWithLabelValues-4                                   0              0              +0.00%
BenchmarkGaugeNoLabels-4                                          0              0              +0.00%
BenchmarkSummaryWithLabelValues-4                                 0              0              +0.00%
BenchmarkSummaryNoLabels-4                                        0              0              +0.00%
BenchmarkHistogramWithLabelValues-4                               0              0              +0.00%
BenchmarkHistogramNoLabels-4                                      0              0              +0.00%
BenchmarkMetricVecWithLabelValuesBasic-4                          0              0              +0.00%
BenchmarkMetricVecWithLabelValues2Keys10ValueCardinality-4        0              0              +0.00%
BenchmarkMetricVecWithLabelValues4Keys10ValueCardinality-4        0              0              +0.00%
BenchmarkMetricVecWithLabelValues2Keys100ValueCardinality-4       0              0              +0.00%
BenchmarkMetricVecWithLabelValues10Keys100ValueCardinality-4      0              0              +0.00%
BenchmarkMetricVecWithLabelValues10Keys1000ValueCardinality-4     0              0              +0.00%

benchmark                                                         old bytes     new bytes     delta
BenchmarkCounterWithLabelValues-4                                 0             0             +0.00%
BenchmarkCounterWithLabelValuesConcurrent-4                       0             0             +0.00%
BenchmarkCounterWithMappedLabels-4                                336           336           +0.00%
BenchmarkCounterWithPreparedMappedLabels-4                        0             0             +0.00%
BenchmarkCounterNoLabels-4                                        0             0             +0.00%
BenchmarkGaugeWithLabelValues-4                                   0             0             +0.00%
BenchmarkGaugeNoLabels-4                                          0             0             +0.00%
BenchmarkSummaryWithLabelValues-4                                 0             0             +0.00%
BenchmarkSummaryNoLabels-4                                        0             0             +0.00%
BenchmarkHistogramWithLabelValues-4                               0             0             +0.00%
BenchmarkHistogramNoLabels-4                                      0             0             +0.00%
BenchmarkMetricVecWithLabelValuesBasic-4                          0             0             +0.00%
BenchmarkMetricVecWithLabelValues2Keys10ValueCardinality-4        0             0             +0.00%
BenchmarkMetricVecWithLabelValues4Keys10ValueCardinality-4        0             0             +0.00%
BenchmarkMetricVecWithLabelValues2Keys100ValueCardinality-4       0             0             +0.00%
BenchmarkMetricVecWithLabelValues10Keys100ValueCardinality-4      0             0             +0.00%
BenchmarkMetricVecWithLabelValues10Keys1000ValueCardinality-4     0             0             +0.00%
```

Signed-off-by: Stephen J Day <stephen.day@docker.com>

Closes #220

### DIFF
--- a/prometheus/vec.go
+++ b/prometheus/vec.go
@@ -31,6 +31,16 @@ type MetricVec struct {
 	newMetric func(labelValues ...string) Metric
 }
 
+// newMetricVec returns an initialized MetricVec. The concrete value is
+// returned for embedding into another struct.
+func newMetricVec(desc *Desc, newMetric func(lvs ...string) Metric) MetricVec {
+	return MetricVec{
+		children:  map[uint64]Metric{},
+		desc:      desc,
+		newMetric: newMetric,
+	}
+}
+
 // Describe implements Collector. The length of the returned slice
 // is always one.
 func (m *MetricVec) Describe(ch chan<- *Desc) {
--- a/prometheus/vec.go
+++ b/prometheus/vec.go
@@ -25,20 +25,27 @@ import (
 // provided in this package.
 type MetricVec struct {
 	mtx      sync.RWMutex // Protects the children.
-	children map[uint64]Metric
+	children map[uint64][]metricLabelValue
 	desc     *Desc
 
 	newMetric func(labelValues ...string) Metric
+	hashAdd   func(h uint64, s string) uint64 // replace hash function for testing collision handling
 }
 
 // newMetricVec returns an initialized MetricVec. The concrete value is
 // returned for embedding into another struct.
 func newMetricVec(desc *Desc, newMetric func(lvs ...string) Metric) MetricVec {
 	return MetricVec{
-		children:  map[uint64]Metric{},
+		children:  map[uint64][]metricLabelValue{},
 		desc:      desc,
 		newMetric: newMetric,
+		hashAdd:   hashAdd,
 	}
+}
+
+type metricLabelValue struct {
+	values []string
+	metric Metric
 }
 
 // Describe implements Collector. The length of the returned slice
@@ -52,8 +59,10 @@ func (m *MetricVec) Collect(ch chan<- Metric) {
 	m.mtx.RLock()
 	defer m.mtx.RUnlock()
 
-	for _, metric := range m.children {
-		ch <- metric
+	for _, metrics := range m.children {
+		for _, metric := range metrics {
+			ch <- metric.metric
+		}
 	}
 }
 
@@ -88,7 +97,7 @@ func (m *MetricVec) GetMetricWithLabelValues(lvs ...string) (Metric, error) {
 	}
 
 	m.mtx.RLock()
-	metric, ok := m.children[h]
+	metric, ok := m.getMetric(h, lvs...)
 	m.mtx.RUnlock()
 	if ok {
 		return metric, nil
@@ -118,11 +127,12 @@ func (m *MetricVec) GetMetricWith(labels Labels) (Metric, error) {
 	}
 
 	m.mtx.RLock()
-	metric, ok := m.children[h]
-	m.mtx.RUnlock()
-	if ok {
-		return metric, nil
+	metrics, ok := m.children[h]
+	if ok && len(metrics) == 1 {
+		m.mtx.RUnlock()
+		return metrics[0].metric, nil
 	}
+	m.mtx.RUnlock()
 
 	lvs := make([]string, len(labels))
 	for i, label := range m.desc.variableLabels {
@@ -178,11 +188,7 @@ func (m *MetricVec) DeleteLabelValues(lvs ...string) bool {
 	if err != nil {
 		return false
 	}
-	if _, ok := m.children[h]; !ok {
-		return false
-	}
-	delete(m.children, h)
-	return true
+	return m.deleteByHash(h, lvs...)
 }
 
 // Delete deletes the metric where the variable labels are the same as those
@@ -203,10 +209,38 @@ func (m *MetricVec) Delete(labels Labels) bool {
 	if err != nil {
 		return false
 	}
-	if _, ok := m.children[h]; !ok {
+
+	var lvs []string
+	for _, k := range m.desc.variableLabels {
+		lvs = append(lvs, labels[k])
+	}
+
+	return m.deleteByHash(h, lvs...)
+}
+
+// deleteByHash removes the metric from the hash bucket h. If there are
+// multiple matches in the bucket, use lvs to select a metric and remove only
+// that metric.
+func (m *MetricVec) deleteByHash(h uint64, lvs ...string) bool {
+	metrics, ok := m.children[h]
+	if !ok {
 		return false
 	}
-	delete(m.children, h)
+
+	if len(metrics) < 2 {
+		delete(m.children, h)
+	}
+
+	i := findMetric(lvs, metrics)
+	if i >= len(metrics) {
+		return false
+	}
+
+	if len(metrics) > 1 {
+		m.children[h] = append(metrics[:i], metrics[i+1:]...)
+	} else {
+		delete(m.children, h)
+	}
 	return true
 }
 
@@ -226,7 +260,7 @@ func (m *MetricVec) hashLabelValues(vals []string) (uint64, error) {
 	}
 	h := hashNew()
 	for _, val := range vals {
-		h = hashAdd(h, val)
+		h = m.hashAdd(h, val)
 	}
 	return h, nil
 }
@@ -241,19 +275,70 @@ func (m *MetricVec) hashLabels(labels Labels) (uint64, error) {
 		if !ok {
 			return 0, fmt.Errorf("label name %q missing in label map", label)
 		}
-		h = hashAdd(h, val)
+		h = m.hashAdd(h, val)
 	}
 	return h, nil
 }
 
 func (m *MetricVec) getOrCreateMetric(hash uint64, labelValues ...string) Metric {
-	metric, ok := m.children[hash]
+	metric, ok := m.getMetric(hash, labelValues...)
 	if !ok {
 		// Copy labelValues. Otherwise, they would be allocated even if we don't go
 		// down this code path.
 		copiedLabelValues := append(make([]string, 0, len(labelValues)), labelValues...)
 		metric = m.newMetric(copiedLabelValues...)
-		m.children[hash] = metric
+		m.children[hash] = append(m.children[hash], metricLabelValue{values: copiedLabelValues, metric: metric})
 	}
 	return metric
+}
+
+// getMetric while handling possible collisions in the hash space. Must be
+// called while holding read mutex.
+func (m *MetricVec) getMetric(h uint64, lvs ...string) (Metric, bool) {
+	metrics, ok := m.children[h]
+	if ok {
+		return m.selectMetric(lvs, metrics)
+	}
+
+	return nil, false
+}
+
+func (m *MetricVec) selectMetric(lvs []string, metrics []metricLabelValue) (Metric, bool) {
+	switch len(metrics) {
+	case 0:
+		return nil, false
+	case 1:
+		// collisions are rare, this should be the fast path.
+		return metrics[0].metric, true
+	}
+
+	i := findMetric(lvs, metrics)
+
+	if i < len(metrics) {
+		return metrics[i].metric, true
+	}
+
+	return nil, false
+}
+
+// findMetric returns the index of the matching metric or len(metrics) if not
+// found.
+func findMetric(lvs []string, metrics []metricLabelValue) int {
+next:
+	for i, metric := range metrics {
+		if len(metric.values) != len(lvs) {
+			continue
+		}
+
+		for j, v := range metric.values {
+			if v != lvs[j] {
+				continue next
+			}
+		}
+		// falling out of the loop here means we have a match!
+
+		return i
+	}
+
+	return len(metrics)
 }
--- a/prometheus/vec.go
+++ b/prometheus/vec.go
@@ -16,6 +16,8 @@ package prometheus
 import (
 	"fmt"
 	"sync"
+
+	"github.com/prometheus/common/model"
 )
 
 // MetricVec is a Collector to bundle metrics of the same name that
@@ -25,25 +27,29 @@ import (
 // provided in this package.
 type MetricVec struct {
 	mtx      sync.RWMutex // Protects the children.
-	children map[uint64][]metricLabelValue
+	children map[uint64][]metricWithLabelValues
 	desc     *Desc
 
-	newMetric func(labelValues ...string) Metric
-	hashAdd   func(h uint64, s string) uint64 // replace hash function for testing collision handling
+	newMetric   func(labelValues ...string) Metric
+	hashAdd     func(h uint64, s string) uint64 // replace hash function for testing collision handling
+	hashAddByte func(h uint64, b byte) uint64
 }
 
 // newMetricVec returns an initialized MetricVec. The concrete value is
 // returned for embedding into another struct.
 func newMetricVec(desc *Desc, newMetric func(lvs ...string) Metric) MetricVec {
 	return MetricVec{
-		children:  map[uint64][]metricLabelValue{},
-		desc:      desc,
-		newMetric: newMetric,
-		hashAdd:   hashAdd,
+		children:    map[uint64][]metricWithLabelValues{},
+		desc:        desc,
+		newMetric:   newMetric,
+		hashAdd:     hashAdd,
+		hashAddByte: hashAddByte,
 	}
 }
 
-type metricLabelValue struct {
+// metricWithLabelValues provides the metric and its label values for
+// disambiguation on hash collision.
+type metricWithLabelValues struct {
 	values []string
 	metric Metric
 }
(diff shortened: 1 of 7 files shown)


prometheus/vec.go
```diff
@@ -108,11 +127,11 @@ func (m *MetricVec) GetMetricWith(labels Labels) (Metric, error) {
...
 	m.mtx.RLock()
-	metric, ok := m.children[h]
-	m.mtx.RUnlock()
-	if ok {
-		return metric, nil
+	metrics, ok := m.children[h]
+	if ok && len(metrics) == 1 {
+		return metrics[0].metric, nil
 	}
+	m.mtx.RUnlock()
```
grobie: How is the lock being released in case of the happy path above?

stevvooe: Fixing.

stevvooe:
Take a peak https://github.com/prometheus/client_golang/pull/221/commits/9b464a48bdf43c423779a9dab1c3a1d5cb0c0d8c. It provides some further fixes around collision handling and greatly increases the unit test coverage.

Zero-alloc behavior is preserved at the cost of some performance.

prometheus/vec.go
```diff
@@ -193,10 +186,29 @@ func (m *MetricVec) Delete(labels Labels) bool {
...
 	}
-	if _, ok := m.children[h]; !ok {
+
+	return m.deleteByHash(h, labels)
+}
+
+// deleteByHash removes the metric from the hash bucket h. If there are
+// multiple matches in the bucket, use lvs to select a metric and remove only
+// that metric.
+func (m *MetricVec) deleteByHash(h uint64, values interface{}) bool {
```
grobie: I'm not really a fan of interface type arguments. We're losing the advantages of the type system. Is this really necessary?

beorn7:
Thanks @stevvooe for the PR, thanks @grobie for the review. I'm back from my travels and will have a look ASAP. Zero-allocation and benchmarks look great.

grobie:
Note: I did not do a full review. Only commented on two things I saw while on mobile.

prometheus/vec.go
stevvooe: To maintain correctness and zero-alloc behavior without repeating code, it is necessary. Any errors are caught during test time.

What is the issue? This is private code. Are you just afraid?

prometheus/vec.go
```diff
@@ -231,19 +243,115 @@ func (m *MetricVec) hashLabels(labels Labels) (uint64, error) {
...
 	}
 	return h, nil
 }
 
-func (m *MetricVec) getOrCreateMetric(hash uint64, labelValues ...string) Metric {
-	metric, ok := m.children[hash]
+// getOrCreateMetric retrieves the metric by hash and label value or creates it
+// and returns the new one.
+//
+// This function takes a lock, so avoid
```
beorn7: Unfinished sentence.

prometheus/vec.go
```diff
@@ -231,19 +243,115 @@ func (m *MetricVec) hashLabels(labels Labels) (uint64, error) {
...
+	for i, metric := range metrics {
+		if m.matchLabels(metric.values, values) {
+			return i
+		}
+	}
+
+	return len(metrics)
+}
+
+func (m *MetricVec) matchLabels(values []string, labels interface{}) bool {
```
beorn7: Naming nit: What was consistently called `values` above is now `labels`, and now the name `values` is used for the new parameter. Perhaps `metricValues` would be less confusing (as those are the label values retrieved from the metric, see L301).

prometheus/vec.go
beorn7: We could, of course, repeat the code. Go isn't really afraid of that. But in this case, it would be a lot of code. I'm fine with the interface usage as it avoids the allocation in L117 of the old code.

It might be good, though, to document the requirements for `values` at the top of the call stack (even if it is an unexported method). I'll leave a comment where I think it should be documented.

prometheus/vec.go
```diff
@@ -25,10 +25,27 @@ import (
...
+func newMetricVec(desc *Desc, newMetric func(lvs ...string) Metric) MetricVec {
+	return MetricVec{
+		children:  map[uint64][]metricLabelValue{},
+		desc:      desc,
+		newMetric: newMetric,
+		hashAdd:   hashAdd,
+	}
+}
+
+type metricLabelValue struct {
```
beorn7: Perhaps call this `metricWithLabelValues`? When reading the code where this type is used, `metricLabelValue` suggest it's a single label value of a metric.

prometheus/vec.go
```diff
@@ -25,10 +25,27 @@ import (
 // provided in this package.
 type MetricVec struct {
 	mtx      sync.RWMutex // Protects the children.
-	children map[uint64]Metric
+	children map[uint64][]metricLabelValue
 	desc     *Desc
 
 	newMetric func(labelValues ...string) Metric
+	hashAdd   func(h uint64, s string) uint64 // replace hash function for testing collision handling
```
beorn7: Would be cool to find a real collision and use that one. Then we could avoid the plumbing.

In any case a comment nit: Please use English capitalization and punctuation, i.e. start with a capital letter and end with a period.

prometheus/vec.go
```diff
@@ -193,10 +186,29 @@ func (m *MetricVec) Delete(labels Labels) bool {
...
 		return false
 	}
-	if _, ok := m.children[h]; !ok {
+
+	return m.deleteByHash(h, labels)
+}
+
+// deleteByHash removes the metric from the hash bucket h. If there are
+// multiple matches in the bucket, use lvs to select a metric and remove only
+// that metric.
```
beorn7: This doc comment is probably a good place to document the requirements for `interface{}` ("... is of type []string or Labels").

prometheus/vec.go
beorn7: And here is the other entry point where the requirements for `interface{}` should be documented.

prometheus/vec.go
```diff
@@ -231,19 +243,115 @@ func (m *MetricVec) hashLabels(labels Labels) (uint64, error) {
...
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	metric, ok = m.getMetric(hash, values)
 	if !ok {
 		// Copy labelValues. Otherwise, they would be allocated even if we don't go
 		// down this code path.
-		copiedLabelValues := append(make([]string, 0, len(labelValues)), labelValues...)
-		metric = m.newMetric(copiedLabelValues...)
-		m.children[hash] = metric
+		labelValues := m.copyLabelValues(values)
```
beorn7: The comment here is not correct anymore ("Copy labelValues..." – it's called `values` now). And perhaps move the comment into the `copyLabelValues` method because it only applies to the case where `values` is a `[]string` (otherwise it needs to be extracted out of the Labels map anyway).

beorn7:
Looks good. Just a few naming and comment nits.

I'll try to dig out a fnv64a hash collision.

beorn7:
`77kepQFQ8Kl` and `!0IC=VloaY` both have a hash value of 0. You can use them in the test as label values so that you don't need to inject the hashing function anymore.

prometheus/vec.go
```diff
@@ -216,7 +228,7 @@ func (m *MetricVec) hashLabelValues(vals []string) (uint64, error) {
 	}
 	h := hashNew()
 	for _, val := range vals {
-		h = hashAdd(h, val)
+		h = m.hashAdd(h, val)
```
beorn7: This should actually be followed by `hashAddByte(h, separatorByte)` (this is a bug in the original code already, hash collisions could be created with label values like `{"foo", "bar"}` and `{"foob", "ar"}`).

prometheus/vec.go
```diff
@@ -231,19 +243,115 @@ func (m *MetricVec) hashLabels(labels Labels) (uint64, error) {
 		if !ok {
 			return 0, fmt.Errorf("label name %q missing in label map", label)
 		}
-		h = hashAdd(h, val)
+		h = m.hashAdd(h, val)
```
beorn7: See above.

prometheus/vec.go
grobie: You're right @stevvooe, I couldn't come up with a typed implementation either without causing more allocations. We'll need to wait until golang's escape analysis becomes better.

prometheus/vec.go
stevvooe: I think the rest of the code uses `lvs` consistently. Will that be sufficient?

prometheus/vec.go
stevvooe: > Unfinished sentence.

Woops. Was trying to get this pushed for review.

prometheus/vec.go
stevvooe: Having _everything_ collide exercises the code paths more completely. We'd have to find 3 or more collisions to test 2+N search case.

I'll use the collision provided in another test to make sure we are falling back correctly.

stevvooe:
> 77kepQFQ8Kl and !0IC=VloaY both have a hash value of 0. You can use them in the test as label values so that you don't need to inject the hashing function anymore.

Thanks for hunting this down, but this won't test the 2+N search case. That extra test caught a number of bugs in both the new and original implementation.

prometheus/vec.go
stevvooe: Good catch.

I'll use the `0x00` to keep sort order of tuple in case we decide to optimize this differently.

prometheus/vec.go
beorn7: Most importantly, the `interface{}` passed throughout the call stack is always the same, so especially in lack of any type hints, it would help to name the function parameter the same in all functions.

prometheus/vec.go
beorn7: Fair point. I can give you more strings with hash value 0, though.

prometheus/vec.go
beorn7: `separatorByte` (= 0xFF) has the advantage that it is not a valid byte value anywhere in a UTF-8 string, while 0x00 is (I believe). It is also consistently used in other hashing in Prometheus code.

Since the hashing is completely internal here, it doesn't really matter. Even if somebody constructed a collision, it doesn't matter anymore as you have implemented the detection here. Still I'd prefer consistency.

beorn7:
More strings with hash value 0:

<pre>
=. hx"iX<;
QvXtM%gt;@Fp%
_"kWk=-v$c
</pre>

See http://www.isthe.com/chongo/tech/comp/fnv/

prometheus/vec.go
stevvooe: > separatorByte (= 0xFF) has the advantage that it is not a valid byte value anywhere in a UTF-8 string, while 0x00 is (I believe). It is also consistently used in other hashing in Prometheus code.

The main problem there is that if you're debugging and print in the wrong place, things explode. Also, if you ever want to use these keys to binary search, which may make sense in certain cases, you'll have to generate one with a different separator that doesn't mess with tuple-oriented sort order.

`0x00` separated records are near standard. I'm really surprised that one would go with anything else. Where else in the code base is this used?

beorn7: This is about hashing, not about tuples. Only add this character while creating the hash, nowhere else.

https://github.com/prometheus/common/blob/master/model/signature.go

stevvooe: Good sir, we are hashing tuples.

beorn7: For certain definition of tuples, certainly. But we are not modifying the "tuple". The separator byte is only added during hashing, there is no way it will ever be printed or used in sorting.
The null byte is a valid UTF-8 character. I could put it legally into a label value. The 0xFF character can never be part of a valid UTF-8 string. That qualifies it as a good separator character during hashing.

prometheus/vec.go
stevvooe: @grobie @beorn7 Just made a polymorphic implementation and confirmed we get allocations.

Up to you, but it might be better to allow allocations and wait for better escape analysis. It would avoid the needless complexity in this code. A simple string slice header allocation on each lookup would let us delete a massive swath of code.

prometheus/vec.go
stevvooe: > The 0xFF character can never be part of a valid UTF-8 string. That qualifies it as a good separator character during hashing.

I don't see this link but I've used the constant `model.SeparatorByte`.

To give you context, we had a [sort error](https://github.com/docker/distribution/pull/1888) in the docker registry API due to the use of a `/` that sorts earlier than a `-`. Changing to `0x00` gives you expected, lexical sort order.

stevvooe:
Most commentary is dealt with. I have not added the new collision tests yet, as I don't see a way to get the coverage needed. If you're fine with getting rid of the extra coverage in, I'll re-tool to use the provided values. If we want to keep it, I'll add a separate test with the colliding values.

prometheus/vec.go
grobie: I guess we're more interested in as little as possible overhead of the client than a nice and short code base. @beorn7 is the gatekeeper of this repository though.

beorn7: Yes, avoiding allocations in this code path is essential. It is very likely to be called very frequently by users.

Current code is the best solution, I guess.

beorn7:
Tests are fine. I'm merging this now, as I want to have it int the release planned for today.

Feel free to add a test with a true collision in a separate PR. I can also do it myself.

Thanks for your contribution.

beorn7:
Only now I get it: the benchmarks in the commit description are superseeded by the following commit.

We indeed have to rework to have no allocations in the `WithLabelValues` case.
But don't worry, I'm already working on a solution.
(And I have also added a test with a "real" collision in #225.)

Thanks again.

stevvooe:
@beorn7 Alright, thanks for taking the PR.

Let me know if you need anything else.

PR #226, opened 2016-08-17
Bring back zero-alloc label-value access for metric vecs

beorn7:
Also, fix mutex copy-by-value bug.

Unfortunately, this did indeed require to duplicate a whole lot of code.

On the other hand, 117 additions and 89 deletions, not that bad... ;)

@grobie


prometheus/vec.go
```diff
@@ -193,21 +193,40 @@ func (m *MetricVec) Delete(labels Labels) bool {
...
 		return false
 	}
 
-	i := m.findMetric(metrics, lvs)
+	i := m.findMetricWithLabelValues(metrics, lvs)
+	if i >= len(metrics) {
+		return false
+	}
+
+	if len(metrics) > 1 {
```
grobie: If we really care about duplicated code, we could even move these 5 lines out in another function.

grobie:
👍

prometheus/vec.go
beorn7: True, but I guess it would not improve readability. So I better leave it as it is.

PR #233, opened 2016-08-29
[WIP] Convert exposition format to proto3.

juliusv:


juliusv:
The CI fails because this depends on https://github.com/prometheus/common/pull/52, which in turn depends on this change here. The two repos import each other circularly :-/

fabxc:
Merging the common PR did not make this one pass.

juliusv:
Right, I should have mentioned that it also depends on https://github.com/prometheus/client_model/pull/14.

beorn7:
Looks good, but all of this has to be changed anyway once the proto3 approach is finalized.

juliusv:
Yep, I changed the PR title to [WIP] for now so that nobody accidentally merges it before I restart all the proto3 stuff with the new insights.

beorn7:
I think the proto3 story (and the protobuf story as a whole) needs to be rewritten. I'll close this PR for now.

PR #234, opened 2016-08-29
update Go version for CI

fabxc:


.travis.yml
```diff
@@ -2,8 +2,8 @@ sudo: false
 language: go
 
 go:
- - 1.5.4
- - 1.6.2
+ - 1.6
```
beorn7: Does Travis use the latest 1.6 in that case, or just the original 1.6.0?

sdurrheimer: If I'm not mistaken `1.6` means `1.6.0` (https://github.com/travis-ci/gimme/blob/master/gimme#L95)

fabxc: The former is what this scheme of providing a version generally implies.
Couldn't find anything explicit in the Travis docs via googling though.

beorn7: It's more a problem with how Go treats versioning.
They call 1.6.0 in fact just 1.6, so I guess we have to explicitly specify the latest bugfix version.

fabxc: Ah, bummer. That doesn't seem like a reasonable pattern to me =/

beorn7:
:+1:

PR #235, opened 2016-09-01
releasing lock before panic to prevent deadlock

249043822:
As func (s *summary) swapBufs(now time.Time) may raise panic, so we should release the locks it held before, or other goroutines will be blocked here

### DIFF
--- a/prometheus/summary.go
+++ b/prometheus/summary.go
@@ -263,7 +263,6 @@ func (s *summary) Desc() *Desc {
 
 func (s *summary) Observe(v float64) {
 	s.bufMtx.Lock()
-	defer s.bufMtx.Unlock()
 
 	now := time.Now()
 	if now.After(s.hotBufExpTime) {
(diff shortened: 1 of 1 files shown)


beorn7:
Did you actually encounter this panic?
If the panic occurs, it's a programming error. You must not try to recover from such a panic.
File a bug about the panic instead, or fix the programming error.

249043822:
Hi beorn7, i didn't encounter this panic, i only read the codes here. And i think the code may be not too strict. As you said, if the program encoutered this panic, it will be very bad,  still if the code was run in a goroutine, the panic only cause the goroutine raising exception, the process may still run, so i thinks the panic should not effect forward processes as of lock unreleasing.  Of cource, the panic may less likely occur, it it really occurs, we do unclock to make the logic more strict yet. Thanks

beorn7:
Thanks for the rationale. But this panic is really an old-style assert. You must not recover from it.
I don't think we should make the code more complicated or more fragile (like not using `defer` and unlocking in less obvious pattern) to cater for a use case (namely recover from panic) that must not be applied anyway.

PR #241, opened 2016-10-13
Add support for Beaer Token Authorization

oscarzhao:
Default Prometheus Server is visited with no authorization. In kubernetes, we can use API Server to proxy requests to prometheus service, API Server often has a token, so I add support to token authorization to the client

### DIFF
--- a/api/prometheus/api.go
+++ b/api/prometheus/api.go
@@ -16,6 +16,7 @@
 package prometheus
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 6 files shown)


mdlayher:
It appears this needs a rebase.

api/prometheus/api.go
```diff
@@ -42,6 +43,11 @@ const (
 	epSeries      = "/series"
 )
 
+const (
+	// ContextBearerTokenKey is used to create a context from Beaer Token
+	ContextBearerTokenKey = "Bearer"
```
mdlayher: Context keys should be unexported types to avoid polluting the global namespace.  See: https://golang.org/pkg/context/#WithValue

beorn7:
Closing because of inactivity. @oscarzhao if you want to pick this up again, please recreate.

PR #244, opened 2016-10-21
Graphite bridge

stuartnelson3:
@beorn7 first pass at writing a graphite bridge. Let's talk about this when you have a chance. I have several questions written in comments, and I'm sure you'll have more to say too.

This PR addresses #197

### DIFF
--- a/prometheus/graphite/bridge.go
+++ b/prometheus/graphite/bridge.go
@@ -14,6 +14,7 @@ package graphite
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -49,15 +50,12 @@ type Bridge struct {
 	url      string
 	interval time.Duration
 	prefix   string
-	stopc    chan struct{}
 
 	g prometheus.Gatherer
 }
 
 func NewBridge(c *Config) (*Bridge, error) {
-	b := &Bridge{
-		stopc: make(chan struct{}),
-	}
+	b := &Bridge{}
 
 	if c.URL == "" {
 		return nil, errors.New("graphite bridge: no url given")
@@ -84,13 +82,8 @@ func NewBridge(c *Config) (*Bridge, error) {
 	return b, nil
 }
 
-// Stop stops the event loop.
-func (b *Bridge) Stop() {
-	b.stopc <- struct{}{}
-}
-
-// Loop starts the event loop that pushes metrics at the configured interval.
-func (b *Bridge) Loop() {
+// Run starts the event loop that pushes metrics at the configured interval.
+func (b *Bridge) Run(ctx context.Context) {
 	ticker := time.NewTicker(b.interval)
 	defer ticker.Stop()
 	for {
@@ -100,7 +93,7 @@ func (b *Bridge) Loop() {
 				// TODO: Use the right logger.
 				log.Printf("%v", err)
 			}
-		case <-b.stopc:
+		case <-ctx.Done():
 			return
 		}
 	}
--- a/prometheus/graphite/bridge.go
+++ b/prometheus/graphite/bridge.go
@@ -148,18 +148,18 @@ func toReader(mfs []*dto.MetricFamily, prefix string, now int64) (bytes.Buffer, 
 			switch *mf.Type {
 			case dto.MetricType_SUMMARY:
 				if summary := m.GetSummary(); summary != nil {
-					_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, "count"), "."), float64(*summary.SampleCount), now))
+					_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, "count"), "."), float64(summary.GetSampleCount()), now))
 					if err != nil {
 						return buf, err
 					}
-					_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, "sum"), "."), *summary.SampleSum, now))
+					_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, "sum"), "."), summary.GetSampleSum(), now))
 					if err != nil {
 						return buf, err
 					}
 
 					for _, q := range summary.GetQuantile() {
 						quantile := fmt.Sprintf("quantile.%g", *q.Quantile*100)
-						_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, quantile), "."), *q.Value, now))
+						_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, quantile), "."), q.GetValue(), now))
 						if err != nil {
 							return buf, err
 						}
--- a/prometheus/graphite/bridge.go
+++ b/prometheus/graphite/bridge.go
@@ -92,6 +92,7 @@ func (b *Bridge) Stop() {
 // Loop starts the event loop that pushes metrics at the configured interval.
 func (b *Bridge) Loop() {
 	ticker := time.NewTicker(b.interval)
+	defer ticker.Stop()
 	for {
 		select {
 		case <-ticker.C:
--- a/prometheus/graphite/bridge.go
+++ b/prometheus/graphite/bridge.go
@@ -148,18 +148,18 @@ func toReader(mfs []*dto.MetricFamily, prefix string, now int64) (bytes.Buffer, 
 			switch *mf.Type {
 			case dto.MetricType_SUMMARY:
 				if summary := m.GetSummary(); summary != nil {
-					_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, "count"), "."), float64(summary.GetSampleCount()), now))
+					_, err = buf.WriteString(fmt.Sprintf(graphiteFormatString, strings.Join(append(parts, "count"), "."), float64(summary.GetSampleCount()), now))
 					if err != nil {
 						return buf, err
 					}
-					_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, "sum"), "."), summary.GetSampleSum(), now))
+					_, err = buf.WriteString(fmt.Sprintf(graphiteFormatString, strings.Join(append(parts, "sum"), "."), summary.GetSampleSum(), now))
 					if err != nil {
 						return buf, err
 					}
 
 					for _, q := range summary.GetQuantile() {
 						quantile := fmt.Sprintf("quantile.%g", *q.Quantile*100)
-						_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, quantile), "."), q.GetValue(), now))
+						_, err = buf.WriteString(fmt.Sprintf(graphiteFormatString, strings.Join(append(parts, quantile), "."), q.GetValue(), now))
 						if err != nil {
 							return buf, err
 						}
@@ -167,18 +167,18 @@ func toReader(mfs []*dto.MetricFamily, prefix string, now int64) (bytes.Buffer, 
 				}
 			case dto.MetricType_HISTOGRAM:
 				if histogram := m.GetHistogram(); histogram != nil {
-					_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, "count"), "."), float64(histogram.GetSampleCount()), now))
+					_, err = buf.WriteString(fmt.Sprintf(graphiteFormatString, strings.Join(append(parts, "count"), "."), float64(histogram.GetSampleCount()), now))
 					if err != nil {
 						return buf, err
 					}
-					_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, "sum"), "."), histogram.GetSampleSum(), now))
+					_, err = buf.WriteString(fmt.Sprintf(graphiteFormatString, strings.Join(append(parts, "sum"), "."), histogram.GetSampleSum(), now))
 					if err != nil {
 						return buf, err
 					}
 
 					for _, b := range histogram.GetBucket() {
 						bucket := fmt.Sprintf("bucket.%g", b.GetUpperBound())
-						_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(append(parts, bucket), "."), float64(b.GetCumulativeCount()), now))
+						_, err = buf.WriteString(fmt.Sprintf(graphiteFormatString, strings.Join(append(parts, bucket), "."), float64(b.GetCumulativeCount()), now))
 						if err != nil {
 							return buf, err
 						}
@@ -188,7 +188,7 @@ func toReader(mfs []*dto.MetricFamily, prefix string, now int64) (bytes.Buffer, 
 				// TODO: Do we want to allow partial writes? i.e., do
 				// we want to attempt to parse later metrics if an
 				// earlier one fails?
-				_, err = buf.WriteString(fmt.Sprintf("%s %g %d\n", strings.Join(parts, "."), getValue(m), now))
+				_, err = buf.WriteString(fmt.Sprintf(graphiteFormatString, strings.Join(parts, "."), getValue(m), now))
 				if err != nil {
 					return buf, err
 				}
@@ -200,6 +200,8 @@ func toReader(mfs []*dto.MetricFamily, prefix string, now int64) (bytes.Buffer, 
 }
 
 var re = regexp.MustCompile("[^a-zA-Z0-9_-]")
+
+const graphiteFormatString = "%s %g %d\n"
 
 func sanitize(s string) string {
 	return re.ReplaceAllString(s, "_")
--- a/prometheus/graphite/bridge.go
+++ b/prometheus/graphite/bridge.go
@@ -16,7 +16,6 @@
 package graphite
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	var z time.Duration
+	if c.Interval == z {
+		b.interval = 15 * time.Second
+	} else {
+		b.interval = c.Interval
+	}
+
+	go b.loop()
```
brian-brazil: In Java/Python the looping is optional.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,224 @@
...
+		for _, m := range mf.GetMetric() {
+			sort.Sort(prometheus.LabelPairSorter(m.GetLabel()))
+
+			parts := []string{prefix, mf.GetName()}
+			for _, lp := range m.GetLabel() {
+				parts = append(parts, sanitize(lp.GetName())+"."+sanitize(lp.GetValue()))
+			}
+
+			switch mf.GetType() {
+			case dto.MetricType_SUMMARY:
```
brian-brazil: It's probably better to convert this into a model.Vector rather than reimplementing all that logic again (you're not handling the case where the +Inf bucket is missing correctly for example).

beorn7:
Thanks for the feedback. @stuartnelson3 and I will do an in-person review today(-ish).

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+	URL string
+
+	// The interval to use for pushing data to Graphite. Defaults to 15 seconds.
+	Interval time.Duration
+
+	// The Gatherer to use for metrics. Defaults to prometheus.DefaultGatherer.
+	Gatherer prometheus.Gatherer
+
+	// The logger that messages are written to. Defaults to log.Base().
+	Logger log.Logger
```
beorn7: This forces users to implement a very specific logger interface. However, you are only using a single method. I suggest to do tho same as in https://github.com/prometheus/client_golang/blob/master/prometheus/promhttp/http.go#L163-L168 , i.e. define a Logger interface here in the package that has just the one method you need. Many loggers (including the stdlib one and the Prometheus common logger) will naturally implement that interface. In that way, you are providing maximal flexibility.

(Note that I suggest to change the one logging function used below.)

Once it's done that way, I wouldn't user the Prometheus common logger as the default. Perhaps no logging as default?

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+// Run starts the event loop that pushes Prometheus metrics to Graphite at the
+// configured interval.
+func (b *Bridge) Run(ctx context.Context) {
+	ticker := time.NewTicker(b.interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			if err := b.Push(); err != nil {
+				b.logger.Errorf("%v", err)
```
beorn7: Since this is the only thing you log, I wouldn't enforce leveled logging on all your users. Just do `b.logger.Println(err)` here and define a `Logger` interface with just `Println` as method. Many loggers can be passed in in that way.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+	// The interval to use for pushing data to Graphite. Defaults to 15 seconds.
+	Interval time.Duration
+
+	// The Gatherer to use for metrics. Defaults to prometheus.DefaultGatherer.
+	Gatherer prometheus.Gatherer
+
+	// The logger that messages are written to. Defaults to log.Base().
+	Logger log.Logger
+
+	// The prefix for your graphite metric. Defaults to empty string.
```
beorn7: graphite → Graphite

This is the only time you address the reader with "you". Perhaps just "Prefix for the pushed Graphite metrics."?

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+	Prefix string
+}
+
+// Bridge pushes metrics to the configured graphite server.
```
beorn7: graphite → Graphite

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+	g      prometheus.Gatherer
+	logger log.Logger
+}
+
+// NewBridge returns a pointer to a new Bridge struct.
+func NewBridge(c *Config) (*Bridge, error) {
+	b := &Bridge{}
+
+	if c.URL == "" {
+		return nil, errors.New("graphite bridge: no url given")
```
beorn7: While I can imagine it's useful in certain situations, we haven't put origin hints into the error string anywhere else. Suggestion:
"missing URL"

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+	b.url = c.URL
+
+	if c.Gatherer == nil {
+		b.g = prometheus.DefaultGatherer
+	} else {
+		b.g = c.Gatherer
+	}
+
+	if c.Logger == nil {
+		b.logger = log.Base()
```
beorn7: See above. I wouldn't pull in the highly opinionated Prometheus common logger into a general purpose library.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+		b.logger = c.Logger
+	}
+
+	if c.Prefix != "" {
+		b.prefix = c.Prefix
+	}
+
+	var z time.Duration
+	if c.Interval == z {
+		b.interval = 15 * time.Second
```
beorn7: Perhaps have that as a constant at the top of the file?

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// Push pushes Prometheus metrics to the configured Graphite server.
+func (b *Bridge) Push() error {
+	mfs, err := b.g.Gather()
+	if err != nil {
+		return err
+	}
+
+	now := time.Now().Unix()
```
beorn7: Use `model.Now()`.

The above will yield seconds, mot milliseconds.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	defer conn.Close()
+
+	_, err = io.Copy(conn, buf)
+	return err
+}
+
+// TODO: Do we want to allow partial writes? i.e., Skip a failed metric, but
+// try to write all metrics?
+func toReader(mfs []*dto.MetricFamily, prefix string, now int64) (*bytes.Buffer, error) {
```
beorn7: `now` should use `model.Time` as type.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+
+// Push pushes Prometheus metrics to the configured Graphite server.
+func (b *Bridge) Push() error {
+	mfs, err := b.g.Gather()
+	if err != nil {
+		return err
+	}
+
+	now := time.Now().Unix()
+	// TODO: Write directly to conn?
```
beorn7: Probably better, as `mfs` already contains all the gathered metrics buffered up. At a first glance, I don't see any reason to buffer again. Any other ideas?

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+	buf, err := toReader(mfs, b.prefix, now)
+	if err != nil {
+		return err
+	}
+
+	// TODO: Should we expose a deadline to the user?
```
beorn7: We should have a timeout for the whole push (as field in the Config struct).

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+		return err
+	}
+	defer conn.Close()
+
+	_, err = io.Copy(conn, buf)
+	return err
+}
+
+// TODO: Do we want to allow partial writes? i.e., Skip a failed metric, but
+// try to write all metrics?
```
beorn7: It's just one big write op, isn't it? It's Graphite's responsibility what to do with that write op not completing.

Or are you referring to something similar to https://github.com/prometheus/client_golang/blob/master/prometheus/promhttp/http.go#L142-L161 ? In that case, we have to decide above after L121, in a similar fashion. Could be an option in Config, i.e. `PushPartial` or `ContinueOnError` or something.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+		}, mfs...),
+		prefix: prefix,
+	}
+
+	return bytes.NewBufferString(vec.String()), nil
+}
+
+// Vector wraps the type model.Vector to implement the String() method for
+// Graphite formatting.
+type Vector struct {
```
beorn7: Not sure if this is really the best approach. It's a lot of "design fluff" to create all those separate types to wrap the orginial model and then funnel everything through re-implemented String() methods. It's only a one-time use, after all. It's also not very resource-efficient to put everything into strings, join them into other strings, to finally write them into a Writer anyway.

How about creating a set of normal functions that convert things and write the result into the Writer directly?

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+	prefix string
+
+	model.Vector
+}
+
+// String implements the Stringer interface.
```
beorn7: golint will not complain if you don't doc-comment `String()`.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,230 @@
...
+// Sample is a type alias for model.Sample to implement the String() method for
+// Graphite formatting.
+type Sample model.Sample
+
+// String implements the Stringer interface.
+func (s Sample) String() string {
+	return fmt.Sprintf("%s %g %d",
+		Metric(s.Metric),
+		s.Value,
+		int64(s.Timestamp),
```
beorn7: s.Timestamp is in ms, but I guess Graphite expects seconds.
The String() method of model,Time is dealing properly with that (but creates a floating point number). Not sure what Graphite likes here.

prometheus/graphite/bridge.go
stuartnelson3: We parse metric by metric and then write, but I think based on the suggestion above to write directly to Graphite that we'll write until error, and leave Graphite to handle any issues with a partial write.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,244 @@
...
+// Push pushes Prometheus metrics to the configured Graphite server.
+func (b *Bridge) Push() error {
+	mfs, err := b.g.Gather()
+	if err != nil {
+		switch b.errorHandling {
+		case AbortOnError:
+			return err
+		case ContinueOnError:
+			if b.logger != nil {
+				b.logger.Printf("continue on error: %v", err)
```
stuartnelson3: Logged this in the event a user would find it interesting, but am happy to remove this.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func writeMetrics(w io.Writer, mfs []*dto.MetricFamily, prefix string, now model.Time) (int, error) {
+	vec := expfmt.ExtractSamples(&expfmt.DecodeOptions{
+		Timestamp: now,
+	}, mfs...)
+
+	var total int
+	for _, s := range vec {
+		n, err := fmt.Fprintf(w, "%s %g %.3f\n", strings.Join([]string{sanitize(prefix), toMetric(s.Metric)}, "."), s.Value, float64(s.Timestamp)/millisecondsPerSecond)
```
stuartnelson3: Let me know if you would like to optimize this further

brian-brazil: As far as I'm aware, Graphite generally takes only integer timestamps (and we can't presume what's on the other end is Graphite)

stuartnelson3: Ah, that explains why the metrics weren't showing up in Graphite.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,244 @@
...
+// configured interval.
+func (b *Bridge) Run(ctx context.Context) {
+	ticker := time.NewTicker(b.interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			if err := b.Push(); err != nil {
+				if b.logger != nil {
+					b.logger.Printf("%v", err)
```
beorn7: Perhaps be a bit more verbose. Also, Printf is not really required:
`b.logger.Println("error pushing to Graphite:", err)`

prometheus/graphite/bridge.go
beorn7: That's fine. I'd just use Println:

`b.logger.Println("continue on error:", err)`

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+	}
+}
+
+// Push pushes Prometheus metrics to the configured Graphite server.
+func (b *Bridge) Push() error {
+	mfs, err := b.g.Gather()
+	if err != nil {
+		switch b.errorHandling {
+		case AbortOnError:
```
beorn7: `|| len(mfs) == 0`

(to not try to push no metrics)

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,244 @@
...
+			return err
+		case ContinueOnError:
+			if b.logger != nil {
+				b.logger.Printf("continue on error: %v", err)
+			}
+		}
```
beorn7: Panic as default here.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func writeMetrics(w io.Writer, mfs []*dto.MetricFamily, prefix string, now model.Time) (int, error) {
+	vec := expfmt.ExtractSamples(&expfmt.DecodeOptions{
+		Timestamp: now,
+	}, mfs...)
+
+	var total int
+	for _, s := range vec {
+		n, err := fmt.Fprintf(w, "%s %g %d\n", strings.Join([]string{sanitize(prefix), toMetric(s.Metric)}, "."), s.Value, int64(s.Timestamp)/millisecondsPerSecond)
```
beorn7: Almost there. :)

Sorry for being so persisting about efficiency here. This is a code path that potentially shovels a lot of date out to Graphite, and it is a client library we ask our users to link into their most precious production binaries. So we really have to be careful to keep a low performance profile.

This line (including its calls) is still treating a lot of throw-away strings.

I would suggest something like:

- No need to track the number of written bytes, you are throwing it away above anyway.
- After all, we should use a `bufio.Writer`. I'm not sure if the `Conn` is buffered, but in any case, the `bufio.Writer` has good methods for writing strings and bytes and such at low overhead.

```Go
buf := bufio.NewWriter(w)

if err := writeSanitized(buf, prefix); err != nil {
    return err
}
if err := buf.WriteByte('.'); err != nil {
    return err
}
if err := writeMetric(buf, s.Metric); err != nil {
    return err
}
if _, err := fmt.Fprintf(buf, " %g %d\n", s.Value, int64(s.Timestamp)/millisecondsPerSecond); err != nil {
    return err
}
if err := buf.Flush(); err != nil {
    return err
}
return nil
```

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,244 @@
...
+		if err != nil {
+			return 0, err
+		}
+		total += n
+	}
+
+	return total, nil
+}
+
+func toMetric(m model.Metric) string {
```
beorn7: As per the suggestion above, this should take a bufio.Writer and stream directly into it.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,244 @@
...
+		return fmt.Sprintf("%s.%s", sanitize(string(metricName)), strings.Join(labelStrings, "."))
+	}
+}
+
+var (
+	reInvalidChars       = regexp.MustCompile("[^a-zA-Z0-9_-]")
+	reRepeatedUnderscore = regexp.MustCompile("_{2,}")
+)
+
+func sanitize(s string) string {
```
beorn7: As per the suggestion above, this should take a bufio.Writer and stream directly into it.

Code would be something like:

```Go
prevUnderscore := false

for _, c := range s {
    c = replaceInvalidRune(c)
    if c == '_' {
        if prevUnderscore {
            continue
        }
        prevUnderscore = true
    } else {
        prevUnderscore = false
    }
    if _, err := buf.WriteRune(c); err != nil {
        return err
    }
}

func replaceInvalidRune(rune c) rune {
    // do numerical comparisons to identify invalid chars.
   // See https://github.com/prometheus/common/blob/master/model/metric.go#L92-L95
}
```

prometheus/graphite/bridge_test.go
```diff
@@ -271,3 +276,38 @@ type mockGraphite struct {
 
 	net.Listener
 }
+
+func ExampleBridge() {
```
stuartnelson3: Also, don't hesitate to re-write this example as you see fit.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,277 @@
...
+
+// HandlerErrorHandling defines how a Handler serving metrics will handle
+// errors.
+type HandlerErrorHandling int
+
+// These constants cause handlers serving metrics to behave as described if
+// errors are encountered.
+const (
+	// Abort the push to Graphite upon the first error encountered.
+	AbortOnError HandlerErrorHandling = iota
```
brian-brazil: Continue should go first so it's the default. A single blip shouldn't take out your monitoring. This is what java/python do.

beorn7:
Thanks everybody. I have a closer look once back from Velocity.

prometheus/graphite/bridge.go
```diff
@@ -0,0 +1,277 @@
...
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+// Push pushes Prometheus metrics to the configured Graphite server.
+func (b *Bridge) Push() error {
+	mfs, err := b.g.Gather()
+	if err != nil || len(mfs) == 0 {
```
beorn7: My comment was incorrectly placed, but this way it's also not what we want, because we would log a nil error in L168. You could add another condition there, but it's probably best to treat things explicitly. Let's just add a condition on L174 to return err if the len(mfs) is 0, i.e. we don't want to connect if there is nothing gathered, and in that case we just return the error (which could be nil if there was nothing collected legitimately.)

prometheus/graphite/bridge_test.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	go func() {
+		// Start the service.
+		if err := http.ListenAndServe(":8080", nil); err != nil {
+			log.Fatal(err)
+		}
+	}()
```
beorn7: I think this is confusing. Let's just have a comment like `// Start something as a goroutine that uses metrics.` (This may or may not be something that uses an HTTP server.)

stuartnelson3:
Nits picked, let me know

prometheus/graphite/bridge_test.go
```diff
@@ -291,7 +291,7 @@ func ExampleBridge() {
 	}
 
 	go func() {
-		// Start the service.
+		// Start something in a goroutine that uses metrics.
 		if err := http.ListenAndServe(":8080", nil); err != nil {
```
beorn7: What I meant is to not actually start an HTTP server here. (It suggests that starting an HTTP server is part of what you have to do to push Graphite metrics, while in fact that's not what you have to do anymore, in contrast to presenting Prometheus metrics in the usual way.)

beorn7:
Final clarifying nit above.

stuartnelson3:
Mk let's see how that one looks then

Issue #219, opened 2016-08-04
ProcessCollector: rework


beorn7:
In particular, see https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector.go#L106

However, even if we can handle errors gracefully now, it will not be the case with the DefaultRegistry. Since a process collector is often added as a best-effort collector to a collections that is otherwise critical to be complete, the process collector should only report errors when requested so. To not have a proliferation of constructors, have just one with ProcessCollectorOpts (pudFn, reportErrors, namespace, with default behavior to not report errors, empty string as namespace, and os.GetPid as pidFn).

Also, the collector should use const metrics.

grobie:
@beorn7 These will be breaking changes. I believe the ProcessCollector is rarely used manually, but if you want to be super strict about it, we'd need to wait until 0.10, right?

beorn7:
No, “mild” breaking changes are OK for 0.9, i.e. those that affect only rare usage patterns.

0.10 is for those breaking changes that will affect almost everybody.

PR #246, opened 2016-10-25
Convert process collector to const metrics

beorn7:
This is needed to remove the deprecated Set method from the Counter.

@grobie 

This contributes to #219.


beorn7:
7 lines less... ;)

PR #249, opened 2016-10-30
Rename LICENSE to LICENSE.md

sparshtekriwal:


beorn7:
Thanks, but this is not following conventions. While the Apache license file renders OK-ish as markdown, it is not meant to be markdown. Even GitHub creates it as `LICENSE` rather than `LICENSE.md` if you create a new GitHub repo and request to prepopulate it with the Apache license.

MarshallOfSound:
@beorn7 Could you tag this as invalid.  It is a spam PR for hacktoberfest

Issue #231, opened 2016-08-19
Add timer helpers


beorn7:
To time a function, the typical code looks like

``` Go
func TimeMe() {
    defer func(begun time.Time) {
        duration.Observe(time.Since(begun).Seconds())
    }(time.Now())
    // Do something.
}
```

A simple utility function can allow something like

``` Go
func TimeMe() {
    defer prometheus.NewTimer()(duration)
    // Do something.
}
```

`NewTimer` returns the function to stop. So some would recommend to write it more explicit:

``` Go
func TimeMe() {
    stop := prometheus.NewTimer()
    defer stop(duration)
    // Do something.
}
```

Issue #223, opened 2016-08-16
Rework type structure


beorn7:
Current issues:
- Some interfaces are identical (`Untyped` and `Gauge`; `Histogram` and `Summary`; all the `XxxFunc` metrics), they are the same for Go (you could assign a Histogram to a Summary variable). The number of interfaces can be reduced a lot here.
- `Metric` can be way less exposed as it is essentially only used internally by the registry. It's probably used by normal using mostly for testing, but it is planned to provide some test tools to make that unnecessary.
- Interfaces like `Gauge` and `Counter` should just have the methods to manipulate the metrics value in instrumentation. `Collector` and `Metric` should be separate, and then have a `GaugeCollector` to return by `NewGauge` and such.
- Consider making the various `XxxVec` interfaces so that we can hide `MetricVec`.

Overall, the number of exported types should be reduced drastically.

PR #250, opened 2016-10-31
Add timer helper function

stuartnelson3:
@beorn7 we didn't discuss this at all so I'm not sure if I tested how you would like or even implemented the timer correctly, I just went off the API you described in #231


beorn7:
Thanks. Reviewing this is at approx. position 10 to 12 of the list of things to do. Sorry for that.

(It might or might not be one of the issues where reviewing takes longer than implementing it myself. We'll see...)

prometheus/timer_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+func TestTimerObserve(t *testing.T) {
+	his := NewHistogram(
+		HistogramOpts{
+			Name: "test_histogram",
+		},
+	)
+
+	stop := NewTimer()
+	stop(his)
```
brian-brazil: The way this looks in Java is

```
static final Summary requestLatency = Summary...

Summary.Timer requestTimer = requestLatency.startTimer();
requestTimer.observeDuration();
```

We should keep to the same naming and structure for consistency.

brian-brazil: I don't believe I've come across code in Go handling this case. I'd expect something like:

```
timer := summary.StartTimer()
defer timer.ObserveDuration()
```

would work well in go.

brian-brazil: > The advantage of this method is that the metric type and the timer are independent.

I'm not sure that is an advantage, it makes discovery harder.

> As long as a metric implements the Observe(int64) method, it works with the timer.

We also need this on gauges for batch job timing, it's startTimer/setDuration in Java.

> The main disadvantage of your example is that each metric that wants to have this helper has to have the additional methods StartTimer and ObserveDuration defined on it.

A bit of extra code in a client to make usage easier to a very common use case seems like a good tradeoff to me.

beorn7:
OK, deep breath. I'll try to explain my thoughts on this (which is one of the issues I had done some research before, I just hadn't found any time yet to turn the results into an implementation).

Let's start with very fundamental Go philosophy…

Go is not really OO. Bundling functions as methods of a type for the sake of discoverability is highly non-idiomatic. Methods in Go are mostly there to implement interfaces. Interfaces in turn are supposed to be as lean as possible, ideally they have exactly one method.

The current `client_golang` code partially honors Go idioms, partially it has flaws in this regard, stemming from the fact that everybody involved was two to three years less Go experienced when the current design was conceived. The recent addition of local registries was following Go practices in a much better way. #223 tackles quite a few issues with interface and type design.

In short: We need more and leaner interfaces, and concrete types that implement multiple interfaces.

And now we can talk about the design of the timer helper here. We must not bloat the interfaces for Histogram/Summary/Gauge even more. (On the contrary, #223 aims towards making those interfaces leaner and separate them from the Collector and Metric part.) What we _could_ do is to create a new interface `Timeable` (or similar) that has a `StartTimer` method. The concrete types implementing Histogram/Summary/Gauge could then _also_ implement that new `Timeable` interface.

We probably depend on #223 to implement that properly, but that's not my main concern about that approach. My main concern is that it doesn't solve one of the problems I want to solve.

Before explaining that in more detail, let's step back for a moment and look again at the code we want to simplify.

This is the recommended code right now for timing with a histogram (or summary) and a gauge:

```Go
var (
	reqDuration prometheus.Histogram
	batchDuration prometheus.Gauge
)

func TimeMe() {
	defer func(begun time.Time) {
		reqDuration.Observe(time.Since(begun).Seconds())
	}(time.Now())
	// ...
}

func Batch() {
	defer func(begun time.Time) {
		batchDuration.Set(time.Since(begun).Seconds())
	}(time.Now())
	// ...
}
```

This is is a bit on the verbose side, but it reads quite nicely. Almost like English: "Using the reqDuration histogram, I observe the time since begun in seconds." And every Go programmer immediately sees and understands what's going on.

With the proposed `Timeable` interface, this would change to

```Go
func TimeMe() {
	timer := reqDuration.StartTimer()
	defer timer.Stop()
	// ...
}

func Batch() {
	timer := batchDuration.StartTimer()
	defer timer.Stop()
	// ...
}
```

Note that only one line is saved, at the cost of additional required knowledge (about the existence of the `StartTimer` method and how to use it.

One could go even more compact with the same approach:

```Go
func TimeMe() {
	defer reqDuration.StartTimer().Stop()
	// ...
}

func Batch() {
	defer batchDuration.StartTimer().Stop()
	// ...
}
```

Arguably less readable.

Some Go devs also like to just return a function (similar to Stuart in this PR) but that's frowned upon by others as it becomes quite cryptic in the one liner case:

```Go
func TimeMe() {
	stopTmer := reqDuration.StartTimer()
	defer stopTimer()
	// ...
}

func Batch() {
	stopTimer := batchDuration.StartTimer()
	defer stopTimer()
	// ...
}
```
 vs.

```Go
func TimeMe() {
	defer reqDuration.StartTimer()()  // Ugh.
	// ...
}

func Batch() {
	defer batchDuration.StartTimer()()  // Ugh.
	// ...
}
```

So far so good. However, I'm kind of lukewarm about doing all of that to essentially removing one line of code and arguably not even improving readability. I would have doubts if it's worth it at all.

But I talked about another problem I want to solve. And that's the more complex case where you want to time something, but only during the execution of a function you will find out what metric you want to use. That's the case if tracking different outcomes in different elements of an HistogramVec. Or if you plainly don't want to track the time if you run into an error during executing the function. Then the code gets really confusing and convoluted. And here some tooling can actually help to make the code both shorter and easier to understand. @stuartnelson3 's approach is already one step into that direction because it sets the histogram/summary/gauge to use for timing only at stop time. However, you ideally want to do the `defer` as soon as possible.

Timing is fundamentally a three-state process:
1. Start the timer.
2. Select the metric to use.
3. Stop the timer.

You want to `defer` step 3, and if no metric was selected in step 2, it implies that you don't want to track the timing for this particular call.

So what I suggest is a `Timer` type that is created with `NewTimer` or perhaps better `StartTimer`, which has three methods:
- `Stop` to stop the timer and call `Observe` or `Set` on the `Observer` or `Gauge` set before. (Nothing happens if no metric is set.)
- `WithObserver` to set a histogram or summary to use. `nil` is possible to unset. Returns the timer for convenience.
- `WithGauge` to set a gauge to use. `nil` is possible to unset. Returns the timer for convenience.

The easy case stays easy:

```Go
func TimeMe() {
	timer := prometheus.StartTimer().WithObserer(reqDuration)
	defer timer.Stop()
	// ...
}

func Batch() {
	timer := prometheus.StartTimer().WithGauge(batchDuration)
	defer timer.Stop()
	// ...
}
```

But you can do labeled Observers, too, or prevent recording of the timing:

```Go
func TimeMePartitionedOnOutcome() error {
	timer := prometheus.StartTimer()  // No timing by default.
	defer timer.Stop()

	result, err := doSomething()
	if err != nil {
		return err
	}
	timer.WithObserver(reqDuration.With(result.Type))
	return nil
}

func TimeMeButNotTheErrors() error {
	timer := prometheus.StartTimer().WithGauge(reqDuration)
	defer timer.Stop()

	result, err := req.Process()
	if err != nil {
		timer.WithObserver(nil)  // Don't time errors.
		return err
	}
	// Do other things, potentially with different return points.
	return nil
}
```

If the timer plumbing is part of the metric, the above wouldn't be possible. So I'm positive we want to separate it. @stuartnelson3 's approach separates it already, but you cannot `defer` the stop if you want different outcomes.

brian-brazil:
> And that's the more complex case where you want to time something, but only during the execution of a function you will find out what metric you want to use. 

I don't think that's a use case we should be supporting, we should be discouraging it if anything as it smells similar to splitting latency based on success/failure.

> Or if you plainly don't want to track the time if you run into an error during executing the function. 

This is already discouraged in our guidelines.


I'm for the explicit defer, without it things are a bit too magic.

```
timer := prometheus.StartTimer().WithObserer(reqDuration)
defer timer.Stop()
```

This seems too wordy to me and it's not quite obvious what this is doing - are we observing the behaviour of the metric? What is it exactly that's being observed? Additionally the names aren't following the patterns set in other clients. Consistency across clients makes life easier for our users.


We should be focusing on the primary encouraged use cases, not esoteric use cases that we're trying to move users away from. If someone wants to do something weird they can do it by hand.

beorn7:
> > And that's the more complex case where you want to time something, but only during the execution of a function you will find out what metric you want to use. 

> I don't think that's a use case we should be supporting, we should be discouraging it if anything as it smells similar to splitting latency based on success/failure.

I had the use case several times. My principal engineer had the use case multiple times. Even the SRE book talks about the importance of tracking successful and error latency separately. (But the functionality I want to offer is not limited to success/failure.) While the common case should be to not pick the metric after the timer has started, the uncommon case is not uncommon enough to not support it.

> > Or if you plainly don't want to track the time if you run into an error during executing the function. 

> This is already discouraged in our guidelines.

How do you define "discouraged"? In the easy case, errors are rare, so it doesn't really matter if you include them. But not all cases are easy. It can be very important to not include errors in the latency measurement. I don't think the case for "always include error latencies" is strong enough to enforce it in the tooling.

> This seems too wordy to me and it's not quite obvious what this is doing - are we observing the behaviour of the metric? What is it exactly that's being observed? Additionally the names aren't following the patterns set in other clients. Consistency across clients makes life easier for our users.

I'm all ears for naming suggestions. However, the functionality we are talking about here is not supported by other clients, if I understand that correctly.

> We should be focusing on the primary encouraged use cases, not esoteric use cases that we're trying to move users away from. If someone wants to do something weird they can do it by hand.

That's OK if the other use cases are only justified under very rare condition. I challenge that assumption.

Should we come to the conclusion that we only want to support the simple case, I am in turn for not offering any timer tooling at all as the additional plumbing cannot really be justified by saving one line of code. See explanation above.

brian-brazil:
> While the common case should be to not pick the metric after the timer has started, the uncommon case is not uncommon enough to not support it.

I think it's rare enough that we shouldn't be changing our APIs to allow for it, especially not at the cost of the use case that'll comprise the vast majority of uses.

> How do you define "discouraged"?

Our exporter guidelines say not to do it.

> However, the functionality we are talking about here is not supported by other clients, if I understand that correctly.

As it happens the Java client supports the not observing, though I'd consider it a bad idea to use it. Java 1.7 has a feature that'd make it less prone to the user forgetting to observe on some code paths, but unfortunately we're stuck on 1.6 so can't offer the better API.

I consider this among the things that it's good that clients don't support or try to make easy.

> I don't think the case for "always include error latencies" is strong enough to enforce it in the tooling.

We're not talking about a hard forcing, it'll always be possible for the user to do it by hand. The cases where it are valid are very rare (I can think of only one I've come across that's even close, and that one doesn't really count).

beorn7:
> I think it's rare enough that we shouldn't be changing our APIs to allow for it, especially not at the cost of the use case that'll comprise the vast majority of uses.

We are changing (or amending) the API here anyway. And I don't see my suggestion at the cost of the majority use case. Adding a method to the metric itself would be problematic in Go terms anyway. 

> Our exporter guidelines say not to do it.

I'm not sure about which part of the guidelines exactly you are talking. Could you give me a pointer?

> We're not talking about a hard forcing, it'll always be possible for the user to do it by hand. The cases where it are valid are very rare (I can think of only one I've come across that's even close, and that one doesn't really count).

You seem to disagree here with a number of different people. What I have seen is that those other people write their tooling themselves, and that's more error-prone than offering a way to do it. (My suggestion is still not dragging you into doing partitioned timers. The simple case is still the simplest pattern.)

In different news: Since using a Gauge for timing is so rare, too, why is that in the libraries, although it would be quite easy to code it by hand for the few cases where it matters?

brian-brazil:
> And I don't see my suggestion at the cost of the majority use case. 

It makes the usage not read as well, and encourages users to mess around with if and which metrics they'll do observations on. I think we should try to avoid making it too easy for users to apply advanced usages inappropriately.

> Adding a method to the metric itself would be problematic in Go terms anyway.

Do you have a reference for this? None of the obvious search terms are turning up anything relevant. I can only find descriptions of how interfaces work, nothing about how functionality in methods should be avoided in favour of utility functions/classes or any other related best practices.

> I'm not sure about which part of the guidelines exactly you are talking. Could you give me a pointer?

https://prometheus.io/docs/instrumenting/writing_exporters/#naming paragraph about failure.

> In different news: Since using a Gauge for timing is so rare, too, why is that in the libraries, although it would be quite easy to code it by hand for the few cases where it matters?

The use case is batch jobs that record their duration on the pushgateway. I'd consider that common in that context, and it'd be incorrect to use a Summary or Histogram there.

beorn7:
> > And I don't see my suggestion at the cost of the majority use case.

> It makes the usage not read as well,

We can improve naming, as said, but structurally, I don't really see how my suggestions read worse. Arguably, the currently recommended way of timing reads best, so we should not provide any tooling.

> and encourages users to mess around with if and which metrics they'll do observations on. I think we should try to avoid making it too easy for users to apply advanced usages inappropriately.

I see partitioned timing as an advanced but important use case. I don't believe we should make life hard for smart users to protect the dumb users. I'm fine with avoiding APIs that _suggest_ to use the advanced way over the basic one, but my suggestion isn't doing that.

> > Adding a method to the metric itself would be problematic in Go terms anyway.

> Do you have a reference for this? None of the obvious search terms are turning up anything relevant. I can only find descriptions of how interfaces work, nothing about how functionality in methods should be avoided in favour of utility functions/classes or any other related best practices.

The problem here is that the timers don't add any unique functionality to the metrics. The timer methods would only interact with the content of the underlying struct via other methods, and the implementation for Histogram and Summary would be exactly the same. Since there is no implementation inheritance and no template methods etc., the implementation needed to be replicated for both types. Which is already a hint that it's not the intended way in Go. By providing a separate `Timer` type that takes an `Observer` implementation at some point, we enable all `Observer` implementations to use the `Timer` without further ado. Would it be a method on the metric type, we would need to change all `Observer` implementation to now also implement the timer method. Practical example: I implement a mock Histogram for testing my code. Now I change my code to use the new timer tooling. With the idiomatic Go approach, all is fine automatically. With a "coerced OOP" approach, I now have to add a method to my mock Histogram implementation.

This kind of utility building is common in the standard lib. Look at https://golang.org/pkg/io/ioutil/#ReadAll . It is not a method of the `io.Reader` interface. It's a separate utility function that will work with any `io.Reader` implementation out there. This is different from asking all `io.Reader` implementers to also implement the `ReadAll` method.

The prime example where functions are just functions that act on a type via its public interface instead of being many many methods on that type is the `strings` package. Any "proper OOP language" would have those as methods on the string type (which Go could even emulate but decided not to do).

> https://prometheus.io/docs/instrumenting/writing_exporters/#naming paragraph about failure.

Thanks. That's talking about counting. We are talking about timing. There is no ratio to calculate.

Timing of errors is important in certain situations (see SRE book as mentioned before, and we had the use case at SC). But it's not even just about errors vs non-errors. If for example a microservice at SC requests the tracks of a user x from a downstream microservice. The downstream service might return "user unknown" or it returns an actual track list, or it returns a "real" error because something is broken. While the real errors are rare, or if they occur, we have a problem anyway, and it's not the latency we are concerned about, it is of great interest to know how long the downstream service needs to find out that a user doesn't exist vs. how long it takes to synthesize an actual tracks list. The upstream service will only be able to distinguish between the two cases after the timer has been started.

> > In different news: Since using a Gauge for timing is so rare, too, why is that in the libraries, although it would be quite easy to code it by hand for the few cases where it matters?

> The use case is batch jobs that record their duration on the pushgateway. I'd consider that common in that context, and it'd be incorrect to use a Summary or Histogram there.

Yes, I agree. I just see the use case of partitioned timing as equally relevant. (Or to be more precise: Partitioned timing by itself might be less relevant, but in turn the effort and complexity of implementing your own solution for it is significant while the timing with a Gauge requires just one additional straight-forward line if you don't have the timer utility we are discussing here. So I would argue it is even _more_ important to provide tooling for partitioned timing than for timing with a gauge.)

beorn7:
@grobie I nominated you before as somebody I would like to consult now and then for `client_golang` questions because you have all of the following:
- Go experience.
- A lot of experience using `client_golang` for both exporters and instrumenting code.
- Experience with the microservice environment at SC.

It would be great if you could contribute your view to this discussion. Perhaps you have a completely different idea (which has happened before), or you can support or debunk arguments made in the discussion.

brian-brazil:
> Arguably, the currently recommended way of timing reads best, so we should not provide any tooling.

The current ways leaves us open to units other than seconds. I think it's worth encouraging best practice behaviour here.

> I'm fine with avoiding APIs that suggest to use the advanced way over the basic one, but my suggestion isn't doing that.

I think your suggestion is doing that, by making the selection of the metric happen after the timer is started. There's nothing to tell the user it's almost always incorrect to not immediately bind to a metric.  In short I believe we always need to bind the metric before the timer starts.

Given what we've seen so far in terms of instrumentation, I'd expect users using this incorrectly to arise at least an order of magnitude often than users using it correctly. 

For most users any form of timing is a new idea, the subtleties of how to best handle timing of optional events and the different approaches of metrics vs. logs to this will be lost of them. (Think of all the users who want to add all possible labels to their instrumented metrics, they'll also probably get this wrong if given a chance).

If someones wishes to not observe a timer, it'd be better to not use defer and to directly call stop/observeDuration on the appropriate code path instead. A smaller scoped timer is possibly more appropriate.

> Since there is no implementation inheritance and no template methods etc., the implementation needed to be replicated for both types. Which is already a hint that it's not the intended way in Go.

It's a principle of Go that a little copying is better than a little dependency. In the same way I don't see an issue with copying a little bit of implementation.

It's my opinion that trying to perfectly factor code to avoid duplication is the cause of some of the worst  abstractions. I know that over in the Java client attempting that has lead to problems, and I've been tempted to switch to copy&paste a few times - and that's for code more involved than what we're discussing here.

> Practical example: I implement a mock Histogram for testing my code. Now I change my code to use the new timer tooling. With the idiomatic Go approach, all is fine automatically. With a "coerced OOP" approach, I now have to add a method to my mock Histogram implementation.

How often is someone actually going to implement a mock Histogram? Unittesting of metrics at that level should be quite rare, so I don't see a problem with making it a little more difficult to benefit the common case.

> Thanks. That's talking about counting. We are talking about timing. There is no ratio to calculate.

It's not as clear in the guidelines as it should be, but for example we removed all metrics of this nature from Prometheus itself.

> The upstream service will only be able to distinguish between the two cases after the timer has been started.

I'm not disputing that the use case exists. I'm arguing that it's so vanishingly rare that it's wrong to base API design decisions on it. In practice you'd probably have the timing down in the track list synthesis function that'd handle this, rather than doing it up in the user API handling function.

> I just see the use case of partitioned timing as equally relevant.

Given I've seen many cases of timing batch jobs, but I can't recall any cases of partitioned timing I must disagree here. Giving the average user a chainsaw to juggle in the form of facilitating partitioned timing is asking for trouble in my opinion, as they will lose a leg.

> the effort and complexity of implementing your own solution for it is significant

It'd be no more difficult than it is now, which you've mentioned it might not even be worth improving for the common case.

stuartnelson3:
The bulk of my comments are dealing with the Go language, since that is where I feel I'm qualified to comment.

>>Since there is no implementation inheritance and no template methods etc., the implementation needed to be replicated for both types. Which is already a hint that it's not the intended way in Go.

>It's a principle of Go that a little copying is better than a little dependency. In the same way I don't see an issue with copying a little bit of implementation.

>It's my opinion that trying to perfectly factor code to avoid duplication is the cause of some of the worst abstractions. I know that over in the Java client attempting that has lead to problems, and I've been tempted to switch to copy&paste a few times - and that's for code more involved than what we're discussing here.

This is a true statement -- a lot of evil has been committed in the name of avoiding duplication. That code usually ends up having switches to shoe-horn several slightly deviating code paths through the same "helper" function. This code is not like that, however. There's a single code path being executed, adding behavior to multiple structs at minimal cost.

The proposed interface code operates as interfaces were designed to be implemented, which I think indicates it is the appropriate solution.

>>Practical example: I implement a mock Histogram for testing my code. Now I change my code to use the new timer tooling. With the idiomatic Go approach, all is fine automatically. With a "coerced OOP" approach, I now have to add a method to my mock Histogram implementation.

>How often is someone actually going to implement a mock Histogram? Unittesting of metrics at that level should be quite rare, so I don't see a problem with making it a little more difficult to benefit the common case.

When a library has as small a surface area as `client_golang`, it's easy to argue against such extensions. By design, `client_golang` is small. While I don't think the unlikely potential of a mock Histogram is the strongest argument for using the interface approach, I don't think it is an argument against it, either. We have pre-existing code that is well extended by Go's preferred method of abstraction. [0]

Regarding partitioned timers and all that -- That's up to you two to discuss. I originally envisioned this as the code currently exists: A timer is started, a metric is bound to it, and at a later that timer is stopped. I defer to wiser monitoring heads for that discussion.

[0] https://www.youtube.com/watch?v=xyDkyFjzFVc

beorn7:
First about the "idiomatic Go" issue:

I foolishly assumed I could explain the reasons for the used Go idioms in a few lines.
Brian's questions are legitimate but if we iteratively answer them until we arrive at first principles, we will have written a book about designing Go code in the end. That's clearly out of scope of a simple PR.

After the discussion so far, I'm even more positive than before that we should not extend the Histogram/Summary/Gauge interface but have a timer type that acts on `Observer` (and `Gauge` in the special case of batch timing). That's my verdict to unstuck this PR without having a general debate about how to write Go code. Happy to git challenged by anybody more qualified in Go than me.

I'll deal with the other topics of the discussion separately.

beorn7:
> > Arguably, the currently recommended way of timing reads best, so we should not provide any tooling.

> The current ways leaves us open to units other than seconds. I think it's worth encouraging best practice behaviour here.

That's a good point. That's indeed adding value to having a timer helper.

> I think your suggestion is doing that, by making the selection of the metric happen after the timer is started. There's nothing to tell the user it's almost always incorrect to not immediately bind to a metric. In short I believe we always need to bind the metric before the timer starts.

That defeats the purpose of making it possible to bind the timer later.

I have pushed a commit to this PR to make my suggestion concrete. We are talking about the difference between the following:

```Go
t := StartTimerWith(his)
defer t.Stop()
```

vs.

```Go
t := StartTimer().With(his)
defer t.Stop()
```

The naive user won't even notice that there is a chance to bind the histogram later. But those that have a valid use case to bind the histogram later can do it easily.

> Given what we've seen so far in terms of instrumentation, I'd expect users using this incorrectly to arise at least an order of magnitude often than users using it correctly.

I doubt that naive users even realize they can bind the histogram later. But even if they do as you expect, I would still be reluctant to punish one of my qualified users for the inability of 10 of my naive users to read the docs.

> If someones wishes to not observe a timer, it'd be better to not use defer and to directly call stop/observeDuration on the appropriate code path instead.

Ironically, exactly the necessity to do that was triggering three separate user requests for a timer helper. In contrast, not a single user requested the simple timer, as users are in general quite satisfied with the way they can implement simple timing with the current library.

By adding the timer with enforced instant binding, we would only help those that don't need help but leave the users that have requested help in the dark. Those users would then implement wrappers (or to be precise, they have already done so), and encouraging the implementation of wrappers is as bad as encouraging wrong usage of the library, as you will most likely agree.

> > The upstream service will only be able to distinguish between the two cases after the timer has been started.

> I'm not disputing that the use case exists. I'm arguing that it's so vanishingly rare that it's wrong to base API design decisions on it. In practice you'd probably have the timing down in the track list synthesis function that'd handle this, rather than doing it up in the user API handling function.

The track list synthesis function is in a downstream microservice. Obviously, instrumenting it there in detail is the right thing to do. However, it is also important to observe latencies as perceived by the frontend.

We could discuss this in all detail, but in the end, we will still end up concluding that in my environtment, I have the use case often enough to justify tooling, even if you haven't seen it that often in your environment.

> Giving the average user a chainsaw to juggle in the form of facilitating partitioned timing is asking for trouble in my opinion, as they will lose a leg.

To stay with your metaphor: Does that mean you want to restrict the sale of chainsaws to certified chainsaw users? Or ban it altogether? Chainsaws are needed. If I don't sell them, then people will built their own that are then ever more dangerous than those provided by expert chainsaw manifacturers. In particular, the home-made chainsaws don't come with instructions how to use them safely.

> > the effort and complexity of implementing your own solution for it is significant

> It'd be no more difficult than it is now, which you've mentioned it might not even be worth improving for the common case.

It would be great if you could try to make sense of my words instead of trying to twist them.

As explained in more detail above, the common case is already easy with the current state of the library. The less common but valid case of partitioned timing is complex enough to warrant the creation of a timer helper. What I suggest is a timer helper that is obvious to use correctly for the common case but can still be used for the partitioned case.,

brian-brazil:
> The naive user won't even notice that there is a chance to bind the histogram later. But those that have a valid use case to bind the histogram later can do it easily.

That's not my expectation. I would expect the type of programmer that tends to be too smart in their code to notice it, and use it. I'd expect this to happen far more often than a valid use case.

Take the `ConstLabels` in the Go client. There is a rare valid use case for them, for which we haven't yet resolved how to handle across clients - I think we're at 3 different approaches thus far with no clear winner yet.

Aside from the rare valid use case, we've also much more often seen them abused to add version labels and the like to all metrics of a library. We've also seen them misused to share a metric across libraries.

Given that, it is unclear that `ConstLabels` overall is a good idea and at this stage I'd be tending towards not. And that's allowing for the fact that without `ConstLabels`, implementing the valid use case is a fair bit more annoying than the sort of thing we're discussing here.

I see this as similar. Even if if there's a rare valid use case, the potential for misuse outweighs the benefit.

> I would still be reluctant to punish one of my qualified users for the inability of 10 of my naive users to read the docs.

I don't have a problem there. In practice it'd be 1 extra line of code for the advanced user (which they need anyway), which with my suggestion wouldn't require reading the docs as it's obvious from the API examples.

> By adding the timer with enforced instant binding, we would only help those that don't need help but leave the users that have requested help in the dark. 

I'd disagree that we're leaving them in the dark, what they want to do is perfectly possible we're just not being overly accommodating for a rare use case that's overall of dubious utility.

> Those users would then implement wrappers (or to be precise, they have already done so), and encouraging the implementation of wrappers is as bad as encouraging wrong usage of the library, as you will most likely agree.

Users have really already created wrappers to save them a line of code for a rare use case? I'd say such users are too far gone to worry about, as no matter what they'll likely end up feeling the need to write a wrapper anyway.

> To stay with your metaphor: Does that mean you want to restrict the sale of chainsaws to certified chainsaw users? Or ban it altogether? Chainsaws are needed. If I don't sell them, then people will built their own that are then ever more dangerous than those provided by expert chainsaw manifacturers. In particular, the home-made chainsaws don't come with instructions how to use them safely.

In this case we leave everyone with a knife. They can make their own chainsaw out of it - but we don't need to hand everyone a chainsaw just because a tiny proportion of people are asking for one.

> As explained in more detail above, the common case is already easy with the current state of the library.

But doesn't encourage seconds, otherwise we'd be okay. If we were talking Python where `time.time` is already what people would use and in seconds (ignoring the possibility of decorators/context managers) then we might get away with not having a helper. That's not the case in Go though.

> The less common but valid case of partitioned timing is complex enough to warrant the creation of a timer helper. What I suggest is a timer helper that is obvious to use correctly for the common case but can still be used for the partitioned case.

I don't think it's that complex. Are you really saying that (ignoring spelling):

```
mytimer1 := summary1.starttimer()
mytimer2 := summary2.starttimer()

if () {
  mytimer1.stop()
} else {
  mytimer1.stop()
}
```

is so complicated that we must offer:

```
mytimer := starttimer()
defer mytimer.stop()

if () {
  mytimer.bind(summary1)
} else {
  mytimer.bind(summary2)
}
```

Looking at it now in pseudo-code, it's not even saving a line of code - it's exactly as complicated syntax wise.

Even if we presume this case is worth adding code for, where are you seeing a saving here?

grobie:
Excuse my late comment, I'm traveling now.

> I don't think it's that complex. Are you really saying that (ignoring spelling):
> 
> ```
> mytimer1 := summary1.starttimer()
> mytimer2 := summary2.starttimer()
> 
> if () {
>   mytimer1.stop()
> } else {
>   mytimer1.stop()
> }
> ```
>
> is so complicated that we must offer:
>
> ``` 
> mytimer := starttimer()
> defer mytimer.stop()
> 
> if () {
>   mytimer.bind(summary1)
> } else {
>   mytimer.bind(summary2)
> }
> ```
>
> Looking at it now in pseudo-code, it's not even saving a line of code - it's exactly as complicated syntax wise.

These examples are not the same. The bind example allows for multiple return paths using a single deferred `stop()` call, while your suggestion will require an explicit `stop()` call in every return path. Having more than one return path is very common in Go, due to all the explicit error handling and `if err != nil { return err }`.

My thoughts on the topic:
- I've always found attaching `set_to_current_time` and `start_timer` methods to metric objects confusing and unexpected. As an API user I'd expect such methods not to be coupled to the metric. As these methods don't share any functionality with the actual metric functions, they lower the cohesion and bloat the interface and responsibility. I see that it might be more convenient due to saving some characters in comparison to calling equivalent library functions, but I'm worried more unrelated functions will be attached out of convenience. This does not only apply to Go, but also more OOP languages.
- On the other side, I've personally also not seen enough use cases for the late binding use case. I agree that designing the API for rare advanced use cases makes it harder for the average user and less clear for beginners. If there enough known user requests for the advanced use case it might be fine to trade readability and beginner experience, but we should be sure it's worth adding the less comprehensible `With(...)` mutator.
- Stop doesn't actually describe the function adequately, as nothing gets stopped but only recorded. For the simple case there is also no need to store a reference to the observer. I guess I'd have done something like `defer prometheus.StartTimer().Record(reqDuration)`, but this would make the advanced use case more complicated.

prometheus/timer.go
```diff
@@ -0,0 +1,51 @@
...
+func (t *Timer) WithGauge(g Gauge) *Timer {
+	t.gauge = g
+	return t
+}
+
+func (t *Timer) Stop() {
+	if t.observer != nil {
+		t.observer.Observe(time.Since(t.begin).Seconds())
+	}
+	if t.gauge != nil {
```
grobie: I don't find this a very clear implementation, the timer shouldn't need to know about gauges, do multiple availability checks or allow to set both a gauge and an `Observer` (with even slightly different timings). What about this instead:

```go
func (t *Timer) WithGauge(g Gauge) *Timer {
  return t.With(ObserveFunc(g.Set))
}

func (t *Timer) Stop() {
  if t.observer != nil {
    t.observer.Observe(time.Since(t.begin).Seconds())
  }
}
```

prometheus/timer.go
```diff
@@ -0,0 +1,51 @@
...
+	Observe(float64)
+}
+
+type Timer struct {
+	begin    time.Time
+	observer Observer
+	gauge    Gauge
+}
+
+func StartTimer() *Timer {
```
grobie: Most exported functions miss the docs comment.

brian-brazil:
> These examples are not the same. 

Do we have a real world example somewhere so we can stop talking past each other in theoreticals?

> This does not only apply to Go, but also more OOP languages.

That's what I personally consider to be "Java style" programming, where there's so many adaptors, interfaces and abstractions that the most practical way to figure out how to plumb A to B is to find the answer on stackoverflow.

I want to have interfaces that are usable in the real world from a quick peek at the docs, over interfaces that are philosophically pure but where it takes at least 10 minutes of pouring over the docs to do anything.

We're already in a state where every time I want to do anything non-trivial with the Go client I have to read large chunks of the source code, and I'd like it not to continue down that path.

> I'm worried more unrelated functions will be attached out of convenience.

I see no problems with adding more if they come up and are common enough to put in the core, rather than relegated to utility functions. Considering Go in progress requests is likely the only other one that'd make sense to add, as I don't think there's anything sane we can do to help in counting err results.

prometheus/timer.go
beorn7: Certainly. This is proof-of-concept code. Once we have decided to go down that alley, I'll add doc comments and proper tests.

prometheus/timer.go
beorn7: Sure, we can do that. Good idea. I added a commit.

I just wouldn't make ObserverFunc exported yet. (If use cases arrive once we leverage the Observer interface more, e.g. for testing, we can still export it.)

beorn7:
Thanks, @grobie , for chiming in.

I'll try to bundle the discussion a bit. The ramifications are getting out of control otherwise.

### Extending the metric interface

@grobie is in support of not extending the interface of Histogram/Summary/Gauge. We could endlessly discuss the pros and cons of that, but as we have seen, it (d)evolves into a fundamental discussion about the design of programs or even programming languages. While we are in a position to redefine how monitoring is done, we are in no position to redefine how to write Go libraries. Unless a Go guru steps in and tells us how to do it differently, let's close this branch of the discussion.

About the "in flight" gauge: That's so straight-forward and idiomatic in Go that I don't see us providing any helper code for it:
```Go
func doSomething() {
    inFlight.Inc()
    defer inFlight.Dec()
}
```

### Vetting potential of abuse vs. enabling advanced use cases

The discussion is not really moving forward here. Pulling in the const-labels is not helping. It's just another example where I'm gaining more and more confidence it is the right way to go while @brian-brazil seems to gain more confidence it's wrong.

I feel my arguments being misunderstood. Then I clarify my argument and feel my clarification misunderstood. Quite possible @brian-brazil feels the same.

I'm tempted to clarify my clarifications, based on @brian-brazil 's latest reply, but then I feel I'm just saying what I said before, and also @brian-brazil has essentially only said what he said before. Expecting to not be misunderstood again might be foolish, and perhaps we just have to conclude that all arguments have been made and we can only agree to disagree.

### Importance and exact nature of the use case

@brian-brazil asked: "Do we have a real world example somewhere so we can stop talking past each other in theoreticals?"

That would be helpful, indeed. One of my three use cases was in a discussion at a Go convention. I don't even remember the details, just that I found it reasonable and that it made me note the request. The other two are internal at SC, so I cannot just copy&paste the code. Furthermore, it needed some explanation of context. Even if I invested the work for that and for sanitizing the code to make it publishable, I'm pretty sure that the best possible outcome would be: "OK, we have two legitimate use cases. But that's only two special cases, not enough to justify the feature." I tried to sketch out the use case here, and the way it was discussed showed me that it wouldn't convince you.

So in turn, I would like to ask: What _would_ be an exit criterion to convince you? My time is too precious for preparing elaborate evidence that will in the end just clarify why and how we disagree.

We could just put this on hold and wait for user requests to come in. But how many would we need?

My line of argument is still that the mental overhead for the naive user is to add `().` is very low, see my example above. Thus, in my view, the bar to justify the feature is quite low.

### "Lines of code saved"

@brian-brazil , thanks for the cod example, which I quote here again for reference:

```Go
mytimer1 := summary1.starttimer()
mytimer2 := summary2.starttimer()

if () {
  mytimer1.stop()
} else {
  mytimer1.stop()
}
```

This example only works for cases where both of the following are true:
1. There are only very few possible outcomes.
2. There is only one return path for each outcome.

While 1. might be true in the majority of cases, even with only two or three possible outcomes, I can already see how many Go programmers will see it as bad design in terms of maintainability (what to do if I need to add a now outcome) as well as efficiency (functions to be timed are often in the hot path).

Condition 2. is an even bigger problem because in Go it is very common to have many return paths out of a function, in particular in the case of errors, as @grobie said, but not restricted to that. Go programmers are also trained to use `defer` for peace of mind. There would be great reluctance to ask them to pay attention to each possible outcome, especially given that the outcome might be a panic.

So counter example:

```Go
func queryBackend() (Result, error) {
	t := prometheus.StartTimer().With(errTiming)
	defer t.Stop()

	response, err := someQueryFromBackend()
	if err != nil {
		return nil, err
	}
	if err := sanityCheck(response); err != nil {
		return nil, err
	}

	t.With(tierTiming.With(prometheus.Labels{"customer_tier", response.CustomerTier}))
	if IsBlacklisted(response.Location) {
		return response.RestrictedResult, nil
	}
	return response.FullResult, nil
}
```

vs. 

```Go
func queryBackend() (Result, error) {
	errT := prometheus.StartTimerWith(errTiming)
	freeT := prometheus.StartTimerWith(tierTiming.With(prometheus.Labels{"customer_tier", "free"}))
	standardT := prometheus.StartTimerWith(tierTiming.With(prometheus.Labels{"customer_tier", "standard"}))
	premiumT := prometheus.StartTimerWith(tierTiming.With(prometheus.Labels{"customer_tier", "premium"}))

	response, err := someQueryFromBackend() // Note that a panic here would not be timed at all.
	if err != nil {
		errT.Stop()
		return nil, err
	}
	if err := sanityCheck(response); err != nil {
		errT.Stop()
		return nil, err
	}

	var result Result // Unidiomatic ndirection required just for timing.
	if IsBlacklisted(response.Location) {
		result = response.RestrictedResult
	} else {
		result = response.FullResult
	}
	switch response.CustomerTier {
	case "free":
		freeT.Stop()
	case "standard":
		standardT.Stop()
	case "premium":
		premiumT.Stop()
	default:
		// Oops, new tier got introduced. Need to change code now...
	}
	return result, nil
}
```

I hope that shows how we are not talking about "saving one line of code".


### Do we need a timer helper for the simple case only?

The point with coercing users into using seconds is a good one, but I'm on the fence if that alone is really worth adding the timer helper.

What turns me against it is that introducing a timer just for the simple case would make it a breaking change to later change the timer helper to supporting late binding of the Histogram. In case we were confident we won't support the late-binding case, I would be willing to consider a simple-case-only timer. But right now it looks more like we will be waiting for more evidence from users.

### Next steps?

To bring it down to one question that needs to be solved right now:

Do you need more legitimate requests from users (how many approx.?), or are you convinced already that a timer helper with a metric binding step separate from starting and stopping is necessary?

I take it that @brian-brazil is obviously not convinced at all and is strong on the "need more evidence" side. I'm equally strong on the "already convinced" side, and @grobie tends towards the former. I don't necessarily propose a formal vote on the subject, but input from others would be welcome.

beorn7:
Forgot about the naming issue @grobie mentioned:

> Stop doesn't actually describe the function adequately, as nothing gets stopped but only recorded. For the simple case there is also no need to store a reference to the observer. I guess I'd have done something like defer prometheus.StartTimer().Record(reqDuration), but this would make the advanced use case more complicated.

I thought the start/stop pair is most intuitive, even if (as an implementation detail) nothing is really started or stopped. I don't feel strongly about the naming. I don't see any problem to bring it in line with the Java client, which would mean `ObserveDuration` instead of `Stop`. As said, I find `Stop` more intuitive to understand, especially given that I'm using `StartTimer` as the name of the constructor (which is in line with the Histogram method in the Java client). But consistency his its merit. So how do you like `ObserveDuration`?

And yeah, as you said, with the histogram as the argument of Stop/Record/ObserverDuration, the problem at hand is not really solved. (I went down that path when I filed the issue but soon noticed that...)

brian-brazil:
> About the "in flight" gauge: That's so straight-forward and idiomatic in Go that I don't see us providing any helper code for it:

Ah yes, had forgotten about defer there. Just that example in the docs will cover that then.

> I hope that shows how we are not talking about "saving one line of code".

I see your point. You'd be looking at doing the timing by hand there with my proposal.

One thing that occurs to me is that Python and Java are already limited in this way, and I've had zero requests to allow for it or make it easier

> Do you need more legitimate requests from users (how many approx.?), or are you convinced already that a timer helper with a metric binding step separate from starting and stopping is necessary?

I'd like to see at the least more potential cases of correct use than mis-use. 

So far it sounds like you have 2 internal SC cases plus one probable other. For mis-uses we've more than that just in the code history of Prometheus itself where we used to have 4-5 of these, plus probably another 5-10 I recall coming across in various code in passing (we still have one in the node exporter).

I get the impression that the mis-use is already moderately common, and am thus wary of adding things that would further enable it.

> What turns me against it is that introducing a timer just for the simple case would make it a breaking change to later change the timer helper to supporting late binding of the Histogram. 

As one option the code as proposed here could always be a 3rd party helper with nice big warnings, with just the simpler case handled in the core. In that case we could also drop the gauge support in the helper, as I have great difficulty seeing late binding coming up in batch jobs.

That keeps things simple for the simple case, while still allowing for the advanced case.

I'd be against putting even something as relatively common as counting bytes sent/received in http processing in the core metrics for the sake of bloat. I don't think it'd be inappropriate to have something as rare as late binding kept well outside the core, particularly given the potential for mis-use. I'm okay with some bloat for basic timing, as it's so common.

> given that I'm using StartTimer as the name of the constructor (which is in line with the Histogram method in the Java client).

I think there's a strong benefit to having the word Timer somewhere in this API, as we regularly get users looking for the "Timer metric".

> But consistency his its merit.

As a historical note, I choose `ObserveDuration` in order to be consistent with the Go client's `Observe`. Similarly with `SetToCurrentTime`.

beorn7:
Giving the timer a test run in a separate package is a possibility. (Perhaps we need a "restricted section of the library"... ;)

I still think the timer should be in the core lib as per my suggestion, so I wouldn't like to introduce a simple timer in the core lib before I am convinced otherwise, with the stated argument that switching the simple timer to the advanced timer would be a breaking change.

I'll add doc comments and such and run the code by people like Sean T. or Peter B. to see if that results in conclusive evidence.

I'll also change the naming to `NewTimer` and `ObserveDuration`. Please let me know if you like the naming better or not.

prometheus/timer.go
```diff
@@ -0,0 +1,99 @@
...
+//
+// With can also be called later, to bind an Observer for the first time or to
+// change the bound Observer. It can be called with nil to unbind a previously
+// bound Observer. Both is useful for recording with different Observers (or
+// none) depending on the outcome of the timed function. Note that timing
+// depending on the outcome can be confusing and should only be used after
+// careful consideration.
+//
+// The ObserveDuration method of the Timer records the duration by calling the
+// Observe method of the bound Observer with the time in seconds as the
```
stuartnelson3: Maybe clarify that the argument is the elapsed time?
```
... with the time elapsed since timer initialization in seconds ...
```

beorn7: Updated. I used a slightly different wording for consistency with other parts of the doc comments.

stuartnelson3:
Looks good to me

beorn7:
Still collecting feedback from Sean & Peter (which, unsurprisingly, is creating even more ramifications…).

In the meantime, question for @brian-brazil to sharpen my understanding for something I vaguely referred to before. On page 60 of the SRE book, the following is written about _latency_ as the first of the "four golden signals":

> It’s important to distinguish between the latency of successful requests and the latency of failed requests. For example, an HTTP 500 error triggered due to loss of connection to a database or other critical backend might be served very quickly; however, as an HTTP 500 error indicates a failed request, factoring 500s into your overall latency might result in misleading calculations. On the other hand, a slow error is even worse than a fast error!. Therefore, it’s important to track error latency, as opposed to just filtering out errors.

If a user of Prometheus reads that and wants to implement it with Prometheus, they are certainly lead to believe they should use a Histogram partitioned by outcome, at least "success" vs. "failure", or even more, e.g. for HTTP request, they might want to separate invalid requests by the clients (4xx) from server-side errors (5xx) because there will be very different latency profiles, and while a lot of 5xx is a problem anyway, many 4xx might be just fine, depending on circumstances. So what would should we tell that user? Is the above bad advice? Or is there a way to take the above into account without regularly partitioning Histograms?

brian-brazil:
> So what would should we tell that user? Is the above bad advice? 

It's good advice if you are an expert in debugging this sort of system, and are aware of the gotchas with this approach. I would expect fleetingly few people to be that level of expert.

For non-experts person this is not good advice, as they tend to run headlong into the gotchas.

> Therefore, it’s important to track error latency, as opposed to just filtering out errors.

This is the key point - don't filter out errors. Users in their dashboards with this approach tend to only graph success. If it's not split by outcome, users can't make this very common mistake,

Then there's more theory here around the three classes of errors (user, system, unclear), the 4xx vs 5xx you mention is heading in that direction. This is all well beyond most users.

> Or is there a way to take the above into account without regularly partitioning Histograms?

There's a few ways you could do it, but it all boils down to roughly the same thing in the end.

As a separate note, adding additional dimensions to a Histogram should be done sparingly as they already have a moderate cardinality.

beorn7:
> > Therefore, it’s important to track error latency, as opposed to just filtering out errors.

> This is the key point - don't filter out errors. Users in their dashboards with this approach tend to only graph success. If it's not split by outcome, users can't make this very common mistake,

But they will then factor in all the errors without any means of separation. The other point in the original quote is: "For example, an HTTP 500 error triggered due to loss of connection to a database or other critical backend might be served very quickly; however, as an HTTP 500 error indicates a failed request, factoring 500s into your overall latency might result in misleading calculations."

To paraphrase, you are saying: "Include errors in the timing. Only if you are one of the few experts, start to think beyond that and time errors separately." The book says: "Including errors in the timing is problematic. But you have to time errors, too. So you have to time them separately, even if you are not an expert."

> Then there's more theory here around the three classes of errors (user, system, unclear), the 4xx vs 5xx you mention is heading in that direction. This is all well beyond most users.

My experience is that reality doesn't really care if is well beyond most users. We had outages that were masked by a large amount of quickly responded 404s. And we had outages with a lot of timed-out requests, where it would have been nice to find out if the successful requests have normal timing (i.e. the timed-out ones are hitting a bad backend, the others are hitting a healthy backend) or if just all requests were close to the timeout in response time, and just a certain fractions happened to cross the threshold.

In those scenarios, the "need an expert" issue was inverted: An expert would have been able to find those issues, even with a non-partitioned histogram. But that was well beyond most users. Most users fell into the trap and thought the site is sufficiently fast in the first case and weren't able to investigate the second case.

I'm gaining confidence that partitioned timing is needed, _especially_ because not everybody is an expert. Restricting ourselves to telling people "this is too dangerous for you, don't do it" appears insufficient to solve the users' problems. And I'm backed by "the book" in this… ;)

> As a separate note, adding additional dimensions to a Histogram should be done sparingly as they already have a moderate cardinality.

Yes, acknowledged. Which tells me even more that we need to address the issue in documented best practices. If we don't give any tooling or advice about how and when to partition properly, people will just partition by status code and let they time series count explode.

beorn7:
Short report from my interviews with Peter and Sean (email and, in the latter case, also in person):

As already hinted at, the answers were not simply yes or no, but added more facets to the problem.

The easy part is that nobody really wanted to add methods to the metric types themselves. Having a separate `Timer` type that acts on an `Observer` (or slight variations thereof) is indeed the way to go. They also showed strong affinity to `defer`.

While both thought partitioned timing is normal and desired, their concerns and use cases were different from mine. Peter hasn't really had the "partition by outcome" case yet. He had a "time in stages" case, which needed separate discussions to vet. Sean was closer with the use case but unhappy that the "late binding" happens somewhere in the function code and cannot be understood by looking at one section of code.

The way they solved their use cases so far gave me some inspiration how we might do it in our provided tooling. I'll update the PR later today with a suggestion.

About partitioned timing as a "normal" need: They might both be mistaken, of course. The SRE book might be mistaken with the general advice to time errors separately. My own judgement as described in the previous post might be mislead. To find "the right thing to do" by a survey is not necessarily a good strategy. But with so many more or less qualified opinions, we needed a really convincing explanation to not make them do it anyway.

beorn7:
OK, new suggestion is pushed. @grobie @stuartnelson3 @brian-brazil please have a look.

brian-brazil:
> Most users fell into the trap and thought the site is sufficiently fast in the first case and weren't able to investigate the second case.

In the second case you need distributed tracing. Trying to slice and dice the data in every possible way ultimately causes cognitive overload and users are going to miss something - too much detail can be worse than too little. When you get to something like this you're fairly deep into debugging, and hopefully a mix of looking at stats for all backends and using profiling/logs will let you figure out what's going on. 

> I'm gaining confidence that partitioned timing is needed, especially because not everybody is an expert. Restricting ourselves to telling people "this is too dangerous for you, don't do it" appears insufficient to solve the users' problems. 

I don't agree with your logic here. If a non-expert can't solve a complex problem no matter what information we give them, throwing them additional information is only going to increase the risk they'll form an incorrect hypothesis for the simpler problems.

Experts in this space are extremely rare, so I don't think we should be optimising for a group that for all practical purposes don't exist.

> If we don't give any tooling or advice about how and when to partition properly, people will just partition by status code and let their time series count explode.

This is the reason I generally encourage quantile-less summaries, much cheaper. It's not practical to do the sort of things you're talking about here with histograms.

> He had a "time in stages" case, which needed separate discussions to vet.

That's multiple quantile-less summaries for each stage with a timer each. Do not use a histogram for this (how to take out your Prometheus in one easy step). This shouldn't require anything additional from the API.

prometheus/example_timer_batch_test.go
```diff
@@ -0,0 +1,48 @@
...
+	batchDuration = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "example_batch_duration_seconds",
+		Help: "Duration of the last batch run.",
+	})
+)
+
+func performBatch() {
+	// The Set method of the Gauge is used to observe the duration.
+	timer := prometheus.NewTimer(prometheus.ObserverFunc(batchDuration.Set))
+	defer timer.ObserveDuration()
```
brian-brazil: The spelling here isn't correct, this is a Set - not an Observe.

prometheus/example_timer_batch_test.go
```diff
@@ -0,0 +1,48 @@
...
+	// See https://prometheus.io/docs/practices/instrumentation/#batch-jobs
+	batchDuration = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "example_batch_duration_seconds",
+		Help: "Duration of the last batch run.",
+	})
+)
+
+func performBatch() {
+	// The Set method of the Gauge is used to observe the duration.
+	timer := prometheus.NewTimer(prometheus.ObserverFunc(batchDuration.Set))
```
brian-brazil: I know this is boilerplate, but it's still a bit wordy for my liking.

Also what other functions might someone pass in here?

prometheus/example_timer_batch_test.go
```diff
@@ -0,0 +1,48 @@
...
+	defer timer.ObserveDuration()
+
+	// Actually perform the batch of work.
+	// ...
+}
+
+func ExampleTimer_batch() {
```
brian-brazil: The example should be for the pushgateway.

For this sort of code you'd be using a summary/historgram - not a gauge.

prometheus/example_timer_complex_test.go
```diff
@@ -0,0 +1,69 @@
...
+	)
+)
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	status := http.StatusOK
+	// The ObserverFunc gets called by the deferred ObserveDuration and
+	// decides wich Histogram's Observe method is called.
+	timer := prometheus.NewTimer(prometheus.ObserverFunc(func(v float64) {
+		switch {
+		case status >= http.StatusInternalServerError:
```
brian-brazil: This would be clearer with the numbers rather than the constants

prometheus/example_timer_complex_test.go
```diff
@@ -0,0 +1,69 @@
...
+		},
+		[]string{"status_class"},
+	)
+)
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	status := http.StatusOK
+	// The ObserverFunc gets called by the deferred ObserveDuration and
+	// decides wich Histogram's Observe method is called.
+	timer := prometheus.NewTimer(prometheus.ObserverFunc(func(v float64) {
```
brian-brazil: This example looks to me like this interface is wildly too generic.

prometheus/example_timer_test.go
```diff
@@ -0,0 +1,40 @@
...
+		Help:    "Histogram for the runtime of a simple example function.",
+		Buckets: prometheus.LinearBuckets(0.01, 0.01, 10),
+	})
+)
+
+func ExampleTimer() {
+	// timer times this example function. It uses a Histogram, but a Summary
+	// would also work, as both implement Observer. Check out
+	// https://prometheus.io/docs/practices/histograms/ for differences.
+	timer := prometheus.NewTimer(requestDuration)
```
brian-brazil: The simple case looks okay.

prometheus/example_timer_batch_test.go
beorn7: The "Example (Complex)" (as it shows up in Godoc) is an example for other functions passed in.

While a bit more verbose, this is good Go code. We are making it explicit that we are using the `Set` method to observe a duration, which is in general not used for that.

prometheus/example_timer_batch_test.go
beorn7: I can change it into a PGW example. However, we are using a Gauge for checkpoint duration in Prometheus itself, and that always appeared very useful to me.

Just yesterday, I ran into a user at SC who used a histogram in a case similar to Prometheus checkpoints, and in that situation, a Gauge would have been much easier to reason with, too.

If something should be timed that happens much less frequent than a scrape, a Gauge as timer should be considered. I would like to see that in the best practices (where it currecntly says, there are only two valid use cases for a Gauge for timing, non-trivial Collectors and "true" batch jobs).

However, to not open up another front of the discussion, I'll change the example to PGW, and we can discuss the Gauge for timing issue later..

prometheus/example_timer_complex_test.go
beorn7: It's covering the use case I want to cover. I'm open for improvements, but I'm not open for not supporting the use case, see general comment I'll write in a sec.

prometheus/example_timer_complex_test.go
beorn7: Makes sense, will change.

beorn7:
About the debate around partitioned timing:

I think we are at a point where all cards are more or loss on the table. I find myself once more in a situation where Brian's arguments in his last post only add a few more details to the discussion over what was already said, while the response I am tempted to write now would likewise add very little over what I have already said before.

After the debate so far, I'm more confident than ever that I want to support the use case of partitioned timing. I'm backed by the SRE book, by experienced Go coders that deal a lot with metrics, by my users here at SC, by my own reasoning and experience, and by the maintainer of the library (not surprising, as that's myself). I appreciate Brian's experience and judgement, but in this case, I cannot follow. In an ideal world, I would enjoy continuing the discussion into any level of detail. I wouldn't even rule out the possibility that one side might convince the other, or that we come up with a completely different solution. However, in real life, our time is limited, and at this point, I would like to draw a line.

The solution we have arrived at now has zero impact on the simple case, it addresses concerns I gathered from experienced Go programmers, and it quite clearly doesn't suggests to naive users that they should go down the path of the advanced use case. I'm happy to improve implementation details but supporting partitioned timing is important enough that, at this point, it would require more than just Brian's resistance to make me drop it.

prometheus/example_timer_batch_test.go
brian-brazil: > Handling those two things in one Timer helper will be a nice gain, and would totally justify to have a dedicated Gauge-Timer (perhaps call it a BatchTimer or something).

The other common case for the current time is recording when you last did something (e.g. processed a record from a queue) to detect when processing has broken. I don't think I'd mix that in with batch jobs, I can see it getting messy when a small bit of extra client code would do the same task in a clearer way.

> So would you suggest two separate metrics for success and failure timestamp?

I'd have a completed, and a success.

beorn7: Thanks. I guess it would be good to add a complete example (in the `push` package) how to instrument a batch job (with run duration, and timestamp for last run and last success). That's pretty involved, and also required the `push.AddCollectors` approach as we don't want to delete the last success from a previous run if the current run has failed.

This all makes me think we should consider a helper for that. This would be a different issue, but it would make it way less common to use a Gauge for timing (provided that the one most common use case is batch jobs).

brian-brazil: > and also required the push.AddCollectors approach as we don't want to delete the last success from a previous run if the current run has failed.

The way I've been approaching that is to only add the success metric to the custom registry at the end.

> it would make it way less common to use a Gauge for timing (provided that the one most common use case is batch jobs).

The other case is exporters/custom collectors, e.g. snmp_scrape_walk_duration_seconds. There's ConstMetrics involved there though, so that'll continue to be done by hand I'd expect.

beorn7: > The way I've been approaching that is to only add the success metric to the custom registry at the end.

That still requires the PushAdd approach as the normal Push will delete all previous metrics from the same group.

For SetToCurrentTime, see https://github.com/prometheus/client_golang/pull/259

beorn7:
OK, tweaked doc comments and examples again, with the following rationale:

Meanwhile, I have added a `SetToCurrentTime()` method to the `Gauge` interface and pushed (no pun intended) a comprehensive example of how to instrument a batch job to the `push` package. See https://godoc.org/github.com/prometheus/client_golang/prometheus/push#example-AddFromGatherer . With the new `SetToCurrentTime()` method, the example reads quite nicely, so that I don't think we urgently need any dedicated helpers for that case. Ironically, the timer helper is not even used there. If you think about it, it's not so surprising, because the whole "time a function" thing only really makes sense for a function that is called from multiple and/or external locations.

As the use case of timing the duration a `Collector` runs is catered to with ConstMetrics, the use case of using a Gauge with the timer helper proves to be quite rare. I have still kept the example, mostly to mention there what it can be used for if at all, but I don't think we should provide a specific helper for timing with Gauges at this point. I have removed the surrounding part of actually pushing to a PGW because the example was incomplete anyway. I have included a pointer to the new PGW example, which should guide users into safety.

@stuartnelson3 @grobie What do you think about the current state here? Are you able to serve as a tiebreaker for the remaining controversy?

prometheus/example_timer_test.go
grobie: I'd actually expect the simple case example to be a one-liner.

grobie:
Short answer as I'm pretty tired and your discussion was long to read.

From a go point of view I find the current interface clear and natural, albeit it's indeed a bit wordy in the gauge case. In applications (not exporters) I've never had use cases for gauge timers though.

When it comes to metrics and whether partitioned histograms should be possible, I can say that I've used the additional information of error vs. success quantiles to a great deal during outages. Not partitioning histograms by default / the simple case sounds good to me, but I'm very skeptical of removing any support for them.

prometheus/example_timer_test.go
beorn7: You _can_ write it as a one-liner:
```Go
defer prometheus.NewTimer(requestDuration).ObserveDuration()
```

This kind of timer pattern is pretty common in Go, and what I have seen multiple times in Go conference talks is the recommendation to write it in two lines for clarity. That's why I did the examples that way. You can still write it in one line if you like.

Another possible simplification is to return a naked function to call for the observation:
```Go
observeDuration := prometheus.NewTimer(requestDuration)
defer observeDuration()
```

Or in one line (quite obfuscated):
```Go
defer prometheus.NewTimer(requestDuration)()
```

To avoid the latter way, you can just tell people to give the returned function a nicely named variable as in the 2nd to last example.

I decided to go the way with an explicit method because it is in better agreement with the client library and makes the one-liner better to read. My impression is that this is in the sweet spot where we are in maximum agreement with most of the community and the Prometheus client library guidelines.

beorn7:
@grobie Thanks for your feedback. Is that a :+1: then? I'd rather like some green light than merging this by decree.

brian-brazil:
I think the explicit function is best.

prometheus/example_timer_test.go
grobie: Oh, my comment was just meant as "given you can write it as a one-line, why did you format it differently". If that's more common for clarity reasons, that's fine.

grobie:
:+1:

beorn7:
Acknowledging that there are remaining concerns, I think the current state is as far as we can get under given circumstances.

PR #254, opened 2016-11-04
Change `*_microseconds` -> `*_seconds`

iamjarvo:
Docs use seconds instead of microseconds. Figured I should change both metrics that use microseconds.

@beorn7

### DIFF
--- a/examples/random/main.go
+++ b/examples/random/main.go
@@ -41,7 +41,7 @@ var (
 	// differentiated via a "service" label.
 	rpcDurations = prometheus.NewSummaryVec(
 		prometheus.SummaryOpts{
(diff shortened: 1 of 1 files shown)


beorn7:
@iamjarvo thanks for the PR. As @grobie said, it's not enough to just rename the metric name. All the other numbers need to be scaled, too. That affects the buckets, but also the default value of the command line parameters (uniform.domain, normal.domain, normal.mean).

iamjarvo:
No problem. I can follow up with the changes in the coming days

beorn7:
To get this off the table, I'll resolve this in a separate PR.

beorn7:
See https://github.com/prometheus/client_golang/pull/257

PR #255, opened 2016-11-06
desc: Allow initial colon in metric name

kamalmarhubi:
Match the documented regex:
    https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

The Java and Python first-party libraries allow the initial colon:
- java: https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/Collector.java#L142
- python: https://github.com/prometheus/client_python/blob/master/prometheus_client/core.py#L26

The Ruby library requires that the name be a symbol, which is less restrictive:
  https://github.com/prometheus/client_ruby/blob/master/lib/prometheus/client/metric.rb#L49

### DIFF
--- a/prometheus/desc.go
+++ b/prometheus/desc.go
@@ -26,7 +26,7 @@ import (
 )
 
 var (
-	metricNameRE = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_:]*$`)
+	metricNameRE = regexp.MustCompile(`^[a-zA-Z_:][a-zA-Z0-9_:]*$`)
 	labelNameRE  = regexp.MustCompile("^[a-zA-Z_][a-zA-Z0-9_]*$")
 )
 


brian-brazil:
Can you add a unittest please?

Note that direct instrumentation should never have colons in it, and having a colon first is not in line with any naming practice.

beorn7:
There is something fishy here.
https://github.com/prometheus/common/blob/master/model/metric.go#L25 does not allow leading colons.
In https://github.com/prometheus/common/blob/master/model/metric.go#L93, which is the validation we actually use, it is allowed.

Before we patch all inconsistent places into allowing leading colons, can we first clarify what the original intention was and then fix everything the right way?

brian-brazil:
My expectation is that colons are allowed anywhere in a metric name, it's only digits that we restrict to not be first for parsing reasons.

I wouldn't be against prohibiting colons completely in direct instrumentation, the rare cases where it's okay to have them would very likely come from custom collectors.

kamalmarhubi:
@brian-brazil 
> Note that direct instrumentation should never have colons in it, and having a colon first is not in line with any naming practice.

> I wouldn't be against prohibiting colons completely in direct instrumentation, the rare cases where it's okay to have them would very likely come from custom collectors.

This should probably be reflected [in the docs](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels). Since I'm not super acquainted, could you or @beorn7 add something there to clarify and maybe give an example where colons *are* used?

If it's all right, I'll hold off the unit test / getting this merged till after that.

beorn7:
OK, this all sounds like the following course of action:
- Allow colon as the 1st character (i.e. declare the current docs and the _actual_ behavior as correct).
- Clarify in the docs when to use colons.
- Fix the MetricNameRE to reflect reality.

I'll create a PR for the latter. @brian-brazil could you look into the doc clarification? @kamalmarhubi feel free to add unit tests to this here to fix client_golang. For other clients, respective maintainers should check the behavior (cc @grobie for Ruby).

beorn7:
See https://github.com/prometheus/common/pull/65

prometheus/desc.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 var (
-	metricNameRE = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_:]*$`)
+	metricNameRE = regexp.MustCompile(`^[a-zA-Z_:][a-zA-Z0-9_:]*$`)
```
beorn7: We should just use https://github.com/prometheus/common/blob/master/model/metric.go#L25 , now that it is fixed (i.e. don't set a variable here but just use `model.MetricNameRE` below).

Same for next line, use `model.LabelNameRE`, now that we are on it.

beorn7:
To get things moving here, I'll create a new PR on my own. Closing this in lieu.

kamalmarhubi:
@beorn7 apologies for the non-response, and thanks for picking it up!

beorn7:
See https://github.com/prometheus/client_golang/pull/256
Thanks for spotting the issue in the first place.

Issue #245, opened 2016-10-24
Update example applications to use standard units


grobie:
The random example uses microseconds instead of seconds. See https://github.com/prometheus/docs/issues/607

PR #257, opened 2016-11-12
Adjust examples to modern usage

beorn7:
- Use seconds instead of microseconds.
- Use promhttp.Handler instead of prometheus.Handler

@grobie @iamjarvo


beorn7:
Fixes #245

PR #259, opened 2016-11-18
Bring Gauges and Counters in line with client library guidelines

beorn7:
- Point from Inc and Dec to Add and Sub in doc comments.

- Deprecate Untyped for direct instrumentation.

- Add a SetToCurrentTime method to Gauge

Note that adding the SetToCurrentTime method is not really following
Go's principle of lean interfaces. However, the Gauge interface is
already quite fat. (The only methods really required are Set and
Add. Everything else could be expressed in terms of those two.) So we
have already quite a few "convenience" methods traditionally, so I
think we should stay consistent here.

The alternatives would be:

- Not support SetToCurrentTime at all (it's only a SHOULD in the
  guidelines).

- A top level function `SetToCurrentTime(Gauge)`.

- Just a helper `CurrentTime()` that returns the curent unix time in
  seconds as a float (which is pretty verbose using the standard
  library, see code in this commit). This would allow
  `myGauge.Set(CurrentTime)`.

Weighing all circumstances, I believe the way in this commit is the
least evil. Issue #223 could be used to rework interfaces more
fundamentally in a breaking change if feasible.

@brian-brazil for guideline compliance
@stuartnelson3 for Go

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -30,7 +30,8 @@ type Counter interface {
 	Metric
 	Collector
 
-	// Inc increments the counter by 1.
+	// Inc increments the counter by 1. Use Add to increment it by arbitrary
+	// non-negative values.
 	Inc()
 	// Add adds the given value to the counter. It panics if the value is <
 	// 0.
--- a/prometheus/gauge_test.go
+++ b/prometheus/gauge_test.go
@@ -19,6 +19,7 @@ import (
 	"sync"
 	"testing"
 	"testing/quick"
+	"time"
 
 	dto "github.com/prometheus/client_model/go"
 )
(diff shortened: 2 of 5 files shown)


prometheus/counter.go
```diff
@@ -30,7 +30,8 @@ type Counter interface {
 	Metric
 	Collector
 
-	// Inc increments the counter by 1.
+	// Inc increments the counter by 1. Use Add to increment it by arbitrary
+	// positive values.
```
brian-brazil: non-negative

Zero is okay.

beorn7: Done.

beorn7:
I'll merge this for now. If radical Gophers have issues with the new method, we can discuss later.

prometheus/gauge_test.go
```diff
@@ -180,3 +181,22 @@ func TestGaugeFunc(t *testing.T) {
...
+	})
+	g.SetToCurrentTime()
+	unixTime := float64(time.Now().Unix())
+
+	m := &dto.Metric{}
+	g.Write(m)
+
+	delta := unixTime - m.GetGauge().GetValue()
+	// This is just a smoke test to make sure SetToCurrentTime is not
+	// totally off. Tests with current time involved are hard...
```
stuartnelson3: I think this approach is fine, especially given the simplicity of the method invoked.

stuartnelson3:
👍

PR #260, opened 2016-11-18
Extend the example for pushing to pushgateway

beorn7:
@brian-brazil Is this meeting your expectations for a typical batch job?

I used the new SetToCurrentTime method here, and the metrics as suggested.

Note the use of `push.AddFromGathere` instead of just `push.FromGatherer` as discussed earlier.

I find this quite clear and concise. My only concern is that the success time and completion time will be slightly different. (Which ironically is a side effect of the SetToCurrentTime method. If it didn't exist, one would probably set a variable to the current timestamp first and then set both gauges to that value.)


brian-brazil:
:+1:

beorn7:
So yeah, I don't think this desperately requires any additional helpers. Looks quite manageable to me.

PR #261, opened 2016-11-22
Make heap_released_bytes a Gauge.

### LINKED ISSUE #229: Go collector: go_memstats_heap_released_bytes_total is not really a counter.
Definitely goes up and down all the time.

While on it, check other memstats counters, too.

beorn7:
It's not a counter after all.

Also, remove a misleading part of the sys_bytes help string.

@stuartnelson3 
Fixes #229

### DIFF
--- a/prometheus/go_collector.go
+++ b/prometheus/go_collector.go
@@ -48,7 +48,7 @@ func NewGoCollector() Collector {
 			}, {
 				desc: NewDesc(
 					memstatNamespace("sys_bytes"),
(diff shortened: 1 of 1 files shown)


stuartnelson3:
Hm, I definitely would have expected a counter.

beorn7:
So did we. My assumption is those are bytes released to the OS but not yet claimed back, i.e. removed from virtual memory.

The memstats stuff is really severely underdocumented.

PR #262, opened 2016-11-23
Allow Summaries with empty objectives and deprecate DefObjectives

### LINKED ISSUE #118: Allow creating summaries without quantiles
From reading the code, it seems like it's not possible to create summaries without any quantiles at the moment:

https://github.com/prometheus/client_golang/blob/master/prometheus/summary.go#L183-L185

If that `len(...) == 0` check was changed to a `... == nil`, then I guess it should work (one would then be able to pass in an empty, but constructed slice of objectives).

The same is true for histograms (although a summary/histogram without quantiles/buckets would behave identically to each other):

https://github.com/prometheus/client_golang/blob/master/prometheus/histogram.go#L184-L186

beorn7:
This also updates all tests and examples to use explicitly set
objectives.

In v0.10, DefObjectives will be completely removed, and the default
Summary will have no objectives then.

Fixes #118 

@stuartnelson3

### DIFF
--- a/prometheus/summary.go
+++ b/prometheus/summary.go
@@ -54,6 +54,9 @@ type Summary interface {
 }
 
 // DefObjectives are the default Summary quantile values.
+//
+// Deprecated: DefObjectives will not be used as the default objectives in
+// v0.10 of the library. The default Summary will have no quantiles then.
 var (
 	DefObjectives = map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001}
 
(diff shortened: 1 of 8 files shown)


prometheus/summary.go
```diff
@@ -54,6 +54,9 @@ type Summary interface {
 }
 
 // DefObjectives are the default Summary quantile values.
+//
+// Deprecated: DefObjectives will be removed in v0.10. The default will be a
```
grobie: Hm, it would be handy to still provide a package variable given how common these quantiles are. I doubt many people have a good feeling what error value to choose and would copy and paste it anyway.

stuartnelson3: Handy to provide these values in v0.10, even if they aren't used internally to the package?

stuartnelson3:
This PR is fine to me, but as @grobie is the guru who has his finger on the pulse of what the users want, I defer to him

prometheus/summary.go
brian-brazil: I agree. While users shouldn't really be using these in the first place, it's a breaking change and this will help quell some of the rumblings it'll cause.

beorn7: Sure, we can leave them in place, but I'd still rename then. Let's decide about that later (when we actually change the default behavior). For now, nothing changes. I have changed the comment to not suggest that they will be removed (rather: not used as default anymore).

beorn7:
As per personal discussion, I have removed the funny comments from the examples. They were grammatically wrong, and Matt taught me to not make examples funny (but relevant). Ironically, those examples were of his own (at least they were not mine, I found them when I entered the project 3y ago...).

beorn7:
@stuartnelson3 Do you think comments are addressed?

stuartnelson3:
looks good

PR #263, opened 2016-11-30
Basic alertmanager API support

tcolgate:
I realize this probably isn't up to snuff, but I needed something for a chatops integration.
What would it take to get this upstreamable?

### DIFF
--- a/api/alertmanager/api.go
+++ b/api/alertmanager/api.go
@@ -223,6 +223,8 @@ func (c apiClient) do(ctx context.Context, req *http.Request) (*http.Response, [
 type AlertAPI interface {
 	// List all the active alerts.
 	List(ctx context.Context) ([]*model.Alert, error)
+	// List all grouped active alerts.
+	ListGroups(ctx context.Context) ([]AlertGroup, error)
 	// Push a list of alerts into the Alertmanager.
 	Push(ctx context.Context, alerts ...*model.Alert) error
 }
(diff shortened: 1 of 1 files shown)


tcolgate:
It has just occurred to me that I didn't actually write any of this, it was a branch in the repo that I just used in my fork. Still, it would be useful to get this, or something like i merged.
What were the objections to the original branch?

beorn7:
@fabxc I'm unsure about the state and intention of the code in `prometheus/client_golang/api/...`. Is what @tcolgate did here in line with your original intentions?

fabxc:
There were no objections to the original branch. I'm just still somewhat unsure whether the API clients should live in client_go or the repos themselves.

As APIs are versioned, client_go should be generally fine – it's just somewhat easy to forget if we make changes/additions to the APIs.

brian-brazil:
I think we decided on client_golang last time around. There'll likely be apis in other languages too (Python and Ruby both seem likely).

tcolgate:
OK, I'll look into adding some tests (I've really tested a HTTP API client
before, so that should be interestnig).

Any thoughts on the AlertGroups not being in common/model?  I guess it's
only used by AlertManager, so may not be relevant elsewhere.

fabxc:
The AlertGroups is this totally dirty hack to get something shown in the UI. We just haven't really figured out a way to present the current internal state to the user.

Note that while the API is versioned as v1, we never explicitly said it's stable and some things might change that would have to be reflected in this package. I think most parts are in an okay state, but certainly not the AlertGroups.

tcolgate:
From the perspective of someone interacting with the API and server. I've gone to the trouble of setting up grouping, and templates etc, unfrotunately the API then effectively just gives me back only the raw data.
  I don't even get the "common labels" and "common annotation", that the templates I've written get to see.
  So the data I can get from the API is two degrees of config and user experience away, from the alerts I'd really like to be able to get at (they don't contain any rendered template content, and don't have the grouping of labels).
  I'm not quite sure what can be done about that.

fabxc:
Any proposal for a stable API from someone with a real use case would be great :)

tcolgate:
I'll have a think. Off the top of my head, having /alerts/groups  return the json equivalent of the struct that is handed to the templates in the config would some help.
  Additionally, separating out the silence/inhibited alerts high up, so that I just have one struct with firing alerts and don't have to dig into it to filter them out. That would probably help a bit too (or maybe, if that's too invasive to the existing AM UI needs, just let us optionally request only  uninhibited and non-silenced alerts?

brian-brazil:
I'd prefer we didn't hand what goes to the templates out via the API (or webhook), things like common annotations are for convenience for templating and I worry about the long term bandiwidth implications of sending them outside the process.

beorn7:
Superseded by #333

PR #277, opened 2017-02-13
goCollector: add thread count gauge in goCollector

ggaaooppeenngg:
@beorn7 I want to add a threads count gauge in the default go collector, but I just make a workaround to pass the unit test, and I am thinking a way to test it.
Signed-off-by: Peng Gao <peng.gao.dut@gmail.com>

### DIFF
--- a/prometheus/go_collector.go
+++ b/prometheus/go_collector.go
@@ -8,9 +8,9 @@ import (
 )
 
 type goCollector struct {
-	goroutines Gauge
-	threads    Gauge
-	gcDesc     *Desc
+	goroutinesDesc *Desc
+	threadsDesc    *Desc
+	gcDesc         *Desc
 
 	// metrics to describe and collect
 	metrics memStatsMetrics
@@ -20,16 +20,14 @@ type goCollector struct {
 // go process.
 func NewGoCollector() Collector {
 	return &goCollector{
-		goroutines: NewGauge(GaugeOpts{
-			Namespace: "go",
-			Name:      "goroutines",
-			Help:      "Number of goroutines that currently exist.",
-		}),
-		threads: NewGauge(GaugeOpts{
-			Namespace: "go",
-			Name:      "threads",
-			Help:      "Number of threads created.",
-		}),
+		goroutinesDesc: NewDesc(
+			"go_goroutines",
+			"Number of goroutines that currently exist.",
+			nil, nil),
+		threadsDesc: NewDesc(
+			"go_threads",
+			"Number of OS threads created",
+			nil, nil),
 		gcDesc: NewDesc(
 			"go_gc_duration_seconds",
 			"A summary of the GC invocation durations.",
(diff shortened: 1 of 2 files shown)


prometheus/go_collector.go
```diff
@@ -236,6 +242,9 @@ func (c *goCollector) Describe(ch chan<- *Desc) {
 func (c *goCollector) Collect(ch chan<- Metric) {
 	c.goroutines.Set(float64(runtime.NumGoroutine()))
 	ch <- c.goroutines
+	n, _ := runtime.ThreadCreateProfile(nil)
+	c.threads.Set(float64(n))
```
grobie: I think we should use NewConstMetric here and for the goroutines.

beorn7: Yes, indeed. This would be a good time to change it.

prometheus/go_collector.go
```diff
@@ -236,6 +242,9 @@ func (c *goCollector) Describe(ch chan<- *Desc) {
 func (c *goCollector) Collect(ch chan<- Metric) {
 	c.goroutines.Set(float64(runtime.NumGoroutine()))
 	ch <- c.goroutines
+	n, _ := runtime.ThreadCreateProfile(nil)
```
beorn7: Are you sure this reliably produces the number of OS threads? From what I can see, it could also return the number of OS threads _created_ but wouldn't be adjusted if an OS thread gets _destroyed_. But I don't know any details. That's why I'm asking.

ggaaooppeenngg: I checked the code of runtime, there are no part dealing with destroy, I think OS threads never destroy but become idle.

prometheus/go_collector.go
ggaaooppeenngg: image

I want to add thoese counts too, I just post a new metric for any suggestion.

prometheus/go_collector.go
beorn7: I did some consistency checks. The threadcreate number indeed looks like the number of OS threads.

prometheus/go_collector.go
beorn7: goroutine already exists. Not sure how much sense it makes to export the others as numbers. Could you make your point what they mean and what they are useful for?

prometheus/go_collector.go
```diff
@@ -24,6 +25,11 @@ func NewGoCollector() Collector {
 			Name:      "goroutines",
 			Help:      "Number of goroutines that currently exist.",
 		}),
+		threads: NewGauge(GaugeOpts{
+			Namespace: "go",
+			Name:      "threads",
+			Help:      "Number of threads created.",
```
beorn7: Perhaps say "Number of OS threads created."

prometheus/go_collector.go
beorn7: And as said, please use `MustNewConstMetric` for this and the goroutine count.

ggaaooppeenngg: I am not sure too, but I think it is helpful in some way and I need some time or experiment to dig it out, and add a threads number may be the first step.

prometheus/go_collector.go
ggaaooppeenngg: OK

beorn7:
:+1:

PR #279, opened 2017-02-16
Replace AUTHORS.md by an updated MAINTAINERS.md

beorn7:
Also, note the unmaintained state of the API client in the README.md

@grobie


README.md
```diff
@@ -30,7 +30,9 @@ The
 [`api/prometheus` directory](https://github.com/prometheus/client_golang/tree/master/api/prometheus)
 contains the client for the
 [Prometheus HTTP API](http://prometheus.io/docs/querying/api/). It allows you
-to write Go applications that query time series data from a Prometheus server.
+to write Go applications that query time series data from a Prometheus
+server. It is in an early stage and further development is currently on hold
```
brian-brazil: I'd be wary of including such a statement in a readme, it may be misconstrued to mean the whole client is unmaintained.

beorn7: What is your suggestion instead?

brian-brazil: I'd mention it's alpha.

beorn7: But how can we express that there will be nobody reviewing PRs?

And how does the "it's alpha" not reflect on the whole client with the same argument you used above? (I don't think this misunderstanding is likely. I'm just trying to understand your line of reasoning.)

brian-brazil: That we're not developing isn't the same as not reviewing PRs. I presume at least one of us has review capacity if someone started fleshing it out.

> And how does the "it's alpha" not reflect on the whole client with the same argument you used above? 

It does, but being alpha is much better than a part of the project being unmaintained and thus dead.

beorn7: OK, I'll remember to assign PRs to you, if they come in... ;)

PR #281, opened 2017-02-22
Adding GCCPUFraction metric to goCollector

fmstephe:
@beorn7 

We would really like to track the `runtime.MemStats.GCCPUFraction` as part of the runtime stats.


beorn7:
Yes, sure. The only reason it is not in there is that it wasn't in runtime.MemStats when this code was written. It has been around for long enough to put it in here.

:+1:

fmstephe:
Great, thanks.

beorn7:
Oops, this should have been merged quite some time ago.

PR #282, opened 2017-02-28
Fix invalid Timer (gauge) example

grobie:
The example method is assumed to be used as main() function. As a main()
function doesn't have any return values, the example doesn't compile and
is invalid.

@beorn7


grobie:
See https://godoc.org/github.com/prometheus/client_golang/prometheus#example-Timer--Gauge

beorn7:
Thanks

PR #286, opened 2017-03-16
RFC: Add ability for user to inject middleware.

tomwilkie:
I want to use the client API to speak to Cortex, which needs basic auth or a token header.  I wondered if adding a generic hook for munging the request was preferable to specific hooks for the various auth schemes.  WDYT?

### DIFF
--- a/api/prometheus/api.go
+++ b/api/prometheus/api.go
@@ -89,6 +89,9 @@ type Config struct {
 	// Transport is used by the Client to drive HTTP requests. If not
 	// provided, DefaultTransport will be used.
 	Transport CancelableTransport
+
+	// Middleware allows you to inject custom auth, monitoring etc.
+	Middleware func(*http.Request)
 }
 
 func (cfg *Config) transport() CancelableTransport {
@@ -115,14 +118,16 @@ func New(cfg Config) (Client, error) {
 	u.Path = strings.TrimRight(u.Path, "/") + apiPrefix
 
 	return &httpClient{
-		endpoint:  u,
-		transport: cfg.transport(),
+		endpoint:   u,
+		transport:  cfg.transport(),
+		middleware: cfg.Middleware,
 	}, nil
 }
 
 type httpClient struct {
-	endpoint  *url.URL
-	transport CancelableTransport
+	endpoint   *url.URL
+	transport  CancelableTransport
+	middleware func(*http.Request)
 }
 
 func (c *httpClient) url(ep string, args map[string]string) *url.URL {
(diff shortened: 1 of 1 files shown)


grobie:
So far we've added _client_ support for Basic Auth + Bearer token to a couple of Prometheus exporters/components already. Does Cortex need something special or would a standard bearer Token suffice?

juliusv:
@grobie That was my thought as well - just support the usual Prometheus auth options directly here?

tomwilkie:
> Does Cortex need something special or would a standard bearer Token suffice?

Basic auth will suffice; the middleware is a little more flexible as it allows me to do things like run behind the filewall (and avoid auth), specifying a user id instead.  Plus, Cortex exposes the query path under /api/prom, so I need to munge the path a little.

juliusv:
Does the URL that you can pass in to `New()` not already allow you to have a custom path suffix? Looks like it does: https://github.com/weaveworks/client_golang/blob/18d6715b27cdf0240231ba3bf3798294898ffe26/api/prometheus/api.go#L118

tomwilkie:
Yeah, good point. 

I'll add basic auth then?

grobie:
I'd prefer to wait with implementing generic interfaces until there are
good reasons to do so. If any of basic auth / bearer token / TLS covers
your needs, let's just add support for it as we do already elsewhere.

juliusv:
@tomwilkie Now we're back at a similar place like in my remote write PR about having a common way of configuring all of the usual HTTP auth and TLS options (https://github.com/prometheus/prometheus/pull/2499/files#r106292482).

There's a general discussion whether we should consolidate that TLS+auth code into a shared library somewhere (proponent: @brian-brazil) or just write a copy of it into each component that needs it (proponent: @fabxc).

Maybe for now just emulate that config struct and the code at https://github.com/prometheus/prometheus/blob/4daffbef1296dd28769e5418376fc12ffaff95de/retrieval/target.go#L70-L105 that uses those options to create a client?

tomwilkie:
Hmm.... `httputil` brings in `prometheus/config`, which we probably don't want to do in this library.

Interestingly, [this comment on the roundtripper](https://golang.org/pkg/net/http/#RoundTripper):
```
        // RoundTrip should not modify the request, except for
        // consuming and closing the Request's Body.
```
Implies we're doing it wrong in httputil.  And it looks like I was the last one to touch this...  Eeek!

Given the current client allows you to inject your own transport and we already have a basic auth injecting transport, I'll just do that and close this PR.

brian-brazil:
> There's a general discussion whether we should consolidate that TLS+auth code into a shared library somewhere (proponent: @brian-brazil)

My concerns around around where there's a config file involved, which doesn't cover this case.

Issue #290, opened 2017-03-31
Support for querying metadata on the client


andrestc:
I've noticed the client is missing methods to fetch for [series metadata](https://prometheus.io/docs/querying/api/#querying-metadata), namely: `GET /api/v1/label/<label_name>/values` and `GET /api/v1/series`. Any particular reason the client isn't exposing those (i noticed it even has the `epLabelValues` and `epSeries` consts)?

I can contribute those since I'll need them anyway.

juliusv:
@andrestc Pretty sure the only reason they're missing because nobody got to that yet - so a PR to add it would be very welcome!

beorn7:
I think both are implemented by now.
Please re-open if not.

PR #248, opened 2016-10-26
Add delete query to api.

roman-vynar:
Implements `curl -g -X DELETE http://localhost:9090/api/v1/series?match[]=metric`

PR #291, opened 2017-03-31
api: reorganize API package and add label values endpoint implementation

andrestc:
This adds a client side implementation for fetching label values from prometheus api.

Would love to get some review on this as this is my first contribution here. Some design questions i may have oversimplified: 
1. return a more descriptive value instead of []string? Maybe model.LabelValues?
2. create a separated interface for metadata api(following the separation on the docs)?

Related to #290.


api/prometheus/api.go
```diff
@@ -280,8 +280,10 @@ func (qr *queryResult) UnmarshalJSON(b []byte) error {
 type QueryAPI interface {
 	// Query performs a query for the given time.
 	Query(ctx context.Context, query string, ts time.Time) (model.Value, error)
-	// Query performs a query for the given range.
+	// QueryRange performs a query for the given range.
 	QueryRange(ctx context.Context, query string, r Range) (model.Value, error)
+	// QueryLabelValues performs a query for the values of the given label.
+	QueryLabelValues(ctx context.Context, label string) ([]string, error)
```
juliusv: To be more consistent with naming we have elsewhere, maybe omit the `Query` and just call this `LabelValues`.

juliusv: And yeah, this should be `model.LabelValues`, since we're using the other `model` types in here as well.

api/prometheus/api.go
```diff
@@ -346,3 +348,15 @@ func (h *httpQueryAPI) QueryRange(ctx context.Context, query string, r Range) (m
 
 	return model.Value(qres.v), err
 }
+
+func (h *httpQueryAPI) QueryLabelValues(ctx context.Context, label string) ([]string, error) {
+	u := h.client.url(epLabelValues, map[string]string{"name": label})
+	req, _ := http.NewRequest(http.MethodGet, u.String(), nil)
```
juliusv: Never ignore returned errors.

andrestc:
@juliusv, addressed your comments. Thanks for the quick review!

api/prometheus/api.go
```diff
@@ -280,8 +280,10 @@ func (qr *queryResult) UnmarshalJSON(b []byte) error {
 type QueryAPI interface {
 	// Query performs a query for the given time.
 	Query(ctx context.Context, query string, ts time.Time) (model.Value, error)
-	// Query performs a query for the given range.
+	// QueryRange performs a query for the given range.
 	QueryRange(ctx context.Context, query string, r Range) (model.Value, error)
+	// LabelValues performs a query for the values of the given label.
+	LabelValues(ctx context.Context, label string) (model.LabelValues, error)
```
grobie: So, I always understood this interface to only include `/api/v1/query*` endpoints. With that, we'd end up with `QueryAPI`, `LabelsAPI`, `AlertmanagerAPI`, etc. interfaces.

I'd say we should either do that, or rename this interface to something like `APIv1` (or a better name). Adding all current v1 functions in a `QueryAPI` interface doesn't make sense though, as not all are about querying.

juliusv: Valid concern, and one that we are violating by the page menu hierarchy under which https://prometheus.io/docs/querying/api/ is embedded as well.

I don't like the idea of `QueryAPI`, `LabelsAPI`, etc. much, but if we also want to add Alertmanager API functionality here eventually, then maybe `Prometheus` and `Alertmanager`? I'm not even sure whether we should keep the `...API` suffix. Or alternatively, `...Client` instead of `...API`.

andrestc: I guess `Prometheus` and `Alertmanager` makes sense since there aren't that many endpoints to justify doing `QueryAPI`, `LabelsAPI` etc.

grobie:
Yet? ;) Alertmanager is difficult though, as there are alertmanager related
endpoints in the Prometheus API and the Alertmanager API itself.

I'm pretty sure we'll come up with a new API version though at some point,
so moving everything into a v1 would be good in general (or include the
version in the types, but less readable and common).

juliusv:
> Yet? ;) Alertmanager is difficult though, as there are alertmanager related endpoints in the Prometheus API and the Alertmanager API itself.

There are alert-related endpoints in Prometheus, but are there Alertmanager-related ones in Prometheus as well?

Agree on versioning, but that should happen for each of {Alertmanager, Prometheus, ...}, as their APIs are versioned separately. Maybe use a package for versioning, like the InfluxDB client does? https://godoc.org/github.com/influxdata/influxdb/client/v2

grobie:
Check `/api/v1/alertmanagers` in https://prometheus.io/docs/querying/api/#alertmanagers. That's what I was referring to above as well, basically either we have one interface per `/api/v1/<group>` thing, or we only have one per API. An interface called `Interface` in a `prometheus/v1` package describing all functions of the API might be enough for now?

andrestc:
Looks good... I'm not so sure about the `Interface` name, maybe name the Interface `API` on `prometheus/v1`.

juliusv:
> Check /api/v1/alertmanagers in https://prometheus.io/docs/querying/api/#alertmanagers.

That's fine though, it's Prometheus's view of what Alertmanagers there are. It's not "an Alertmanager API", or part of it. It's just part of the Prometheus API. I'd be for `prometheus/v1` (grouping all functionality under Prometheus's `/api/v1/...` endpoints) and `alertmanager/vXXX` (same for AM) packages. The two binaries are separate components with separate APIs and URLs.

In each package, I would expect an `API` type that gives access to all the endpoints of one particular binary. The `httpClient` and maybe even the `apiClient` wrapper around it could be shared between Prometheus and Alertmanager API packages.

andrestc:
LGTM. If everyone agrees on that, should I update this PR with the new
package/interface structure or open a new issue/pr for that?

Em 4 de abr de 2017 19:53, "Julius Volz" <notifications@github.com>
escreveu:

grobie:
Completely cool with that. I don't care about a new PR or continue using this one (but changing the name). Just make sure to use separate commits to first change the package structure and then another one to add your new feature/endpoint.

juliusv:
@andrestc Oh yeah sorry, I meant to reply to this PR saying that we should wait until @grobie is back from his hiking to give him chance to object, but I forgot. Now we know he's fine with it, so go ahead :)

grobie:
@andrestc Are you cool with the changes? This should unblock #248 as well.

andrestc:
Yeah! I will update the PR with those, i'll probably have time for that on the next few days

andrestc:
Hey @juliusv, @grobie. I've pushed some of the changes we discussed. 

I've left the http client on a separated package so it can be reused across Prometheus and Alertmanagers APIs(including future versions?).

Once you guys are OK with the changes I can rebase and make sure this last change comes before the label values query change.

api/client.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/url"
+	"path"
+	"strings"
+	"time"
+
+	"golang.org/x/net/context"
```
grobie: The context package has been integrated into the stdlib, let's use that instead.

api/client.go
```diff
@@ -0,0 +1,136 @@
...
+	"golang.org/x/net/context/ctxhttp"
```
grobie: Not longer necessary, the stdlib http package has native support for context now.

api/client.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+// CancelableTransport is like net.Transport but provides
+// per-request cancelation functionality.
+type CancelableTransport interface {
```
grobie: Not longer necessary with the new http package. Will also fix #292.

api/client.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	u := *c.endpoint
+	u.Path = p
+
+	return &u
+}
+
+func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte, error) {
+	resp, err := ctxhttp.Do(ctx, &http.Client{Transport: c.transport}, req)
```
grobie: I think we'll now have to use context.WithCancel here. I guess we should leave that to the user.

andrestc:
@grobie I've addressed your comments by removing the CancelableTransport and by passing down the ctx to the request (should I create a child context?).

We will need to drop go 1.6.x from travis and add go 1.8.x.

grobie:
> We will need to drop go 1.6.x from travis and add go 1.8.x.

Oh, @beorn7 will need to say if that's actually acceptable as requirement for client_golang users. What's the go1.8 feature? The new context package in stdlib and request.WithCancel() was added in go1.7 I thought?

andrestc:
Oh, I said to add go 1.8 just to keep the last two major versions on travis, we are fine on 1.7+.

beorn7:
@grobie We are talking only about the `api` package here, which is not widely used (and arguably experimental/unmaintained). the client_galang/prometheus/... packages should still be fine with older Go versions.

grobie:
Once we stop running all tests with go1.6 it becomes very easy to
unknowingly break compatibility in other packages. Although, if you're fine
with such a setup, let's go forward here.

beorn7:
OK, I see.

Yeah, that would be bad if the tests for an experimental package had an impact on the ones for a production package in frequent use.

Can we use build tags to not run the api tists on older go versions?

grobie:
Yes, this should be possible, but we might need to put the build tags in
both the test+api files.

andrestc:
I've pushed the change that includes the build tag to the api package files.

api/client.go
```diff
@@ -0,0 +1,131 @@
...
+// Client is the interface for an API client.
+type Client interface {
+	URL(ep string, args map[string]string) *url.URL
+	Do(context.Context, *http.Request) (*http.Response, []byte, error)
+}
+
+// New returns a new Client.
+//
+// It is safe to use the returned Client from multiple goroutines.
+func New(cfg Config) (Client, error) {
```
grobie: Actually I think we should call this NewClient now, as the package is called API but we're not returning a type called api.

andrestc:
Done! @grobie @juliusv

api/prometheus/v1/api.go
```diff
@@ -234,6 +70,16 @@ type Range struct {
 	Step time.Duration
 }
 
+// API provides bindings the Prometheus's v1 API.
```
juliusv: bindings the -> bindings for

andrestc:
Done, @juliusv

juliusv:
Thanks!

tboerger:
The build tag prevents the usage with go 1.8 :(

grobie:
@tboerger Are you sure that's the problem? The documentation explicitly states `- "go1.7", from Go version 1.7 onward` https://golang.org/pkg/go/build/.

tboerger:
I have created https://github.com/prometheus/client_golang/pull/298 now to allow go 1.8 as well.

PR #288, opened 2017-03-31
Http client

stuartnelson3:
@beorn7 adding this here for to start discussion, it has been heavily informed by the middleware approach in #285 

Current API:

```go
client := *http.DefaultClient
client.Timeout = 500 * time.Millisecond

inFlightGauge := prometheus.NewGauge(prometheus.GaugeOpts{
	Name: "in_flight",
	Help: "An in-flight gauge",
})

counter := prometheus.NewCounterVec(
	prometheus.CounterOpts{
		Name: "test_counter",
		Help: "A test counter.",
	},
	[]string{"code", "method"},
)

histVec := prometheus.NewHistogramVec(
	prometheus.HistogramOpts{
		Name:    "latency",
		Help:    "A histogram of latency.",
		Buckets: prometheus.DefBuckets,
	},
	[]string{"event"},
)

prometheus.MustRegister(counter, histVec, inFlightGauge)

promclient := promhttp.InFlightC(inFlightGauge,
	promhttp.CounterC(counter,
		promhttp.ClientTrace(histVec, &client),
	),
)

resp, err := promclient.Get("example.com")
// And so forth ...
```

PR #295, opened 2017-04-25
Instrument RoundTripper via middleware

stuartnelson3:
Closing #288 as it was easier to open a new pull request than deal with the rebase/merge conflicts.


.travis.yml
```diff
@@ -2,8 +2,8 @@ sudo: false
 language: go
 
 go:
- - 1.6.3
```
grobie: We just had a discussion about supported versions in #291 and the decision was to test against 1.6 but to just build tags for 1.7+ features.

prometheus/promhttp/client.go
```diff
@@ -0,0 +1,246 @@
...
+	return rt(r)
+}
+
+// InstrumentTrace is used to offer flexibility in instrumenting the available
+// httptrace.ClientTrace hook functions. Each function is passed a float64
+// representing the time in seconds since the start of the http request. A user
+// may choose to use separately buckets Histograms, or implement custom
+// instance labels on a per function basis.
+type InstrumentTrace struct {
+	GotConn, PutIdleConn, GotFirstResponseByte, Got100Continue, DNSStart, DNSDone, ConnectStart, ConnectDone, TLSHandshakeStart, TLSHandshakeDone, WroteHeaders, Wait100Continue, WroteRequest func(float64)
```
grobie: I'd prefer to list one struct member per line, even if that means repeating the value, than having such a long line.

stuartnelson3: Totally cool with me

.travis.yml
stuartnelson3: How do we want to handle testing features that are only available in 1.7, like httptrace?

grobie:
I guess the whole rountripper becomes a go1.7+ feature, e.g. adding the
build tag to both files you added here.

beorn7:
On my review queue. Will get to it once the conference permits.

prometheus/promhttp/client.go
```diff
@@ -0,0 +1,138 @@
...
+// expensive and should be used judiciously.
+//
+// If the wrapped RoundTripper panics, no values are reported.
+func InstrumentRoundTripperDuration(obs prometheus.ObserverVec, next http.RoundTripper) RoundTripperFunc {
+	code, method := checkLabels(obs)
+
+	return RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		var (
+			start     = time.Now()
+			resp, err = next.RoundTrip(r)
```
beorn7: Not sure if the otherwise neat `var` block isn't a bit confusing here. The work is now done in the `var` block, which is not where people might be looking for it. Also, the order is critical here. Final point: you didn't use the `var` block in InstrumentHandlerDuration.

prometheus/promhttp/client.go
```diff
@@ -0,0 +1,138 @@
...
+// The ObserverVec must have zero, one, or two labels. The only allowed label
+// names are "code" and "method". The function panics if any other instance
+// labels are provided. The Observe method of the Observer in the ObserverVec
+// is called with the request duration in seconds. Partitioning happens by HTTP
+// status code and/or HTTP method if the respective instance label names are
+// present in the ObserverVec. For unpartitioned observations, use an
+// ObserverVec with zero labels. Note that partitioning of Histograms is
+// expensive and should be used judiciously.
+//
+// If the wrapped RoundTripper panics, no values are reported.
```
beorn7: ... or returns a non-nil error...

prometheus/promhttp/client.go
```diff
@@ -0,0 +1,138 @@
...
+		)
+		if err != nil {
+			return nil, err
+		}
+		obs.With(labels(code, method, r.Method, resp.StatusCode)).Observe(time.Since(start).Seconds())
+		return resp, err
+	})
+}
+
+func checkEventLabel(c prometheus.Collector) {
```
beorn7: This isn't used anywhere.

prometheus/promhttp/client_1_8.go
```diff
@@ -0,0 +1,141 @@
...
+// RoundTripper and reports times to hook functions provided in the
+// InstrumentTrace struct. Hook functions that are not present in the provided
+// InstrumentTrace struct are ignored. Times reported to the hook functions are
+// time since the start of the request. Note that partitioning of Histograms
+// is expensive and should be used judiciously.
+//
+// See the example for ExampleInstrumentRoundTripperDuration for example usage.
+func InstrumentRoundTripperTrace(it *InstrumentTrace, next http.RoundTripper) RoundTripperFunc {
+	return RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		var (
```
beorn7: I really wouldn't use a var block for a single initialization.

prometheus/promhttp/client_1_8_test.go
```diff
@@ -0,0 +1,193 @@
...
+			Buckets: prometheus.DefBuckets,
+		},
+		[]string{"method"},
+	)
+
+	prometheus.MustRegister(counter, tlsLatencyVec, dnsLatencyVec, histVec, inFlightGauge)
+
+	trace := &InstrumentTrace{
+		DNSStart: func(t float64) {
+			dnsLatencyVec.WithLabelValues("DNSStart")
```
beorn7: In general, we prefer snake_case for names in Prometheus.

Applicable to label names below, too.

prometheus/promhttp/client_1_8.go
```diff
@@ -0,0 +1,141 @@
...
+	"net/http"
+	"net/http/httptrace"
+	"time"
+)
+
+// InstrumentTrace is used to offer flexibility in instrumenting the available
+// httptrace.ClientTrace hook functions. Each function is passed a float64
+// representing the time in seconds since the start of the http request. A user
+// may choose to use separately buckets Histograms, or implement custom
+// instance labels on a per function basis.
```
beorn7: Perhaps document somehow that the various trace functions that get an error passed (PutIdleConn and others) don't trigger a call of the function given here if that error is non-nil.

stuartnelson3:
Hello from the other side ... let me know how it looks :)

beorn7:
I'm also on the other side. :-)

Will have a look ASAP.

prometheus/promhttp/instrument_client.go
```diff
@@ -0,0 +1,101 @@
...
+// InstrumentRoundTripperInFlight is a middleware that wraps the provided
+// http.RoundTripper. It sets the provided prometheus.Gauge to the number of
+// requests currently handled by the wrapped http.RoundTripper.
+//
+// See the example for ExampleInstrumentRoundTripperDuration for example usage.
+func InstrumentRoundTripperInFlight(gauge prometheus.Gauge, next http.RoundTripper) RoundTripperFunc {
+	return RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		gauge.Inc()
+		defer gauge.Dec()
+		resp, err := next.RoundTrip(r)
```
beorn7: I think it's better to just directly return here, i.e. `return next.RoundTrip(r)`. Should RountTrip return a non-nil error together with a non-nil Response, we probably want to keep that behavior.

prometheus/promhttp/instrument_client.go
```diff
@@ -0,0 +1,101 @@
...
+// is not incremented.
+//
+// See the example for ExampleInstrumentRoundTripperDuration for example usage.
+func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.RoundTripper) RoundTripperFunc {
+	code, method := checkLabels(counter)
+
+	return RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		resp, err := next.RoundTrip(r)
+		if err != nil {
+			return nil, err
```
beorn7: Again, I think we want to return whatever RoundTrip returned in any case, i.e. `return resp, err`.

Which means you can just increment the counter in a `if err == nil` clause and have only one return line.

prometheus/promhttp/instrument_client.go
```diff
@@ -0,0 +1,101 @@
...
+// If the wrapped RoundTripper panics or returns a non-nil error, no values are
+// reported.
+func InstrumentRoundTripperDuration(obs prometheus.ObserverVec, next http.RoundTripper) RoundTripperFunc {
+	code, method := checkLabels(obs)
+
+	return RoundTripperFunc(func(r *http.Request) (*http.Response, error) {
+		start := time.Now()
+		resp, err := next.RoundTrip(r)
+		if err != nil {
+			return nil, err
```
beorn7: Same concern as above.

beorn7:
@stuartnelson3 Have you seen my review? It's only minor changes that are still needed. Should be easy to update.

(In different news, I'll have a look at https://github.com/mwitkow/go-httpwares/pull/12 and see if we can align with each other somehow. But that will be a separate thing from this PR.)

stuartnelson3:
Sorry for the delay, I missed your comments. Pushed!

beorn7:
Thanks. I'll close the related issues, but will also check out https://github.com/mwitkow/go-httpwares/pull/12 to see where we can align. And of course, I have to think about #299 .

PR #298, opened 2017-05-09
api: added go1.8 build tag

tboerger:
Otherwise the api package works with go 1.7 only.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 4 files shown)


beorn7:
The go1.7 tag includes all versions higher than 1.7, too. This should not be needed.

beorn7:
I think this got clarified on IRC. Closing here. Please re-open if there is still anything unclear.

tboerger:
Sorry, looks like I just used the wrong import path.

PR #302, opened 2017-05-10
Improve promhttp middleware

### LINKED ISSUE #299: promhttp handlers panic with "metric partitioned with non-supported labels" when providing a ConstLabel
The [example](https://godoc.org/github.com/prometheus/client_golang/prometheus/promhttp#example-InstrumentHandlerDuration) for the promhttp handlers mention adding a `ConstLabel {"handler": "some_handler"}`.
When you do this, `checkLabels` will panic. I think it would make sense to not look at the `ConstLabel` when checking for labels. Otherwise you have to encode the handler name into the metric names to get sensible metrics for your handlers.

beorn7:
See commit descriptions.

Fixes #299 .

@stuartnelson3

### DIFF
--- a/prometheus/promhttp/instrument_client_1_8_test.go
+++ b/prometheus/promhttp/instrument_client_1_8_test.go
@@ -28,6 +28,8 @@ func TestClientMiddlewareAPI(t *testing.T) {
 	client := http.DefaultClient
 	client.Timeout = 1 * time.Second
 
+	reg := prometheus.NewRegistry()
+
 	inFlightGauge := prometheus.NewGauge(prometheus.GaugeOpts{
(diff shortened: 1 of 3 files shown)


beorn7:
As it says: It's magic! Nothing to explain, it just works.

In fact, it's quite easy to explain: The problem was that we want to allow arbitrary const labels but we couldn't recognize them in the metric protobuf. All labels look the same there. So what I'm doing here is to set the label value of the one or two normal labels (which must be `code` or `method`) to the magic string, i.e. a string that will never occur otherwise. Then, all the labels with the magic string as a label are guaranteed to not be const labels. From there on, we can do the check as normal.

PR #304, opened 2017-05-11
Add time to write header handler middleware

stuartnelson3:
@beorn7 first stab at writing the "time to writing header" middleware.

Things to note are:

- I've forgone branching between "has the code label" and "doesn't have the code label". They paths are too similar to warrant it IMO.
- As mentioned, this is just based off what I saw in https://github.com/mwitkow/go-httpwares/pull/12. No strong opinions one way or another, but this seemed like a simple way to implement it. I didn't think it necessary to generalize quite like that PR did with a slice of callback functions.

### DIFF
--- a/prometheus/promhttp/instrument_server.go
+++ b/prometheus/promhttp/instrument_server.go
@@ -120,9 +120,6 @@ func InstrumentHandlerCounter(counter *prometheus.CounterVec, next http.Handler)
 // labels. Note that partitioning of Histograms is expensive and should be used
 // judiciously.
 //
-// If the wrapped Handler does not set a status code via WriteHeader, no value
-// is reported.
-//
 // If the wrapped Handler panics before calling WriteHeader, no value is
 // reported.
 //
--- a/prometheus/promhttp/instrument_server.go
+++ b/prometheus/promhttp/instrument_server.go
@@ -63,7 +63,7 @@ func InstrumentHandlerDuration(obs prometheus.ObserverVec, next http.Handler) ht
 	if code {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			now := time.Now()
-			d := newDelegator(w)
+			d := newDelegator(w, nil)
 			next.ServeHTTP(d, r)
 
 			obs.With(labels(code, method, r.Method, d.Status())).Observe(time.Since(now).Seconds())
@@ -96,7 +96,7 @@ func InstrumentHandlerCounter(counter *prometheus.CounterVec, next http.Handler)
 
 	if code {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			d := newDelegator(w)
+			d := newDelegator(w, nil)
 			next.ServeHTTP(d, r)
 			counter.With(labels(code, method, r.Method, d.Status())).Inc()
 		})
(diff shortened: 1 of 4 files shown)


beorn7:
On my review queue…

prometheus/promhttp/instrument_server.go
```diff
@@ -108,6 +108,40 @@ func InstrumentHandlerCounter(counter *prometheus.CounterVec, next http.Handler)
...
+// function panics if any other instance labels are provided. The Observe
+// method of the Observer in the ObserverVec is called with the request
+// duration in seconds. Partitioning happens by HTTP status code and/or HTTP
+// method if the respective instance label names are present in the
+// ObserverVec. For unpartitioned observations, use an ObserverVec with zero
+// labels. Note that partitioning of Histograms is expensive and should be used
+// judiciously.
+//
+// If the wrapped Handler does not set a status code via WriteHeader, no value
+// is reported.
```
beorn7: See https://golang.org/pkg/net/http/#ResponseWriter . I understand this as WriteHeader being called eventually in any case, so we should be fine and don't need this note.

prometheus/promhttp/instrument_server.go
```diff
@@ -108,6 +108,40 @@ func InstrumentHandlerCounter(counter *prometheus.CounterVec, next http.Handler)
...
+// If the wrapped Handler panics before calling WriteHeader, no value is
+// reported.
+//
+// See the example for InstrumentHandlerDuration for example usage.
+func InstrumentHandlerTimeToWriteHeader(obs prometheus.ObserverVec, next http.Handler) http.HandlerFunc {
+	code, method := checkLabels(obs)
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		now := time.Now()
+		d := &wroteHeaderDelegator{
```
beorn7: This breaks interface upgrades (as the type of the embedded delegator is just the plain `delegator` interface, and the check for interface upgrades only type-switches on the outer type, not the type of the embedded struct element, see https://play.golang.org/p/uuqlJpUG7G as proof).

I suggest to chenge the general `responseWriterDelegator` implementation to include an `observeWriteHeader` function, which we check for non-nil in the `WriteHeader` method.

stuartnelson3: K, I updated it, not sure if this was what you imagined. Since `newDelegator` returns an interface type, I just went one more layer deep on what gets sent into that function so I could attach the `observeWriteHeader` method.

prometheus/promhttp/instrument_server.go
```diff
@@ -132,12 +129,12 @@ func InstrumentHandlerTimeToWriteHeader(obs prometheus.ObserverVec, next http.Ha
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		now := time.Now()
-		d := &wroteHeaderDelegator{
-			delegator: newDelegator(w),
+		d := newDelegator(&responseWriterDelegator{
```
beorn7: This will again kill the interface upgrade. I think the way to go is to change `newDelegator` to also accept the observeWriteHeader func (which can be `nil`) and set it in the created `responseWriterDelegator`. It doesn't look beautiful, but at least `newDelegator` is not exported.

stuartnelson3: Roger, pushed.

beorn7:
This should work. Thanks.

PR #305, opened 2017-05-22
Import "log", it is required for log.Fatal(...

nobse:


prometheus/doc.go
```diff
@@ -27,6 +27,7 @@
 //
 //    import (
 //    	"net/http"
+//      "log"
```
beorn7: The space just before the double quote needs to be a TAB. And please have this line before the previous one (lexicographical sorting of imports).

beorn7:
Thanks. :+1:

beorn7:
Thanks. :+1:

PR #307, opened 2017-05-25
Add process_open_fds_percent metric

jesseolsen:
Add process_open_fds_percent metric calculated
from existing process_open_fds and process_max_fds.
This makes it possible to create alarms like:
process_open_fds_percent > 90

@beorn7

### DIFF
--- a/prometheus/process_collector.go
+++ b/prometheus/process_collector.go
@@ -16,13 +16,13 @@ package prometheus
 import "github.com/prometheus/procfs"
 
 type processCollector struct {
-	pid             int
-	collectFn       func(chan<- Metric)
-	pidFn           func() (int, error)
-	cpuTotal        *Desc
-	openFDs, maxFDs *Desc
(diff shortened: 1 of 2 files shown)


grobie:
Thank you for your contribution. Such calculations are best left to metric and alerting rules, as it's easier to change and manage them there. Your approach would also create more time series, even in situations where people don't want to create an alert on fd usage.

Why do you want to create this metric instead of making the calculation on the fly?

```
ALERT ProcessHighFileDescriptorUsage
IF process_open_fds / process_max_fds * 100 > 90
LABELS { ... }
ANNOTATIONS { ... }
```

beorn7:
Thanks @jesseolsen , but @grobie is right. It's the very essence of Promethean philosophy to instrument very low level metrics and let the Prometheus server do the logic. (One of the Prometheus proverbs: _Instrument first, ask questions later._)

Issue #272, opened 2017-01-25
HTTP instrumentation doesn't work properly with http/2


fd:
* [Tests](https://github.com/fd/httpmiddlewarevet/blob/master/vet_handler_go18.go)
* [Results](https://travis-ci.org/fd/httpmiddlewarevet/jobs/195139656#L191-L193)

The `http.ResponseWrite` passed to the inner handler doesn't implement `http.Flusher`, `http.CloseNotifier` and `http.Pusher` when using http/2.

beorn7:
InstrumentHandler is massively deprecated. There won't be any fixes or maintenance on it. We are already working on a replacement. I'll add this concern to the doc comment.

fd:
@beorn7 thanks. Is there a place where I can investigate the replacement?

beorn7:
It will show up in branches/PRs ASAP. The tracking issue is #224 .

jeromegn:
The tracked issue has been closed, but I'm still seeing issues.

```
2017/05/31 14:22:43 http2: panic serving 127.0.0.1:59913: interface conversion: *promhttp.pushDelegator is not http.CloseNotifier: missing method CloseNotify
goroutine 148 [running]:
net/http.(*http2serverConn).runHandler.func1(0xc420162058, 0xc42050bfaf, 0xc4203348c0)
  /usr/local/opt/go/libexec/src/net/http/h2_bundle.go:4592 +0x190
panic(0x15d9be0, 0xc4201db440)
  /usr/local/opt/go/libexec/src/runtime/panic.go:489 +0x2cf
github.com/<my-repo>/vendor/github.com/bugsnag/bugsnag-go.(*Notifier).AutoNotify(0xc420194020, 0xc4202ce1a0, 0x2, 0x2)
  /Users/jerome/go/src/github.com/<my-repo>/vendor/github.com/bugsnag/bugsnag-go/notifier.go:72 +0x138
panic(0x15d9be0, 0xc4201db440)
  /usr/local/opt/go/libexec/src/runtime/panic.go:489 +0x2cf
github.com/<my-repo>/vendor/github.com/romainmenke/pusher/link.(*responseWriter).CloseNotify(0xc4202ce060, 0x19823a0)
  /Users/jerome/go/src/github.com/<my-repo>/vendor/github.com/romainmenke/pusher/link/response_writer_go18.go:70 +0x43
[...]

PR #308, opened 2017-06-02
promhttp: Bite the bullet and implement all 32 possible interface combos

beorn7:
@stuartnelson3 This is crazy…

As it turned out, it's not that esay to guess "common" combination of
interface upgrades. So I decided to just implement all 32 possible
combination of interface upgrades. (Only 16 with Go 1.7 and earlier.)

Clearly, this calls for code generation. But right now, we still need
to find out what's the best form of the code. For later additions,
implementing code generation might be useful.

Note that newDelegator is called for each HTTP request. Thus, this
commit aims to make the upgrade selection quick. (After the type
checks, it's just directly accessing an element in a slice.)

### DIFF
--- a/prometheus/promhttp/delegator_1_8.go
+++ b/prometheus/promhttp/delegator_1_8.go
@@ -20,54 +20,162 @@ import (
 	"net/http"
 )
 
-// newDelegator handles the four different methods of upgrading a
-// http.ResponseWriter to delegator.
+type pusherDelegator struct{ *responseWriterDelegator }
+
+func (d *pusherDelegator) Push(target string, opts *http.PushOptions) error {
+	return d.ResponseWriter.(http.Pusher).Push(target, opts)
+}
+
+func init() {
+	pickDelegator[pusher] = func(d *responseWriterDelegator) delegator { // 16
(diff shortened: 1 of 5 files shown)


beorn7:
Fixes #272.

prometheus/promhttp/delegator_1_8.go
```diff
@@ -20,54 +20,162 @@ import (
...
+		id += flusher
+	}
+	if _, ok := w.(http.Hijacker); ok {
+		id += hijacker
+	}
+	if _, ok := w.(io.ReaderFrom); ok {
+		id += readerFrom
+	}
+	if _, ok := w.(http.Pusher); ok {
+		id += pusher
```
stuartnelson3: clever

beorn7:
> Thanks for doing this. I doubt the overhead is very much, since it is only incrementing a number and then indexing into the slice of delegators that's created initially.

Yes, that's the idea: Minimize the overhead at runtime by having a lot of static code. What I did manually here is similar to things the C++ compiler does based on templates, so you can imagine where the infamous binary code bloat of C++ comes from.

PR #311, opened 2017-06-15
calculate size of headers outside of gorouting

yangyuqian:
@beorn7 

Calculate headers concurrently in `computeApproximateRequestSize` concurrently might lead to crash due to limitation in
hashmap of Go.

For instance, in `src/net/http/httputil/reverseproxy.go`, headers will
be modified to ensure consistency, and if calculate(iterate on) headers
inside a separated goroutine, program will crash due to concurrently
read/write on hashmaps.

```
copiedHeaders := false
for _, h := range hopHeaders {
	if outreq.Header.Get(h) != "" {
		if !copiedHeaders {
			outreq.Header = make(http.Header)
			copyHeader(outreq.Header, req.Header)
			copiedHeaders = true
		}
		outreq.Header.Del(h)
	}
}
```

### DIFF
--- a/prometheus/http.go
+++ b/prometheus/http.go
@@ -325,17 +325,35 @@ func computeApproximateRequestSize(r *http.Request) <-chan int {
 		s += len(r.URL.String())
 	}
 
+	// Since headers can be modified in http handlers, for instance,
+	// if iterate headers in a separated gorouting, the program might crash
+	// when headers(hashmap) is write/read concurrently.
+	// ```
+	// # src/net/http/httputil/reverseproxy.go@1.7.5
+	// copiedHeaders := false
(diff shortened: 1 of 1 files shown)


yangyuqian:
@beorn7 This PR helps to use prometheus metrics in the reverse proxy I'm working on, and below is the screenshot of error message:

image

brian-brazil:
This bit of the library is deprecated for reasons like this. It's recommended to use the promhttp package.

yangyuqian:
I see, trying the recommend lib instead, thank you!

beorn7:
Indeed. We'll remove this code once 0.9 has been released.

yangyuqian:
Thank you!

Issue #222, opened 2016-08-16
Rework Desc and ConstMetrics


beorn7:
* Desc needs to also include bucket layout and type.
* Desc creation needs to be easier. Consider creating a Desc from an Opts instance, give it a Desc() method. Give Desc a method AddLabelNames.
* Remove NewInvalidDesc, not needed anywhere (especially with the option of not returning a Desc in Describe at all).
* ConstMetric creation can then be streamlined: no type parameter needed anymore. Could also take Opts directly but that has a higher performance cost as a Desc has some things precalculated. Needs vetting.

beorn7:
This should, after all, go into 0.10, as it will break a lot of code.

beorn7:
With the bucket layout in the Desc, validating the bucket layout can be part of validating the Desc, i.e. invalid bucket layouts will show up at registration time. Currently, a panic is triggered during NewHistogram (which is OK) but in a HistogramVec only at the point a new histogram element in the histogram vector is created, i.e. during the WithLabels etc. call. That's sometimes hard to spot, especially if it occurs in the HTTP serving path where panics are recovered by the HTTP library.
@stuartnelson3 FYI

brian-brazil:
I'd suggest a look at https://github.com/prometheus/client_python#custom-collectors CounterMetricFamily and friends that Python/Java has. It makes the simple cases really easy for custom collectors.

beorn7:
Note to self to consider #516 to make Desc's useful for generating documentation.

Issue #47, opened 2015-01-12
Allow more flexible metric exposition.


beorn7:
With the current state of affairs, a collector has to know exactly the names of the metrics (including the label names for each) it will expose. The only way around that is the SetMetricFamilyInjectionHook function. However, it requires the user to assemble a dto.MetricFamily protobuf, i.e. it is very low-level.

Two approaches come to my mind (which we might want to implement both):

(1) A "prefix" registration: A collector can register a whole namespace (by prefix, could be simply implemented by putting "prefix*" as the name in the descriptor). It would still register for a fixed set of label names. That registration would disallow any other registration matching the prefix. With this approach, full consistency can still be guaranteed. The planned scrape of selected metrics would be a bit more expensive as we have to call the collector whenever a metric is requested that matches the prefix, but that's fundamentally unavoidable if we have dynamic metric names.

(2) An "unchecked" registration: A collector is allowed to collect _anything_ (possibly with the restriction on a name prefix aka namespace). Since metrics with any label set could be collected, consistency guarantees are gone. This would be similar to the SetMetricFamilyInjectionHook, but on a higher level, i.e. the user can collect normal Metric implementations and does not have to bother with low-level protobufs.

beorn7:
@brian-brazil Does that match your requirements?

beorn7:
@grobie @discordianfish You had a use-case for that in the HAProxy expoter, didn't you?

brian-brazil:
That sounds good. I'd suggest combining 1) and 2) by allowing an empty prefix.

Is there any way to let the collector know which names are being requested? That could allow it optimise if if needs to (e.g. if this was ever added for the JMX exporter, you'd definitely want that).

beorn7:
Definitely allow empty prefixes, but that's not automatically a combination of 1 and 2. 1 enforces a fixed set of label names, while 2 allows leaves it to the responsibility of the user to create consistent label name sets (i.e. the same metric name only ever having the same set of label names). In other word: The difference between 1 and 2 is not 'prefix or not' but 'fixed set of label names or not'. Both solution will deal with prefixes including the empty prefix.

Currently no plans to give the Collector a list of names to collect. That would require the collector to replicate logic that currently is (and has to remain) in the registry. So far, I would simply call all the collectors that _could_ expose a metric of the requested name, and then filter them before returning them over the wire. If you want to avoid needless collector calls, you must register collectors that have only one descriptor with a fully-qualified name (i.e. non-prefix).

[...]

PR #326, opened 2017-07-20
Add getter methods for Desc() struct

jacksontj:
The fields of the Desc struct are unexported so other packages can't
reach in and change the value, but sometimes it is useful to *read* the
value of these fields (such as writing a registry/gatherer/etc.)

### DIFF
--- a/prometheus/desc.go
+++ b/prometheus/desc.go
@@ -194,6 +194,13 @@ func (d *Desc) String() string {
 	)
 }
 
+// Expose some number of internal fields through methods. This allows us to 
+// keep these fields as unexported (so others cannot change them) -- but allows
+// others to view them (for creating other registries/collectors/etc.)
+func (d *Desc) FQName() string { return d.fqName }
+func (d *Desc) Help() string   { return d.help }
+func (d *Desc) Err() error     { return d.err }
+
 func checkLabelName(l string) bool {
 	return model.LabelName(l).IsValid() &&
 		!strings.HasPrefix(l, reservedLabelPrefix)


jacksontj:
cc: @beorn7

beorn7:
I'm reworking this for https://github.com/prometheus/client_golang/issues/222 . I don't want to expose an internal structure in the main repo that is doomed to get changed in a couple of months.

I hope you understand. If you really need this in the meantime, feel free to work off a fork.

beorn7:
I have read a bit of the context now.

Perhaps it is interesting to know that I'm also working on https://github.com/prometheus/client_golang/issues/47 to allow more lenient metrics exposition. However, the whole library is designed to not ever allow the exposition of inconsistent metrics. There are various ways of dealing with those errors, but if this PR is meant to facilitate writing registries that allow that, then I'm afraid that's completely against the basic principle of this library and Prometheus as a whole.

jacksontj:
Adding getters() here facilitates creating *any* registry that isn't part of the prometheus client library package. Now if you want to implement the `Gatherer` interface you have to maintain a map of `Desc` -> Metric, since the `Gather()` method needs to return `([]*dto.MetricFamily, error)`

As for the comment of

> then I'm afraid that's completely against the basic principle of this library and Prometheus as a whole.

This is not necessarily true. The goal here is to facilitate migrations from other metrics systems (granted with other rules) to prometheus. The statsd_exporter (which I'm primarily doing this work for) is a wonderful mechanism to get people into prometheus-- as its a low-cost, low-effort mechanism to try out prometheus.


Thanks for the quick reply, hopefully we can get this (or something like it) in. If not I can work around this in the statsd_exporter (the PR is linked). IMO this PR is helpful if anyone wants to implement the Gatherer/Registerer interfaces-- which presumably is "supported" since they are interfaces.

brian-brazil:
> The goal here is to facilitate migrations from other metrics systems (granted with other rules) to prometheus.

You should not need to make any changes to a client library's Registry to do this, the existing Collector interface should cover you as it permits ingestion of arbitrary valid data. Usually when someone feels the need to rewrite the Registry, they've missed this aspect of the API design.

beorn7:
There are many ways (and use cases) to implement `Gatherer` and `Registerer` without a need to meddle with `Desc` internals. If you want to go all low level, you could even create `dto.MetricFamily` protobuf messages on your own and return them.

However, my immediate concern here is that the `Desc` internals are not stable at the moment. Once they are, we can revisit your use case and find the best way to deal with it.

PR #327, opened 2017-07-20
Make public

jacksontj:


beorn7:
I think I even had this exported some time ago, but then no use case outside of the package showed up. This library suffers from a lot of types and is quite confusing already. If there is really a need for this to be in a library, I'd go for a separate utility package for "power user" needs.

jacksontj:
So (to summarize what I'm reading) -- the preference would be to move those into a sub-package here inside this client? If so-- definitely can (and will) do so.

beorn7:
Since I made it unexported because nobody was using it, how about first creating your own package somewhere. Then we'll see how it is used, and if it turns out to be generally useful after all, we'll move it into this repo. Perhaps by then, there will be a number of "power user" utilities that we can bundle. Right now, I'd be cautious to create a separate package with one type with one somewhat dubious use case (see the other discussions around the statsd_exporter issue).

The more general picture here is that I'm trying to streamline client_golang a bit. It's quite confusing right now, but there are also many partially conflicting use cases or just modes of using it, which each having their own justification and need to be catered for one way or the other. It's hard at this time to also add the more fringe use-cases to the mix. I'd like to keep those separate for now and reconsider once the core has stabilized. (And believe me, this is a lengthy process with a lot of context. I'm trying to find the right trade-offs since 2013...)

beorn7:
Closing this for now.

PR #332, opened 2017-08-03
supports batch add oberserations to histogram and summary

fortrue:


brian-brazil:
What is your goal here?

We need to know the individual values.

fortrue:
I have a proxy(https://github.com/fortrue/predixy) for redis, the proxy record latency like histogram in prometheus.
For example:
LatencyMonitorName:all
            latency(us)   sum(us)           counts
<=          100              3769836            91339 91.34%
<=          200               777185             5900 97.24%
<=          300               287565             1181 98.42%
<=          400               185891              537 98.96%
<=          500               132773              299 99.26%
<=          600                85050              156 99.41%
<=          700                85455              133 99.54%
<=          800                40088               54 99.60%
<=         1000                67788               77 99.68%
>          1000               601012              325 100.00%
T            60              6032643           100001
The last line is total summary, 624 is average latency(us)
Now, I want to write an exporter for the proxy, I can't observe every value individual, it's too slow and waste CPU. So, ObserveBatch is good idea.

brian-brazil:
What you want here is a ConstMetric to pass that histogram on to Prometheus. The direct instrumentation histogram API isn't appropriate for this use case.

fortrue:
@brian-brazil Thank you so much.
ConstHistogram is work for my case.

PR #333, opened 2017-08-11
Alert Manager API

au-phiware:
This is a reboot of the 9 month old #263 PR. It builds on that work by adding the `/status` endpoint and reorganises the code so that it is a little DRYer and the package name matches the import path.

Feedback welcome.

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -38,6 +38,8 @@ const (
 	epQueryRange  = apiPrefix + "/query_range"
 	epLabelValues = apiPrefix + "/label/:name/values"
 	epSeries      = apiPrefix + "/series"
+
+	epAlertManagers = apiPrefix + "/alertmanagers"
 )
 
 // ErrorType models the different API error types.
@@ -70,6 +72,11 @@ type Range struct {
 	Step time.Duration
 }
 
+type AlertManager struct {
+	// URL of alert manager's alerts endpoint
+	URL string `json:"url"`
+}
+
 // API provides bindings for Prometheus's v1 API.
 type API interface {
 	// Query performs a query for the given time.
@@ -78,6 +85,9 @@ type API interface {
 	QueryRange(ctx context.Context, query string, r Range) (model.Value, error)
 	// LabelValues performs a query for the values of the given label.
 	LabelValues(ctx context.Context, label string) (model.LabelValues, error)
+
+	// AlertManagers retrieves the list of active alert managers.
+	AlertManagers(ctx context.Context) (AlertManagersResult, error)
 }
 
 // queryResult contains result data for a query.
@@ -87,6 +97,11 @@ type queryResult struct {
 
 	// The decoded value.
 	v model.Value
+}
+
+// AlertManagersResult contains result data for a request to alertmanagers
+type AlertManagersResult struct {
+	ActiveAlertManagers []AlertManager `json:"activeAlertmanagers"`
 }
 
 func (qr *queryResult) UnmarshalJSON(b []byte) error {
@@ -142,7 +157,7 @@ func (h *httpAPI) Query(ctx context.Context, query string, ts time.Time) (model.
 
 	u.RawQuery = q.Encode()
 
-	req, err := http.NewRequest("GET", u.String(), nil)
+	req, err := http.NewRequest(http.MethodGet, u.String(), nil)
 	if err != nil {
 		return nil, err
 	}
@@ -175,7 +190,7 @@ func (h *httpAPI) QueryRange(ctx context.Context, query string, r Range) (model.
 
 	u.RawQuery = q.Encode()
 
-	req, err := http.NewRequest("GET", u.String(), nil)
+	req, err := http.NewRequest(http.MethodGet, u.String(), nil)
 	if err != nil {
 		return nil, err
 	}
@@ -204,6 +219,21 @@ func (h *httpAPI) LabelValues(ctx context.Context, label string) (model.LabelVal
 	var labelValues model.LabelValues
 	err = json.Unmarshal(body, &labelValues)
 	return labelValues, err
+}
+
+func (h *httpAPI) AlertManagers(ctx context.Context) (AlertManagersResult, error) {
+	var amResult AlertManagersResult
+	u := h.client.URL(epAlertManagers, nil)
+	req, err := http.NewRequest(http.MethodGet, u.String(), nil)
+	if err != nil {
+		return amResult, err
+	}
+	_, body, err := h.client.Do(ctx, req)
+	if err != nil {
+		return amResult, err
+	}
+	err = json.Unmarshal(body, &amResult)
+	return amResult, err
 }
 
 // apiClient wraps a regular client and processes successful API responses.
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -42,28 +42,6 @@ const (
 	epAlertManagers = apiPrefix + "/alertmanagers"
 )
 
-// ErrorType models the different API error types.
-type ErrorType string
-
-// Possible values for ErrorType.
-const (
-	ErrBadData     ErrorType = "bad_data"
-	ErrTimeout               = "timeout"
-	ErrCanceled              = "canceled"
-	ErrExec                  = "execution"
-	ErrBadResponse           = "bad_response"
-)
-
-// Error is an error returned by the API.
-type Error struct {
-	Type ErrorType
-	Msg  string
-}
-
-func (e *Error) Error() string {
-	return fmt.Sprintf("%s: %s", e.Type, e.Msg)
-}
-
 // Range represents a sliced time range.
 type Range struct {
 	// The boundaries of the time range.
@@ -245,7 +223,7 @@ type apiClient struct {
 type apiResponse struct {
 	Status    string          `json:"status"`
 	Data      json.RawMessage `json:"data"`
-	ErrorType ErrorType       `json:"errorType"`
+	ErrorType api.ErrorType   `json:"errorType"`
 	Error     string          `json:"error"`
 }
 
@@ -258,8 +236,8 @@ func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, [
 	code := resp.StatusCode
 
 	if code/100 != 2 && code != statusAPIError {
-		return resp, body, &Error{
-			Type: ErrBadResponse,
+		return resp, body, &api.Error{
+			Type: api.ErrBadResponse,
 			Msg:  fmt.Sprintf("bad response code %d", resp.StatusCode),
 		}
 	}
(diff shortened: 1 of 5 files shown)


beorn7:
@fabxc you were commenting on the original PR.
@stuartnelson3 you might have an opinion on this.

api/alertmanager/v1/alertmanager/api.go
```diff
@@ -0,0 +1,280 @@
...
+		method string
+	)
+
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(sil); err != nil {
+		return 0, err
+	}
+
+	// Talk to different endpoints depending on whether its a new silence or not.
+	if sil.ID != 0 {
```
stuartnelson3: This will panic if `sil == nil`

api/alertmanager/v1/alertmanager/api.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+type httpStatusAPI struct {
+	client api.Client
+}
+
+func (h *httpStatusAPI) Get(ctx context.Context) (*ServerStatus, error) {
+	u := h.client.URL(epStatus, nil)
+
+	req, _ := http.NewRequest("GET", u.String(), nil)
```
stuartnelson3: You're using `http.MethodGet` below, would be nice to be consistent with that.

api/alertmanager/v1/alertmanager/api.go
```diff
@@ -0,0 +1,280 @@
...
+
+func (h *httpAlertAPI) Push(ctx context.Context, alerts ...*model.Alert) error {
+	u := h.client.URL(epAlerts, nil)
+
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(alerts); err != nil {
+		return err
+	}
+
+	req, _ := http.NewRequest("POST", u.String(), &buf)
```
stuartnelson3: same thing, stdlib const would be nice

api/alertmanager/v1/alertmanager/api.go
```diff
@@ -0,0 +1,280 @@
...
+type httpSilenceAPI struct {
+	client api.Client
+}
+
+func (h *httpSilenceAPI) Get(ctx context.Context, id uint64) (*model.Silence, error) {
+	u := h.client.URL(epSilence, map[string]string{
+		"id": strconv.FormatUint(id, 10),
+	})
+
+	req, _ := http.NewRequest("GET", u.String(), nil)
```
stuartnelson3: stdlib const

api/alertmanager/v1/alertmanager/api.go
```diff
@@ -0,0 +1,280 @@
...
+
+	return &sil, err
+}
+
+func (h *httpSilenceAPI) Del(ctx context.Context, id uint64) error {
+	u := h.client.URL(epSilence, map[string]string{
+		"id": strconv.FormatUint(id, 10),
+	})
+
+	req, _ := http.NewRequest("DELETE", u.String(), nil)
```
stuartnelson3: stdlib const

api/alertmanager/v1/alertmanager/api.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// Talk to different endpoints depending on whether its a new silence or not.
+	if sil.ID != 0 {
+		u = h.client.URL(epSilence, map[string]string{
+			"id": strconv.FormatUint(sil.ID, 10),
+		})
+		method = "PUT"
+	} else {
+		u = h.client.URL(epSilences, nil)
+		method = "POST"
```
stuartnelson3: const for these two

api/alertmanager/v1/alertmanager/api.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	err = json.Unmarshal(body, &res)
+
+	return res.SilenceID, err
+}
+
+func (h *httpSilenceAPI) List(ctx context.Context) ([]*model.Silence, error) {
+	u := h.client.URL(epSilences, nil)
+
+	req, _ := http.NewRequest("GET", u.String(), nil)
```
stuartnelson3: here also

api/alertmanager/v1/alertmanager/api.go
au-phiware: I'm of the opinion that `sil` ought to be a value (not a pointer). Imagine if some goroutine also had the `sil` address and was messing with its properties during the `Encode` above; it would leave the request in an inconsistent state (i.e. not in the user's intention at the point that `Set` was called). Agreed?

stuartnelson3: No complaints here. Feel free to update the code.

au-phiware:
Apologies for the extended delay in making these small changes. I don't understand the problem with the go1.6.3 travis build... I'm happy to take any other suggestions to improve this PR.

beorn7:
The 1.6.3 error is caused by the `context` package only being part of the standard library in 1.7+.

I don't think we have very clear guidelines how far back the `client_golang` repo should support Go versions. But so far, we were testing the last three (1.6, 1.7, 1.8). Now that 1.9 is out, we could drop the 1.6 support and go with the `context` package from the standard library. What does everybody think? I'll create a separate PR for discussion.

Besides this meta issue, I leave the detailed review to @stuartnelson3 .

beorn7:
If you rebase or merge this on top of master, the CI test should succeed.

au-phiware:
Okay, rebased onto master, all tests are passing. Sorry about the churn on the commits, would you prefer if I squashed them together?

stuartnelson3:
They can be squashed when we merge

api/alertmanager/v1/alertmanager/api.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// SilenceAPI provides bindings the Alertmanager's silence API.
+type SilenceAPI interface {
+	// Get returns the silence associated with the given ID.
+	Get(ctx context.Context, id uint64) (*model.Silence, error)
+	// Set updates or creates the given silence and returns its ID.
+	Set(ctx context.Context, sil model.Silence) (uint64, error)
+	// Del deletes the silence with the given ID.
+	Del(ctx context.Context, id uint64) error
```
grobie: I think this should be `Delete`.

api/prometheus/v1/api.go
```diff
@@ -206,6 +199,21 @@ func (h *httpAPI) LabelValues(ctx context.Context, label string) (model.LabelVal
 	return labelValues, err
 }
 
+func (h *httpAPI) ActiveAlertManagers(ctx context.Context) ([]AlertManager, error) {
```
grobie: Hmm, I feel about unsure about adding this method instead of a plain `AlertManagers`. As soon as that API endpoint returns more than just the active alert managers, we'll have to add that method anyway. While I agree that the current API response is a bit cumbersome, adding `ActiveAlertmanagers` unnecessarily clutters the interface.

au-phiware:
Thanks for your feedback @grobie (also thanks for noticing the history of the package name, I see the rationale now and will change the name to v1). I'm happy to incorporate all your feedback, but you seem unclear on your last point. My opinion is that it should match the http endpoint (i.e. ` alertmanagers`) with maybe a case change (i.e. `AlertManager`).

grobie:
Thanks for all the changes @au-phiware!

I see that my wording was not clear, sorry for that. IMHO we should just follow the naming decisions of the Prometheus API and have a method `AlertManagers` returning a struct. I'm fine with your `AlertManagersResult` naming.

An `ActiveAlertManagers` function would increase the `API` interface signature and be the first function in that interface not being related to an API endpoint/path. I'd avoid that unless really necessary.

Regarding squashing, we def. try to have a clean commit history in Prometheus projects. While @stuartnelson3 is right that we Github offers the option to squash&merge, that means your changes to the Prometheus API and alertmanager API would be squashed together. Please don't spend too much time on this, but in case you're able to quickly squash together your changes into a few meaningful commits (maybe move the error type to the main api package, add the alertmanager API, add the alertmanager function to the prometheus API?), that would be greatly appreciated.

au-phiware:
No problems, I'm happy to contribute. My only concern is that there's no tests over the new functionality... I guess that will come with time.

stuartnelson3:
@grobie do you want to handle this PR? Seems like you've got it mostly how you want it.

stuartnelson3:
Had a quick chat with @grobie, it's still on me.

stuartnelson3:
@au-phiware tests are definitely something that will be required for this PR. Since you're contributing the code, you're also on the hook for the tests :)

au-phiware:
> Since you're contributing the code, you're also on the hook for the tests :)

Naw, a little help from @fabxc wouldn't go astray, that's where most of the code has originated :)

stuartnelson3:
it's very unlikely @fabxc will have time to contribute to this since he's focusing on prom2.0 at the moment.

We're happy to see this PR rebooted, and would really appreciate it if you could finish it and contribute some tests.

au-phiware:
Yes, it's no problem, I just can't be sure when I can find the time.

I did take a quick look at api_test.go and I'm a little confused: is `apiClientTest` testing the client?

I'll use `apiTestClient` as examples to do the mocking, fixtures and test table. Sound good?

stuartnelson3:
Yeah, it looks like `apiClientTest` is holding the both the data that gets fed to the `Do()` method (https://github.com/RS3000/client_golang/blob/b8b705b64ed375ec8a378e135ef5b945b9bd9297/api/prometheus/v1/api_test.go#L242), and also expected results arising from it.

It is pretty confusing ... if you think of a way to write some tests that are easier to follow, I won't complain.

josedonizetti:
@stuartnelson3 @beorn7 Is this something you want to maintain on this project?

I want to clean up the amtool by having a proper alertmanager client tested, the amtool would access this client instead of writing the access code itself (untested). My first goal was to code the client for alertmanager on AM itself because I don't see much use cases for an external client lib for it. @stuartnelson3 pointed me out this PR.

So I could either help out on here by writing tests so this can be merged. 
Or we can move the code to AM (and write tests), and not mix the alertmanager client stuff on this prometheus lib. What do you think?

stuartnelson3:
From my side, I think it would be cool to have the client code live here, and then amtool imports the package. @beorn7 is the official maintainer so I defer to his judgment.

beorn7:
In very general, I'm super relaxed about the exact location of those API Go packages. It's certainly not unreasonable to let them all live in the prometheus/client_golang GitHub repo. However, that doesn't mean I would be able to maintain them all. We already have the problem with `api/prometheus/`, which  is borderline abandonware, and where I can just try to delegate questions and PRs on a best-effort basis.

@stuartnelson3 If you could take some responsibility for `api/alertmanager`, I'm all up for having it here.

stuartnelson3:
I think that makes sense, and any issues can be directed to me.

beorn7:
@stuartnelson3 then feel free to finish this review and merge at your own discretion.

beorn7:
Superseded by #333

beorn7:
Obviously, not self-superseding. That should have gone into the other PR.

In different news, @stuartnelson3 , did you drop the ball here?

beorn7:
@stuartnelson3 friendly one-month-later ping…

stuartnelson3:
thanks for the ping, i lost track of this over the holidays

stuartnelson3:
Ok, I took a look and tried using the library. It's using the wrong domain types for interacting with alertmanager (it's using `common/model` instead of `alertmanager/types`), and this coming very close to implementing the api for the amtool cli.

I think it would make sense to have the amtool use the lib for its interactions. The question I'm trying to answer now is if it makes more sense to move this library layer into alertmanager, or have it here and have `amtool` import it (which I feel kind of weird about, since this code will be importing types from alertmanager).

@beorn7 do you have a strong vote one way or another, as a library author? My inclination is to move this into alertmanager after having taken a look and given it a modicum of thought.

beorn7:
I don't have a strong opinion here. It wasn't my idea in the first place to put API clients in this repo.

What you said above smells like running into circular dependencies. Intuitively, I'd prefer having it all in the alertmanager repo, but I don't feel I'm the most qualified person to make that call.

stuartnelson3:
>What you said above smells like running into circular dependencies.

Pretty much my concern. I'll re-resurrect this in alertmanager.

PR #335, opened 2017-08-19
Check validity of label values

### LINKED ISSUE #274: Check validity of label values.
Label values can be any UTF-8 string. However, not any arbitrary byte sequence is a valid UTF-8 string. (FF bytes are always illegal, as an example.)

The Prometheus server validates label values upon ingestion. Thus, the client library should prevent invalid label from being exposed.

https://github.com/prometheus/snmp_exporter/issues/117 exposed this problem.

marco-jantke:
This PR fixes #274.

@beorn7 I was not sure whether I should cover all similar implementation with individual unit tests. Maybe have a look first and tell me if you think we need more tests or on different places. 

In regards to the `Gather`ing part, at the moment an invalid label value will make the whole `Gather` run fail. Might it be better to be not so restrictive and only remove the invalid metrics or should it be this strict in your opinion?


beorn7:
Looks good at a first glance. Will do a more detailed review ASAP.

prometheus/registry.go
```diff
@@ -23,6 +23,8 @@ import (
 
 	"github.com/golang/protobuf/proto"
 
+	"unicode/utf8"
```
beorn7: Standard library imports should go into the first block.

marco-jantke:
Ok, feedback is implemented.

prometheus/labels.go
```diff
@@ -0,0 +1,57 @@
...
+	for name, val := range labels {
+		if !utf8.ValidString(val) {
+			return fmt.Errorf("label %s: value %q is not valid UTF-8", name, val)
+		}
+	}
+
+	return nil
+}
+
+func validateValuesInLabels(vals []string, expectedNumberOfValues int) error {
```
beorn7: You confused the two functions here. This was called `validateLabelValues`, which was a fine name. `validateValuesInLabels` should be the new name of `validateLabels` above.

prometheus/labels.go
```diff
@@ -0,0 +1,57 @@
...
+// create a Desc.
+type Labels map[string]string
+
+// reservedLabelPrefix is a prefix which is not legal in user-supplied
+// label names.
+const reservedLabelPrefix = "__"
+
+var errInconsistentCardinality = errors.New("inconsistent label cardinality")
+
+func validateLabels(labels Labels, expectedNumberOfValues int) error {
```
beorn7: This should be called `validateValuesInLabels`. See next comment below.

marco-jantke:
Ah gotcha, thanks for the notice!

beorn7:
Cool. Thank you. Merging now…

PR #340, opened 2017-08-29
Remove the deprecated uses of Untyped metrics

beorn7:
I couldn't find any use of the removed identifiers on sourcegraph.com
(outside of this repo itself).

@grobie Just a tiny review.


brian-brazil:
:+1:

PR #342, opened 2017-08-31
http.go: incorrect error message displayed when no metrics encoded error

wrouesnel:
The "No metrics encoded" error was erroneously displayed the value of err, not lastErr.

It looks like this could potentially have caused a panic in this error condition, since `err` is never set in scope.


brian-brazil:
Thanks!

PR #343, opened 2017-08-31
Fix error reporting bug

beorn7:
That's the same bug fixed in #342 (which is the deprecated code only
replicated in the `prometheus` package until v0.10 is out).

@brian-brazil @wrouesnel


brian-brazil:
:+1:

PR #345, opened 2017-09-05
Correct DefaultRegistry to DefaultRegisterer

mattbostock:
`DefaultRegistry` no longer exists as of a6321dd0b.

/cc @beorn7


beorn7:
Thanks. (Reviewing while listening to your PromCon talk... Weird. ;)

PR #346, opened 2017-09-06
Drop Go1.6 support, add test for Go1.9

beorn7:
Also, document the version requirement.

Triggered by #333 . @stuartnelson3 @grobie @fabxc any concerns about dropping 1.6 support?
I'd say support for three minor version ought to be enough for anybody. Supporting too many versions will be quite a maintenance overhead, as demonstrated in #333. (We also have separate code for 1.7 and 1.8+ right now, which can be dropped once Go1.10 is out.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ sudo: false
 language: go
 
 go:
- - 1.6.3
- - 1.7
- - 1.8.1
+ - 1.7.x
+ - 1.8.x
+ - 1.9.x
 
 script:
  - go test -short ./...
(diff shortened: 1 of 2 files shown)


stuartnelson3:
3 minor versions definitely seems like a long spread to me, as the 6month release cycle means users have a large time buffer. I would even be fine with just 2, personally. It seems like language change frequency is decreasing over time, so the overhead of supporting 3 vs. 2 versions is decreasing.

grobie:
:+1:

Issue #347, opened 2017-09-06
Push the simple and random examples to docker hub


dlespiau:
It'd be nice to be able to install those two examples with docker invocations so both an instrumented daemon and prometheus can be started this way. This would allow a full end to end example with two `docker run` and with a config file bound mounted, which seems fairly minimal.

The labels could be:
- `prometheus/example-golang-simple`
- `prometheus/example-golang-random`

I can prepare a PR building the docker images, but would be good to know if people are happy with the idea.

beorn7:
I don't feel strongly. @discordianfish does this make sense?

discordianfish:
Sounds good!

bergquist:
Whats the state on this? I would love to use those images in our Grafana dev environment. 
So far I can only find the simple example https://hub.docker.com/r/prom/client_golang_examples_simple/

beorn7:
See discussion on https://github.com/prometheus/client_golang/pull/348 . It looks we need to implement what was suggested last in the discussion on that (already merged) PR. I wanted to do it, but I have to admit I have way too many things on my plate that are much more important.
Happy to accept anything that improves things.

PR #348, opened 2017-09-11
Create Docker images for the examples

dlespiau:
When those are published on Dockerhub, people will be able to start those
examples in fewer steps than the current documentation:

  https://prometheus.io/docs/introduction/getting_started/#starting-up-some-sample-targets

While relying on Docker may not be always wished, giving the possibility to
quickly start a daemon that exposes metrics can be useful in many tutorials.

This commit tries really hard to limit the number of per-example maintenance
work by sharing most of the Makefile to build the Docker images. The tiny top
level Makefiles in each example directory are strictly identical.

I've also copied the way the main prometheus repo tags its Dockerimage in the Makefile, I assume something else is givin its final tag by overriding DOCKER_IMAGE_NAME and DOCKER_IMAGE_TAG somewhere.

Updates: https://github.com/prometheus/client_golang/issues/347

### DIFF
--- a/examples/random/Dockerfile
+++ b/examples/random/Dockerfile
@@ -1,0 +1,20 @@
+# This Dockerfile builds an image for a client_golang example.
+#
+# Use as (from the root for the client_golang repository):
+#    docker build -f examples/$name/Dockerfile -t prometheus/golang-example-$name .
+
+# Builder image, where we build the example.
+FROM golang:1.9.0 AS builder
+WORKDIR /go/src/github.com/prometheus/client_golang
+COPY . .
+WORKDIR /go/src/github.com/prometheus/client_golang/prometheus
+RUN go get -d
+WORKDIR /go/src/github.com/prometheus/client_golang/examples/random
+RUN CGO_ENABLED=0 GOOS=linux go build -a -tags netgo -ldflags '-w'
+
+# Final image.
+FROM scratch
+LABEL maintainer "The Prometheus Authors <prometheus-developers@googlegroups.com>"
+COPY --from=builder /go/src/github.com/prometheus/client_golang/examples/random .
+EXPOSE 8080
+ENTRYPOINT ["/random"]
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,10 @@
 _obj
 _test
 
+# Examples
+/examples/simple/simple
+/examples/random/random
+
 # Architecture specific extensions/prefixes
 *.[568vq]
 [568vq].out
--- a/examples/simple/Dockerfile
+++ b/examples/simple/Dockerfile
@@ -1,0 +1,20 @@
+# This Dockerfile builds an image for a client_golang example.
+#
+# Use as (from the root for the client_golang repository):
+#    docker build -f examples/$name/Dockerfile -t prometheus/golang-example-$name .
+
+# Builder image, where we build the example.
+FROM golang:1.9.0 AS builder
+WORKDIR /go/src/github.com/prometheus/client_golang
+COPY . .
+WORKDIR /go/src/github.com/prometheus/client_golang/prometheus
+RUN go get -d
+WORKDIR /go/src/github.com/prometheus/client_golang/examples/simple
+RUN CGO_ENABLED=0 GOOS=linux go build -a -tags netgo -ldflags '-w'
+
+# Final image.
+FROM scratch
+LABEL maintainer "The Prometheus Authors <prometheus-developers@googlegroups.com>"
+COPY --from=builder /go/src/github.com/prometheus/client_golang/examples/simple .
+EXPOSE 8080
+ENTRYPOINT ["/simple"]


examples/random/Makefile
```diff
@@ -0,0 +1,5 @@
+EXAMPLE_DIRECTORY := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
+name = $(patsubst %/,%,$(dir $1))
+EXAMPLE := $(notdir $(call name,$(EXAMPLE_DIRECTORY)))
+
+include $(EXAMPLE_DIRECTORY)/../Makefile
```
beorn7: Newline missing.

examples/simple/Makefile
```diff
@@ -0,0 +1,5 @@
+EXAMPLE_DIRECTORY := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
+name = $(patsubst %/,%,$(dir $1))
+EXAMPLE := $(notdir $(call name,$(EXAMPLE_DIRECTORY)))
+
+include $(EXAMPLE_DIRECTORY)/../Makefile
```
beorn7: Newline missing.

beorn7:
In case that wasn't clear above: I think you are using the right approach here. But happy to hear @discordianfish about those things.

dlespiau:
Forgot to mention that the examples are build as static binaries and packaged as
single binaries in Docker images.

PR Updated: Adding the missing end of line character at the end of 2 Makefiles.

dlespiau:
I'm not sure how you guys want to handle publishing to Dockerhub (automated vs manual) and if it's a good idea to push directly from travis (I would say, yes!). Can try to add something to push to Dockerhub in the Makefile when it's a bit clearer how publishing should be done.

discordianfish:
We could also just add a Dockerfile to build the examples and add this repo as automated build. That would be much simpler. With https://docs.docker.com/engine/userguide/eng-image/multistage-build/ we could also have a image with just the binary.

beorn7:
In general, I'd say let's keep it as simple as possible. The aim here is to allow Docker users to easily run the two examples. I don't think we need to update the build on each commit. (Perhaps on each release? If that's easily possible. But we will only have a normal release cadence once we are at 1.0+.)

If we did anything more involved, we should just do it in the same way as in other Prometheus repos with binary releases. However, we would then go down the same rabbit hole we went down before, adding vendoring just for the examples and such, which we did before and confused people because a library suddenly had vendoring.

discordianfish:
Agree on all points. I think adding a Dockerfile, then adding this repo to the Docker hub and/or Quay as automated build is the easiest way to do. It will get rebuild on every commit but why not? It's a no-op for us.

beorn7:
Sounds good. @dlespiau does this make sense to you?

dlespiau:
Yup! will rework the PR accordingly

examples/Makefile
```diff
@@ -0,0 +1,39 @@
...
+all: $(EXAMPLE) docker
+
+$(EXAMPLE): main.go Makefile
+	@echo ">> fetching dependencies"
+	@go get -d
+	@echo ">> building $@"
+	@CGO_ENABLED=0 GOOS=linux go build -a -tags netgo -ldflags '-w'
+
+define DockerfileTemplate
+FROM scratch
```
sdurrheimer: `scratch` works, but generally we tend to prefer to have a shell available, that's why we use busybox based images.

beorn7: On the other hand, this is just a tiny example binary...

dlespiau:
Finally took the time to get to this PR again.

I had to use two Dockerfiles (one for each example) but they are now self-contained. I don't believe there's a way to generate (and more importantly tag) two final images from the same Dockerfile. So we end up with a bit of duplication. To build the images:

```
$ docker build -t prom/golang-example-random:latest -f examples/random/Dockerfile .
```

I've kept `FROM scratch` instead of `FROM prom/busybox` because they're a some size difference. It's not that big though so it's debatable which base image should be taken:

```
random example, FROM scratch: 6MB
random example, FROM prom/busybox: 8.56MB
```

examples/random/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# This Dockerfile builds an image for a client_golang example.
+#
+# Use as (from the root for the client_golang repository):
+#    docker build -f examples/$name/Dockerfile -t prometheus/golang-example-$name .
+
+# The base image contains go, client_golang and client_golang dependencies to
```
discordianfish: I'd only use two images, the builder image to build everything and the final `from scratch` image.
Docker should still reuse pre-existing layers.

dlespiau:
Good point, changed it to use only 2 images.

examples/random/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# This Dockerfile builds an image for a client_golang example.
+#
+# Use as (from the root for the client_golang repository):
+#    docker build -f examples/$name/Dockerfile -t prometheus/golang-example-$name .
+
+# Builder image, where we build the example.
+FROM golang:1.9.0 AS builder
```
discordianfish: Maybe I miss something but wouldn't this be simpler?
```
FROM golang:1.9.0 AS builder
COPY . /go/src/github.com/prometheus/client_golang
WORKDIR /go/src/github.com/prometheus/client_golang/examples/random
RUN CGO_ENABLED=0 GOOS=linux go build -a -tags netgo -ldflags '-w'
```

discordianfish:
Beside that comment above, you could also create one Dockerfile in examples/ and just build and add both examples. But I think it's fine either way and would leave this to you and @beorn7 to decide.

examples/random/Dockerfile
dlespiau: That was to share the `go get -d` layer between images.

discordianfish: Ah I see what you mean.. Well, guess then it's fine. I personally would prefer a simpler, less optimized Dockerfile but really shouldn't matter.

discordianfish: So overall LGTM and leaving the final decision over that to Björn

beorn7:
I think it's fine now. Let's not overengineer this simple problem.

beorn7:
@discordianfish To create an automated build, I need to link the Docker-Hub account to a Gittub account. Since I'm not sure what the exact permissions implications are, I'd like to get your recommendation what combination of Docker Hub and Github accounts would make most sense.

beorn7:
@discordianfish I just went forward with using the prombot account for both.

However, I'm running into problems, see log below:

```
Building in Docker Cloud's infrastructure...
Cloning into '.'...

KernelVersion: 4.4.0-93-generic
Arch: amd64
BuildTime: 2017-08-17T22:50:04.828747906+00:00
ApiVersion: 1.30
Version: 17.06.1-ce
MinAPIVersion: 1.12
GitCommit: 874a737
Os: linux
GoVersion: go1.8.3
Starting build of index.docker.io/prombot/client_golang:latest...
Step 1/12 : FROM golang:1.9.0 AS builder

 ---> 1cdc81f11b10

Step 2/12 : WORKDIR /go/src/github.com/prometheus/client_golang

 ---> 008731d54420

Removing intermediate container 862206312b15

Step 3/12 : COPY . .

 ---> 92b5b82b1899

Removing intermediate container a67cad3717a3

Step 4/12 : WORKDIR /go/src/github.com/prometheus/client_golang/prometheus

 ---> 7ac9250bea0e

Removing intermediate container 256b7f9c81c4

Step 5/12 : RUN go get -d

 ---> Running in 708808b49e1b

[91mcan't load package: package github.com/prometheus/client_golang/prometheus: no Go files in /go/src/github.com/prometheus/client_golang/prometheus
[0m
Removing intermediate container 708808b49e1b

The command '/bin/sh -c go get -d' returned a non-zero code: 1
```

I'm fishing in the dark as I don't really have experience with this kind of things. My questions:
- Why are there no Go files in /go/src/github.com/prometheus/client_golang/prometheus? Didn't it properly clone the repo?
- Why is the `go get -d` in that directory anyway? Shouldn't it be done in the `examples/simple` or `examples/random` directory? (But of course, none of this will help if the repo is not cloned properly.)

discordianfish:
@beorn7 You need to run the docker build from the root of the repo. Manually that would be:

```
client_golang$ docker build -f examples/random/Dockerfile .
```

Similar setting should be in quay and docker hub.

beorn7:
My assumption was that building from the root of the repo is what an automated build is doing, because there is no setting for the directory to run Docker from.

Perhaps the build context is the same as the directory where the Dockerfile is?

Reading of docs didn't really get my anywhere. @discordianfish do you perhaps have any idea while looking at https://hub.docker.com/r/prom/client_golang_examples_simple/ (logged in as prombot)?

discordianfish:
The docker hub is so confusing.. Yes, it looks like it's not possible to specify the working directory in the 'classic' docker hub. But it should be possible in the docker 'store': https://store.docker.com/community/images/prom/client_golang_examples_simple
See: https://forums.docker.com/t/docker-cloud-release-notes-11-8-2016/24537

I would check myself but I don't seem to have acccess to any shared passwords (yet). @grobie Can you look into that? I've signed up when you sent around that mail back then but doesn't look like any secrets are shared with me.

grobie:
@discordianfish you're part of the Prometheus group and should have access. there isn't anything I can do here unfortunately.

discordianfish:
We figured the access out and I check but it really looks like that 'feature' isn't available on the docker hub but only on the docker store. And it seems like we need to sign up for the docker store which might make sense since it's the successor of the docker hub.
Alternatively we could merge the Dockerfile and put one in / which builds both examples.

dlespiau:
The problem with merging the dockerfiles is that we'll only be able to produce one tagged image (IIUC). So we'll have to put the two examples in the same image on two different ports?

beorn7:
That might be the easiest way. It also has the advantage of only one repo, not two different ones for a tiny example, each.

PR #349, opened 2017-09-12
Change summary docs

povilasv:
Improving docs to warn users about deprecation of `DefObjectives` in Summary type

@beorn7


beorn7:
Thanks for the catch.

Issue #230, opened 2016-08-19
Add test tools based on the new type structure


beorn7:
Once #223 is done, we can make use of it for better testing tools (like mocking helpers for the various interfaces we end up with).

marco-m:
Hello @beorn7 , I stumbled upon this ticket investigating about testability of Prometheus instrumentation. Are there any news/plans related to this ticket and #223 ? Thanks.

beorn7:
This is all planned for the v0.10 (soon to be renamed v2) milestone. I still need to finalize the grand design of it to enable detailed work like this one.

Have you looked at https://godoc.org/github.com/prometheus/client_golang/prometheus/testutil ? It should satisfy most of the current needs.

marco-m:
I will look at testutil, thanks for the prompt answer!

PR #350, opened 2017-09-15
Add the func of get value of gauge and counter

zcytop:
I think it's can be helpful to make tests or something.


beorn7:
We deliberately don't want getters on metrics, mostly because people would inevitable abuse metrics as goroutine-safe variables. Even the usefulness of getters for testing is somewhat disputed. See #223 and #230 and past discussions on the mailing list.

When resolving the bugs above, I'm planning to document them with some testing best practices.

Issue #275, opened 2017-02-09
histogram.Write creates invalid result when racing with observe


lmb:
Seems like the library will create invalid histograms when histogram.Observe races with histogram.Write.

```
$ go build && ./promrace
found race
sample_count: 9590
sample_sum: 18220.999999999192
bucket: <
  cumulative_count: 0
  upper_bound: 1
>
bucket: <
  cumulative_count: 9591
  upper_bound: 2
>
```

Note that sample_count is lower than cumulative_count of the highest bucket.

See https://gist.github.com/lmb/3caaeb79efc6d6e0d2bdde8a4865e636 to reproduce. It's possible to fix this by changing the order of the atomic adds, but that will then overstate the +Inf bucket by one in some cases.

beorn7:
Yeah, this is kind of by design. If we wanted to update all three, `sample_count`, `sample_sum`, and the respective bucket, we needed way more sophisticated locking. Which might be costly in terms of performance.

By incrementing the `sample_count` first, as you considered, we would at least make sure no bucket has a higher count than `sample_count`. Results look more consistent then, but are still incorrect.

I have some vague ideas with buffer swaps that could get proper atomicity of observation without excessive locking cost... will continue to think about it.

lmb:
Sorry for not replying earlier. To me it seems like changing the behaviour would be desirable, since then we at least generate histograms which are valid (but still not correct).

beorn7:
Yes, this is on the 0.9 milestone.

PR #359, opened 2017-11-09
add lock for value operation.

zhanglp92:
When I concurrent running program, has some wrong. I don‘t use metex with my code.
I think locks should be added here.

### DIFF
--- a/prometheus/value.go
+++ b/prometheus/value.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"math"
 	"sort"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


metalmatze:
Is this still necessary when using [atomic](https://golang.org/pkg/sync/atomic/) :thinking:

grobie:
No. client_golang is concurrency safe and uses the atomic package to reduce the footprint. Mutex are not only unnecessary here, they would incur a huge performance penalty.

Björn gave a talk explaining the design at the GopherCon 2015, here are the [slides](https://docs.google.com/presentation/d/1T_wdd2KGOPEBhHN6ubendIi87RhtadHOOSAwK0M-Exs/edit?usp=sharing).

beorn7:
The only concurrency issue I know of is https://github.com/prometheus/client_golang/issues/275 (for which I have a solution in my head, I just need to find time to write it down).

zhanglp92:
If you do not use the lock, there will be concurrency issue. 
eg: I concurrent implementation of the program.

func test() {
    gauge.Inc()
    defer gauge.Dec()
}

image

beorn7:
Could you provide the full code to replicate the problem? The above is not sufficient to do that.

PR #360, opened 2017-11-16
Omit zero value query time

icholy:


beorn7:
This looks sane to me, as https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries reads: “The current server time is used if the time parameter is omitted.”
@andrestc @fabxc As those having touched this code last, do you agree?

beorn7:
Thanks everyone.

PR #362, opened 2017-12-06
Added a faster counter

smcquay:
We would like to use something in the prometheus ecosystem with comparable performance to `github.com/rcrowley/go-metrics`. It seems as though the use of [`atomic.CompareAndSwap`](https://github.com/prometheus/client_golang/blob/master/prometheus/value.go#L97) slows things down a bit, so we've made a counter that stores and uses an `int64` directly and leans on `atomic.*Int64` for locking.

This branch introduces the `FixedPrecisionCounter` type and I'd like some feedback on the approach generally (and name) before I plumb through the vector version of this counter.

As to how I benchmarked performance I used the following test file:

```go
package fpc

import (
        "testing"

        metrics "github.com/rcrowley/go-metrics"
)

var fpc prometheus.Counter
var p prometheus.Counter

func init() {
        fpc = prometheus.NewFixedPrecisionCounter(prometheus.Opts{
                Name: "foo",
                Help: "helpful message about foo",
        }, 0)
        p = prometheus.NewCounter(prometheus.CounterOpts{Name: "pfoo", Help: "pfoo help"})
        prometheus.MustRegister(fpc)
        prometheus.MustRegister(p)
}

func BenchmarkGoMetrics(b *testing.B) {
        c := metrics.NewCounter()
        b.RunParallel(func(pb *testing.PB) {
                for pb.Next() {
                        c.Inc(1)
                }
        })
}

func BenchmarkPromCounter(b *testing.B) {
        b.RunParallel(func(pb *testing.PB) {
                for pb.Next() {
                        p.Inc()
                }
        })
}

func BenchmarkFixedPrecisionCounter(b *testing.B) {
        b.RunParallel(func(pb *testing.PB) {
                for pb.Next() {
                        fpc.Inc()
                }
        })
}
```

We get the following performance numbers:

```
go test -cpu=64 -benchtime=5s -v -bench=. github.com/smcquay/fpc
goos: darwin
goarch: amd64
pkg: github.com/smcquay/fpc
BenchmarkGoMetrics-64                   500000000               13.2 ns/op
BenchmarkPromCounter-64                 100000000               62.1 ns/op
BenchmarkFixedPrecisionCounter-64       300000000               22.5 ns/op
PASS
ok      github.com/smcquay/fpc  23.261s
```

### DIFF
--- a/prometheus/value.go
+++ b/prometheus/value.go
@@ -45,7 +45,6 @@ type value struct {
 	// to go first in the struct to guarantee alignment for atomic
 	// operations.  http://golang.org/pkg/sync/atomic/#pkg-note-BUG
 	valBits uint64
-
 	// valInt is used to store values that are exact integers
 	valInt int64
 
@@ -99,8 +98,9 @@ func (v *value) Dec() {
 }
 
 func (v *value) Add(val float64) {
-	if math.Trunc(val) == val {
-		atomic.AddInt64(&v.valInt, int64(val))
+	ival := int64(val)
+	if float64(ival) == val {
+		v.add(ival)
 		return
 	}
 
--- a/prometheus/value.go
+++ b/prometheus/value.go
@@ -46,6 +46,9 @@ type value struct {
 	// operations.  http://golang.org/pkg/sync/atomic/#pkg-note-BUG
 	valBits uint64
 
+	// valInt is used to store values that are exact integers
+	valInt int64
+
 	selfCollector
 
 	desc       *Desc
@@ -82,15 +85,25 @@ func (v *value) SetToCurrentTime() {
 	v.Set(float64(time.Now().UnixNano()) / 1e9)
 }
 
+// add adjusts the underlying int64
+func (v *value) add(delta int64) {
+	atomic.AddInt64(&v.valInt, delta)
+}
+
 func (v *value) Inc() {
-	v.Add(1)
+	v.add(1)
 }
 
 func (v *value) Dec() {
-	v.Add(-1)
+	v.add(-1)
 }
 
 func (v *value) Add(val float64) {
+	if math.Trunc(val) == val {
+		atomic.AddInt64(&v.valInt, int64(val))
+		return
+	}
+
 	for {
 		oldBits := atomic.LoadUint64(&v.valBits)
 		newBits := math.Float64bits(math.Float64frombits(oldBits) + val)
@@ -105,7 +118,10 @@ func (v *value) Sub(val float64) {
 }
 
 func (v *value) Write(out *dto.Metric) error {
-	val := math.Float64frombits(atomic.LoadUint64(&v.valBits))
+	fval := math.Float64frombits(atomic.LoadUint64(&v.valBits))
+	ival := atomic.LoadInt64(&v.valInt)
+	val := fval + float64(ival)
+
 	return populateMetric(v.valType, val, v.labelPairs, out)
 }
 
--- a/prometheus/counter_test.go
+++ b/prometheus/counter_test.go
@@ -126,3 +126,87 @@ func expectPanic(t *testing.T, op func(), errorMsg string) {
 
 	op()
 }
+
+func TestCounterAddInf(t *testing.T) {
+	counter := NewCounter(CounterOpts{
+		Name: "test",
+		Help: "test help",
+	}).(*counter)
+
+	counter.Inc()
+	if expected, got := 0.0, math.Float64frombits(counter.valBits); expected != got {
+		t.Errorf("Expected %f, got %f.", expected, got)
+	}
+	if expected, got := int64(1), counter.valInt; expected != got {
+		t.Errorf("Expected %f, got %f.", expected, got)
+	}
+
+	counter.Add(math.Inf(1))
+	if expected, got := math.Inf(1), math.Float64frombits(counter.valBits); expected != got {
+		t.Errorf("valBits expected %f, got %f.", expected, got)
+	}
+	if expected, got := int64(1), counter.valInt; expected != got {
+		t.Errorf("valInts expected %d, got %d.", expected, got)
+	}
+
+	counter.Inc()
+	if expected, got := math.Inf(1), math.Float64frombits(counter.valBits); expected != got {
+		t.Errorf("Expected %f, got %f.", expected, got)
+	}
+	if expected, got := int64(2), counter.valInt; expected != got {
+		t.Errorf("Expected %d, got %d.", expected, got)
+	}
+
+	m := &dto.Metric{}
+	counter.Write(m)
+
+	if expected, got := `counter:<value:inf > `, m.String(); expected != got {
+		t.Errorf("expected %q, got %q", expected, got)
+	}
+}
+
+func TestCounterAddLarge(t *testing.T) {
+	counter := NewCounter(CounterOpts{
+		Name: "test",
+		Help: "test help",
+	}).(*counter)
+
+	// large overflows the underlying type and should therefore be stored in valBits
+	large := float64(math.MaxInt64 + 1)
+	counter.Add(large)
+	if expected, got := large, math.Float64frombits(counter.valBits); expected != got {
+		t.Errorf("valBits expected %f, got %f.", expected, got)
+	}
+	if expected, got := int64(0), counter.valInt; expected != got {
+		t.Errorf("valInts expected %d, got %d.", expected, got)
+	}
+
+	m := &dto.Metric{}
+	counter.Write(m)
+
+	if expected, got := fmt.Sprintf("counter:<value:%0.15e > ", large), m.String(); expected != got {
+		t.Errorf("expected %q, got %q", expected, got)
+	}
+}
+
+func TestCounterAddSmall(t *testing.T) {
+	counter := NewCounter(CounterOpts{
+		Name: "test",
+		Help: "test help",
+	}).(*counter)
+	small := 0.000000000001
+	counter.Add(small)
+	if expected, got := small, math.Float64frombits(counter.valBits); expected != got {
+		t.Errorf("valBits expected %f, got %f.", expected, got)
+	}
+	if expected, got := int64(0), counter.valInt; expected != got {
+		t.Errorf("valInts expected %d, got %d.", expected, got)
+	}
+
+	m := &dto.Metric{}
+	counter.Write(m)
+
+	if expected, got := fmt.Sprintf("counter:<value:%0.0e > ", small), m.String(); expected != got {
+		t.Errorf("expected %q, got %q", expected, got)
+	}
+}
--- a/prometheus/counter_test.go
+++ b/prometheus/counter_test.go
@@ -28,11 +28,25 @@ func TestCounterAdd(t *testing.T) {
 		ConstLabels: Labels{"a": "1", "b": "2"},
 	}).(*counter)
 	counter.Inc()
-	if expected, got := 1., math.Float64frombits(counter.valBits); expected != got {
+	if expected, got := 0.0, math.Float64frombits(counter.valBits); expected != got {
+		t.Errorf("Expected %f, got %f.", expected, got)
+	}
+	if expected, got := int64(1), counter.valInt; expected != got {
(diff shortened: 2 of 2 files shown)


brian-brazil:
Prometheus only supports float64 internally, so this won't work.

beorn7:
@smcquay thanks for your contribution. I was tempted to add something like that long ago but then decided against it because of the additional cognitive load for the user for a performance improvement that is only relevant in very specific scenarios. Note that this had to be implemented for both counters and gauges (you incidentally implemented a fixed-precision gauge, not a counter), which adds a lot more top-level types to the already quite confusing package. And then users can easily get things wrong in subtle ways by picking the fixed-precision version when they should not have.

I would much prefer to first try our best to optimize within the bounds of the existing interfaces. For example, you could just count `Inc` calls in a separate integer and only add them to the final value in the `Write` call. `Add` calls with an effective integer (e.g. `Add(42.0)`) could use the same intermediate integer.

smcquay:
@beorn7 Thanks for the feedback. I am going to work a branch that adds the optimization you've suggested, but I'm happy moving this implementation into a separate package where people can use it if they want this flavor of optimization. However I'd like to use this new implementation with the prometheus client vector implementations, however due to unexported types I cannot conveniently use this external type with `metricVec`.

What are your thoughts about exporting this type? I understand the desire to keep the documentation as narrow as possible for users, and perhaps this is why `metricVec` is unexported, but would you think it reasonable to expose this type so that I can make my own `*Vec` types?

This approach would allow users to create their own completely decoupled `prometheus.Metric` implementations and still gain the benefits of using those within this client ecosystem.

Thoughts?

beorn7:
> What are your thoughts about exporting this type? I understand the desire to keep the documentation as narrow as possible for users, and perhaps this is why metricVec is unexported, but would you think it reasonable to expose this type so that I can make my own *Vec types?
>
> This approach would allow users to create their own completely decoupled prometheus.Metric implementations and still gain the benefits of using those within this client ecosystem.

I understand your need. And it is exactly the reason why `metricVec` used to be an exported type for quite some time. However, nobody ever used it, but I got feedback about how unwieldy the package is with the many exported type visible to the naive user (that only wants to do a few simple direct instrumentation and doesn't need most of them). So I started a multi-step effort to reduce the number of exported types. Which made `metricVec` the way it is.

Perhaps there is a way to create a separate package that would be some kind of implementation helper for developers like you. But it's not entirely trivial without creating circular package dependencies… I'd be delighted if you have an idea along these lines.

smcquay:
@beorn7 --

> Perhaps there is a way to create a separate package that would be some kind of implementation helper for developers like you. But it's not entirely trivial without creating circular package dependencies… I'd be delighted if you have an idea along these lines.

I wouldn't mind taking a crack at it, but have concerns about doing the work to have that PR declined. My design goals for this lift would be to keep [top-level docs](https://godoc.org/github.com/prometheus/client_golang/prometheus) same size (or smaller) and expose the functionality I need immediately (probably just `MetricVec`) to demonstrate feasibility of the effort. 

Can you please validate that the approach works for you? Anything else you'd like me to keep an eye on as I do this refactor?

smcquay:
I'm starting to gain an appreciation for the difficulty of the cyclic package dependency problem :)

A possible approach: how about we have all of the useful functionality exposed in a sub package, perhaps as it was earlier. Then we could expose a filtered view on that large package by having the `github.com/prometheus/client_golang/prometheus` package simply reference (and document) the types within the package that has the superset of exported objects? Something of the form:

```go
package prometheus

// ...

// A Metric models a single sample value with its meta data being exported to
// Prometheus. Implementations of Metric in this package are Gauge, Counter,
// Histogram, Summary, and Untyped.
type Metric prometheus.Metric

// ...

```

I'm guessing you've considered this approach already and we'd be trading the ability to use types externally (as I'd like to do with `MetricVec`) for additional work of duplicating documentation.

beorn7:
We might go down that way one day. However, that would result in a major refactoring, which might be more appropriate once the changes planned for v0.10 are in, which are by itself quite invasive.

My experience so far is that there are a _lot_ of sometimes contradicting ideas and requirements users of this package have. It is hard to optimize for satisfying as many of them as possible. I initially valued the option to implement own `Metric` and `MetricVec` types pretty highly but then it was never needed by anybody (until now). So I changed directions and weighed other properties stronger.

Given the huge backlog of more concrete demand for other features, I'm hesitant to commit to a larger restructuring to satisfy a relatively rare requirement at this point. It even looks that most of the gain for your use case can be had by implementing the transparent optimizations I have contemplated above.

Perhaps it's really the easiest if you keep your specialized counters in a friendly fork of this repo for now and we take a raincheck a bit further down the path.

In any case, I'd be happy to review and accept any under-the-hood optimizations that are transparent for the user (see above for ideas).

smcquay:
@beorn7 I agree with your assessment. Thanks for helping me reason through this. I'm taking a whack at an under-the-hood optimization.

smcquay:
Alright had a whack at it and it seems to be within striking distance:

```
 ⌘ go test -cpu=64 -benchtime=5s -v -bench=. github.com/smcquay/counters/bench
goos: darwin
goarch: amd64
pkg: github.com/smcquay/counters/bench
BenchmarkGoMetrics-64                   500000000               15.5 ns/op
BenchmarkPromCounter-64                 500000000               18.6 ns/op
BenchmarkFixedPrecisionCounter-64       300000000               24.5 ns/op
PASS
ok      github.com/smcquay/counters/bench       30.160s
```

prometheus/value.go
```diff
@@ -82,15 +85,25 @@ func (v *value) SetToCurrentTime() {
...
 func (v *value) Dec() {
-	v.Add(-1)
+	v.add(-1)
 }
 
 func (v *value) Add(val float64) {
+	if math.Trunc(val) == val {
+		atomic.AddInt64(&v.valInt, int64(val))
+		return
+	}
```
beorn7: This has edge cases that won't end well, e.g. +Inf or -Inf. Or just very large numbers that would overflow an int64 but would still work fine with float64 (at least approximately).

How about this:
```Go
ival := int64(val)
if float64(ival) == val {
	v.add(ival)
	return
}
```

It would also be good to have some of those edge cases in the tests (e.g. `Add(+Inf)` or `Add(1e99)`).

prometheus/value.go
```diff
@@ -46,6 +46,9 @@ type value struct {
 	// operations.  http://golang.org/pkg/sync/atomic/#pkg-note-BUG
 	valBits uint64
 
+	// valInt is used to store values that are exact integers
```
beorn7: Let's not have a blank line before this line. valBits and valInt are kind of related.

smcquay:
Implemented your review feedback and am considering how to attack the concurrent `Set`/`Write` problem without introducing locks. The naive approach that comes to mind is to use a `RWMutex` in both functions, but that seems heavy-handed. Please do let me know if you come up with a better idea.

beorn7:
Yeah, an `RWMutex` would defeat the whole purpose of lock-free mutation of the metric. We tried that originally, and it is very slow in highly concurrent scenarios. (And it's in any case a “many writes, few reads” scenario, so an `RWMutex` would perform even worse than a normal `Mutex`.)

Rough ideas so far:

We are concurrently mixing calls to `Set`, calls to `Add` with an int, calls to `Add` with a float, and the `Write` call that writes out the total value of the counter/gauge. `Set` and `Add` are racing anyway, but the user knows that, i.e. if the user calls `Set(42)` in one goroutine and `Add(1)` in another, the result may by 42 or 43. That's OK.

Since `Add` only touches either `valBits` or `valInt`, mixing those in is not problematic. The problematic case is if `Set` and `Write` are mixed, as they touch both variables. Let's assume `Set(x)` first sets the `valBits` to `x`, then sets `valInt` to 0. `Write`, in turn, first reads `valInt`, then `varBits`. I think this order simplifies things.

What can now go wrong? Four possibilities:

1. Set/Write/Set/Write:
    1. `Set` sets `valBits` to `x`.
    2. `Write` reads old non-zero `valInt`.
    3. `Set` sets `valInt` to 0.
    4.  `Write` reads `x` from `valBits`
    → wrong result (`x` + old `valInt`)
1. Set/Write/Write/Set:
    1. `Set` sets `valBits` to `x`.
    1. `Write` reads old non-zero `valInt`.
    1. `Write` reads `x` from `valBits`
    1. `Set` sets `valInt` to 0.
    → wrong result (`x` + old `valInt`)
1. Write/Set/Write/Set:
    1. `Write` reads old non-zero `valInt`.
    1. `Set` sets `valBits` to `x`.
    1. `Write` reads `x` from `valBits`
    1. `Set` sets `valInt` to 0.
    → wrong result (`x` + old `valInt`)
1. Write/Set/Set/Write:
    1. `Write` reads old non-zero `valInt`.
    1. `Set` sets `valBits` to `x`.
    1. `Set` sets `valInt` to 0.
    1. `Write` reads `x` from `valBits`
    → wrong result (`x` + old `valInt`)

Write/Write/Set/Set and Set/Set/Write/Write are the two happy cases.

The good thing about the above: All possible disastrous courses of events result in the same result. If `valInt` is 0, we know all is good. (The nasty case here is if `valInt` was -1 and we had an `Add(1)` in between, setting it to 0 at some point. But then `valInt` is legitimately 0, so we don't really have a problem.)

In other words: We can simply retry if `valInt` is not 0.

_However_, how do we find out if `valInt` is legitimately != 0 and no `Set` is involved? We could increment another atomic counter (let's call it the set counter) upon entering and exiting `Set`. `Write` could then check before and after starting to read `valBits` and `valInt`. If `valInt` is 0 _or_ the set counter hasn't changed, we are good and can proceed. Otherwise we retry. In theory, this strategy could deadlock if one goroutine is permanently calling `Set(1)`, another one is permanently calling `Add(1)`, and while that is happening, we call `Write`. With a bit of luck, the `Add` will permanently set `valInt` to 1 just before we read it, and the `Set` will execute with the right timing to increment the set counter. Of course, calling `Set` and `Add` in a permanent race is pretty nonsensical from the user's perspective (and in practice, I don' t think we would really deadlock). So that might be acceptable.

As said, a rough thought. Perhaps you have more ideas.

jacksontj:
Another approach would be to pull the actual values themselves behind another atomic (implemented here: https://github.com/smcquay/client_golang/pull/1)

From running the benchmarks this seems to similar (if not the same) performance
```
$ go test -cpu=64 -benchtime=5s -v -bench=.
goos: linux
goarch: amd64
pkg: github.com/smcquay/counters/bench
BenchmarkGoMetrics-64                	500000000	        13.4 ns/op
BenchmarkPromCounter-64              	500000000	        14.8 ns/op
BenchmarkFixedPrecisionCounter-64    	300000000	        23.8 ns/op
PASS
ok  	github.com/smcquay/counters/bench	26.509s
```

beorn7:
Yes, that should work. However, it will create an alloc on each call of `Set`. Scenarios with frequently set gauges will suffer a lot from that.

Since counters don't even have a `Set` method, while the `Set` method is presumably the most commonly used method of a gauge, how about separating the implementations of `Counter` and `Gauge`? the `Counter` implementation would work with the current state of this PR (as there is no `Set`). Gauges arguably require a fast `Set` more than a fast `Add`/`Inc`, so we could just leave the Gauge implementation in the original state.

jacksontj:
I think that'd be an okay approach. From cpuprofiling I see that the time is actually spent doing the atomic pointer swap *not* the alloc (roughly 2 to 1):

image

Even in the case of a gauge, I'm not sure how *frequently* its being called -- only because that would be a calculated number. In the case of counters they are Inc/Dec each request (for example) but sets() are usually done on some interval.

IMO if we are going to have a separate counter implementation, I think there'd definitely be a place for a fixed-precision counter as well.

beorn7:
Allocs in the hot path is a big no-no in Go. It's not so much about the time the alloc takes but the load on the GC.

Since the `Counter` interface doesn't have the `Set` method, we can go with the implementation in this PR and get the best of both worlds without making it more complicated for the user, so no need for an explicit fixed-precision counter.

beorn7:
To get things going here, I have implemented what I had suggested above in a new PR, which is based on this one, see #367 .

Issue #320, opened 2017-06-29
Currying for metric vecs


beorn7:
See #204 for some details about the difference between metric vecs and the metrics themselves, and how the current state is to represent them as different types to at least make use of the limited power of the Go type system as far as possible.

This whole problem has so far made my shy away from "partial specification" of labels because the `With` method should always return the metric rather than a (partially specified or “curried”) metric vec. (A powerful type system could figure that out. A vec with three labels would return a vec with two labels after the first `With` call, than one with one label, then the metric, and it would all trackable at compile time.)

The way to do the 'partial specification' was to use const labels, and the `promhttp` package has an example using that, which I quote here in slightly reduced form (to only demonstrate the partial specification):

```go
func ExampleInstrumentHandlerDuration() {
	// pushVec and pullVec are partitioned by the HTTP method and use custom
	// buckets based on the expected request duration. ConstLabels are used
	// to set a handler label to mark pushVec as tracking the durations for
	// pushes and pullVec as tracking the durations for pulls. Note that
	// Name, Help, and Buckets need to be the same for consistency, so we
	// use the same HistogramOpts after just modifying the ConstLabels.
	histogramOpts := prometheus.HistogramOpts{
		Name:        "request_duration_seconds",
		Help:        "A histogram of latencies for requests.",
		Buckets:     []float64{.25, .5, 1, 2.5, 5, 10},
		ConstLabels: prometheus.Labels{"handler": "push"},
	}
	pushVec := prometheus.NewHistogramVec(
		histogramOpts,
		[]string{"method"},
	)
	histogramOpts.ConstLabels = prometheus.Labels{"handler": "pull"}
	pullVec := prometheus.NewHistogramVec(
		histogramOpts,
		[]string{"method"},
	)

	pushHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte("Push"))
	})
	pullHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte("Pull"))
	})

	prometheus.MustRegister(pullVec, pushVec)

	instrumentedPush := InstrumentHandlerDuration(pushVec, pushHandler)
	instrumentedPull := InstrumentHandlerDuration(pullVec, pullHandler)

	http.Handle("/metrics", Handler())
	http.Handle("/push", instrumentedPush)
	http.Handle("/pull", instrumentedPull)

	if err := http.ListenAndServe(":3000", nil); err != nil {
[...]

PR #365, opened 2017-12-22
 Allow currying of metric vec's

beorn7:
The idea behind it is described in detail in #320 .

This commit also updates the example given in
promhttp/instrument_server_test.go , which nicely illustrates the
benefit of this change.

So far, currying could be emulated by creating different metric vec's
with different values in their ConstLabels. This was quite difficult
to grasp - which is essentially what was done in the example mentioned
above. Now that this use case can be solved without ConstLabels, we
can safely declare ConstLabels as rarely used. (Perhaps we can
deprecate them entirely one day, but I'll take a raincheck on that
when the changes of v0.10 have materialized.) This commit thus also
updates the ConstLabel doc comments in the various Opts. (It contained
fairly outdated stuff anyway.)

@grobie @stuartnelson3 

@dominikschulz and @metalmatze, IIRC you were keen on this to make the new middleware to instrument HTTP handlers (see promhttp package) more convenient.

### DIFF
--- a/prometheus/metric.go
+++ b/prometheus/metric.go
@@ -79,20 +79,12 @@ type Opts struct {
 	// with the same fully-qualified name must have the same label names in
 	// their ConstLabels.
 	//
-	// Note that in most cases, labels have a value that varies during the
-	// lifetime of a process. Those labels are usually managed with a metric
-	// vector collector (like CounterVec, GaugeVec, UntypedVec). ConstLabels
-	// serve only special purposes. One is for the special case where the
-	// value of a label does not change during the lifetime of a process,
-	// e.g. if the revision of the running binary is put into a
-	// label. Another, more advanced purpose is if more than one Collector
-	// needs to collect Metrics with the same fully-qualified name. In that
-	// case, those Metrics must differ in the values of their
-	// ConstLabels. See the Collector examples.
-	//
-	// If the value of a label never changes (not even between binaries),
-	// that label most likely should not be a label at all (but part of the
-	// metric name).
+	// ConstLabels are only used rarely. In particular, do not use them to
+	// attach the same labels to all your metrics. Those use cases are
+	// better covered by target labels set by the scraping Prometheus
+	// server, or by one specific metric (e.g. a build_info or a
+	// machine_role metric). See also
+	// https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels,-not-static-scraped-labels
 	ConstLabels Labels
 }
 
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -152,14 +152,22 @@ func (m *CounterVec) GetMetricWith(labels Labels) (Counter, error) {
 // error, WithLabelValues allows shortcuts like
 //     myVec.WithLabelValues("404", "GET").Add(42)
 func (m *CounterVec) WithLabelValues(lvs ...string) Counter {
-	return m.metricVec.withLabelValues(lvs...).(Counter)
+	c, err := m.GetMetricWithLabelValues(lvs...)
+	if err != nil {
+		panic(err)
+	}
+	return c
 }
 
 // With works as GetMetricWith, but panics where GetMetricWithLabels would have
 // returned an error. By not returning an error, With allows shortcuts like
 //     myVec.With(Labels{"code": "404", "method": "GET"}).Add(42)
 func (m *CounterVec) With(labels Labels) Counter {
-	return m.metricVec.with(labels).(Counter)
+	c, err := m.GetMetricWith(labels)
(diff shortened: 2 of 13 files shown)


beorn7:
Originally, I planned to have a "curry free" implementation under the hood for performance reasons, but the performance impact without currying is minimal. In particular, the whole curry handling is alloc-free.

Before:
```
BenchmarkCounterWithLabelValues-4                                       10000000               115 ns/op               0 B/op          0 allocs/op
BenchmarkCounterWithLabelValuesConcurrent-4                             20000000                94.0 ns/op             0 B/op          0 allocs/op
BenchmarkGaugeWithLabelValues-4                                         20000000               112 ns/op               0 B/op          0 allocs/op
BenchmarkSummaryWithLabelValues-4                                        2000000               786 ns/op               0 B/op          0 allocs/op
BenchmarkHistogramWithLabelValues-4                                     10000000               143 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValuesBasic-4                                10000000               140 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValues2Keys10ValueCardinality-4              10000000               153 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValues4Keys10ValueCardinality-4               5000000               251 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValues2Keys100ValueCardinality-4             10000000               164 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValues10Keys100ValueCardinality-4             2000000               607 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValues10Keys1000ValueCardinality-4            2000000               689 ns/op               0 B/op          0 allocs/op
```

After:
```
BenchmarkCounterWithLabelValues-4                                       10000000               127 ns/op               0 B/op          0 allocs/op
BenchmarkCounterWithLabelValuesConcurrent-4                             20000000                99.8 ns/op             0 B/op          0 allocs/op
BenchmarkGaugeWithLabelValues-4                                         10000000               121 ns/op               0 B/op          0 allocs/op
BenchmarkSummaryWithLabelValues-4                                        2000000               772 ns/op               0 B/op          0 allocs/op
BenchmarkHistogramWithLabelValues-4                                     10000000               148 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValuesBasic-4                                10000000               146 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValues2Keys10ValueCardinality-4              10000000               162 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValues4Keys10ValueCardinality-4               5000000               265 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValues2Keys100ValueCardinality-4             10000000               166 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValues10Keys100ValueCardinality-4             2000000               617 ns/op               0 B/op          0 allocs/op
BenchmarkMetricVecWithLabelValues10Keys1000ValueCardinality-4            2000000               662 ns/op               0 B/op          0 allocs/op
```

prometheus/metric.go
```diff
@@ -79,20 +79,12 @@ type Opts struct {
...
-	//
-	// If the value of a label never changes (not even between binaries),
-	// that label most likely should not be a label at all (but part of the
-	// metric name).
+	// ConstLabels are only used rarely. In particular, do not use them to
+	// attach the same labels to all your metrics. Those use cases are
+	// better covered by target labels set by the scraping Prometheus
+	// server, or by one specific metric (e.g. a build_info or a
+	// machine_role metric). More details in this blog post:
+	// https://www.robustperception.io/target-labels-are-for-life-not-just-for-christmas/
```
brian-brazil: https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels,-not-static-scraped-labels would be the right link I think.

beorn7:
> I've wondered about is if users mix maps and lists, that could get confusing. It might be an idea to advise against doing do.

The order stays the same. Curried labels just disappear from it. We already advise against using the variadic functions because of ordering concerns: “Note that for more than one label value, this method is prone to mistakes caused by an incorrect order of arguments. Consider GetMetricWith(Labels) as an alternative to avoid that type of mistake.”

beorn7:
I have clarified the doc comment about the ordering of remaining labels.

dominikschulz:
Thanks a lot, this looks like a very good approach that could help us. @zwopir

beorn7:
I still need to add some tests for the `promhttp` middleware interaction.

PR #367, opened 2018-01-19
Make Counter faster for simple Inc or Add with an integer

beorn7:
@grobie 
@smcquay This is implementing what I had suggested on top of your change.

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -15,6 +15,10 @@ package prometheus
 
 import (
 	"errors"
+	"math"
+	"sync/atomic"
+
+	dto "github.com/prometheus/client_model/go"
 )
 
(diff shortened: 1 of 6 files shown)


beorn7:
The existing `BenchmarkCounterNoLabels` covers this:

BEFORE:
```
BenchmarkCounterNoLabels-4   100000000               16.4 ns/op             0 B/op          0 allocs/op
```

AFTER:
```
BenchmarkCounterNoLabels-4   200000000                8.10 ns/op            0 B/op          0 allocs/op
```

beorn7:
@smcquay had more sophisticated benchmarks, I believe. If they could be contributed easily, I would appreciate it.

smcquay:
I will dig them up and submit them.

PR #373, opened 2018-02-01
Fix delegator return value for single hijacker

tsandall:
Previously, the pickDelegator function was not returning a
*hijackerDelegator so the return value did not implement the Hijacker
interface. As a result, code that attempts to hijack the connection
would fail when using a type assertion.

All the other cases returned the hijackerDelegator correctly.


tsandall:
@beorn7 PTAL

beorn7:
:+1: Thanks. There might be similar errors elsewhere. I'll check…

PR #377, opened 2018-02-07
Provide an InstrumentedHandler

### LINKED ISSUE #316: promhttp: Provide a sane instrumented handler
Similar to the (by default) instrumented `prometheus.Handler`, a `promhttp.InstrumentedHandler` should be provided. The problem is to find sane defaults. The old one was questionable in various ways.

beorn7:
See https://github.com/prometheus/client_golang/issues/316 for details.

Fixes #316 . @grobie @brian-brazil

### DIFF
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -67,19 +67,32 @@ func giveBuf(buf *bytes.Buffer) {
 	bufPool.Put(buf)
 }
 
-// Handler returns an HTTP handler for the prometheus.DefaultGatherer. The
-// Handler uses the default HandlerOpts, i.e. report the first error as an HTTP
-// error, no error logging, and compression if requested by the client.
+// Handler returns an http.Handler for the prometheus.DefaultGatherer, using
+// default HandlerOpts, i.e. it reports the first error as an HTTP error, it has
+// no error logging, and it applies compression if requested by the client.
 //
-// If you want to create a Handler for the DefaultGatherer with different
-// HandlerOpts, create it with HandlerFor with prometheus.DefaultGatherer and
-// your desired HandlerOpts.
+// The returned http.Handler is already instrumented using the
+// InstrumentMetricHandler function and the prometheus.DefaultRegisterer. If you
+// create multiple http.Handlers by separate calls of the Handler function, the
+// metrics used for instrumentation will be shared between them, providing
+// global scrape counts.
+//
+// This function is meant to cover the bulk of basic use cases. If you are doing
+// anything that requires more customization (including using a non-default
+// Gatherer, different instrumentation, and non-default HandlerOpts), use the
+// HandlerFor function. See there for details.
 func Handler() http.Handler {
-	return HandlerFor(prometheus.DefaultGatherer, HandlerOpts{})
+	return InstrumentMetricHandler(
+		prometheus.DefaultRegisterer, HandlerFor(prometheus.DefaultGatherer, HandlerOpts{}),
+	)
 }
 
-// HandlerFor returns an http.Handler for the provided Gatherer. The behavior
-// of the Handler is defined by the provided HandlerOpts.
+// HandlerFor returns an uninstrumented http.Handler for the provided
+// Gatherer. The behavior of the Handler is defined by the provided
+// HandlerOpts. Thus, HandlerFor is useful to create http.Handlers for custom
+// Gatherers, with non-default HandlerOpts, and/or with custom (or no)
+// instrumentation. Use the InstrumentMetricHandler function to apply the same
+// kind of instrumentation as it is used by the Handler function.
 func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		mfs, err := reg.Gather()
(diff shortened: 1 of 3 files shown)


prometheus/promhttp/http.go
```diff
@@ -52,7 +52,21 @@ const (
...
-var bufPool sync.Pool
+var (
+	bufPool sync.Pool
+
+	scrapesCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "promhttp_metric_handler_requests_total",
+			Help: "Total number of scrapes by HTTP status code.",
+		},
+		[]string{"code"},
```
brian-brazil: If you are going to use a label, initialise the potential label values. Personally I'd do this without a label.

prometheus/promhttp/http.go
```diff
@@ -78,6 +92,56 @@ func Handler() http.Handler {
...
+// InstrumentedHandler is meant to cover most of the simple cases. If you need
+// anything special, like other metrics or non-default HandlerOpts, or you are
+// not using the DefaultRegisterer, you have to use the various
+// InstrumentHandler… middlewares provided by this package to do the
+// instrumentation explicitly.
+//
+// InstrumentHandler panics if the registration of the metrics fails. If metrics
+// of the same name are already registered, the already registered ones are
+// tried, again panicking upon failure.
+func InstrumentedHandler() http.Handler {
```
brian-brazil: I don't think this will be a lot of use if users specifically have to turn it on, the benefit of these metrics is that they are there automatically.

prometheus/promhttp/http.go
beorn7: The label is important here, partially because of upcoming features like limited number of requests in flight.
We cannot really initialize all potential status codes here. Would you prefer to initialize a few common ones?

prometheus/promhttp/http.go
beorn7: Users of the default registry pick either `Handler` or `InstrumentedHandler`. There is nothing more or less automatic picking the one over the other. I always hated to have an `UninstrumentedHandler` in the previous approach. The naming now is way more natural, readable, and understandable.

prometheus/promhttp/http.go
brian-brazil: There's a few we expect to see, like 200 and whatever the duplicate labels one is.

prometheus/promhttp/http.go
brian-brazil: My point is that users should not have to pick, this should be just one more metric users of the default registry get automatically like process_cpu_seconds_total. We should only have one Handler, and it should be instrumented - otherwise there's not much point having instrumentation as users won't use it.

prometheus/promhttp/http.go
beorn7: > There's a few we expect to see, like 200 and whatever the duplicate labels one is.

I'm well aware of that, and it wasn't my point. My point was if it makes sense to initialize a few (but not all) label values. (This is a true question, where I want to know your opinion.)

prometheus/promhttp/http.go
beorn7: I don't follow the argument of "users won't use it". Users who can't get as far as seeing that there is `Handler` and `InstrumentedHandler` will hardly be able to make use of the provided handler metrics anyway.

However, I get the consistency point: The default registry already has process and go metrics, so the default handler should also have handler metrics. It's easy enough to create an uninstrumented handler with `HandlerFor` (and more involved use cases won't use the global registry anyway).

So yeah, let's just have one global `Handler`, which is instrumented. Unless @grobie has other thoughts, I'll change the commit later today.

prometheus/promhttp/http.go
brian-brazil: Ah yes, it makes sense to initialise what we know. That will avoid most of the potential breakage.

beorn7:
OK, this got a bit more involved. Initialization for code 200 and 500 (the only ones that should usually occur right now) are added. Also, the normal `Handler()` function now returns the instrumented handler.

However, since most users of a non-global registry will also like to instrument their handler in the same way, but now have to use `HandlerFor` and obviously need to register the collectors with their custom registry, I have created a separate `InstrumentMetricHandler`, which takes the Registerer as an argument. This also helped to avoid the global registry in the test ,win-win.

beorn7:
> I'm a little unsure about InstrumentMetricHandler being public.

But that's the whole point. How would you otherwise conveniently instrument a handler for a custom registry?

brian-brazil:
> But that's the whole point. How would you otherwise conveniently instrument a handler for a custom registry?

I'd expect it to use it's own instrumentation and metrics, https://github.com/prometheus/snmp_exporter/blob/master/main.go#L29 for example.

beorn7:
Which would inevitably lead to inconsistent instrumentation for scrape metrics. The goal here is that everybody who wants a scrape count uses the same metric name and dimensionality.

brian-brazil:
I see the case for custom registries as for SNMP/Blackbox style exporters where there's additional labels so you have to use different metrics.

Even if it's a non-scrape-specific custom registry such as the kubelet has for cadvisor metrics, you'd want to differentiate that from the metrics for the default registry handler as they have quite different performance characteristics.

beorn7:
I have said that before, but just FTR once more: Go programmers and subsequently Go programs tend to avoid global state. Any Go program of at least moderate complexity will not use the global registry at all. So there is nothing to differentiate. `InstrumentMetricHandler` will be used to get the normal instrumentation for the metric handler.

Issue #376, opened 2018-02-07
Make `process_…` metrics work on MS Windows


maxtar:
All process metrics doesn't provided in windows.
I've wrote exporter for our system and couldn't find metrics started from `process_`. After that I moved exporter to linux server and all these metrics were appeared in prometheus server.

beorn7:
Yeah, that's a know issue, but undocumented. We should definitely document that the process collector currently only works on OSs that have a proc filesystem, i.e. where https://github.com/prometheus/procfs works.
It should be relatively straight forward to wire in MS Windows support. Pull requests welcome. :o)

We needed another processCollect method (see https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector.go#L113 ), which could then be set [here](https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector.go#L90). Currently, just none is set if procfs cannot be found.

beorn7:
We have just documented the not so satisfying state of affairs. It would still be great if somebody could write code for the process collector that works on MS Windows. We could leave this issue open to fish for a contribution.

PR #378, opened 2018-02-08
Document that the process collector only works on Linux

beorn7:
@maxtar

To address the documentation issue raised in #376 .

### DIFF
--- a/prometheus/process_collector.go
+++ b/prometheus/process_collector.go
@@ -26,8 +26,11 @@ type processCollector struct {
 }
 
 // NewProcessCollector returns a collector which exports the current state of
-// process metrics including cpu, memory and file descriptor usage as well as
-// the process start time for the given process id under the given namespace.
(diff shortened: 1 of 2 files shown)


prometheus/process_collector.go
```diff
@@ -26,20 +26,20 @@ type processCollector struct {
...
 	)
 }
 
-// NewProcessCollectorPIDFn returns a collector which exports the current state
-// of process metrics including cpu, memory and file descriptor usage as well
-// as the process start time under the given namespace. The given pidFn is
-// called on each collect and is used to determine the process to export
-// metrics for.
+// NewProcessCollectorPIDFn works like NewProcessCollector but the process ID is
+// determined on each collect anew by calleng the given pidFn function.
```
brian-brazil: calling

beorn7: thanks

brian-brazil:
:+1:

Issue #372, opened 2018-01-30
support for basic authentication in the pushgateway


alitari:
As far as I have seen there is no opportunity to send metrics to a pushgateway which is secured with basic authentication.

beorn7:
Yeah, I guess that could be solved in the same way as the custom `Client` problem in #341.

PR #381, opened 2018-02-13
Add completely new push syntax

### LINKED ISSUE #341: Add the ability to use a custom HTTP Client to the push
Currently, the push to gateway function uses http.DefaultClient to make the request to the pushgateway.
By abling to use a custom client, there can be more flexibility on how requests are sent.
For my specific usage when I encoutered the need, I am using the client from GAE, which doesn't allow at all using the http.DefaultClient, which makes the library unusable in GAE

beorn7:
This allows adding more options in elegant ways, showcased here by
HTTP basic auth and by injecting a custom http.Client.

Fixes #341 and #372.

@grobie @alitari @shlimp

### DIFF
--- a/prometheus/push/example_add_from_gatherer_test.go
+++ b/prometheus/push/example_add_from_gatherer_test.go
@@ -11,12 +11,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Copyright (c) 2013, The Prometheus Authors
-// All rights reserved.
-//
-// Use of this source code is governed by a BSD-style license that can be found
-// in the LICENSE file.
-
 package push_test
 
 import (
--- a/prometheus/push/example_add_from_gatherer_test.go
+++ b/prometheus/push/example_add_from_gatherer_test.go
@@ -53,10 +53,14 @@ func performBackup() (int, error) {
 	return 42, nil
 }
 
-func ExampleAddFromGatherer() {
+func ExamplePusher_Add() {
+	// We use a registry here to benefit from the consistency checks that
+	// happen during registration.
 	registry := prometheus.NewRegistry()
 	registry.MustRegister(completionTime, duration, records)
-	// Note that successTime is not registered at this time.
+	// Note that successTime is not registered.
+
+	pusher := push.New("http://pushgateway:9091", "db_backup").Gatherer(registry)
 
 	start := time.Now()
 	n, err := performBackup()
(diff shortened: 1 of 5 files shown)


beorn7:
See #341 in particular to see the idea behind this.

prometheus/push/example_add_from_gatherer_test.go
```diff
@@ -67,18 +71,16 @@ func ExampleAddFromGatherer() {
 	if err != nil {
 		fmt.Println("DB backup failed:", err)
 	} else {
-		// Only now register successTime.
-		registry.MustRegister(successTime)
+		// Add succesTime to pusher only in case of success.
```
brian-brazil: success

prometheus/push/example_add_from_gatherer_test.go
```diff
@@ -67,18 +71,16 @@ func ExampleAddFromGatherer() {
...
+		// We could as well register it with the registry.
+		// This example, however, demonstrates that you can
+		// mix Gatherers and Collectors when handling a Pusher.
+		pusher.Collector(successTime)
```
brian-brazil: Its not clear to me at first glance exactly what this call is doing, is it pushing or is it registering?
`AddCollector` may be clearer

prometheus/push/push.go
```diff
@@ -44,57 +52,164 @@ import (
...
+		if strings.Contains(value, "/") {
+			p.error = fmt.Errorf("value of grouping label %s contains '/': %s", name, value)
+			return p
 		}
+		p.grouping[name] = value
+	}
+	return p
+}
+
+// HostnameGrouping adds a label pair with “instance” as the name and the
```
brian-brazil: The reason this existed was to allow for past mis-uses of the pgw when the pgw api changed quite some time ago. Now might be the time to remove it.

prometheus/push/example_add_from_gatherer_test.go
beorn7: Fixed.

prometheus/push/example_add_from_gatherer_test.go
beorn7: I had considered that name, too, but in the end I decided to go for the approach to use a verb only for the methods that are actually performing the push to the PGW (Push/Add) and use nouns for the  methods that configure the push. This also makes the methods shorter. Furthermore, using the verb "add" specifically will easily get confused with the `Add` method.

Hence, I do thing the current naming is the clearest and easiest to handle. Does somebody else have an opinion on this?

prometheus/push/push.go
beorn7: Personally, I don't feel strongly. I vaguely remember that some people did feel strongly about it. @grobie @juliusv I think you were part of the debate.

If nobody speaks up now, I'll happily remove it.

juliusv: I don't remember anything, but I'm for removing it. As the author of https://prometheus.io/docs/practices/pushing/, I really want to discourage this use :)

beorn7: OK, that's quorum. I have removed `HostnameGrouping`, adjusted the tests and examples, and edited the deprecation notice on the old `HostnameGroupingKey`.

PR #382, opened 2018-02-14
Add a link to godoc from the README

brian-brazil:


README.md
```diff
@@ -1,5 +1,6 @@
 # Prometheus Go client library
 
+[![GoDoc](http://img.shields.io/badge/go-documentation-blue.svg?style=flat-square)](http://godoc.org/github.com/prometheus/client_golang)
```
beorn7: Could you use the same style and order as for the GoDoc badges further down? (Either make it here the same as below or change below to this style, if you like it more.)

brian-brazil: Ah, I didn't realise they were there. No need for this PR then.

grobie:
I'd actually like a link to the godocs at the top of the document. It's not obvious right now and I had to tell people more than once to scroll further to find a link to the documentation. A full sentence would be even better than just a badge.

beorn7:
That is what I thought @brian-brazil intended.

I do not, however, like the idea of a text link _and_ the badge. There is enough content in the README.md already. I'm fine with adding a top link as we have now multiple proven incidents of people not seeing the links further down. Either a badge or text (but not both).

brian-brazil:
Just the badge is normal for other projects, and it's easy to scan for.

beorn7:
So let's have a badge at the top pointing to the root of the repo? That's easy to spot. I want to leave the individual badges below in place because they point to actual packages, while there is no Go package in the root of the repo.

PR #384, opened 2018-02-15
Add Godoc link at the top of the README

brian-brazil:
For those following that link, make it more obvious
which subpackage they probably want.

@beorn7 @grobie 

Following on for #382. I don't think the new doc text is the best, but it is better.


beorn7:
:+1:

PR #385, opened 2018-02-15
Add a package promauto that provides auto-registering metrics

beorn7:
Finally, I found an easy solution to provide the "evil"
auto-registration without getting death threats from the wardens of Go
purity. The reasoning can be found in the package's doc comment.

@brian-brazil I think you will like this.
@grobie next time you give an instrumentation introduction, you can use this package and ignore the registry with its Registerer and Gatherer interfaces altogether.

### DIFF
--- a/prometheus/promauto/auto.go
+++ b/prometheus/promauto/auto.go
@@ -1,0 +1,204 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package promauto provides constructors for the usual Prometheus metrics that
+// return them already registered with the global registry
+// (prometheus.DefaultRegisterer). This allows very compact code, avoiding any
+// references to the registry altogether, but all the constructors in this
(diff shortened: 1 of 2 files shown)


grobie:
Oh nice!

prometheus/promauto/auto.go
```diff
@@ -0,0 +1,197 @@
...
+// choose a local Registerer (usually created with prometheus.NewRegistry, but
+// there are other scenarios, e.g. testing).
+//
+// The second issue is that registration may fail, e.g. if a metric inconsistent
+// with the newly to be registered one is already registered. But how to signal
+// and handle a panic in the automatic registration with the default registry?
+// The only way is panicking. While panicking on invalid input provided by the
+// programmer is certainly fine, things are a bit more subtle in this case: You
+// might just add another package to the program, and that package (in its init
+// function) happens to register a metric with the same name as your code. Now,
```
brian-brazil: I'd change this to cover it happening at runtime due to loading some new code somehow (plugin?). Crashing out at startup isn't that bad, as it'll be pretty obvious from unittests. When you're already up and running however it's an outage.

beorn7: I'll add something along the lines "even worse, during runtime". But I want to keep the panic-at-startup scenario in there as well as many proficient Go devs feel strongly about it, too.

beorn7:
> Another option that I've been pondering recently is adding a MustRegister method to Counter&friends which then returns the Counter.

Which I had considered earlier, but rejected it, as idiomatic Go tends to reduce the number of methods in an interface. You would essentially pollute the interface for everybody who doesn't want to deal with global state anyway. That's why I'm so happy with the package solution here. Whoever wants to go down the "easy" path, just uses this package. Slightly different name when calling the constructors, that's very little friction, while the purists will never even notice that the package exists.

beorn7:
Improved the package doc comment to include comments above.

brian-brazil:
:+1:

PR #393, opened 2018-03-17
promauto: add NewCounterFunc and NewGaugeFunc

athomason:
This adds the Func variants of NewCounter and NewGauge to the handy promauto package.

@beorn7


beorn7:
Thanks. (The test errors are unrelated, I'll fix them ASAP in a different commit.)

PR #395, opened 2018-03-28
promhttp: fix DNSDone should call Done, not Start

LukeHandle:
When tracing, it appears `DNSDone` is instead calling the supplied `it.DNSStart` function - _NOT_ `it.DNSDone`. This looks like it would result in no reporting for completed DNS requests.

Per the example code, we would never log "dns_done".
```
	trace := &InstrumentTrace{
		DNSStart: func(t float64) {
			dnsLatencyVec.WithLabelValues("dns_start")
		},
		DNSDone: func(t float64) {
			dnsLatencyVec.WithLabelValues("dns_done")
		},
		TLSHandshakeStart: func(t float64) {
			tlsLatencyVec.WithLabelValues("tls_handshake_start")
		},
		TLSHandshakeDone: func(t float64) {
			tlsLatencyVec.WithLabelValues("tls_handshake_done")
		},
	}
```

Please forgive if I am missing anything obvious here... :)

### DIFF
--- a/prometheus/promhttp/instrument_client_1_8.go
+++ b/prometheus/promhttp/instrument_client_1_8.go
@@ -81,8 +81,8 @@ func InstrumentRoundTripperTrace(it *InstrumentTrace, next http.RoundTripper) Ro
 				}
 			},
 			DNSDone: func(_ httptrace.DNSDoneInfo) {
-				if it.DNSStart != nil {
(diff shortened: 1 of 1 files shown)


beorn7:
Looks like an obvious fix to me. Paging @stuartnelson3 for a final sanity check. (He wrote this code.)

PR #398, opened 2018-04-06
Add support for API admin methods

### LINKED ISSUE #337: Add delete series API
#248 needs to be reworked.

bobmshannon:
Fixes #337. Adds support for `DeleteSeries`, `CleanTombstones`, and `Snapshot` as documented [here](https://prometheus.io/docs/prometheus/latest/querying/api/#tsdb-admin-apis).

@beorn7

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -34,10 +34,13 @@ const (
 
 	apiPrefix = "/api/v1"
 
-	epQuery       = apiPrefix + "/query"
-	epQueryRange  = apiPrefix + "/query_range"
-	epLabelValues = apiPrefix + "/label/:name/values"
-	epSeries      = apiPrefix + "/series"
+	epQuery           = apiPrefix + "/query"
+	epQueryRange      = apiPrefix + "/query_range"
+	epLabelValues     = apiPrefix + "/label/:name/values"
+	epSeries          = apiPrefix + "/series"
+	epSnapshot        = apiPrefix + "/admin/tsdb/snapshot"
+	epDeleteSeries    = apiPrefix + "/admin/tsdb/delete_series"
+	epCleanTombstones = apiPrefix + "/admin/tsdb/clean_tombstones"
 )
 
 // ErrorType models the different API error types.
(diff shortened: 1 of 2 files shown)


beorn7:
Thanks @bobmshannon !

@gouthamve I guess you are most familiar with those endpoints. Does this make sense this way?
@andrestc You are probably most familiar with the current state of the API client code. Would appreciate your opinion.

api/prometheus/v1/api.go
```diff
@@ -238,6 +253,73 @@ func (h *httpAPI) Series(ctx context.Context, matches []string, startTime time.T
...
+	}
+
+	return nil
+}
+
+func (h *httpAPI) CleanTombstones(ctx context.Context) error {
+	u := h.client.URL(epCleanTombstones, nil)
+	q := u.Query()
+
+	u.RawQuery = q.Encode()
```
gouthamve: You don't need to encode anything, I guess. You can send an empty body.

andrestc:
I think in https://github.com/prometheus/client_golang/pull/291#issuecomment-291663377 we decided to have a single type holding the API for v1. But that was before we had a `admin` api, so I'm not sure if that still applies. Just thought I should've brought that up. Either way, this LGTM.

bobmshannon:
Thanks for the context. I think the single interface is preferable otherwise users will need to instantiate multiple API clients, e.g. one to access the admin methods, another for the status methods, etc. Looking at the previous PR I had assumed that a decision had been made otherwise somewhere along the line. Let me update the PR.

api/prometheus/v1/api.go
```diff
@@ -238,6 +253,70 @@ func (h *httpAPI) Series(ctx context.Context, matches []string, startTime time.T
...
+	q.Set("end", endTime.Format(time.RFC3339Nano))
+
+	u.RawQuery = q.Encode()
+
+	req, err := http.NewRequest(http.MethodPost, u.String(), nil)
+	if err != nil {
+		return err
+	}
+
+	_, _, err = h.client.Do(ctx, req)
```
simonpasquier: L297-302 can be shorten:

```
_, _, err = h.client.Do(ctx, req)
return err
```

api/prometheus/v1/api.go
```diff
@@ -238,6 +253,70 @@ func (h *httpAPI) Series(ctx context.Context, matches []string, startTime time.T
...
+
+func (h *httpAPI) CleanTombstones(ctx context.Context) error {
+	u := h.client.URL(epCleanTombstones, nil)
+
+	req, err := http.NewRequest(http.MethodPost, u.String(), nil)
+	if err != nil {
+		return err
+	}
+
+	if _, _, err = h.client.Do(ctx, req); err != nil {
```
simonpasquier: ditto

bobmshannon:
@beorn7 @simonpasquier I've addressed the PR comments.

beorn7:
Thank you very much. However, I'm terribly sorry that we have just added another hoop to jump through: We need the DCO sign-off (see recent post on prometheus-developers@). Perhaps, while adding the sign-off, squash the commits into one (which you can then sign off).
Thanks again.

beorn7:
And of course, the _Details_ link above next to the failed DCO test explains it, too.

bobmshannon:
@beorn7 Okay -- should be signed off now and all status checks are now passing.

beorn7:
Thanks again, everyone!

PR #401, opened 2018-04-09
Fix unprotected write in metricMap

erock2112:
@beorn7


beorn7:
Thanks. Sadly, this isn't exposed by the tests even if run with `go test -race`. Would probably be fairly easy to add something involving concurrency and deletion. I'll merge this now anyway, as it is fairly obvious. If you fancy adding a test, a PR would be highly appreciated.

PR #402, opened 2018-04-10
Implement AlertManager, Status, and Target methods

### LINKED ISSUE #400: Missing methods in Prometheus API client
The API client is missing the following non-admin methods:

```
GET /api/v1/targets
GET /api/v1/alertmanagers
GET /api/v1/status/config
GET /api/v1/status/flags
```

I think it would be useful to include these in the client -- the other methods are already implemented which should make this relatively straightforward.

bobmshannon:
Fixes #400 and should provide full API coverage as of v2.2.1. Re-organized the code slightly by re-arranging things in alphabetical order for readability.

cc @beorn7 

@gouthamve @simonpasquier @andrestc may also be interested in this too


api/prometheus/v1/api.go
```diff
@@ -75,21 +81,79 @@ type Range struct {
...
+	// AlertManagers returns an overview of the current state of the Prometheus alert manager discovery.
+	AlertManagers(ctx context.Context) (AlertManagersResult, error)
+	// CleanTombstones removes the deleted data from disk and cleans up the existing tombstones.
+	CleanTombstones(ctx context.Context) error
+	// Config returns the current Prometheus configuration.
+	Config(ctx context.Context) (ConfigResult, error)
+	// DeleteSeries deletes data for a selection of series in a time range.
+	DeleteSeries(ctx context.Context, matches []string, startTime time.Time, endTime time.Time) error
+	// Flags returns the flag values that Prometheus was launched with.
+	Flags(ctx context.Context) (FlagsResult, error)
```
simonpasquier: It could return a `map[string]string` directly.

api/prometheus/v1/api.go
```diff
@@ -75,21 +81,79 @@ type Range struct {
...
+
+// AlertManagersResult contains the result from querying the alertmanagers endpoint.
+type AlertManagersResult struct {
+	Active  []AlertManager `json:"activeAlertManagers"`
+	Dropped []AlertManager `json:"droppedAlertManagers"`
+}
+
+// AlertManager models a configured Alert Manager.
+type AlertManager struct {
+	URL *url.URL `json:"url"`
```
simonpasquier: Having `URL` as `string` feels more natural to me.

api/prometheus/v1/api.go
```diff
@@ -75,21 +81,79 @@ type Range struct {
...
+}
+
+// ActiveTarget models an active Prometheus scrape target.
+type ActiveTarget struct {
+	DiscoveredLabels model.LabelSet      `json:"discoveredLabels"`
+	Labels           model.LabelSet      `json:"labels"`
+	ScrapeURL        *url.URL            `json:"scrapeUrl"`
+	LastError        string              `json:"lastError"`
+	LastScrape       time.Time           `json:"lastScrape"`
+	Health           scrape.TargetHealth `json:"health"`
```
simonpasquier: Again `string` instead of `scrape.TargetHealth` so the code doesn't have to pull `github.com/prometheus/prometheus/scrape`.

.travis.yml
```diff
@@ -2,9 +2,8 @@ sudo: false
 language: go
 
 go:
- - 1.7.x
- - 1.8.x
- - 1.9.x
```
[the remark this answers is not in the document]
bobmshannon: This was an artifact from vendoring the main prometheus repo for the health target status constants. Since we're going to maintain that data in the API package instead, I've reverted this change.

api/prometheus/v1/api.go
bobmshannon: Done.

api/prometheus/v1/api.go
bobmshannon: Done.

api/prometheus/v1/api.go
bobmshannon: Done.

bobmshannon:
@simonpasquier Thanks for taking a look and the suggestions, I've pushed up another commit which addresses the comments.

.travis.yml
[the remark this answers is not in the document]
simonpasquier: Can you please revert any change to this file? It seems that the indentation has changed which isn't wrong but unnecessary.

api/prometheus/v1/api.go
simonpasquier: I don't see the change.

bobmshannon: @simonpasquier `FlagResult` is now defined as a `map[string]string` whereas before it was defined as a custom struct with the flags data contained inside the `Flags` field.

Not sure what the idiomatic thing to do here is, but figured that the former option and recent change might be more descriptive. I can change it to return `map[string]string` directly if that's more preferable.

simonpasquier: No strong opinion from my side. Lets get inputs from other reviewers.

beorn7: The reason for using types is that you can manipulate the JSON marshalling. In your case, you should probably add a tag to get a lower-case `flags` in the JSON. See my remark below.

api/prometheus/v1/api.go
```diff
@@ -75,21 +81,79 @@ type Range struct {
...
+}
+
+// ConfigResult contains the result from querying the config endpoint.
+type ConfigResult struct {
+	YAML string `json:"yaml"`
+}
+
+// FlagsResult contains the result from querying the flag endpoint.
+type FlagsResult struct {
+	Flags Flags
```
beorn7: Add a tag `json:"flags"` ?

bobmshannon: @beorn7 The problem is that the JSON in the response is structured like this so we'd need a custom unmarshaller to make that struct as declared work:

```
{
    "alertmanager.notification-queue-capacity": "10000",
    "alertmanager.timeout": "10s",
    "log.level": "info",
    "query.lookback-delta": "5m",
    "query.max-concurrency": "20"
}
```

And that can definitely be done. I guess the open question is whether we prefer that approach or not. The main benefit to this approach is that I believe we'll be able to add additional fields in a non-breaking way should the API change in the future.

beorn7: @bobmshannon Pardon my lack of JSON understanding, but does the automatic JSON marshalling really work with the current embedded `Flags Flags` structure? Naively, I would have assumed `FlagsResult` itself needed to be a `map[string]string` to make it work.

beorn7: And yes, my initial remark won't definitely not work. I just wasn't aware that the result JSON did not contain a `"flags": {...}`

bobmshannon: The current revision actually uses a `map[string]string` [here](https://github.com/prometheus/client_golang/pull/402/files#diff-4f30009d9ef316a0175312cc1739f05fR132) so the automatic marshalling does work. Seems like we're OK with this then?

bobmshannon:
Friendly ping on this. @beorn7, from your comments it looks like we're actually on the same page and the latest revision is doing what we want here.

beorn7:
Yes, indeed. Merging now.

PR #405, opened 2018-04-17
Add support for lifecycle actions in client API

bobmshannon:
- Adds support for reloading and shutting down Prometheus via the client API, if the lifecycle endpoints are enabled.

cc @beorn7

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -32,19 +32,25 @@ import (
 const (
 	statusAPIError = 422
 
-	apiPrefix = "/api/v1"
+	v1APIPrefix        = "/api/v1"
+	lifecycleAPIPrefix = "/-/"
 
-	epAlertManagers   = apiPrefix + "/alertmanagers"
-	epQuery           = apiPrefix + "/query"
-	epQueryRange      = apiPrefix + "/query_range"
-	epLabelValues     = apiPrefix + "/label/:name/values"
-	epSeries          = apiPrefix + "/series"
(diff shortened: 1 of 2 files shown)


beorn7:
As usual, despite being formally the maintainer of the API client, too, I'm not overly qualified or invested here.

I'm not sure if lifecycle actions should be integrated with the v1 API in this way.

@gouthamve @simonpasquier @andrestc @grobie @stuartnelson3 for various reasons, any or all of you might have a better founded opinion on this. Your input will be appreciated.

bobmshannon:
Happy to make any revisions here, should we decide that it would be best to move it over into another package or similar. These methods are not technically under the v1 namespace, and are also implemented in other prometheus components, so maybe that would be preferable.

beorn7:
@simonpasquier Yeah, I had similar thoughts, too.

I would also vote for not implementing a client API wrapper for calls with neither a request nor a response body. (I could be convinced otherwise by people more experienced with those API clients, but so fare, only @simonpasquier showed up supporting my gut feeling.)

andrestc:
Yeah, I'm not sure those endpoints belong here as they are not part of the v1 API (at least by looking at their URIs)? If they are added to the client, I would expect to find them somewhere else.

stuartnelson3:
Agreeing with @simonpasquier.

beorn7:
Looks like quorum. @bobmshannon I hope you understand the general sentiment here.

bobmshannon:
Sounds good to me, thanks everyone for chiming in.

PR #406, opened 2018-04-18
Fix a gofmt error and add gofmt in TravisCI

JordanP:
Let's block any further Gofmt regression thanks to TravisCI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,5 @@ go:
 
 script:
  - go test -short ./...
+ - go tool vet .
+ - diff -u <(echo -n) <(gofmt -d .)
--- a/api/prometheus/v1/api_test.go
+++ b/api/prometheus/v1/api_test.go
@@ -246,7 +246,7 @@ func TestAPIs(t *testing.T) {
 				"end":   []string{testTime.Format(time.RFC3339Nano)},
 			},
 			res: []model.LabelSet{
-				model.LabelSet{
+				{
 					"__name__": "up",
 					"job":      "prometheus",
 					"instance": "localhost:9090",


.travis.yml
```diff
@@ -8,3 +8,5 @@ go:
 
 script:
  - go test -short ./...
+ - go tool vet .
+ - diff -u <(echo -n) <(gofmt -d .)
```
beorn7: Other Prometheus repos just use `gofmt -d | grep '^'` (via Makefile.common, but we don't have a Makefile here as it is a library). I think it's better to do the fmt check in the same way everywhere.

.travis.yml
```diff
@@ -8,3 +8,5 @@ go:
 
 script:
  - go test -short ./...
+ - go tool vet .
```
beorn7: Other Prometheus repos (via Makefile.common, see next comment) use `staticcheck`. What do you think about that? We needed a `go get -u honnef.co/go/tools/cmd/staticcheck` first, but that would bring the check level in line.

.travis.yml
brian-brazil: The plan is to include the Makefile in all the Go repos, including this one. Not all of it will make sense obviously, but at least these things will be consistent.

@krasi-georgiev

JordanP: @beorn7 gofmt -d . alone will not work as it seems that this command return 0 (error_code 0) even if the code is not properly formatted. "man diff" says "Exit status is 0 if inputs are the same, 1 if different, 2 if trouble." that's what we leverage here.

@brian-brazil I am all in favor a makefile but my makefile mafic sucks. In the mean time, this PR could be a step in the right direction.

.travis.yml
JordanP: That sounds a good idea. But for now `staticcheck ./...` fails with

```
prometheus/examples_test.go:166:14: prometheus.UninstrumentedHandler is deprecated: Use promhttp.Handler instead. See there for further documentation.  (SA1019)
prometheus/examples_test.go:198:26: prometheus.Handler is deprecated: Please note the issues described in the doc comment of InstrumentHandler. You might want to consider using promhttp.InstrumentedHandler instead.  (SA1019)
```
I am too new to `prometheus/client_golang` to come with a proper fix for this 2 warnings. So not sure what to do w.r.t this PR.

.travis.yml
krasi-georgiev: why not use the `style` target of the common makefile?
https://github.com/prometheus/prometheus/blob/master/Makefile.common


only targets that use the `pkgs          = ./...` will fail when used with golang prior the 1.9 version
https://github.com/prometheus/prometheus/blob/4a4e8a7d3bab177aa446e8b0fd4de2c41bf5298d/Makefile.common#L34

JordanP:
I will.

beorn7:
@JordanP do you think you will continue to work on this eventually?
Otherwise, I'll close this here and create an issue for moving to the common Makefile.

JordanP:
No unfortunately I have switched priority :( You can close this issue.

Issue #187, opened 2015-12-07
Consider allowing setting a timestamp for ConstMetrics.


beorn7:
With a lot of caveats….

beorn7:
Use-case: Monitored client cares about an exact timestamp, and the timestamp is close to the time of the scrape (e.g. it is between beginning and end of a scrape, which implies no problem with staleness).

brian-brazil:
That's a highly advanced use case that should generally be discouraged, as having different timestamps for metrics from the same scrape will cause oddness and may prevent some features from working correctly.

bits01:
Another use case case is for hardware devices that generate counters with their own time stamps. These are retrieved and pushed to some server that is then scraped by Prometheus. It helps to use the source timestamp to get an accurate rate.

ccassar:
@bits01 above is applicable to network devices too. Would I be right in saying, that, without this enhancement, a gateway implemented in go and terminating telemetry streams from network devices, and wishing to persist to prometheus while preserving original timestamps must push via pushgw to do so. It would be great if the telemetry gateway could be scraped directly and network device timestamps could be preserved.

brian-brazil:
Those are the okay use cases (presuming you aren't talking about SNMP). 

However offering any functionality like this before https://github.com/prometheus/prometheus/issues/398 is complete would be reckless, as the vast majority of users looking to use timestamps it turned out that they should either have used normal pull or the pushgateway without timetamps. Had those users used timestamps, they would have run into a number of problems (as the few users that figured out how discovered).

ccassar:
Indeed. The interesting case is streaming telemetry not SNMP.  The context is a network pushing timestamped content. I would like to push the timestamped content into prometheus. We can accommodate collapsing multiple updates in the same series to the latest within a scrape period, but we would like to preserve the original timestamp. We are currently achieving something close using the pushgateway, but that has its own downsides.

beorn7:
Coming from the discussion about the collectd_exporter, an exporter that does something equivalent to federation, i.e. it exposes regularly updated timestamped metrics from another monitoring system, would be a good example for a use case.

I'll mark this with the v0.10 milestone because the API to create const metrics will be reworked in v0.10, and it is the right time to think about how the timestamp would be passed in. At least I want to avoid a breaking change for this feature.

brian-brazil:
[...]

PR #407, opened 2018-04-24
Add timestamp support for ConstMetric

simonpasquier:
This PR addresses #187 since @brian-brazil mentioned that staleness handling is now ok.

Eager to hear from @beorn7 too. Maybe you had something else in mind regarding how to expose the timestamps in the client API.

### DIFF
--- a/prometheus/value.go
+++ b/prometheus/value.go
@@ -79,15 +79,15 @@ func (v *valueFunc) Write(out *dto.Metric) error {
 // the Collect method. NewConstMetric returns an error if the length of
 // labelValues is not consistent with the variable labels in Desc.
 func NewConstMetric(desc *Desc, valueType ValueType, value float64, labelValues ...string) (Metric, error) {
-	if err := validateLabelValues(labelValues, len(desc.variableLabels)); err != nil {
-		return nil, err
-	}
-	return &constMetric{
-		desc:       desc,
-		valType:    valueType,
-		val:        value,
-		labelPairs: makeLabelPairs(desc, labelValues),
-	}, nil
+	return newConstMetric(desc, valueType, value, nil, labelValues...)
+}
+
+// NewConstMetricWithTimestamp returns a metric similar to NewConstMetric
+// except that the timestamp is set. It should be used only for very specific
+// use cases such as exposing metrics coming from push-based systems and for
+// which it it important to retain the original timestamp.
+func NewConstMetricWithTimestamp(desc *Desc, valueType ValueType, value float64, ts int64, labelValues ...string) (Metric, error) {
+	return newConstMetric(desc, valueType, value, &ts, labelValues...)
 }
 
 // MustNewConstMetric is a version of NewConstMetric that panics where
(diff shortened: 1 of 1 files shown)


beorn7:
This looks doable to me. However, the ConstMetric creation is a bit clunky at the moment, and I am planning to redo it completely in v0.10. Thus, my grand plan was to introduce the timestamped metrics at that time, so that we don't introduce something now that we break six months later (or so... my scheduling is unfortunately determined by external factors).

On the other hand, if people really need it _now_ (and possibly also without migrating to v0.10), we can do something like in this PR.

Let's see what others think.

brian-brazil:
OpenMetrics will also break this a bit, as we'll probably want to allow for more resolution than milliseconds there.

beorn7:
I'm currently aiming for aligning the OpenMetrics adjustments with the v0.10 changes.

The questions here really is how much it is worth to have any possibility for timestamped metrics in v0.9 (which will then surely experience a breaking change when moving to the bright new v0.10/OpenMetrics universe).

@simonpasquier is this something you have concrete need for or did you just feel it's low hanging fruit to introduce?

simonpasquier:
I don't have an urgent need for it. I know of some internal projects at $work that would benefit from it but they are still at the prototype stage. Another candidate would be collectd_exporter but as collectd supports Prometheus natively since 5.7 and exposes the timestamps, this isn't a compelling case either.

Personally I'm fine with waiting for 0.10 unless other users need it now.

beorn7:
OK, let's keep this PR open for a while to see if others chime in.

luqasn:
We would really like to use this for some of our custom CloudWatch exporters, as the official `cloudwatch-exporter` now uses `ConstMetric` with timestamps and our own golang-exporters still have the timestamps offset, which makes them weird to work with.

beorn7:
@luqasn OK, you got counted. :o)

Would it be possible for you to just merge this PR into your local fork of client_golang? If there were a handful of people requesting this, I'd consider it valuable to introduce a function now that we will break anyway in v0.10. If it's only you, it might be better in sum if you work from a slightly modified version until v0.10 is there.

beorn7:
OK, let's do it then. However, I have a better idea how to do it in detail. The current PR only handles the "plain" metrics, not histogram or summary. And I'd like to avoid the multiplicity of new functions. Stay tuned for a day or tow, and I'll present an alternative PR.

brian-brazil:
OpenMetrics is going for up to nanosecond resolution, accordingly I'd suggest using a `time.Time`.

beorn7:
I have implemented my idea in #443. Please have a look everybody. I'll close this as superseded.

jojimt:
@beorn7 With this change, it is possible to feed two samples of a series into a collector with two different timestamps. That's quite normal when the timestamp is attached by a hw or driver and there is some kind of buffering before the samples are collected. Is that supported? Or it will result in the "was collected before with the same name and label values" error?

beorn7:
Generally speaking, this is against the scrape model of Prometheus. It might work by accident, but it might as well create an error.

It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a (closed!) GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

jojimt:
Thanks. I will take it to the mailing list (it was sort of related to the timestamp).

PR #409, opened 2018-05-19
Document the stop-the-world implications of the Go collector

### LINKED ISSUE #403: Document stop-the-world implication of the Go collector and thus the default registry for older Go versions
Hi,

The `GoCollector` which is registered by default in the default registry calls[1] `runtime.ReadMemStats` which has the side effect of stopping the world[2] (releasing all the running threads to the OS...).
Obviously this can have a negative impact on performance.
The main problem is you can't opt out from this if you use the default registry, since the collector is registered in the default registry on package init.
A workaround is to have a package that registers a new default registry in init and import it before everything else in your main binary, but it doesn't feel sane and you have to get the other stats from a different collector that doesn't call `runtime.ReadMemStats`.

[1] https://github.com/prometheus/client_golang/blob/master/prometheus/go_collector.go#L273
[2] https://github.com/golang/go/blob/eb4f33243ece5874d2690c4b9bc72cb47f8da772/src/runtime/mstats.go#L457

beorn7:

@gwik

Fixes #403

### DIFF
--- a/prometheus/go_collector.go
+++ b/prometheus/go_collector.go
@@ -17,8 +17,12 @@ type goCollector struct {
 	metrics memStatsMetrics
 }
 
-// NewGoCollector returns a collector which exports metrics about the current
-// go process.
+// NewGoCollector returns a collector which exports metrics about the current Go
+// process. This includes memory stats. To collect those, runtime.ReadMemStats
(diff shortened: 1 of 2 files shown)


gwik:
Thanks!

PR #414, opened 2018-05-31
Decode JSON body when Prometheus returns 400 to an api call

bboreham:
400 and 422 are documented error codes from Prometheus, so we should attempt to parse the error returned for both of them.  Discussed briefly on IRC last night.

Needed to change a test that was requiring the old behaviour - made it use 500 instead of 400.

@beorn7 as maintainer

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -455,6 +455,11 @@ type apiResponse struct {
 	Error     string          `json:"error"`
 }
 
+func apiError(code int) bool {
+	// These are the codes that Prometheus sends when it returns an error.
+	return code == statusAPIError || code == http.StatusBadRequest
+}
+
 func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte, error) {
 	resp, body, err := c.Client.Do(ctx, req)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


beorn7:
Thanks @bboreham . My maintainer role for the `api` part of this repo is more a formality mishap than a real thing. It would be helpful to get the people chime in here that were part of the discussion. (@brian-brazil IIRC.)

In general, this makes sense to me. HTTP as I understand it may always give you something useful in the body. Thus, I'm a fan of trying to decode the body by default and only not do it if it is for sure (part of the API contract or something) that the body can or should be ignored.

bboreham:
FWIW the contributor instructions say to @-mention the maintainer.

brian-brazil:
I think this makes sense, we should pass back error information when we can.

beorn7:
> FWIW the contributor instructions say to @-mention the maintainer.

@bboreham you did nothing wrong. What's wrong is the situation. The `api` subdir was dropped here a while ago by important Prometheus developer A as an experiment and without any intentions of maintaining it themself.

When we formalized maintainership a bit more, I wanted to add a note that the `api` subdir is an unmaintained experience. But then important Prometheus developer B said that would appear unprofessional or something…

I will bring this up with the team again.

api/prometheus/v1/api.go
```diff
@@ -455,6 +455,11 @@ type apiResponse struct {
 	Error     string          `json:"error"`
 }
 
+func apiError(code int) bool {
+	// These are the codes that Prometheus sends when it returns an error
```
beorn7: Nit: End sentence with a period.

bboreham:
Nit fixed.

beorn7:
Thanks again.

Issue #418, opened 2018-06-13
too restrictive invalid characters for metric name


neeral:
I'd like to understand the logic in `bridge.go:replaceInvalidRune(rune)`. It seems to be very restrictive. In particular, you can't have `-` in the metric name, which is common in our use case. Another pain point is `.` are disallowed. The use case we have is for the `prefix`. It seems that only a single top-level prefix is supported. A multi-level prefix would be useful.

To give some background, we're using the bridge to convert stats from from one application to graphite format. We already have a lot of graphite metrics which are already arranged with prefixes.
e.g. An example is `node-name.service.metric_name`. Note the node name has a `-`, and `node-name.service` is a prefix.

I'm happy to help contribute to making this change, as long as there wasn't a technical reason why it wasn't done in the first place.

simonpasquier:
The metric names have indeed some restrictions which are documented here: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

It is very unlikely that this will change in the near future. What people usually do is to  replace the unsupported characters (for instance `.` to `:`).

brian-brazil:
This is about the Graphite bridge, which does not have to follow those rules.

beorn7:
First of all: Whatever comes from a Prometheus metric name or a Prometheus label name cannot have `.` or `-` anyway. It simply cannot be produced by the Prometheus instrumentation as those characters would be invalid in Prometheus. A different story are the label _values_, where Prometheus allows all UTF-8 characters, but Graphite doesn't. I _guess_ the `replaceInvalidRune` was meant to conservatively not allow anything that would not work for Graphite. Via Prometheus labels, you could sneak in `.` and `-` for sure into the resulting Graphite metric name. (Note that `.` is used as a separator between the Graphite name components that have been label pairs on the Prometheus side.)

@stuartnelson3 wrote the bridge. He might know a rationale why he was so restrictive here.

Without further context, I would assume we could allow all characters that are legal in Graphite metric names. Possibly without the `.` if we want to create a clear mapping from Prometheus label pairs to Graphite name components (e.g. if a `.` appears in the Graphite metrics name, it should always be a label-pair separator but not coming from a label value).

brian-brazil:
> Without further context, I would assume we could allow all characters that are legal in Graphite metric names.

I believe that varies based on the implementation and at times the underlying filesystem. In Python I do no sanitisation of the prefix, as I see it as at that point you are the end user/sysadmin and know what your Graphite can handle.

[...]

PR #419, opened 2018-06-22
Do not sanitize prefix in Graphite bridge

neeral:
Refer to #418.

Signed-off-by: neeral <neeral@users.noreply.github.com>


beorn7:
Thanks.

PR #420, opened 2018-06-27
Update DefaultRoundTripper to DialContext because Dial is deprecated

alileza:
As per https://golang.org/pkg/net/http/#Transport.Dial documentation says

```
// Deprecated: Use DialContext instead, which allows the transport
// to cancel dials as soon as they are no longer needed.
// If both are set, DialContext takes priority.
```


beorn7:
Makes sense to me, but I defer final judgement to @krasi-georgiev (who is by now the maintainer of the `api` directory).

krasi-georgiev:
LGTM 

thanks @alileza

Issue #422, opened 2018-07-03
Include testing package


brancz:
Within the Kubernetes project there have lately been multiple instances where people wanted to test metric exposition. At one point @fabxc and I extracted testing code from the node-exporter (if I recall correctly) to [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics/blob/2e30678a4d4d055ce13a095c2ec8d14eef353570/pkg/collectors/testutils/testutils.go) where it has been extremely helpful. Now there are other components such as the [Kubernetes nginx-ingress-controller](https://github.com/kubernetes/ingress-nginx/pull/2726#issuecomment-401816568) and the [Kubernetes project itself](https://github.com/kubernetes/kubernetes/pull/65632) that want to vendor/re-use that package. I felt that naturally a testing package would fit into the client library, so I'd like to see what everyone thinks regarding adding this or a package like this to this project.

@beorn7 @brian-brazil @grobie

brian-brazil:
The python/java clients have getSampleValue: http://prometheus.github.io/client_java/io/prometheus/client/CollectorRegistry.html#getSampleValue-java.lang.String- for unit testing.

beorn7:
I like the idea of a testing package (and #58 has been on the plate forever…).

How about you create a PR with the testutil package as you would like to see it, and then we discuss details on the PR?

beorn7:
Let's track this is #58 to avoid issue duplication.

Issue #58, opened 2015-01-27
Improve testability


beorn7:
First of all, there is the public Registry interface tracked in https://github.com/prometheus/client_golang/issues/46

But there is more:

Two things come to my mind here, from the email thread:

(1)
Another utility function could compare two collectors (or a collector
to a list of template metrics) and verify that they return the same
set of metrics. In that way, you could simply create the expected
metrics and verify that the collector to be tested collects the same
metrics.
[...]
But one of my suggestions was to have a comparison utility function
that would abstract away the order (as the Prometheus registry does
it, too).
Something like

```
testutil.Collects(myCollector, []prometheus.Metric{
    prometheus.MustNewConstMetric(...),
}
```

(2)

``` go
func readCounter(m prometheus.Counter) float64 {
    pb := &dto.Metric{}
    m.Write(pb)
    return pb.GetCounter().GetValue()
}
```

Could be done in a generic way, like this:

``` go
package testutil

func ToFloat64(m prometheus.Metric) float64 {
    pb := &dto.Metric{}
    m.Write(pb)
    if pb.Gauge != nil {
        return pb.Gauge.GetValue()
    }
    if pb.Counter != nil {
        return pb.Counter.GetValue()
    }
    if pb.Untyped != nil {
        return pb.Untyped.GetValue()
    }
    return math.NaN()
}
```

brian-brazil:
https://github.com/prometheus/client_java/blob/master/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L102 is one way that the java_client makes things easier to test

grobie:
:+1:

beorn7:
This represents the first part of test tools, those that can be implemented without the more invasive changes in 0.10. For the rest, see #230

beorn7:
Plan is to adopt the [kube-state-metrics testutils package](https://github.com/kubernetes/kube-state-metrics/blob/2e30678a4d4d055ce13a095c2ec8d14eef353570/pkg/collectors/testutils/testutils.go). @brancz will create a PR. See #422 for details.

PR #423, opened 2018-07-03
[WIP] Add testutils package

brancz:
This is mostly a copy and paste from the testutils package that we have originally extracted from the node-exporter test code (I think). I opened #422 about this and @beorn7 suggested me to open a PR and we start discussing. I haven't added documentation yet in case we want to change any of this. If we are happy with this I am more than happy to add more documentation.

Note that this primarily revolves around testing (custom) collectors, it is less useful for standard Counter/Gauge/Histogram/Summary metrics, unless the tested entity is a wrapping object around those (I suppose it could also be useful for single metrics but I see it being most useful for collectors).

Also related: #58 

Signed-off-by: Frederic Branczyk <fbranczyk@gmail.com>

@beorn7 @brian-brazil @grobie

### DIFF
--- a/testutils/testutils.go
+++ b/testutils/testutils.go
@@ -1,0 +1,182 @@
+/*
+Copyright 2018 The Prometheus Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package testutils
+
+import (
+	"bytes"
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+)
+
+// GatherAndCompare retrieves all metrics exposed by a collector and compares it
+// to an expected output in the Prometheus text exposition format.
+// metricNames allows only comparing the given metrics. All are compared if it's nil.
+func GatherAndCompare(c prometheus.Collector, expected string, metricNames ...string) error {
+	expected = removeUnusedWhitespace(expected)
+
+	reg := prometheus.NewPedanticRegistry()
+	if err := reg.Register(c); err != nil {
+		return fmt.Errorf("registering collector failed: %s", err)
+	}
+	metrics, err := reg.Gather()
+	if err != nil {
+		return fmt.Errorf("gathering metrics failed: %s", err)
+	}
+	if metricNames != nil {
+		metrics = filterMetrics(metrics, metricNames)
+	}
+	var tp expfmt.TextParser
+	expectedMetrics, err := tp.TextToMetricFamilies(bytes.NewReader([]byte(expected)))
+	if err != nil {
+		return fmt.Errorf("parsing expected metrics failed: %s", err)
+	}
+
+	if !reflect.DeepEqual(metrics, normalizeMetricFamilies(expectedMetrics)) {
+		// Encode the gathered output to the readbale text format for comparison.
+		var buf1 bytes.Buffer
+		enc := expfmt.NewEncoder(&buf1, expfmt.FmtText)
+		for _, mf := range metrics {
+			if err := enc.Encode(mf); err != nil {
+				return fmt.Errorf("encoding result failed: %s", err)
+			}
+		}
+		// Encode normalized expected metrics again to generate them in the same ordering
+		// the registry does to spot differences more easily.
+		var buf2 bytes.Buffer
+		enc = expfmt.NewEncoder(&buf2, expfmt.FmtText)
+		for _, mf := range normalizeMetricFamilies(expectedMetrics) {
+			if err := enc.Encode(mf); err != nil {
+				return fmt.Errorf("encoding result failed: %s", err)
+			}
+		}
+
+		return fmt.Errorf(`
+metric output does not match expectation; want:
+
+%s
+
+got:
+
+%s
+`, buf2.String(), buf1.String())
+	}
+	return nil
+}
+
+func filterMetrics(metrics []*dto.MetricFamily, names []string) []*dto.MetricFamily {
+	var filtered []*dto.MetricFamily
+	for _, m := range metrics {
+		drop := true
+		for _, name := range names {
+			if m.GetName() == name {
+				drop = false
+				break
+			}
+		}
+		if !drop {
+			filtered = append(filtered, m)
+		}
+	}
+	return filtered
+}
+
+func removeUnusedWhitespace(s string) string {
+	var (
+		trimmedLine  string
+		trimmedLines []string
+		lines        = strings.Split(s, "\n")
+	)
+
+	for _, l := range lines {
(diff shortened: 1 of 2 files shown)


grobie:
"wants" ;-) But I'll try to give it a look over the weekend.

beorn7:
FTR: Travis test was flaky, I restarted it, now it is green.

testutils/testutils.go
```diff
@@ -0,0 +1,182 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+)
+
+// GatherAndCompare retrieves all metrics exposed by a collector and compares it
+// to an expected output in the Prometheus text exposition format.
+// metricNames allows only comparing the given metrics. All are compared if it's nil.
+func GatherAndCompare(c prometheus.Collector, expected string, metricNames ...string) error {
```
beorn7: Yes, that's exactly what I had in mind for this package.

testutils/testutils.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	// The Prometheus metrics representation parser expects an empty line at the
+	// end otherwise fails with an unexpected EOF error.
+	return strings.Join(trimmedLines, "\n") + "\n"
+}
+
+// The below sorting code is copied form the Prometheus client library modulo the added
+// label pair sorting.
+// https://github.com/prometheus/client_golang/blob/ea6e1db4cb8127eeb0b6954f7320363e5451820f/prometheus/registry.go#L642-L684
```
beorn7: Which raises the question if we should move this code to an internal package…

testutils/testutils.go
```diff
@@ -0,0 +1,182 @@
...
+import (
+	"bytes"
+	"fmt"
+	"reflect"
+	"sort"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
```
beorn7: Nit: Traditional Prometheus grouping here would be:

```go

import (
￼ 	"bytes"
￼ 	"fmt"
￼ 	"reflect"
￼ 	"sort"
￼ 	"strings"
￼ 

￼ 	dto "github.com/prometheus/client_model/go"

￼ 	"github.com/prometheus/client_golang/prometheus"
￼)
```

beorn7:
As in the comments (besides the nits), this is what I had in mind as one of the helpers in this upcoming package. So yes, full agreement we should have this here.

Suggested course of action:
1. Let's wait for @grobie to chime in whether this is what he wants, too.
2. Improve code-level details (started with a few nits from my side already, see comments).
3. Merge this.
4. I will then add the single-value retrieval helper suggested in #58 in a follow-up PR.

(Reading the following is optional: Very fundamentally, I think this single-value retrieval is not the right (tm) approach to testing, but we can/should still provide it here for now. The "gather all the metric and compare it to a golden set of metrics" is a nice end-to-end test. For unit-testing, the only thing I want to test is if, for example, an `Inc()` call has happened on a `Counter`. However, retrieving the counter value before and after the `Inc` call to verify the increment by one involves the whole counter machinery (which I don't want to test) and doesn't play well with concurrency. It would be better to use test mocks of metric types that directly report if the call has happened and do nothing else. However, with the current state of client_golang, the interfaces are less clear than they could be, and not everything is an interface that one could implement with test mocks. Thus, this part will follow later, cf. #230, and for now I'll provide that single-value retrieval helper, but at least it is nicely locked away in this package, and it can be deprecated without breaking anything in the main packages.)

testutils/testutils.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+// GatherAndCompare retrieves all metrics exposed by a collector and compares it
+// to an expected output in the Prometheus text exposition format.
+// metricNames allows only comparing the given metrics. All are compared if it's nil.
+func GatherAndCompare(c prometheus.Collector, expected string, metricNames ...string) error {
+	expected = removeUnusedWhitespace(expected)
+
+	reg := prometheus.NewPedanticRegistry()
+	if err := reg.Register(c); err != nil {
```
beorn7: How about leaving these two lines to the caller? I.e. make `GatherAndCompare` accept a `Gatherer` rather than a `Collector`. This would also let the name of the function appear more consistent.

With this approach, a caller could act on the registry level, which allows testing even if metrics are coming from a library that does not allow access to individual collectors, but a registry with all the collectors already registered is readily available. Also, the caller can decide if a pedantic registry should be used or not.

If a caller wants to test a single Collector, it's easy to wrap `GatherAndCompare` with these two lines. Or wo could even provide a `CollectAndCompare`.

testutils/testutils.go
```diff
@@ -0,0 +1,182 @@
...
+	for _, l := range lines {
+		trimmedLine = strings.TrimSpace(l)
+
+		if len(trimmedLine) > 0 {
+			trimmedLines = append(trimmedLines, trimmedLine)
+		}
+	}
+
+	// The Prometheus metrics representation parser expects an empty line at the
+	// end otherwise fails with an unexpected EOF error.
```
beorn7: That's misleading. It expects the last line to end on a `\n`, but that line does not need to be empty.

testutils/testutils.go
beorn7: As a concrete example, right now you couldn't test many Collectors because the "cheat" by providing a dummy Desc, which the pedantic registry would flag. (#47 will allow to implement those Collectors cleanly, and I'll work in this issue this evening (coincidentally).)

testutils/testutils.go
grobie: The interface and implementation couple two independent concerns: verifying the given collector and formatting a custom input format. Instead of this coupling and assuming all users will format their expectation in this special way, using the `io.Reader` interface would leave it to the user how to provide the expectation. For example, this allows projects with extensive end to end tests to simply provide a File reader to such fixtures.

The current custom sanitizer implementation in `removeUnusedWhitespace` can be provided as another public function which will satisfy the reader interface.

testutils/testutils.go
grobie: :+1:

testutils/testutils.go
beorn7: It certainly makes sense to use an `io.Reader` instead of a `string` as input.

I don't quite follow the line of argument with the `removeUnusedWhitespace`, though. In fact, now that I have looked more closely, I got the suspicion the `removeUnusedWhitespace` is not even needed. The provided text for "expected" is parsed with the text format parses. (Thus, the parser should eat all the spurious whitespace.) For the diff output, we then re-create the text format from the protobuf representation and get a sanitized output anyway.

beorn7:
Another point: Since these testutils are only for testing the exposition client, I'd make it a sub-package of client_golang/prometheus rather than client_golang directly.

testutils/testutils.go
grobie: Even better if we don't need it.

beorn7: I played with the tests, and it really looks to me we can just remove `removeUnusedWhitespace`. :o)

beorn7:
@brancz will you be able to address the comments given? I think the changes will be quite straight forward, and then we can merge this. If you are busy right now, I can take over from here.

brancz:
Yeah I have a big deadline that is keeping me busy at least for today and next week, feel free to take over! :slightly_smiling_face:

beorn7:
I was a bit optimistic. Preparations for my grand cycle tour took precedence, and now I'm on said cycle tour, with not much time for anything else. @brancz if you find time, please go ahead to work on this. Should I find time before that, I'll let you know here.

brancz:
I'm still busy as well, but might be able to get to it next week.

beorn7:
Superseded by #445 .

PR #424, opened 2018-07-08
Move both examples into single Dockerfile for Docker Hub (fixes #347)

### LINKED ISSUE #347: Push the simple and random examples to docker hub

pedrosland:
This PR should be the last step to get the examples on the Docker Hub. It:

* merges the Dockerfile for the two examples into one
* puts it in the root of the repo for Docker Hub
* updates to Go 1.10.3 to build the examples

The two commands to run the examples with this change could be:
```sh
docker run -ti prom/golang-examples /simple
docker run -ti prom/golang-examples /random
```

I have tested this on the hub and it does build:
```sh
docker run -ti -p 8080:8080 pedrosland/prom-test:test /simple
```

See issue #347 and the discussion on #348.

Please let me know @beorn7 if you would like any changes. I can also make a PR to the docs repo to add this to the getting started guide.


beorn7:
This appears to me exactly the thing we need. However, since my Docker foo is still mediocre, it would be great if somebody else could chime in for a sanity check, for example @discordianfish . :pray:

Dockerfile
```diff
@@ -4,17 +4,20 @@
 #    docker build -f examples/$name/Dockerfile -t prometheus/golang-example-$name .
 
 # Builder image, where we build the example.
-FROM golang:1.9.0 AS builder
+FROM golang:1.10.3 AS builder
```
grobie: Golang 1 has a stable and guaranteed interface, using `golang:1` would remove the need to constantly update this version with every new release.

beorn7: If that works, great.

pedrosland: Thanks for the feedback. Sounds good. I've updated the PR.

discordianfish:
LGTM

beorn7:
Thanks everybody.

PR #425, opened 2018-07-08
Introduce unchecked Collectors

### LINKED ISSUE #47: Allow more flexible metric exposition.

beorn7:
Fixes #47 . See there for more detailed discussion.

@grobie this removes the need to return "dummy" descriptors as it has been done by countless exporters so far.

### DIFF
--- a/prometheus/collector.go
+++ b/prometheus/collector.go
@@ -29,24 +29,35 @@ type Collector interface {
 	// collected by this Collector to the provided channel and returns once
 	// the last descriptor has been sent. The sent descriptors fulfill the
 	// consistency and uniqueness requirements described in the Desc
-	// documentation. (It is valid if one and the same Collector sends
-	// duplicate descriptors. Those duplicates are simply ignored. However,
-	// two different Collectors must not send duplicate descriptors.) This
-	// method idempotently sends the same descriptors throughout the
-	// lifetime of the Collector. If a Collector encounters an error while
-	// executing this method, it must send an invalid descriptor (created
-	// with NewInvalidDesc) to signal the error to the registry.
+	// documentation.
+	//
+	// It is valid if one and the same Collector sends duplicate
+	// descriptors. Those duplicates are simply ignored. However, two
+	// different Collectors must not send duplicate descriptors.
+	//
+	// Sending no descriptor at all marks the Collector as “unchecked”,
+	// i.e. no checks will be performed at registration time, and the
+	// Collector may yield any Metric it sees fit in its Collect method.
+	//
+	// This method idempotently sends the same descriptors throughout the
+	// lifetime of the Collector.
+	//
+	// If a Collector encounters an error while executing this method, it
+	// must send an invalid descriptor (created with NewInvalidDesc) to
+	// signal the error to the registry.
 	Describe(chan<- *Desc)
 	// Collect is called by the Prometheus registry when collecting
 	// metrics. The implementation sends each collected metric via the
 	// provided channel and returns once the last metric has been sent. The
-	// descriptor of each sent metric is one of those returned by
-	// Describe. Returned metrics that share the same descriptor must differ
-	// in their variable label values. This method may be called
-	// concurrently and must therefore be implemented in a concurrency safe
-	// way. Blocking occurs at the expense of total performance of rendering
-	// all registered metrics. Ideally, Collector implementations support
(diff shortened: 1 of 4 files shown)


prometheus/registry.go
```diff
@@ -68,7 +67,8 @@ func NewRegistry() *Registry {
 
 // NewPedanticRegistry returns a registry that checks during collection if each
 // collected Metric is consistent with its reported Desc, and if the Desc has
-// actually been registered with the registry.
+// actually been registered with the registry. Uncheckd Collectors (those whose
```
matthiasr: Typo: unchecked

prometheus/registry.go
```diff
@@ -405,64 +422,109 @@ func (r *Registry) Gather() ([]*dto.MetricFamily, error) {
...
+	// are collected.
 	go func() {
 		wg.Wait()
-		close(metricChan)
+		close(checkedMetricChan)
+		close(uncheckedMetricChan)
 	}()
 
-	// Drain metricChan in case of premature return.
+	// Drain checkedMetricChan and uncheckdMetricChan in case of premature return.
```
matthiasr: uncheckd

prometheus/collector.go
```diff
@@ -29,24 +29,33 @@ type Collector interface {
...
+	// method. (Since the library will still ensure metric consistency,
+	// errors will thus only happen at scrape time. Thus, the Collector is
+	// still responsible to return metrics that are consistent with all the
+	// other metrics. Unchecked Collectors are useful if the exact metrics
+	// collected later cannot be predicted while consistency can still be
+	// ensured by the implementation.) This method idempotently sends the
+	// same descriptors throughout the lifetime of the Collector. If a
+	// Collector encounters an error while executing this method, it must
+	// send an invalid descriptor (created with NewInvalidDesc) to signal
+	// the error to the registry.
```
grobie: The comment is quite long and hard to read and digest. Using paragraphs for the different topics (consistency, idempotence, etc.) would be helpful already. Cutting down on the prose and moving tips and tricks somewhere else would be good as well.

prometheus/registry.go
beorn7: Fixed.

prometheus/registry.go
beorn7: Fixed. (Once emacs remembers that spelling, you run into it all the time...)

prometheus/registry_test.go
```diff
@@ -34,7 +34,20 @@ import (
...
+}
+
+func (u uncheckedCollector) Describe(_ chan<- *prometheus.Desc) {}
+func (u uncheckedCollector) Collect(c chan<- prometheus.Metric) {
+	u.c.Collect(c)
+}
+
 func testHandler(t testing.TB) {
+	// TODO(beorn7): This test is a bit too "end-to-end". The moving parts
+	// it tests could be tested separately. I good moment to restructure
```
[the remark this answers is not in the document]
beorn7: I reworded, hoping to make it clearer.

prometheus/collector.go
beorn7: I have introduced paragraphs and moved the explanation of the high-level concept to the package doc comment.

beorn7:
Updates pushed. PHAL.

PR #427, opened 2018-07-11
Make delegators wrapping responseWriterDelegator have value receivers

edaniels:
Without doing so, they don't actually implement the interfaces they claim to.

EDIT: They do implement the interfaces in actual usage but can now more universally do so regardless of how they are references (value or pointer)

### DIFF
--- a/prometheus/promhttp/delegator.go
+++ b/prometheus/promhttp/delegator.go
@@ -76,16 +76,16 @@ type flusherDelegator struct{ *responseWriterDelegator }
 type hijackerDelegator struct{ *responseWriterDelegator }
 type readerFromDelegator struct{ *responseWriterDelegator }
 
-func (d *closeNotifierDelegator) CloseNotify() <-chan bool {
+func (d closeNotifierDelegator) CloseNotify() <-chan bool {
 	return d.ResponseWriter.(http.CloseNotifier).CloseNotify()
 }
-func (d *flusherDelegator) Flush() {
+func (d flusherDelegator) Flush() {
 	d.ResponseWriter.(http.Flusher).Flush()
 }
-func (d *hijackerDelegator) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+func (d hijackerDelegator) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 	return d.ResponseWriter.(http.Hijacker).Hijack()
 }
-func (d *readerFromDelegator) ReadFrom(re io.Reader) (int64, error) {
+func (d readerFromDelegator) ReadFrom(re io.Reader) (int64, error) {
 	if !d.wroteHeader {
 		d.WriteHeader(http.StatusOK)
 	}
--- a/prometheus/promhttp/delegator.go
+++ b/prometheus/promhttp/delegator.go
@@ -102,34 +102,34 @@ func init() {
 		return d
 	}
 	pickDelegator[closeNotifier] = func(d *responseWriterDelegator) delegator { // 1
-		return &closeNotifierDelegator{d}
+		return closeNotifierDelegator{d}
 	}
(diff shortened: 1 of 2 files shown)


beorn7:
Hmm, I thought we had fixed this by returning pointers in `pickDelegator` (cf. https://github.com/edaniels/client_golang/commit/e87046a87e6fa0385d2ff32b11e7ed938664d814#diff-b02864f99c25aa21436b5820f58b399a and https://github.com/edaniels/client_golang/commit/44a8a1218c17b46b21adbc6cab151e172e748ca3#diff-b02864f99c25aa21436b5820f58b399a ).

Looking at your solution, it's probably better as the delegators wrap a pointer anyway. However, I would like to understand where you ran into the problem.

Independent of the above, we need you to add the DCO sign-off. https://github.com/probot/dco#how-it-works 

Thanks!

edaniels:
Amended my commit for the DCO (sorry about that). So my code looks a little like this:
```
	return instrumentHandlerCounterIn(reqInCounter,
		promhttp.InstrumentHandlerCounter(reqOutCounter,
			instrumentHandlerInFlight(inFlight,
				promhttp.InstrumentHandlerDuration(requestDurationObs, next))))
```

I noticed that `InstrumentHandlerCounter` correctly gave me a `hijackerDelegator` but not a pointer to it. This is due to the picker func:

```
	pickDelegator[hijacker] = func(d *responseWriterDelegator) delegator { // 4
		return hijackerDelegator{d}
	}
```

So I could have fixed it there with a ref instead but yeah as you said, it seems more sensical to have a value struct wrap pointers, not a pointer to a struct with pointers.

beorn7:
I believe you have looked at an older version of the library. In master, the code you have quoted above, looks like this:

```Go
	pickDelegator[hijacker] = func(d *responseWriterDelegator) delegator { // 4
		return &hijackerDelegator{d}
	}
```

So there is no bug in master.
But you are right that using a value receiver makes more sense in the first place.

But then we should also remove all the now unneeded `&` from the code. And we should do the same with the `pusherDelegator` in delegator_1_8.go.

Would you mind doing that in this PR? If that's too much effort right now, I can also do it.

edaniels:
Indeed. You are correct. I was using an older version of the library. Removed the references to the delegators.

prometheus/promhttp/delegator.go
```diff
@@ -147,45 +147,45 @@ func init() {
 			*responseWriterDelegator
 			io.ReaderFrom
 			http.CloseNotifier
-		}{d, &readerFromDelegator{d}, &closeNotifierDelegator{d}}
+		}{d, &readerFromDelegator{d}, closeNotifierDelegator{d}}
```
beorn7: `&readerFromDelegator` should now also work as simply `readerFromDelegator`, shouldn't it? Same for the other occurrences below.

prometheus/promhttp/delegator_1_8.go
```diff
@@ -22,58 +22,58 @@ import (
...
-		}{d, &pusherDelegator{d}, &flusherDelegator{d}, &closeNotifierDelegator{d}}
+		}{d, pusherDelegator{d}, flusherDelegator{d}, closeNotifierDelegator{d}}
 	}
 	pickDelegator[pusher+hijacker] = func(d *responseWriterDelegator) delegator { // 20
 		return struct {
 			*responseWriterDelegator
 			http.Pusher
 			http.Hijacker
-		}{d, &pusherDelegator{d}, &hijackerDelegator{d}}
+		}{d, pusherDelegator{d}, &hijackerDelegator{d}}
```
beorn7: `&hijackerDelegator` should be `hijackerDelegator`, no? A few more occurrences below.

beorn7:
Thanks again. All good now. Merging…

PR #433, opened 2018-08-02
Unify Representation of Counter Values

valarauca:
### Summery

In short, why use two fields when you can use one.

### Motivation

1. Equating the cast results of floating points is a [non-trivial problem](https://randomascii.wordpress.com/2017/06/19/sometimes-floating-point-math-is-perfect/) and it shouldn't be relied on this.
2. Using 2 load fences will generally always be less performant then 1.
3. My company is dropping metrics in production, while investigating the problem I realized very small values in the `Add()` field could be dropped (see: `1`)

### Changes

* Removes the `bitsInt` field from `counter`
* The old `Inc()` function is now wrapping `Add(1)`

### Additional Changes

I updated the test suite to have the correct values.

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -37,6 +37,7 @@ type Counter interface {
 	// Inc increments the counter by 1. Use Add to increment it by arbitrary
 	// non-negative values.
 	Inc()
+
 	// Add adds the given value to the counter. It panics if the value is <
 	// 0.
 	Add(float64)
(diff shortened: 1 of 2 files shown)


valarauca:
So I'm not sure how to DCO happy, I follow the instructions. 

One of my commit messages me stating I agree to the ToS and including the entire ToS. So let me know what I need to do.

davecheney:
@valarauca you need to use `git commit -s` to sign the commit.

I recommend

1. squash all your commits back to one, then you only need to sign one commit
2. then, `git commit -s --amend`
3. `git push -f`

valarauca:
Yeah that "how-it-works" is functionally useless as following its instructions normally results in an error. 

Also squashing from the CLI is extremely difficult.

PR #434, opened 2018-08-02
Unify Representation of Counter Values

valarauca:
### Summery

In short, why use two fields when you can use one.

### Motivation

1. Equating the cast results of floating points is a [non-trivial problem](https://randomascii.wordpress.com/2017/06/19/sometimes-floating-point-math-is-perfect/) and it shouldn't be relied on this.
2. Using 2 load fences will generally always be less performant then 1.
3. My company is dropping metrics in production, while investigating the problem I realized very small values in the `Add()` field could be dropped (see: `1`)

### Changes

* Removes the `bitsInt` field from `counter`
* The old `Inc()` function is now wrapping `Add(1)`

### Additional Changes

I updated the test suite to have the correct values.

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -37,6 +37,7 @@ type Counter interface {
 	// Inc increments the counter by 1. Use Add to increment it by arbitrary
 	// non-negative values.
 	Inc()
+
 	// Add adds the given value to the counter. It panics if the value is <
 	// 0.
 	Add(float64)
@@ -72,7 +73,6 @@ type counter struct {
 	// in the struct to guarantee alignment for atomic operations.
 	// http://golang.org/pkg/sync/atomic/#pkg-note-BUG
 	valBits uint64
-	valInt  uint64
 
 	selfCollector
 	desc *Desc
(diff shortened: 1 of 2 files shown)


beorn7:
Thanks for your contribution.

Note that you are undoing what was introduced in #367 and discussed excessively in #362 . The change was motivated to improve the performance of the counter. Did you actually observe better performance with the old implementation prior to #367 (which is the same you are suggesting now)? In which use cases? How significant was the performance change?

WRT adding a very small number to a counter: Note that a counter that already has reached a value of a certain size will naturally drop small increments as the Prometheus data model is based on IEEE 64bit floating point values. However, if the code that is supposed to find out if the argument of `Add` can be represented as an integer is buggy (as you might be suggesting above), that would be a real bug. If I'm not mistaken, you did not add a test case for that to the test file. If we had a test case, we could try to improve the code for detecting if a float64 can be truthfully represented by an int, which would solve the problem.

beorn7:
I played a bit with the code, and even `Add(math.SmallestNonzeroFloat64)` behaved correctly for me, i.e. the code recognized it's a non-integer increment and used the float part of the internal counter state.

I cannot really see a way where the current code would drop an increment that a vanilla float64-only implementation would not drop. Happy to see code that proves me wrong.

PR #437, opened 2018-08-07
go routine -> goroutine

lidalei:
Signed-off-by: Dalei Li <dalei.li@icloud.com>


beorn7:
Thanks. Could you add the [DCO](https://github.com/probot/dco#how-it-works) to the commit message so that we can merge this?

lidalei:
Sorry, I don't how to modify it. So I will create a new one.

PR #438, opened 2018-08-07
Add process_maximum_virtual_memory_bytes metric

simonpasquier:
Follow-up of https://github.com/prometheus/prometheus/pull/4418#issuecomment-407789154


brian-brazil:
Any new standard exports should be added to https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics, as these are a standard across all client libraries in the ecosystem.

The name should also be process_virtual_memory_max_bytes per standard naming, and to be consistent with process_max_fds.

grobie:
> The name should also be process_virtual_memory_max_bytes per standard naming

@brian-brazil I wasn't aware there are any standardized patterns around max/min/limit/etc. I'm fine with your suggestion, but couldn't find anything in that regard on https://prometheus.io/docs/practices/naming/. Is that documented anywhere else?

brian-brazil:
The general rule is to go from less to more specific, so related metrics have the same prefix.

beorn7:
Thanks everybody for the hints and applying them.

PR #439, opened 2018-08-07
go routine to goroutine

lidalei:
go routine to goroutine
Signed-off-by: Dalei Li <dalei.li@icloud.com>


beorn7:
:+1: Thanks again.

PR #443, opened 2018-08-18
Add a wrapper to add a timestamp to a metric

beorn7:
Supersedes #407 .

@simonpasquier 

### DIFF
--- a/prometheus/metric.go
+++ b/prometheus/metric.go
@@ -15,7 +15,9 @@ package prometheus
 
 import (
 	"strings"
+	"time"
 
+	"github.com/gogo/protobuf/proto"
 	dto "github.com/prometheus/client_model/go"
 )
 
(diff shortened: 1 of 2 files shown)


prometheus/metric.go
```diff
@@ -142,3 +145,31 @@ func NewInvalidMetric(desc *Desc, err error) Metric {
...
+	t time.Time
+}
+
+func (m timestampedMetric) Write(pb *dto.Metric) error {
+	e := m.Metric.Write(pb)
+	pb.TimestampMs = proto.Int64(m.t.Unix()*1000 + int64(m.t.Nanosecond()/1000000))
+	return e
+}
+
+// NewMetricWithTimestamp returns a new Metric wrapping the provided Metric in a
```
brian-brazil: This sounds like it works with direct instrumenation, I think we should try to limit this to MustNewConstMetric. Everyone I've seen trying to do timestamps on direct instrumentation was otherwise using the library incorrectly.

beorn7: The usual use case is already described in the doc comment. Artificially limiting this to the output of `MustNewConstMetric` (and `MustNewConstSummary` and `MustNewConstHistogram` and `NewConstMetric` and `NewConstSummary` and `NewConstHistogram`) would be weird and cumbersome. Using this wrapper is arcane enough to not happen accidentally. Whoever elects to use such a contraption without reading the doc comments is on their own.

beorn7: Additional point: Since the timestamp is immutable, people that incorrectly wrap metrics from direct instrumentation will quickly notice that they are exposing the same timestamp forever.

brian-brazil: That's true.

prometheus/metric.go
```diff
@@ -142,3 +145,31 @@ func NewInvalidMetric(desc *Desc, err error) Metric {
...
 func (m *invalidMetric) Write(*dto.Metric) error { return m.err }
+
+type timestampedMetric struct {
+	Metric
+	t time.Time
+}
+
+func (m timestampedMetric) Write(pb *dto.Metric) error {
+	e := m.Metric.Write(pb)
+	pb.TimestampMs = proto.Int64(m.t.Unix()*1000 + int64(m.t.Nanosecond()/1000000))
```
luqasn: Probably does not make a difference, but I’d return early in case of an error

simonpasquier:
:+1: very nice!

prometheus/metric.go
beorn7: Somehow my comment here got lost.

It probably really doesn't make a difference, but I find it a bit more robust to set the timestamp even in case of an error. We don't explicitly document to not make any use of the protomessage after an error is returned, so there might be a future use where the error conveys some message that looking at the proto message still makes sense.

Right now, of course, nobody should ever look at the proto message after an error, but then it doesn't matter if we set the timestamp or not.

atyronesmith:
Many thanks for this enhancement!  This allows for fine-grain event correlation and opens Prometheus to a much wider set of applications.

PR #446, opened 2018-08-23
Add prometheus common makefile

### LINKED ISSUE #430: Introduce common Prometheus Makefile
See #406 for context where this is useful.

glefloch:
This pull request aims to close #430. 
@beorn7, other prometheus projects use the following make command, should I set it here too? 

     make check_license style unused test staticcheck

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,19 @@
+# Copyright 2018 The Prometheus Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include Makefile.common
+
+.PHONY: get_dep
+getdep:
+	@echo ">> getting dependencies" 
+	$(GO) get -t ./...
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,9 @@
 include Makefile.common
 
 .PHONY: get_dep
-getdep:
+get_dep:
 	@echo ">> getting dependencies" 
 	$(GO) get -t ./...
+
+.PHONY: test
+test: get_dep common-test
--- a/Makefile.common
+++ b/Makefile.common
@@ -1,0 +1,132 @@
+# Copyright 2018 The Prometheus Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include Makefile.common
+
+.PHONY: get_dep
+getdep:
```
beorn7: That's spelled different here than anywhere else (`get_dep`). In different news, could you turn this into a real (Make-style) dependency of the Make targets that depend on getting the dependencies? (Must go deeper...)

I'm not a Make guru myself, but I guess there is some way to override the dependency structure here in `Makefile` to add dependencies to `Makefile.common`. (In doubt, let's page @matthiasr .)

beorn7:
`make check_license style unused test staticcheck` sounds like a good choice. If you make `get_dep` a dependencies of targets in need of it, you could take it verbatim.

Makefile.common
```diff
@@ -0,0 +1,132 @@
...
+DOCKER_IMAGE_TAG        ?= $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))
+DOCKER_REPO             ?= prom
+
+.PHONY: all
+all: style staticcheck unused build test
+
+# This rule is used to forward a target like "build" to "common-build".  This
+# allows a new "build" target to be defined in a Makefile which includes this
+# one and override "common-build" without override warnings.
+%: common-% ;
```
beorn7: Look, this seems to be the way to go. You can override the targets that need `go_dep` and let them depend on `go_dep` and `common-xxx`.

glefloch: You are right, this looks really cleaner: 

    .PHONY: test
    test: get_dep common-test

grobie:
staticcheck only supports the last two minor versions. The go project itself does not provide bug fixes for older versions, and security fixes only for the last version before the current one https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance

I see the following options:
1) client_golang follows the go project and does not run tests against versions older than current-1
2) remove staticcheck
3) check travisci for a way to run staticcheck only for 1.9 and 1.10

I vote for 1.

glefloch:
@grobie, you mean removing version `1.7.x` and `1.8.x` from .travis.yml?

beorn7:
Hmmm, given the widespread use of this library, I prefer to make sure it still runs with 1.7 over making sure it doesn't have any style violations. We had multiple incidents where people that were forced to run older Go versions bumped into requirements of newer Go versions by this library. Obviously, we don't have to support 1.7 forever, but only supporting the last minor release will be a bit harsh.

It would be great if we could make (3) possible. Otherwise, I'll vote for (2). We can still run staticcheck manually on PRs and such.

glefloch:
@beorn7, after having a look into travis documentation, it appears that travis expose go version used for the build in `TRAVIS_GO_VERSION` variable. I thinks the following script should work, would it be acceptable ?

    script:
        - make check_license style unused test
        -  if [[ $TRAVIS_GO_VERSION =~ ^1\.(9|10)\.[0-9]+$ ]]; then make staticcheck; fi

grobie:
Looks good to me @glefloch! I'd just suggest to blacklist the non-working go versions rather than whitelisting the currently working ones, so that we won't need to touch that statement when we add go1.11 for example.

beorn7:
Cool. Thank you very much.

beorn7:
I'll have a look at the duration of the test runs. We can go back to `-short` if the tests tend to run for too long.

Issue #387, opened 2018-02-23
promhttp: Sanitize method to ALLCAPS


beorn7:
`sanitizeMethod` is currently sanitizing methods to lower-case. However, is is way more common to use all-caps. See also [the RFC](https://tools.ietf.org/html/rfc7231#section-4.3 uses all-caps).

As we are breaking everything with 0.10 anyway, let's rectify this and sanitize to all-caps, using the [constants in the Go http package](https://golang.org/pkg/net/http/#pkg-constants).

metalmatze:
But isn't it way more common in Prometheus metrics to just lowercase everything?

brian-brazil:
For the metric and label names yes, there's not that convention on label values. In general you want what the underlying data is.

beorn7:
I looked around a bit and found that most people use upper-case for HTTP method names in a Prometheus label _value_. Except, of course, they use the client_golang tooling. Looking at the RFC and the constants in the Go http package, it really looks that the lower-casing (introduced very early in the history of client_golang) was a mistake.

metalmatze:
Alright. I guess this change seems good then.
Thanks for the clarification! :ok_hand:

beorn7:
It _is_ a breaking change and is therefore in the v0.10 milestone.

beorn7:
Fixed in https://github.com/prometheus/client_golang/tree/dev-0.10

PR #447, opened 2018-08-27
Sanitize http method to uppercase

glefloch:
This pull request aims to close issue #387. As this is part of version 0.10.0, should it be merge in master or in another branch? 

### DIFF
--- a/prometheus/promhttp/instrument_server.go
+++ b/prometheus/promhttp/instrument_server.go
@@ -322,23 +322,23 @@ func computeApproximateRequestSize(r *http.Request) int {
 func sanitizeMethod(m string) string {
 	switch m {
 	case "GET", "get":
-		return "get"
+		return "GET"
 	case "PUT", "put":
-		return "put"
+		return "PUT"
 	case "HEAD", "head":
-		return "head"
+		return "HEAD"
 	case "POST", "post":
-		return "post"
+		return "POST"
 	case "DELETE", "delete":
-		return "delete"
+		return "DELETE"
 	case "CONNECT", "connect":
-		return "connect"
+		return "CONNECT"
 	case "OPTIONS", "options":
-		return "options"
+		return "OPTIONS"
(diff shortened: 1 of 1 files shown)


glefloch:
Yes for sure, no problem

GheRivero:
LGTM. small nit

beorn7:
I changed the base branch and will merge now. Thanks.

prometheus/promhttp/instrument_server.go
```diff
@@ -322,23 +322,23 @@ func computeApproximateRequestSize(r *http.Request) int {
...
+		return "CONNECT"
 	case "OPTIONS", "options":
-		return "options"
+		return "OPTIONS"
 	case "NOTIFY", "notify":
-		return "notify"
+		return "NOTIFY"
 	default:
-		return strings.ToLower(m)
+		return strings.ToUpper(m)
```
GheRivero: The default rule will work for any case. No need to duplicate code. 
Or made the default case to error: non permitted HTTP method.

GheRivero (commented):
.

prometheus/promhttp/instrument_server.go
beorn7: > The default rule will work for any case. No need to duplicate code.

I don't understand. What code is duplicated here?

> Or made the default case to error: non permitted HTTP method.

The goal here is not to validate. If people really want to, they could come up with fantasy HTTP methods (or there might even be more added officially). The goal here is to make sure that the method is always upper case.

GheRivero: If you remove all the cases, except the default one, the output will be the same. Or even more simple:
```
func sanitizeMethod(m string) string {
    return strings.ToUpper(m)
}
```

GheRivero: Anyway, it's just a cosmetic suggestion for clarity, so just ignore it if it's not worth the time.

PR #448, opened 2018-08-27
Using the recommended label syntax for maintainer in Dockerfile.

### LINKED ISSUE #449: MAINTAINER instruction in Dockerfile is deprecated
See https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

Use the following format

`LABEL maintainer="..."`

tariq1890:
Fixes #449

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ RUN CGO_ENABLED=0 GOOS=linux go build -a -tags netgo -ldflags '-w'
 
 # Final image.
 FROM prom/busybox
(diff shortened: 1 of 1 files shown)


tariq1890:
Looks like an unrelated test case failure. Is that a flakey test?

simonpasquier:
Probably a flake, I've restarted the CI.

beorn7:
Thanks.

WRT the flaky tests: This might be because we dropped the restriction on short tests recently. I'll check if we should restrict them again.

PR #452, opened 2018-09-02
Create an internal package

beorn7:
This is for types we don't want to export but which are used in
different packages within client_golang.

Currently, that's only NormalizeMetricFamilies (used in the prometheus
package and in the testutil package). More to be added as needed.

@grobie


brancz:
lgtm :+1:

grobie:
:+1:

PR #453, opened 2018-09-03
Un-export prometheus.LabelPairSorter

beorn7:
The only known external usage of it was in prometheus/pushgateway,
where it was removed by
https://github.com/prometheus/pushgateway/pull/200 .

Originally, the expectation was that users would implement the Metric
interface now and then. As we know now, neither it is happening, nor
would it make a lot of sense. (Users implement the Collector interface
instead.) By now, LabelPairSorter is essentially noise in the already
quite cluttered namespace in the prometheus package.

### DIFF
--- a/prometheus/desc.go
+++ b/prometheus/desc.go
@@ -156,7 +156,7 @@ func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *
 			Value: proto.String(v),
 		})
 	}
-	sort.Sort(LabelPairSorter(d.constLabelPairs))
+	sort.Sort(labelPairSorter(d.constLabelPairs))
 	return d
 }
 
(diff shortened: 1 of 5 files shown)


aledbf:
@beorn7 this is also being used here https://github.com/kubernetes/kube-state-metrics/blob/master/pkg/collectors/testutils/testutils.go

beorn7:
Hmm, the Sourcegraph search doesn't seem to have good coverage...
There are two options now:
1. Use the very similar [`testutil` package](https://godoc.org/github.com/prometheus/client_golang/prometheus/testutil) we have now here in client_golang.
2. It's just very trivial boilerplate code, so you could simply replicate it as I did for the Pushgateway.

PR #454, opened 2018-09-04
Add a single-value extractor and general package doc

beorn7:
@grobie 

This concludes #58. Note that more test utils will be added as per #230 with the improved type stucture planned for v0.10.

### DIFF
--- a/prometheus/testutil/testutil_test.go
+++ b/prometheus/testutil/testutil_test.go
@@ -20,6 +20,104 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+type untypedCollector struct{}
+
+func (u untypedCollector) Describe(c chan<- *prometheus.Desc) {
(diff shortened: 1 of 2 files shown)


prometheus/testutil/testutil_test.go
```diff
@@ -20,6 +20,104 @@ import (
...
+			panics:    true,
+		},
+	}
+
+	for n, s := range scenarios {
+		func() {
+			defer func() {
+				r := recover()
+				if r == nil && s.panics {
+					t.Errorf("%s: expected panic", n)
```
grobie: Instead of having to prefix every error message with the scenario name, it would be easier and imho more idiomatic to use go's subtests instead of the plain `func()`.

beorn7: You are totally right. Since sub-tests are supported since Go1.7, there is no reason to avoid them.
I'll transform this as an exercise for myself…

PR #456, opened 2018-09-06
Rework process collector

### LINKED ISSUE #219: ProcessCollector: rework

beorn7:
This unifies both constructors in one with an options argument.

The options argument allows to switch on error reporting, as discussed
in #219.

The change of the contructor signature is breaking, but only mildly
so. Plus, the process collector is rarely used explicitly. I used
Sourcegraph to search for public usages, with the following results:

- 2 occurrences of NewProcessCollectorPIDFn, once in @discordianfish's
  glimpse, once in @fabxc's etcd_exporter (deprecated anyway). Both
  are Prom veterans and will simply do the one line change if needed.
- 8 occurrences of NewProcessCollector, of which 7 are of the form
    NewProcessCollector(os.Getpid(), "")
  Thus, it's a very easy change, which I even hinted at in the doc
  comment.

Fixes #219.

### DIFF
--- a/prometheus/process_collector_test.go
+++ b/prometheus/process_collector_test.go
@@ -17,12 +17,15 @@ package prometheus
 
 import (
 	"bytes"
+	"errors"
 	"os"
 	"regexp"
 	"testing"
 
 	"github.com/prometheus/common/expfmt"
 	"github.com/prometheus/procfs"
+
+	dto "github.com/prometheus/client_model/go"
 )
 
 func TestProcessCollector(t *testing.T) {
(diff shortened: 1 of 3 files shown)


prometheus/process_collector_test.go
```diff
@@ -72,4 +78,23 @@ func TestProcessCollector(t *testing.T) {
...
+		ReportErrors: true,
+	})
+
+	ch := make(chan Metric)
+	go func() {
+		brokenProcessCollector.Collect(ch)
+		close(ch)
+	}()
+	for m := range ch {
+		fmt.Println("feep")
```
grobie: What is this?

beorn7: Errr… probably just to test the reviewer. :o)

beorn7: Will remove before merging.

PR #458, opened 2018-09-11
Wrappable registerers

beorn7:
Essentially middleware for Registerers!

@juliusv @fabxc @brancz you all requested this feature. Sorry for the long delay.

@stuartnelson3 as my local main reviewer while @grobie is on vacations.

### DIFF
--- a/prometheus/example_clustermanager_test.go
+++ b/prometheus/example_clustermanager_test.go
@@ -17,18 +17,18 @@ import "github.com/prometheus/client_golang/prometheus"
 
 // ClusterManager is an example for a system that might have been built without
 // Prometheus in mind. It models a central manager of jobs running in a
-// cluster. To turn it into something that collects Prometheus metrics, we
-// simply add the two methods required for the Collector interface.
+// cluster. Thus, we implement a custom Collector called
+// ClusterManagerCollector, which collects information from a ClusterManager
+// using its provided methods and turns them into Prometheus Metrics for
+// collection.
 //
 // An additional challenge is that multiple instances of the ClusterManager are
 // run within the same binary, each in charge of a different zone. We need to
-// make use of ConstLabels to be able to register each ClusterManager instance
-// with Prometheus.
+// make use of wrapping Registerers to be able to register each
+// ClusterManagerCollector instance with Prometheus.
 type ClusterManager struct {
-	Zone         string
-	OOMCountDesc *prometheus.Desc
-	RAMUsageDesc *prometheus.Desc
-	// ... many more fields
+	Zone string
+	// Contains many more fields not listed in this example.
 }
 
 // ReallyExpensiveAssessmentOfTheSystemState is a mock for the data gathering a
@@ -50,10 +50,30 @@ func (c *ClusterManager) ReallyExpensiveAssessmentOfTheSystemState() (
 	return
 }
 
-// Describe simply sends the two Descs in the struct to the channel.
-func (c *ClusterManager) Describe(ch chan<- *prometheus.Desc) {
-	ch <- c.OOMCountDesc
-	ch <- c.RAMUsageDesc
+// ClusterManagerCollector implements the Collector interface.
+type ClusterManagerCollector struct {
+	ClusterManager *ClusterManager
+}
+
+// Descriptors used by the ClusterManagerCollector below.
+var (
+	oomCountDesc = prometheus.NewDesc(
+		"clustermanager_oom_crashes_total",
+		"Number of OOM crashes.",
+		[]string{"host"}, nil,
+	)
+	ramUsageDesc = prometheus.NewDesc(
+		"clustermanager_ram_usage_bytes",
+		"RAM usage as reported to the cluster manager.",
+		[]string{"host"}, nil,
+	)
+)
+
+// Describe is implemented with DescribeByCollect. That's possible because the
+// Collect method will always return the same two metrics with the same two
+// descriptors.
+func (cc ClusterManagerCollector) Describe(ch chan<- *prometheus.Desc) {
+	prometheus.DescribeByCollect(cc, ch)
 }
 
 // Collect first triggers the ReallyExpensiveAssessmentOfTheSystemState. Then it
@@ -61,11 +81,11 @@ func (c *ClusterManager) Describe(ch chan<- *prometheus.Desc) {
 //
 // Note that Collect could be called concurrently, so we depend on
 // ReallyExpensiveAssessmentOfTheSystemState to be concurrency-safe.
-func (c *ClusterManager) Collect(ch chan<- prometheus.Metric) {
-	oomCountByHost, ramUsageByHost := c.ReallyExpensiveAssessmentOfTheSystemState()
+func (cc ClusterManagerCollector) Collect(ch chan<- prometheus.Metric) {
+	oomCountByHost, ramUsageByHost := cc.ClusterManager.ReallyExpensiveAssessmentOfTheSystemState()
 	for host, oomCount := range oomCountByHost {
 		ch <- prometheus.MustNewConstMetric(
-			c.OOMCountDesc,
+			oomCountDesc,
 			prometheus.CounterValue,
 			float64(oomCount),
 			host,
@@ -73,7 +93,7 @@ func (c *ClusterManager) Collect(ch chan<- prometheus.Metric) {
 	}
 	for host, ramUsage := range ramUsageByHost {
 		ch <- prometheus.MustNewConstMetric(
-			c.RAMUsageDesc,
+			ramUsageDesc,
 			prometheus.GaugeValue,
 			ramUsage,
 			host,
@@ -81,38 +101,27 @@ func (c *ClusterManager) Collect(ch chan<- prometheus.Metric) {
 	}
 }
 
-// NewClusterManager creates the two Descs OOMCountDesc and RAMUsageDesc. Note
-// that the zone is set as a ConstLabel. (It's different in each instance of the
-// ClusterManager, but constant over the lifetime of an instance.) Then there is
-// a variable label "host", since we want to partition the collected metrics by
-// host. Since all Descs created in this way are consistent across instances,
-// with a guaranteed distinction by the "zone" label, we can register different
-// ClusterManager instances with the same registry.
-func NewClusterManager(zone string) *ClusterManager {
-	return &ClusterManager{
+// NewClusterManager first creates a Prometheus-ignorant ClusterManager
+// instance. Then, it creates a ClusterManagerCollector for the just created
+// ClusterManager. Finally, it registers the ClusterManagerCollector with a
+// wrapping Registerer that adds the zone as a label. In this way, the metrics
+// collected by different ClusterManagerCollectors do not collide.
+func NewClusterManager(zone string, reg prometheus.Registerer) *ClusterManager {
+	c := &ClusterManager{
 		Zone: zone,
-		OOMCountDesc: prometheus.NewDesc(
-			"clustermanager_oom_crashes_total",
-			"Number of OOM crashes.",
-			[]string{"host"},
-			prometheus.Labels{"zone": zone},
-		),
-		RAMUsageDesc: prometheus.NewDesc(
-			"clustermanager_ram_usage_bytes",
-			"RAM usage as reported to the cluster manager.",
-			[]string{"host"},
-			prometheus.Labels{"zone": zone},
-		),
 	}
+	cc := ClusterManagerCollector{ClusterManager: c}
+	prometheus.WrapWith(prometheus.Labels{"zone": zone}, reg).MustRegister(cc)
+	return c
 }
 
 func ExampleCollector() {
-	workerDB := NewClusterManager("db")
-	workerCA := NewClusterManager("ca")
-
 	// Since we are dealing with custom Collector implementations, it might
 	// be a good idea to try it out with a pedantic registry.
 	reg := prometheus.NewPedanticRegistry()
-	reg.MustRegister(workerDB)
-	reg.MustRegister(workerCA)
+
+	// Construct cluster managers. In real code, we would assign them to
+	// variables to then do something with them.
+	NewClusterManager("db", reg)
+	NewClusterManager("ca", reg)
 }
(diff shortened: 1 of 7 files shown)


prometheus/wrap.go
```diff
@@ -0,0 +1,172 @@
...
+// registered with the returned Registerer will be registered with the wrapped
+// Registerer in a modified way. The modified Collector adds the provided Labels
+// to all Metrics it collects (as ConstLabels). The Metrics collected by the
+// unmodified Collector must not have any of those labels on their own.
+//
+// WrapWith provides a way to add fixed labels to a subset of Collectors. It
+// should not be used to add fixed labels to all metrics exposed.
+//
+// The Collector example demonstrates a use of WrapWith.
+func WrapWith(labels Labels, reg Registerer) Registerer {
```
brancz: I don't think the "Wrap" terminology is necessary and the "wrapped" thing should probably be the first argument, then the labels can be variadic which is a more natural interface I think. I'd suggest to rename the function to `WithLabels`.

prometheus/wrap.go
```diff
@@ -0,0 +1,172 @@
...
+
+// WrapWithPrefix returns a Registerer wrapping the provided Registerer. Collectors
+// registered with the returned Registerer will be registered with the wrapped
+// Registerer in a modified way. The modified Collector adds the provided prefix
+// to the name of all Metrics it collects.
+//
+// WrapWithPrefix is useful to have one place to prefix all metrics of a
+// sub-system. To make this work, register metrics of the sub-system with the
+// wrapping Registerer returned by WrapWithPrefix.
+func WrapWithPrefix(prefix string, reg Registerer) Registerer {
```
brancz: Similar suggestion to `WrapWith`, the registerer should be the first argument and the function name I'd suggest to just be `WithPrefix`.

prometheus/wrap.go
```diff
@@ -0,0 +1,172 @@
...
+		// No wrapping labels.
+		return nil
+	}
+	for ln, lv := range m.labels {
+		out.Label = append(out.Label, &dto.LabelPair{
+			Name:  proto.String(ln),
+			Value: proto.String(lv),
+		})
+	}
+	sort.Sort(labelPairSorter(out.Label))
```
brancz: I realize that we recommend this, but above all we want a stable label order, no? I wonder whether this might be over doing it and potentially causing problems in performance in rendering of the metrics output.

prometheus/example_clustermanager_test.go
```diff
@@ -50,69 +50,78 @@ func (c *ClusterManager) ReallyExpensiveAssessmentOfTheSystemState() (
...
-// NewClusterManager creates the two Descs OOMCountDesc and RAMUsageDesc. Note
-// that the zone is set as a ConstLabel. (It's different in each instance of the
-// ClusterManager, but constant over the lifetime of an instance.) Then there is
-// a variable label "host", since we want to partition the collected metrics by
-// host. Since all Descs created in this way are consistent across instances,
-// with a guaranteed distinction by the "zone" label, we can register different
-// ClusterManager instances with the same registry.
-func NewClusterManager(zone string) *ClusterManager {
-	return &ClusterManager{
+// NewClusterManager creates first a Prometheus-ignorant ClusterManager
```
juliusv: "creates first" -> "first creates"

prometheus/wrap.go
juliusv: I mostly agree with these comments, although it's *slightly* weird to have `WithXXX()` functions on the package level rather than on a concrete type. Since this already returns us a non-global registry anyway, we might as well define the method on a registry itself and then people can call it on the default registry if they want? `myRegistry.WithLabels(...)`. Like contexts do it.

juliusv: (ok contexts were a stupid example, they define package-level functions instead of methods, but that is because the context type is the only one that's relevant in that whole package)

prometheus/wrap.go
stuartnelson3: Given how relatively infrequent scrapes are in comparison to other actions done by the client, I would guess that over all this has a relatively small effect on performance.

prometheus/wrap.go
brancz: I agree it would make sense to have them just on the registry itself, but throws up the question whether they should be part of the interface (which seems odd) or part of the registry types, which would cause type switches/casts to wrap already wrapped registries. It seems to me that the package level functions are actually a reasonable tradeoff.

prometheus/wrap.go
brancz: The problem I'm seeing is that we sort potentially multiple times per sample as we sort at every stage of "unwrapping", I don't see the value add of sorting, when we already have a stable order. I realize the stable order may not apply for custom `prometheus.Collector` structs, but that should be part of the interface not the infrastructure enforcing it.

beorn7: Traditionally, client_golang has always provided "perfectly" sorted output, despite not being required by the exposition format spec. I would like to keep it that way for now to avoid surprises. In this concrete examples, the allocation of the LabelPair proto message will probably need more resources than the sorting. When we move to a faster protobuf library or to no protobuf at all, we can think about performance tradeoffs of sorting. But that's again a v0.10 issue to be discussed later.

prometheus/example_clustermanager_test.go
beorn7: Reworded. PHAL.

beorn7:
Thanks for all your feedback. Changed comments as suggested.

WRT the API:

I'm fairly confident about the argument order and that we have to use top-level "middleware-like" functions. Also, the `With(Labels)` is consistent with other uses and reads nicer than `WithLabels(Labels)`. However, I'm open to suggestions for the naming, i.e. use a different verb than `Wrap` or use a longer name like `WrapRegistererWith`. I couldn't come up with better names myself, though, so I need concrete suggestions.

beorn7:
Any more ideas or comments?

beorn7:
@juliusv So yeah, let's err on the explicit side. I have renamed `WrapWith...` to `WrapRegistererWith...`.

brancz:
I personally think Registerer and Wrap in the function name does not add to the expressiveness. Just “With” is an otherwise used pattern in golang, and seems in my opinion perfectly fine. I have the bigger issue with having two methods here, one regarding labels one regarding a prefix. I don’t have data to back this up, but my feeling is that prefixing for namespacing is going to be most widely used, so I’m tending towards “With” doing a prefix, and “WithLabels” (simply because overloading is not possible, probably for the better :) )

That said I’m incredibly happy that you’re working on this and that naming is the only thing we seem to need to decide on :)

beorn7:
Avoiding stutter seems to be a big deal in the Go community, thus I would really like to avoid the `WithLabels(Labels{"foo": "bar"})` stutter (while `WithPrefix("pre_")` reads perfectly fine and arguably better than `With("pre_")`.

About the bare `With`: I think that would only be justified if we had a separate `registry` package. But the `prometheus` package has just too many things in it (incidentally even a `With(Labels)` method, which has a much different meaning (but at least it's a method, not a top-level function)).

brancz:
That makes a lot of sense. Seems like there is a dilemma unless we create the registry package and type alias, which also sounds a bit messy as the registry type is rather essential to the prometheus package. But maybe the right choice in the long run?

beorn7:
I would postpone a package restructuring to v0.10, which will anyway break many things (perhaps even everything... :-o ). (I tried out some crazy ideas already. However, we should also keep in mind that too many packages are confusing, too, and that we are catering for a relatively advanced use case here. Everybody will use a registry, but only few will need to wrap one. So we should not make things more complicated for everyone to make naming a bit more concise for a few.)

brancz:
That’s fair. Given all the constraints my personal favorites are WithLabels and WithPrefix. As the API being

WithPrefix(r Registerer, prefix string) Registerer
WithLabels(r Registerer, labels ...Label) Registerer

Adding “registerer” to the Namen doesn’t add to expressiveness as the registerer is an argument of the function and the WithLabels using a variadic Parameter removes the stutter. Both pkg/errors and go-kit/log use the subject as the first argument, my feeling is that it would make sense to stay consistent with them, even if I understand your argument.

beorn7:
I would like to prefer consistency _within_ this package and its sub-packages over consistency with other packages. And we already have tons of http.Handler-wrapping middleware here that use the thing to wrap last, mostly for better stacking, which is also valid here, as already argued above.

About the variadic labels: This is again about consistency. As mentioned above, we aren't using variadic arguments for labels anywhere else (only for label _values_). We _do_ have a `Labels` type already but no `Label` type. (Again, all of this is up for discussion with the v0.10 release. But for now, I would like to stay consistent within the package and not introduce new concepts at this time.)

Part of the above is already repeating arguments.

If that means all arguments are on the table now, then my conclusion is that the current state of this PR is the least bad and my preference. From the above, I'd assume @juliusv and @stuartnelson3 are in broad agreement. If there are other arguments to consider or if anybody wants to support @brancz's suggestion, please let us know.

juliusv:
@beorn7 Yeah, I don't feel strongly either way (my favorite thing to say), definitely can see the argument about consistency within the existing package.

brancz:
I'm not sure how much I like the free-form string lists, I think the compiler can be useful :slightly_smiling_face: . With the label struct a non variadic API is also not so bad:

```go
type Labels []Label

type Label struct {
        Name, Value string
}

---

With(Labels{                                                                                                                                                                  
    {"test_key1", "test-value1"},                                                                                                                                                    
    {"test_key2", "test-value2"},                                                                                                                                                    
}, reg)
```

juliusv:
Are there more upsides to variadic arguments than the original "then the labels can be variadic which is a more natural interface I think"?

Since labels are fundamentally a map, I think keeping it as a map might make most sense? Also automatically ensures key uniqueness.  And as @beorn7 mentioned, going the struct route will cause linter warnings unless you key all struct fields.

brancz:
Sorry about that, it was mostly extending the thought into what may be possible with it, but I felt it was unrealistic to begin with at this point (I think I went to far with the bike-shedding).

Probably my fault, but I've lost sight of what the current proposal is, or whether that differs from what is currently proposed in this pull request. Could you comment on that?

I still feel that `WrapRegisterer` doesn't add to expressiveness, but can live with the ordering of parameters. In that sense I think my proposal would now be:

```
prometheus.With(prometheus.Labels{"label_key": "label_value"}, reg)
prometheus.WithPrefix("pre_", reg)
```

I actually don't think it's inconsistent with the other use of `With` in the package, as they are both constructs to fill labels.

prometheus/wrap.go
```diff
@@ -0,0 +1,173 @@
...
+
+// WrapRegistererWith returns a Registerer wrapping the provided
+// Registerer. Collectors registered with the returned Registerer will be
+// registered with the wrapped Registerer in a modified way. The modified
+// Collector adds the provided Labels to all Metrics it collects (as
+// ConstLabels). The Metrics collected by the unmodified Collector must not
+// duplicate any of those labels.
+//
+// WrapRegistererWith provides a way to add fixed labels to a subset of
+// Collectors. It should not be used to add fixed labels to all metrics exposed.
```
brian-brazil: I think there should be a similar warning for metric prefixes. The use cases for these are quite rare, while users trying to have target labels as metric prefixes are common.

beorn7:
> I still feel that `WrapRegisterer` doesn't add to expressiveness,

I do not doubt that statement. (And I understood it the first time you said it.) The problem is not expressiveness for the caller. The problem is that a bare `With` steals a large part of the namespace. If Go were capable of overloading, we could totally do that. But it is not, and thus we would now hijack `With` for the entire `prometheus` package for the one purpose of wrapping a Registerer with labels. If we ever introduced another feature for which `With` would be an equally or even better suited name, we would need to use a less appropriate name there.

This would all be different if we had a `registry` package. `registry.With(prometheus.Labels{"foo": "bar"}, reg)` reads beautifully because we are actually doing something with a registry. We are not polluting the namespace because the `With` is doing something that is kind of obvious from the package name and thus most likely the best fit. But what is `prometheus.With` supposed to mean? We are doing something "with Prometheus"? This is a hint that we might grab too much namespace. Perhaps it helps to think about `math.Float64frombits(b uint64) float64`. The `Float64` part doesn't add to expressiveness. `math.FromBits(b uint64) float64` would be shorter and mean the same, but since it's not in a hypothetical `float` package, it already reads a bit weird. And then, of course, there is `math.Float32frombits(b uint32) float32`. One could argue that float32 is rarely used and the general `math.FromBits` name should be given to the float64 version. But that didn't happen. In our case, one might argue that we don't have another case for `With` yet, but that argument might backfire if we one day want to introduce one for which a bare `With` would be equally or even better suited.

One more example from the standard library before I stop (I feel we are investing too much time into this): `os.NewFile(fd uintptr, name string) *File` could just be `os.New(fd uintptr, name string) *File`. There is no competing `New` function in the package. But `New` is just not specific enough for files (same as `With` is not specific enough to wrapping Registerers), so it is not used despite the return value clearly telling us that we are creating a file. In contrast, there is a naked `os.Open`, because opening is something you usually do with files.

> I actually don't think it's inconsistent with the other use of `With` in the package, as they are both constructs to fill labels.

The problem is no the "Label" part. That's indeed consistent, and consequently, I quite explicitly do not want to put the "Label" part in the more descriptive function name (in contrast to your original suggestion). The problem is the part that I _do_ miss in the name: The first `With` creates a new child in a metric vector while the other one wraps a Registerer. _Those_ are harshly different, but a naive user could easily get confused. The actually existing method `With` is fine, as it is a method of a metric vector. (It is "in the metric vector's namespace", kind of…) And so would be `With` for wrapping a registry if it were in the registry's namespace, i.e. in a hypothetical `registry` package. But for now it is in the very broad `prometheus` namespace, so I do want to play safe and make it more specific. (In fact, @juliusv convinced me to make it even more specific than initially suggested. My original idea was half-baked.)

prometheus/wrap.go
beorn7: OK, but we need to phrase this carefully as it is common to prefix all the application specific metrics (e.g. `node_...`, `prometheus_...`), which is a fine use-case. I'll try to come up with something tomorrow.

brancz:
Fair enough, given the circumstances this seems like a reasonable trade off.

This lgtm :+1:

PR #460, opened 2018-09-17
Detect invalid Descs in const metrics, but also allow empty Help strings.

### LINKED ISSUE #459: NewConstMetric does not check desc.Err
Currently, `NewConstMetric` does not check the passed `Desc.Err`, meaning for example that desc:s with missing help texts _seem_ to work, in that they result in metrics being output. Here's an example I found in some code I looked at recently:

```golang
ch <- prometheus.MustNewConstMetric(
	prometheus.NewDesc(metricName+"_total", "", nil, labels),
	prometheus.GaugeValue,
	metricValue.Total,
)
```

This gives no indication of error, and the metric shows up as expected. However, without the labels, since those are set after checking the help string.

It seems reasonable to me that the error be checked and reported, or are there reasons (other than backwards compatibility, of course) for this behaviour? Overall, why doesn't `NewDesc` return `Desc, error`?

beorn7:
Fixes #459 .

@carlpett @brian-brazil

### DIFF
--- a/prometheus/desc.go
+++ b/prometheus/desc.go
@@ -67,7 +67,7 @@ type Desc struct {
 
 // NewDesc allocates and initializes a new Desc. Errors are recorded in the Desc
 // and will be reported on registration time. variableLabels and constLabels can
-// be nil if no such labels should be set. fqName and help must not be empty.
+// be nil if no such labels should be set. fqName must not be empty.
 //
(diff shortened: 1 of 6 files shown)


carlpett:
LGTM :+1: 
Thanks for fast fix!

PR #467, opened 2018-09-25
fix exponential buckets failure message

daveworth:
it appears there is a copy/paste error in the exponential buckets test failure message which is fixed here.  @beorn7 does this seem right?


beorn7:
Very true, luckily only in the testing code.

Thanks!

PR #469, opened 2018-09-28
Add opentracing support for Prometheus client.

tomwilkie:
Basic instrumentation to support clients who want to propagate Opentracing context with this client library.

Should be a noop if clients don't have any Opentracing configured.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -25,16 +25,21 @@ import (
 	"path"
 	"strings"
 	"time"
+
+	"github.com/opentracing-contrib/go-stdlib/nethttp"
+	opentracing "github.com/opentracing/opentracing-go"
 )
 
 // DefaultRoundTripper is used if no RoundTripper is set in Config.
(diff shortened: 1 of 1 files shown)


simonpasquier:
I guess that Grafana has no problem with it :smirk: but it is worth noting that there are a few other projects importing the `v1` package (see https://godoc.org/github.com/prometheus/client_golang/api?importers) which would have to import these additional dependencies on their next vendoring update. Is that really inconvenient to delegate this instrumentation to the caller? Or create another opt-in package with opentracing support?

metalmatze:
Maybe it's possible to have these few functions calls as an local interface? :thinking: 
People could then decide on their own what to pass.

tomwilkie:
@simonpasquier I think thats a reasonable request, let me look into it.

tomwilkie:
Yeah we can do this all in our own code with no modification needed.

PR #476, opened 2018-10-09
Do not parse json on 204 responses

adamjaso:
This pull request fixes a problem where 204 responses are parsed as JSON. This incorrectly results in an error of `bad_response...`, since HTTP 204 is "No Content" and represents a successful response.

Signed-off-by: Adam Jaso <2285656+adamjaso@users.noreply.github.com>

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -477,10 +477,12 @@ func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, [
 
 	var result apiResponse
 
-	if err = json.Unmarshal(body, &result); err != nil {
-		return resp, body, &Error{
-			Type: ErrBadResponse,
-			Msg:  err.Error(),
+	if 204 != code {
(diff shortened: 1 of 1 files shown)


api/prometheus/v1/api.go
```diff
@@ -477,10 +477,12 @@ func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, [
 
 	var result apiResponse
 
-	if err = json.Unmarshal(body, &result); err != nil {
-		return resp, body, &Error{
-			Type: ErrBadResponse,
-			Msg:  err.Error(),
+	if 204 != code {
```
andrestc: Maybe use `http.StatusNoContent` instead of plain `204` here.

adamjaso: Good call. I've fixed that.

beorn7:
Paging @krasi-georgiev for final verdict.

krasi-georgiev:
Thanks

PR #478, opened 2018-10-10
 Add a concurrent end-to-end test for observe-register-gather

beorn7:
@douglas-reid @bobbytables this should test what you are doing in istio. Failure to fail here tells me the bug is either already fixed (and should be fixed in istio once using the current version of client_golang) or something else is happening in the Istio code.


beorn7:
Precisely. I could of course rebase this branch on top of master, but then everything else I had branched off this branch would be even more of a mess.

PR #480, opened 2018-10-14
Release v0.9.0

beorn7:
@grobie @stuartnelson3 let's finally do this! I'll tag once this is merged.
@krasi-georgiev Since the API client is completely independent from the instrumentation library, but both are in one repo, we will share the tagging. I thus also tried to put API client changes into the CHANGELOG.md. Hope that works for you.

This ends a long time of not really tagging versions (as it didn't
make a lot of sense - we could have tagged every commit with a new
minor release essentially). I still worked through the git log and GH
issues to create a nice CHANGELOG.md.

As dependency management tools really love version tags these days, I
plan to tag more often from now on. The ideas behind the pre-1.0
version numbers are explained in the README.md as changed in this
commit.

I also changed the issue template accordingly.


README.md
```diff
@@ -22,38 +22,42 @@ declaring something 1.0.0 doesn't make it 1.0.0. Instead, we are working
...
+  be included in those. We do not plan any breaking changes from one v0.9.x
+  release to any later v0.9.y release, but nothing is guaranteed. Since the
+  master branch will eventually be switched over to track the upcoming v0.10
+  (see below), we recommend to tell your dependency management tool of choice
+  to use the latest v0.9.x release, at least for your production software. In
+  that way, you should get bug fixes and non-invasive, low-risk new features
+  without the need to change anything on your part.
+- v0.10 is a planned release that will have a _lot_ of breaking changes
+  (despite being only a “minor” release in the Semantic Versioning terminology,
+  but as sayd, pre-1.0.0 means nothing is guaranteed). Essentially, we have
```
stuartnelson3: `/s/sayd/said/`

krasi-georgiev:
> @krasi-georgiev Since the API client is completely independent from the instrumentation library, but both are in one repo, we will share the tagging. I thus also tried to put API client changes into the CHANGELOG.md. Hope that works for you.

no objections on my side.

README.md
beorn7: fixed

Issue #366, opened 2018-01-17
Consider using https://github.com/NYTimes/gziphandler for compression


tinytub:
I have use client_golang to implement a subservice like pushgateway.
Every 30s scrape 500k - 700k line metrics from one subservice.
And it seems gzip cause very high cpu usage.
![image](https://user-images.githubusercontent.com/1871460/35047362-a954c596-fbd4-11e7-9047-6851010fd5ae.png)

With this large data, gzip didn't make the request much more quick

curl -w %{time_connect}:%{time_starttransfer}:%{time_total} -I -H 'Accept-Encoding: gzip' http://k4637v:6060/metrics
HTTP/1.1 200 OK
Content-Encoding: gzip
Content-Length: 4935721
Content-Type: text/plain; version=0.0.4
Date: Wed, 17 Jan 2018 14:25:26 GMT

0.002:8.412:8.413                                                                                                               

curl -w %{time_connect}:%{time_starttransfer}:%{time_total} -I  http://k4637v:6060/metrics
HTTP/1.1 200 OK
Content-Length: 70412653
Content-Type: text/plain; version=0.0.4
Date: Wed, 17 Jan 2018 14:25:52 GMT

0.002:10.068:10.068

beorn7:
We are using the normal way of the Go standard library to server gzip'd HTTP content. At least that's our intention.

Do you think we are doing something wrong here? If not, this would be an upstream issue of the Go standard library. I would assume, though, that gzip'ing just takes that much CPU. If you don't like compression, you can disable it in HandlerOpts: https://godoc.org/github.com/prometheus/client_golang/prometheus/promhttp#HandlerOpts

beorn7:
Closing this for lack of response. Please respond with more information if you think this still needs to be addressed on our side.

bobrik:
Similar issue: https://github.com/prometheus/blackbox_exporter/issues/270

beorn7:
I guess this would boil down to trying out https://github.com/NYTimes/gziphandler . Reopening and adjusting the title.

dswarbrick:
@beorn7 The nytimes gziphandler is actually pretty straightforward, and I'm not sure if it warrants adding as a dependency. I did some brief experiments last week with pooled gzip writers, and whilst it was an improvement (about 10% CPU time reduction), it was also a bit underwhelming. Clearly there are other areas for improvement.

beorn7:
@dswarbrick Thanks for sharing your concerns.

[...]

PR #481, opened 2018-10-17
Use sync.Pool for gzipWriter

glefloch:
This PR aims to close issue #366 by adding a sync.Pool for `gzipWriter`. 
As mentionned by @dswarbrick, the `NYTime gzipHandler` did not improve performance compared to a `sync.Pool`. 

### DIFF
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -59,6 +59,12 @@ var gzipPool = sync.Pool{
 	},
 }
 
+var bufPool = sync.Pool{
+	New: func() interface{} {
+		return &bytes.Buffer{}
+	},
+}
+
 // Handler returns an http.Handler for the prometheus.DefaultGatherer, using
 // default HandlerOpts, i.e. it reports the first error as an HTTP error, it has
 // no error logging, and it applies compression if requested by the client.
@@ -103,7 +109,6 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 				return
 			}
 		}
-
 		mfs, err := reg.Gather()
 		if err != nil {
 			if opts.ErrorLog != nil {
@@ -124,8 +129,11 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 		}
 
 		contentType := expfmt.Negotiate(req.Header)
-		buf := &bytes.Buffer{}
+		buf := bufPool.Get().(*bytes.Buffer)
+		buf.Reset()
 		enc := expfmt.NewEncoder(buf, contentType)
+
+		defer bufPool.Put(buf)
 
 		var lastErr error
 		for _, mf := range mfs {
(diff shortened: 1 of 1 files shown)


dswarbrick:
Is there any performance (or GC) difference having the gzipHandler as an anonymous function, vs. declared normally?

prometheus/promhttp/http.go
```diff
@@ -133,10 +124,9 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 		}
 
 		contentType := expfmt.Negotiate(req.Header)
-		buf := getBuf()
-		defer giveBuf(buf)
-		writer, encoding := decorateWriter(req, buf, opts.DisableCompression)
-		enc := expfmt.NewEncoder(writer, contentType)
+		buf := &bytes.Buffer{}
```
beorn7: Why did you remove the sync.Pool for buf? This will create a lot of allocations in high-scrape-rate scenarios.

beorn7: Or to put it another way: For the same reason we want the gzip writer in a sync.Pool, we want this intermediate buffer in a sync.Pool.

prometheus/promhttp/http.go
```diff
@@ -155,29 +145,50 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
...
-			closer.Close()
-		}
+
 		if lastErr != nil && buf.Len() == 0 {
 			http.Error(w, "No metrics encoded, last error:\n\n"+lastErr.Error(), http.StatusInternalServerError)
 			return
 		}
-		header := w.Header()
-		header.Set(contentTypeHeader, string(contentType))
-		header.Set(contentLengthHeader, fmt.Sprint(buf.Len()))
```
beorn7: This commit removes this header (which SHOULD be used as per https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13 ). That's probably OK for gzip'd case because we avoid another buffering layer, but it could still be set in the cases where we don't compress. This is complicated here because of the double-wrapping, see comment below.

prometheus/promhttp/http.go
```diff
@@ -155,29 +145,50 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
...
+
+		w.Header().Set(contentTypeHeader, string(contentType))
+
 		if _, err := w.Write(buf.Bytes()); err != nil && opts.ErrorLog != nil {
 			opts.ErrorLog.Println("error while sending encoded metrics:", err)
 		}
 		// TODO(beorn7): Consider streaming serving of metrics.
 	})
 
+	gzipHandler := http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
```
beorn7: It is helpful to have two `http.HandlerFunc`s here where one just wraps the other? Couldn't this just be inlined?

beorn7:
Thanks for doing this. The idea to gzip the completely rendered response instead of the decorated writer approach so far looks quite neat.

I left a few comments we need to clarify.

prometheus/promhttp/http.go
dswarbrick: Nested handlers is the normal way to implement gzip middleware in Go, especially if you want it to be optional / settable by some flag.

dswarbrick: I'm not a fan of it being an anonymous function either, as I suspect this will cause a lot of allocations / GC. The NYTimes implementation is worth studying. And bear in mind that the content-length header should be the length of the gzipped output, not the original input.

beorn7: WRT content length: Yes, I had that in mind. That's why I suggested it's probably OK to not add a content length when gzipping. But we should (or even SHOULD :-) still add the content length when _not_ gzipping. And that's not possible with the approach here where the decision is made within the gzipHandler local function but the uncompressed content length is only available in the h local function. In this specific case, where we don't want to provide a generically usable gzip handler, I'd probably just inline everything into one local function. At least for now. We can still refactor into something nicer once we have something that works.

glefloch:
@beorn7 Thanks for the comment, I will update the code.

beorn7:
Hmm, thinking about the need (or not-need) of content length, the changes proposed in your PR, and the current performance testing I'm doing for kube-state-metrics, I'm wondering if we should perhaps ditch that intermediate buffer and do streaming exposition, at least for the encoding part.

I'll play with it and create a PR based on yours.

beorn7:
See https://github.com/prometheus/client_golang/pull/482

PR #482, opened 2018-10-19
Use gzip from a pool and stream while encoding metrics

beorn7:
@glefloch Here the promised joint PR.

### DIFF
--- a/prometheus/http.go
+++ b/prometheus/http.go
@@ -15,9 +15,7 @@ package prometheus
 
 import (
 	"bufio"
-	"bytes"
 	"compress/gzip"
-	"fmt"
 	"io"
 	"net"
 	"net/http"
@@ -41,19 +39,10 @@ const (
 	acceptEncodingHeader  = "Accept-Encoding"
 )
 
-var bufPool sync.Pool
-
-func getBuf() *bytes.Buffer {
-	buf := bufPool.Get()
-	if buf == nil {
-		return &bytes.Buffer{}
-	}
-	return buf.(*bytes.Buffer)
-}
-
-func giveBuf(buf *bytes.Buffer) {
-	buf.Reset()
-	bufPool.Put(buf)
+var gzipPool = sync.Pool{
+	New: func() interface{} {
+		return gzip.NewWriter(nil)
+	},
 }
 
 // Handler returns an HTTP handler for the DefaultGatherer. It is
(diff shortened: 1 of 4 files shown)


glefloch:
This looks pretty neat! We're you able to measure memory and CPU gain?

beorn7:
I'm about to try it out with our kube-state-metrics setup at SoundCloud. It spits out ~150MiB of metrics (uncompressed text format) on each scrape… It's for a cluster with about 10k pods. Will report back once I have results.

glefloch:
Well, do you think this is only due to compression or due to the fact there is no more intermediate buffer?

beorn7:
I compared compression to compression and no-compression to no-compression. Each time without this PR vs. with this PR. But I think the weird results were because my pod landed on a slow node. I redeployed the whole shebang, and now the metrics look more consistent. But we should take it as a reminder that on a cluster, things can vary a lot, not only due to different hardware of node but also different load state of nodes etc. So taking all the results with a grain of salt, I'd say that this PR reduces scrape duration and CPU usage by broadly 20% without compression and broadly 10% with compression. Since this is high-volume low-frequency scraping, the sync.Pool will have no real gain but leaving out the extra buffer is a net saving.  It makes sense that the net gain is lower with enabled compression as previously, we wrote the _compressed_ metrics into the extra buffer, which is much smaller than the uncompressed metrics.

To prove the use of the sync.Pool, I'll try a load test ASAP.

glefloch:
@beorn7, this looks good to me, I have nothing to add.

glefloch:
@beorn7, this looks good to me, I have nothing to add.

glefloch:
@beorn7, this looks good to me. I have nothing to add.

beorn7:
Load test of the "random" example with 5000 requests in 20s, median and 99th percentile latency, CPU usage,
* before, no gzip: 12ms, 72ms, 140%
* before, with gzip: 11ms, 76ms, 130%
* after, no gzip: 10ms, 77ms, 130%,
* before, with gzip: 9ms, 79ms, 120%

While a small improvement is visible, this is very close to noise.

Overall, I'd still say the change is worth it. @glefloch let me know if you have anything to add or change here. Otherwise, I'll also propagate this into the legacy HTTP code in the `prometheus` package before merging.

beorn7:
GH seemed to have issues last night. I get a mail from travis that the test were passing after my final commit.

beorn7:
FTR: comment repeats were happening because of a GH outage.

All tests green now. I'll merge.

Thanks everybody for contributing to the discussion and the code here and in #481 .

beorn7:
Added a commit to have the same improvements in the deprecated HTTP stuff in the `prometheus` package.

PR #488, opened 2018-10-31
Add more info to the inconsistent cardinality errors

### LINKED ISSUE #487: Add additional information to the inconsistent cardinality error
At the moment the error message that's returned doesn't tell you much about the source of the error. It would be useful if the error included the name, label values and any other piece of data that could help identify the source.

peterj:
Fixes #487

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -136,7 +136,8 @@ func NewCounterVec(opts CounterOpts, labelNames []string) *CounterVec {
 	return &CounterVec{
 		metricVec: newMetricVec(desc, func(lvs ...string) Metric {
 			if len(lvs) != len(desc.variableLabels) {
-				panic(errInconsistentCardinality)
+				err := fmt.Errorf("%s: labels and values for '%s' don't match: labels %v, values %v", errInconsistentCardinality, desc.fqName, desc.variableLabels, lvs)
+				panic(err)
 			}
 			result := &counter{desc: desc, labelPairs: makeLabelPairs(desc, lvs)}
 			result.init(result) // Init self-collection.
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -15,6 +15,7 @@ package prometheus
 
 import (
 	"errors"
+	"fmt"
 	"math"
 	"sync/atomic"
 
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -137,8 +137,7 @@ func NewCounterVec(opts CounterOpts, labelNames []string) *CounterVec {
 	return &CounterVec{
 		metricVec: newMetricVec(desc, func(lvs ...string) Metric {
 			if len(lvs) != len(desc.variableLabels) {
-				err := fmt.Errorf("%s: labels and values for '%s' don't match: labels %v, values %v", errInconsistentCardinality, desc.fqName, desc.variableLabels, lvs)
-				panic(err)
+				panic(makeInconsistentCardinalityError(desc.fqName, desc.variableLabels, lvs))
 			}
 			result := &counter{desc: desc, labelPairs: makeLabelPairs(desc, lvs)}
 			result.init(result) // Init self-collection.
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -15,7 +15,6 @@ package prometheus
 
 import (
 	"errors"
-	"fmt"
 	"math"
 	"sync/atomic"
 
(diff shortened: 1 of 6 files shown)


peterj:
@beorn7 can you take a look when you get a chance?

prometheus/counter.go
```diff
@@ -136,7 +137,8 @@ func NewCounterVec(opts CounterOpts, labelNames []string) *CounterVec {
 	return &CounterVec{
 		metricVec: newMetricVec(desc, func(lvs ...string) Metric {
 			if len(lvs) != len(desc.variableLabels) {
-				panic(errInconsistentCardinality)
+				err := fmt.Errorf("%s: labels and values for '%s' don't match: labels %v, values %v", errInconsistentCardinality, desc.fqName, desc.variableLabels, lvs)
```
beorn7: This will expand to something like
```
inconsistent label cardinality: labels and values for 'requests_total' don't match: labels [foo bar], values [baz]
```

which is not as precise as it could be and has a bit of stutter with two colons and such.

How about the following:

```Go
panic(fmt.Errorf(
    "%s: %q has %d variable labels named %q but %d values %q were provided",
    errInconsistentCardinality, desc.fqName,
    len(desc.variableLabels), desc.variableLabels,
    len(lvs), lvs,
))
```

That would result in:
```
inconsistent label cardinality: "requests_total" has 2 variable labels named ["foo" "bar"] but only 1 values ["baz"] were provided
```

Besides the usual singular/plural oddness, I think this tells you more precisely what the actual issue is and has nicer quoting.

prometheus/counter.go
```diff
@@ -136,7 +137,8 @@ func NewCounterVec(opts CounterOpts, labelNames []string) *CounterVec {
...
+				panic(err)
```
beorn7: You are not using the `err` variable anywhere else. Why not just `panic` directly without the `err` indirection? (See my previous comment for how that would look like in code.)

prometheus/gauge.go
```diff
@@ -147,7 +148,8 @@ func NewGaugeVec(opts GaugeOpts, labelNames []string) *GaugeVec {
 	return &GaugeVec{
 		metricVec: newMetricVec(desc, func(lvs ...string) Metric {
 			if len(lvs) != len(desc.variableLabels) {
-				panic(errInconsistentCardinality)
+				err := fmt.Errorf("%s: labels and values for '%s' don't match: labels %v, values %v", errInconsistentCardinality, desc.fqName, desc.variableLabels, lvs)
+				panic(err)
```
beorn7: See above for both lines.

prometheus/histogram.go
```diff
@@ -165,7 +165,8 @@ func NewHistogram(opts HistogramOpts) Histogram {
 
 func newHistogram(desc *Desc, opts HistogramOpts, labelValues ...string) Histogram {
 	if len(desc.variableLabels) != len(labelValues) {
-		panic(errInconsistentCardinality)
+		err := fmt.Errorf("%s: labels and values for '%s' don't match: labels %v, values %v", errInconsistentCardinality, desc.fqName, desc.variableLabels, labelValues)
+		panic(err)
```
beorn7: See above for both lines.

prometheus/summary.go
```diff
@@ -181,7 +181,8 @@ func NewSummary(opts SummaryOpts) Summary {
 
 func newSummary(desc *Desc, opts SummaryOpts, labelValues ...string) Summary {
 	if len(desc.variableLabels) != len(labelValues) {
-		panic(errInconsistentCardinality)
+		err := fmt.Errorf("%s: labels and values for '%s' don't match: labels %v, values %v", errInconsistentCardinality, desc.fqName, desc.variableLabels, labelValues)
+		panic(err)
```
beorn7: See above for both lines. (As we have this four times, you might consider a helper function for creating the error, and then just write `panic(makeInconsistentCardinalityError(desc.variableLabels, labelValues))`

prometheus/labels.go
```diff
@@ -39,7 +39,7 @@ var errInconsistentCardinality = errors.New("inconsistent label cardinality")
 
 func validateValuesInLabels(labels Labels, expectedNumberOfValues int) error {
 	if len(labels) != expectedNumberOfValues {
-		return errInconsistentCardinality
+		return fmt.Errorf("%s: number of labels and values don't match: expected %d, actual %d, labels %v", errInconsistentCardinality, expectedNumberOfValues, len(labels), labels)
```
beorn7: Suggestion (mostly just leveraging `%#v`, making it more concise, and formatting):
```Go
return fmt.Errorf(
    "%s: expected %d label values but got %d in %#v",
    errInconsistentCardinality, expectedNumberOfValues,
    len(labels), labels,
)
```

prometheus/counter.go
peterj: Much better -thanks!

prometheus/summary.go
peterj: Created the function in `labels.go`

peterj:
Thanks for the review @beorn7! I updated the code and the failing test.

beorn7:
Thanks again. Merging now…

PR #489, opened 2018-10-31
WriteToTextfile function for node exporter textfiles

sevagh:
@beorn7

This is my attempt to add (similar, and implementation copied from, [client_python](https://github.com/prometheus/client_python/blob/master/prometheus_client/exposition.py#L186)) WriteToTextfile for a Registry.

This is so we can write Go tools that emit Node Exporter textfiles.

My strategy is to create the same output from the promhttp HTML website.

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -533,6 +533,50 @@ func (r *Registry) Gather() ([]*dto.MetricFamily, error) {
 	return internal.NormalizeMetricFamilies(metricFamiliesByName), errs.MaybeUnwrap()
 }
 
+func (r *Registry) WriteToTextfile(path string) error {
+	metricFamilies, err := r.Gather()
+	if err != nil {
+		return err
+	}
+	output := []string{}
+	for _, metricFamily := range metricFamilies {
+		output = append(output, fmt.Sprintf("# HELP %s %s", metricFamily.GetName(), metricFamily.GetHelp()))
+		output = append(output, fmt.Sprintf("# TYPE %s %s", metricFamily.GetName(), metricFamily.GetType().String()))
+		for _, metric := range metricFamily.GetMetric() {
+			labelString := ""
+			if metric.GetLabel() != nil {
+				labelStrings := []string{}
+				for _, labelPair := range metric.GetLabel() {
+					labelStrings = append(labelStrings, fmt.Sprintf("%s=\"%s\"", labelPair.GetName(), labelPair.GetValue()))
+				}
+				labelString = fmt.Sprintf("{%s}", strings.Join(labelStrings, ","))
+			}
+			timestampString := ""
+			if metric.TimestampMs != nil {
+				timestampString = fmt.Sprintf(" %d", int(float64(metric.GetTimestampMs())*1000))
+			}
+			var value float64
+			switch metricFamily.GetType() {
+			case dto.MetricType_COUNTER:
+				value = metric.GetCounter().GetValue()
+			case dto.MetricType_GAUGE:
+				value = metric.GetGauge().GetValue()
+			case dto.MetricType_SUMMARY:
+				//value = metric.GetSummary().GetValue()
+				//what to do here
+			case dto.MetricType_HISTOGRAM:
+				//same
+				//value = metric.GetHistogram().GetValue()
+			case dto.MetricType_UNTYPED:
+				value = metric.GetUntyped().GetValue()
+			}
+			output = append(output, fmt.Sprintf("%s%s %f%s", metricFamily.GetName(), labelString, value, timestampString))
+		}
+	}
+	fmt.Println(strings.Join(output, "\n"))
+	return nil
+}
+
 // processMetric is an internal helper method only used by the Gather method.
 func processMetric(
 	metric Metric,
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -16,6 +16,7 @@ package prometheus
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"runtime"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


sevagh:
Usage looks like:
```
shanssian:disk_exporter $ cat dummy.go
package main

import (
)

var (
        fakeGauge = prometheus.NewGaugeVec(
                prometheus.GaugeOpts{
                        Name: "fake_metric",
                        Help: "this is so fake",
                },
                []string{"name", "foo", "bar"},
        )
)

func main() {
        registry := prometheus.NewRegistry()
        registry.MustRegister(fakeGauge)

        fakeGauge.With(prometheus.Labels{"name": "hello", "foo": "xyz", "bar": "abc"}).Set(65.99)

        registry.WriteToTextfile("./helloworld.txt")
}
shanssian:disk_exporter $
shanssian:disk_exporter $
shanssian:disk_exporter $ go run dummy.go
# HELP fake_metric this is so fake
# TYPE fake_metric GAUGE
fake_metric{bar="abc",foo="xyz",name="hello"} 65.990000
```

sevagh:
I'm also working on the Histogram/Summary code.

sevagh:
Will add tests in my next commit - also the actual tmpfile logic.

sevagh:
I'm also stuck somewhere kinda dumb. The Python client does `tmppath = '%s.%s.%s' % (path, os.getpid(), threading.current_thread().ident)`

`threading.current_thread()` - the equivalent in Go I assume is the current goroutine ID, which is either hacky, discouraged, or impossible to get (based on some reading I'm doing).

sevagh:
In my test, the output I compare to (https://github.com/prometheus/client_golang/compare/master...sevagh:master#diff-7c4dacb8aec4877f9e6a1fd1768ce969R878) is copy-pasted from the `promhttp` output, since I figure the best test is to see that the HTTP metrics page and text file output are the same.

I also had to manually create a "+Inf" bucket (https://github.com/prometheus/client_golang/compare/master...sevagh:master#diff-6223e8ae8180a02a4024a006b8e5f2f8R597) in my output string. Not sure if it's expected for `Histogram.GetBucket() -> []*Bucket` to not return the `+Inf` bucket.

sevagh:
I tried to think of ways of making the code less ugly, but it's hard, given that the functions for `GetHistogram().GetBucket()` and `GetSummary().GetQuantile()`, and `GetGauge/Counter/Untyped().GetValue()` are _similar_ but there are no interfaces or abstractions.

beorn7:
This is indeed a cool feature. So cool that it already exists, kind of…

This is the version of your initial example program, using the existing functionality:

```Go
package main

import (
	"fmt"
	"io/ioutil"
	"os"
	"path/filepath"

)

var (
	fakeGauge = prometheus.NewGaugeVec(
		prometheus.GaugeOpts{
			Name: "fake_metric",
			Help: "this is so fake",
		},
		[]string{"name", "foo", "bar"},
	)
)

func WriteToTextfile(g prometheus.Gatherer, filename string) error {
	tmp, err := ioutil.TempFile(filepath.Dir(filename), filepath.Base(filename))
	defer os.Remove(tmp.Name())

	mfs, err := g.Gather()
	if err != nil {
		return err
	}
	for _, mf := range mfs {
		if _, err := expfmt.MetricFamilyToText(tmp, mf); err != nil {
			return err
		}
	}
	return os.Rename(tmp.Name(), filename)
}

func main() {
	registry := prometheus.NewRegistry()
	registry.MustRegister(fakeGauge)
	fakeGauge.With(prometheus.Labels{"name": "hello", "foo": "xyz", "bar": "abc"}).Set(65.99)

	if err := WriteToTextfile(registry, "test.prom"); err != nil {
		fmt.Println("Textfile creation failed:", err)
	}
}
```

As you can see, it's only nine lines of code. However, this is probably providing enough convenience and encodes the operational concern of first writing into a temp file that it warrants the creation of an own function in the client library. It should, however, not be a method on the `Registry` type but a top level function so that it works on any implementation of `Gatherer`.

In other words: We could just add my `WriteToTextfile` as above and use your test from this PR to test it. What do you think? Would you like to change your PR accordingly?

(I think there is no need to replicate the PID way of naming the temp file as we have `ioutil.TempFile` in the Go standard library.)

beorn7:
Improved version of my suggested function:

```Go
func WriteToTextfile(filename string, g prometheus.Gatherer) error {
	tmp, err := ioutil.TempFile(filepath.Dir(filename), filepath.Base(filename))
	if err != nil {
		return err
	}
	defer os.Remove(tmp.Name())

	mfs, err := g.Gather()
	if err != nil {
		return err
	}
	for _, mf := range mfs {
		if _, err := expfmt.MetricFamilyToText(tmp, mf); err != nil {
			return err
		}
	}
	if err := tmp.Close(); err != nil {
		return err
	}
	return os.Rename(tmp.Name(), filename)
}
```

- Same order of arguments as in the Pythos client library.
- Proper error handling.
- Actually closing the file.

sevagh:
Sounds good to me - I used your code, and adapted the test. I also wrote a little doc string for the new function:
>// WriteToTextfile formats the metrics of the provided Gatherer interface and
// emits them in text format to the path. Intended for use with the node_exporter
// textfile collector

prometheus/registry.go
```diff
@@ -533,6 +537,31 @@ func (r *Registry) Gather() ([]*dto.MetricFamily, error) {
 	return internal.NormalizeMetricFamilies(metricFamiliesByName), errs.MaybeUnwrap()
 }
 
+// WriteToTextfile formats the metrics of the provided Gatherer interface and
+// emits them in text format to the path. Intended for use with the node_exporter
+// textfile collector
```
beorn7: Almost perfect. I'd add the fact that the node exporter wants the filename to look like *.prom.

Suggested modified text (with a few more tweaks):

WriteToTextfile calls Gather on the provided Gatherer, encodes the result in the Prometheus text format, and writes it to a temporary file. Upon success, the temporary file is renamed to the provided filename. This is intended for use with the textfile collector of the node exporter. Note that the node exporter expects the filename to be suffixed with ".prom".

sevagh:
Not sure if this is related to how Python and Go write/create files differently, I'm running into this:
```
shanssian@systest-8:/usr/share/ansible$ stat -c '%a %n' /opt/node-exporter-sb/textfiles/smart.prom
644 /opt/node-exporter-sb/textfiles/smart.prom
shanssian@systest-8:/usr/share/ansible$ stat -c '%a %n' /opt/node-exporter-sb/textfiles/disks.prom
600 /opt/node-exporter-sb/textfiles/disks.prom
```
`smart.prom` is generated using Python `write_to_textfile` (in a cronjob running as root), `disks.prom` is using this pull request.
node_exporter can't collect it:
```
node-exporter-sb/textfiles/disks.prom\": open /opt/node-exporter-sb/textfiles/disks.prom: permission denied","source":"textfile.go:204","time":"2018-11-02T09:42:34-07:00"}
```

sevagh:
I think it's from the Python code manually creating a tempfile, while Go ioutil uses OS tempfile specifics with tighter permissions.

beorn7:
I guess 644 makes sense in this context. Users could still tweak the permissions of the directory if they want to lock things down.

beorn7:
Thanks for the new feature. :o)

PR #492, opened 2018-11-05
The metric consistency check should take in account the timestamp if it exists. If samples are provided out-of-order it will be dropped by prometheus

topine:
When writing an exporter that will collect metrics with some delay that can couse an "overlap" this is useful to avoid gaps in the samples.

storage_avg_read_io_ops_per_second{,name="SVCAAA-11",phase="",type="storageSystem"} 250 1541429110000
storage_avg_read_io_ops_per_second{,name="SVCAAA-11",phase="",type="storageSystem"} 500 1541429411000

@beorn7 Could you please review ?

Thank  you

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"unicode/utf8"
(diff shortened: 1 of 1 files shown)


brian-brazil:
> To my knowledge, it is an invalid exposition if the same metric is contained multiple times with a different timestamp.

It's undefined behaviour what happens with the current format if there's multiple timestamps.

beorn7:
I think a client library should not emit anything that is ingested in an undefined way.

topine:
Hello,

In this case I will stop emitting the metrics with multiple timestamps and assume that in case of "overlap" the sample will be missing.


Thanks a lot for the help.

PR #497, opened 2018-11-13
Fix testutil metric comparison

### LINKED ISSUE #494: Bug: testutil package returning unexpected results
I'm new to working with Prometheus so apologies if this is a misunderstanding on my part. 

I noticed when writing a test for a Counter without any ConstLabels defined in the CounterOpts struct, the test will fail and provide an error message saying two identical strings do not match.

This can be replicated by modifying slightly one of the tests in the `testutil_test.go` file, by just removing the labels, example:

```go
func TestCollectAndCompareBug(t *testing.T) {
	const metadata = `
		# HELP some_total A value that represents a counter.
		# TYPE some_total counter
	`

	c := prometheus.NewCounter(prometheus.CounterOpts{
		Name: "some_total",
		Help: "A value that represents a counter.",
	})
	c.Inc()

	expected := `

		some_total 1
	`

	if err := CollectAndCompare(c, strings.NewReader(metadata+expected), "some_total"); err != nil {
		t.Errorf("unexpected collecting result:\n%s", err)
	}
}
```

This results in the following output:

```bash

--- FAIL: TestCollectAndCompareBug (0.00s)
    testutil_test.go:165: unexpected collecting result:
        
        metric output does not match expectation; want:
        
        # HELP some_total A value that represents a counter.
        # TYPE some_total counter
        some_total 1
        
        
        got:
        
        # HELP some_total A value that represents a counter.
        # TYPE some_total counter
        some_total 1
        
FAIL

Process finished with exit code 1
```

I haven't dug in too deeply but if this is a bug and anyone have some pointers to fix I can take it on as soon as I have some free cycles.

beorn7:
Fixes #494.

@PhilipGough thanks for reporting, please have a look.

@brancz I think the "upstream" code for the testutil package is gone from KSM by now. So no action required there, I guess.

### DIFF
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -37,7 +37,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"reflect"
 
 	"github.com/prometheus/common/expfmt"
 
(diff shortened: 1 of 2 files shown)


PhilipGough:
@beorn7 great thanks. That seems to fix the original bug which I had resolved by just using `ToFloat64` function. However I also just noticed that the `CollectAndCompare` function does not seem to work with some other types either. 

I can confirm at least a `*HistogramVec`, has the same behaviour where matching strings are reported unequal. Perhaps it is to do with the DeepEqual on the pointer?

 I can update the original issue or open a new one. If the latter then this looks to have resolved the original problem 👍

beorn7:
With this PR, there is no DeepEqual anymore.
I'll merge this, and if you still see problems with this merged, please let me know.

PR #501, opened 2018-11-15
*: support Go modules

simonpasquier:


api/prometheus/v1/api.go
```diff
@@ -56,12 +56,12 @@ type HealthStatus string
...
-	ErrExec                  = "execution"
-	ErrBadResponse           = "bad_response"
-	ErrServer                = "server_error"
-	ErrClient                = "client_error"
+	ErrTimeout     ErrorType = "timeout"
+	ErrCanceled    ErrorType = "canceled"
+	ErrExec        ErrorType = "execution"
+	ErrBadResponse ErrorType = "bad_response"
+	ErrServer      ErrorType = "server_error"
+	ErrClient      ErrorType = "client_error"
```
beorn7: Where is this change coming from? Is it needed/intended?

go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/prometheus/client_golang
+
+require (
+	github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973
+	github.com/gogo/protobuf v1.1.1
```
beorn7: Inclusion of this happens via prometheus/wrap_test.go, but it's an accident from my side.
I'll send you a PR to fix, and then this should be updated to not include this line.

api/prometheus/v1/api.go
simonpasquier: It was spotted by the newer version of staticcheck that is compatible with Go modules:

```
api/prometheus/v1/api.go:58:2: only the first constant in this group has an explicit type (SA9004)
```

https://github.com/dominikh/go-tools/blob/master/cmd/staticcheck/docs/checks/SA9004

beorn7: Ah cool, thanks.

simonpasquier:
@beorn7 I've rebased on top of master and `go mod tidy` properly detected that gogo/protobuf isn't needed anymore.

PR #508, opened 2018-11-21
Add support for fetching rules with API client

bobmshannon:
This PR adds support for fetching [rules](https://prometheus.io/docs/prometheus/latest/querying/api/#rules) via the `/api/v1/rules` endpoint using the API client. Currently, the API client exposes no way to do this and it would be nice to have for external systems that wish to retrieve this information.

cc @beorn7

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -40,6 +40,7 @@ const (
 	epLabelValues     = apiPrefix + "/label/:name/values"
 	epSeries          = apiPrefix + "/series"
 	epTargets         = apiPrefix + "/targets"
+	epRules           = apiPrefix + "/rules"
 	epSnapshot        = apiPrefix + "/admin/tsdb/snapshot"
 	epDeleteSeries    = apiPrefix + "/admin/tsdb/delete_series"
 	epCleanTombstones = apiPrefix + "/admin/tsdb/clean_tombstones"
@@ -47,13 +48,27 @@ const (
 	epFlags           = apiPrefix + "/status/flags"
 )
 
+// AlertState models the state of an alert.
+type AlertState string
+
 // ErrorType models the different API error types.
 type ErrorType string
 
 // HealthStatus models the health status of a scrape target.
 type HealthStatus string
 
+// RuleType models the type of a rule.
+type RuleType string
+
+// RuleHealth models the health status of a rule.
+type RuleHealth string
+
 const (
+	// Possible values for AlertState.
+	AlertStateFiring   AlertState = "firing"
+	AlertStateInactive AlertState = "inactive"
+	AlertStatePending  AlertState = "pending"
+
 	// Possible values for ErrorType.
 	ErrBadData     ErrorType = "bad_data"
 	ErrTimeout     ErrorType = "timeout"
@@ -67,6 +82,15 @@ const (
 	HealthGood    HealthStatus = "up"
 	HealthUnknown HealthStatus = "unknown"
 	HealthBad     HealthStatus = "down"
+
+	// Possible values for RuleType.
+	RuleTypeRecording RuleType = "recording"
+	RuleTypeAlerting  RuleType = "alerting"
+
+	// Possible values for RuleHealth.
+	RuleHealthGood    = "ok"
+	RuleHealthUnknown = "unknown"
+	RuleHealthBad     = "err"
 )
 
 // Error is an error returned by the API.
@@ -111,6 +135,8 @@ type API interface {
 	// Snapshot creates a snapshot of all current data into snapshots/<datetime>-<rand>
 	// under the TSDB's data directory and returns the directory as response.
 	Snapshot(ctx context.Context, skipHead bool) (SnapshotResult, error)
+	// Rules returns a list of alerting and recording rules that are currently loaded.
+	Rules(ctx context.Context) (RulesResult, error)
 	// Targets returns an overview of the current state of the Prometheus target discovery.
 	Targets(ctx context.Context) (TargetsResult, error)
 }
@@ -137,6 +163,38 @@ type FlagsResult map[string]string
 // SnapshotResult contains the result from querying the snapshot endpoint.
 type SnapshotResult struct {
 	Name string `json:"name"`
+}
+
+// RulesResult contains the result from querying the rules endpoint.
+type RulesResult struct {
+	Groups []RuleGroup `json:"groups"`
+}
+
+type RuleGroup struct {
+	Name     string  `json:"name"`
+	File     string  `json:"file"`
+	Rules    []Rule  `json:"rules"`
+	Interval float64 `json:"interval"`
+}
+
+type Rule struct {
+	Alerts      []Alert        `json:"alerts",omitempty`
+	Annotations model.LabelSet `json:"annotations",omitempty`
+	Labels      model.LabelSet `json:"labels",omitempty`
+	Duration    float64        `json:"duration",omitempty`
+	Health      RuleHealth     `json:"health"`
+	Name        string         `json:"name"`
+	Query       string         `json:"query"`
+	LastError   string         `json:"lastError,omitempty"`
+	Type        RuleType       `json:"type"`
+}
+
+type Alert struct {
+	ActiveAt    time.Time `json:"activeAt"`
+	Annotations model.LabelSet
+	Labels      model.LabelSet
+	State       AlertState
+	Value       float64
 }
 
 // TargetsResult contains the result from querying the targets endpoint.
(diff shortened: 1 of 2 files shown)


api/prometheus/v1/api.go
```diff
@@ -139,6 +165,38 @@ type SnapshotResult struct {
...
 
+// RulesResult contains the result from querying the rules endpoint.
+type RulesResult struct {
+	Groups []RuleGroup `json:"groups"`
+}
+
+type RuleGroup struct {
+	Name     string  `json:"name"`
+	File     string  `json:"file"`
+	Rules    []Rule  `json:"rules"`
```
bobmshannon: Since a Rule can be either `recording` or `alerting`, I'm not sure if we actually want to use a stronger type here with two fields `RecordingRules` and `AlertingRules`, or put everything in the same list to preserve the rule order like https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L721 does.

beorn7:
That one is for @krasi-georgiev . :smile:

api/prometheus/v1/api.go
krasi-georgiev: I will let @simonpasquier 	and @mxinden comment on that.

api/prometheus/v1/api.go
```diff
@@ -67,6 +82,15 @@ const (
 	HealthGood    HealthStatus = "up"
 	HealthUnknown HealthStatus = "unknown"
 	HealthBad     HealthStatus = "down"
+
+	// Possible values for RuleType.
+	RuleTypeRecording RuleType = "recording"
+	RuleTypeAlerting  RuleType = "alerting"
+
+	// Possible values for RuleHealth.
+	RuleHealthGood    = "ok"
```
mxinden: Why not import them from here? https://github.com/prometheus/prometheus/blob/master/rules/manager.go#L46

api/prometheus/v1/api.go
```diff
@@ -67,6 +82,15 @@ const (
 	HealthGood    HealthStatus = "up"
 	HealthUnknown HealthStatus = "unknown"
 	HealthBad     HealthStatus = "down"
+
+	// Possible values for RuleType.
+	RuleTypeRecording RuleType = "recording"
```
mxinden: Mhh, its a bummer that I didn't expose these as constants when I added this. Sorry about that. Feel free to propose a patch on prometheus/prometheus.

api/prometheus/v1/api.go
```diff
@@ -139,6 +165,38 @@ type SnapshotResult struct {
 	Name string `json:"name"`
 }
 
+// RulesResult contains the result from querying the rules endpoint.
+type RulesResult struct {
+	Groups []RuleGroup `json:"groups"`
+}
+
+type RuleGroup struct {
```
mxinden: Why not use https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L718 ?

api/prometheus/v1/api.go
```diff
@@ -139,6 +165,38 @@ type SnapshotResult struct {
...
+	Name     string  `json:"name"`
+	File     string  `json:"file"`
+	Rules    []Rule  `json:"rules"`
+	Interval float64 `json:"interval"`
+}
+
+type Rule struct {
```
mxinden: The name `Rule` is misleading, as recording rules do not expose e.g. `Alerts`.

mxinden: Again, could we import structs from prometheus/prometheus instead?

api/prometheus/v1/api.go
bobmshannon: The primary reasons for not not importing `RuleGroup` from the `v1` API package referenced above:
-`RuleGroup` models the list of rules as `interface{}` - this requires the client to use type assertion on a private type to determine whether a given rule is a recording or alerting rule.
- It adds an external dependency that the client will need to import in order to use this API client package but I think we can encapsulate the types to mitigate this concern.

api/prometheus/v1/api.go
bobmshannon: I think we should just create two new types here: `RecordingRule` and `AlertingRule`. Is there any reason why order needs to be preserved by putting both types of rules in the same array?

If we create these two new types, I think we'll need some changes to the imported structs, for example `recordingRule` and `alertingRule` are not exported:
https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L730
https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L743

bobmshannon:
I'm happy make the changes to this PR such that the types are imported from `prometheus/prometheus`, however I'll also need to make a patch to export some types so that they can be consumed here. Additionally, I think I'd prefer to keep the types encapsulated as much as possible so that clients don't need to import a whole bunch of external packages just to use the API client. Let me know what you all think.

As for using OpenAPI to generate the client code, agreed it would be a very nice place to get to.

bobmshannon:
@mxinden Do you mind having another look at this one? I addressed some of the comments and added new fields `AlertRules` and `RecordingRules` to the `RuleGroup` struct to provide stronger typing and reduce ambiguity from a client perspective.

I think that importing the types at this point may be more trouble than it's worth, and we should probably just plan to supercede this client with the OpenAPI generated one, however I could be wrong about this.

api/prometheus/v1/api.go
```diff
@@ -169,6 +240,63 @@ type queryResult struct {
...
+
+	rg.Name = v.Name
+	rg.File = v.File
+	rg.Interval = v.Interval
+
+	var alertingRules []AlertingRule
+	var recordingRules []RecordingRule
+
+	for _, rule := range v.Rules {
+		// Because both recording and alerting rules are stored in the same
```
mxinden: They are stored in the same JSON array as the ordering among them matters (See https://github.com/prometheus/prometheus/pull/4318#discussion_r202675025 for more details). That way users can understand which rules depend on which.

I am not sure this is a blocker. But probably worth mentioning in a comment somewhere. What do you think?

mxinden:
On the one hand given the fact that this client is already tightly coupled with the `prometheus/prometheus` structs (e.g. when we change something in `prometheus/prometheus` this will likely break this client), I think this dependency should also be expressed in the code, hence importing it.

On the other hand according to the README of this repository we don't give any stability guarantees in regards to the client. Thereby I am fine with going either way.

api/prometheus/v1/api.go
bobmshannon: Thanks for pointing me to that.

To keep things consistent, I've modified the client to preserve rule order as returned by the API, and added some documentation on this behavior regarding how to use a type switch to access each recording or alerting rule.

api/prometheus/v1/api.go
```diff
@@ -139,6 +166,63 @@ type SnapshotResult struct {
...
+// Rule types can be determined using a type switch:
+//   switch v := rule.(type) {
+//   case RecordingRule:
+//   	fmt.Print("got a recording rule")
+//   case AlertingRule:
+//   	fmt.Print("got a alerting rule")
+//   default:
+//   	fmt.Printf("unknown rule type %s", v)
+//   }
+type Rules []interface{}
```
mxinden: Where is this type used? Am I missing something?

api/prometheus/v1/api.go
```diff
@@ -169,6 +253,56 @@ type queryResult struct {
...
+
+	for _, rule := range v.Rules {
+		// Because both recording and alerting rules are stored in the same
+		// JSON array, each rule is first encoded into JSON and then decoded
+		// into either a RecordingRule or AlertingRule.
+		t, ok := rule["type"]
+		if !ok {
+			return errors.New("rule has no type field")
+		}
+		ruleJSON, err := json.Marshal(rule)
```
mxinden: I am a bit confused here. You are marshaling a rule to unmarshal it?

How about unmarshaling the rule into something like this first:

```golang
struct {
  type string
}
```

and depending on `type` unmarshal as a recording or an alerting rule?

bobmshannon: With that approach I think we'd still need to get the complete byte or JSON encoding of each individual rule (by using MarshalJSON) in order to unmarshal them into a typed recording or alerting rule. That said, I think most of the complexity here is from decoding the mixed type slice and there may not be a super elegant way to do it in Go.

api/prometheus/v1/api.go
bobmshannon: Nice catch, fixed.

api/prometheus/v1/api.go
bobmshannon: I've made an update to this PR that should be easier to read and is more in-line with the approach suggested in this comment.

bobmshannon:
@mxinden I've made some adjustments to this PR based on the most recent comments, please let me know what you think.

bobmshannon:
@beorn7 @mxinden Just wanted to circle back on this one, is there anything I can do to help push this one through? It would be really nice to have this functionality in the API client for my use-case and for any external system that fetches alert and rules data from Prometheus. Thanks in advanced. :)

beorn7:
I'd like to see @mxinden and @krasi-georgiev to have the final look here.

krasi-georgiev:
yeah waiting for Max for the final input.

bobmshannon:
Great, thanks!

krasi-georgiev:
Thanks

PR #510, opened 2018-12-03
Return observed duration from Timer.ObserveDuration

### LINKED ISSUE #509: Timer.ObserveDuration should returned the observed duration
The following usage example is given in the docs:
```go
func TimeMe() {
    timer := NewTimer(myHistogram)
    defer timer.ObserveDuration()
    // Do actual work.
}
```

I find that I typically want to also emit durations in log statements. Therefore it would be useful if `ObserveDuration()` returned the observed duration (probably as a `time.Duration`, but I could see a case being made for the value as a `float64`), which would allow code like the following:

```go
func TimeMe() {
    timer := NewTimer(myHistogram)
    defer func() { log.Printf("finished in %f seconds", timer.ObserveDuration().Seconds()) }()
    // Do actual work.
}
```

beorn7:
Fixes #509.

### DIFF
--- a/prometheus/timer.go
+++ b/prometheus/timer.go
@@ -39,13 +39,16 @@ func NewTimer(o Observer) *Timer {
 
 // ObserveDuration records the duration passed since the Timer was created with
 // NewTimer. It calls the Observe method of the Observer provided during
-// construction with the duration in seconds as an argument. ObserveDuration is
-// usually called with a defer statement.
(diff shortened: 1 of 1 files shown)


beorn7:
@eklitzke

beorn7:
I'll take @metalmatze's :+1: as a review for this (fairly small) change.

PR #513, opened 2018-12-06
Fix the race in the registry

### LINKED ISSUE #512: Race in checkMetricConsistency
`checkMetricConsistency` relies on labels being sorted. If they are not, it is sorting them, which is modifying a slice that is assumed to be immutable elsewhere. Even if they are sorted, `sort.Sort` might temporarily change order (even if no equal elements are in the slice) if there are more than 12 labels (where the implementation switches from ShellSort to HeapSort or QuickSort).

Just using `sort.Stable` is not a solution as the immutable slice will still be mutated if it is not sorted (which can only happen with home-grown metric families, but it can still happen).

See https://github.com/prometheus/client_golang/pull/511 for discussion . Thanks to @douglas-reid for identifying the bug and its cause.

beorn7:
This fixes #512 and subsequently https://github.com/istio/istio/issues/10236 .

@douglas-reid , this incorporates your findings. Please have a look.

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -909,8 +909,8 @@ func checkDescConsistency(
 	}
 
 	// Is the desc consistent with the content of the metric?
-	lpsFromDesc := make([]*dto.LabelPair, 0, len(dtoMetric.Label))
-	lpsFromDesc = append(lpsFromDesc, desc.constLabelPairs...)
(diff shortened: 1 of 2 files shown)


douglas-reid:
@beorn7 thanks for seeing this through. really appreciate it. any chance there will be a new tag soon that we can use to grab this update ?

beorn7:
I'll cut v0.9.2 today.

PR #517, opened 2018-12-18
Adjust to breaking change in prometheus/common

beorn7:
This updates the Go modules and it will then also fix the pre-Go1.11 builds (which always build from head of the dependencies).


simonpasquier:
:+1: nice example of why Go modules can be useful ;-)

brian-brazil:
:+1:

PR #519, opened 2018-12-21
Fix typo

gkze:
situtatios => situations


prometheus/doc.go
```diff
@@ -128,7 +128,7 @@
 // errors. Thus, with unchecked Collectors, the responsibility to not collect
 // metrics that lead to inconsistencies in the total scrape result lies with the
 // implementer of the Collector. While this is not a desirable state, it is
-// sometimes necessary. The typical use case is a situatios where the exact
+// sometimes necessary. The typical use case is a situations where the exact
```
beorn7: I guess it should be “situation”.

gkze:
Good point :)

PR #524, opened 2018-12-30
add vendor directory to source control

tariq1890:
Why this PR?

1. This stays consistent with other prometheus golang repos where the vendor directory has been added to source control
2. Test cases don't work in my IDE. They pass only if I turn on the GO111MODULE env var and run it via the CLI. (It seems to depend on that to get the latest version of `prometheus/common`.) This is less than ideal in terms of DX IMO.


beorn7:
The reason why this repo doesn't contain the vendor directory is that it is a library. The other repos are resulting in stand alone binaries.

With Go modules, vendoring isn't really necessary anymore, but some wanted it nevertheless, so it was added as an option later in the process. So far, we kept vendoring in Prometheus repos that already have them. However, vendoring in library repos is, in my understanding, now even less appropriate as it was before.

About the setting of the env variable: That's all done by the Makefile. Even if you don't use the Makefile and don't set the env variable, things will work if you check out the dependencies at the version defined in the go.mob file. (Right now, test cases work with the most recent version of prometheus/common.)

In summary, I'm more confident than ever that we don't want a vendor folder in this repo. Thus, I'm closing this, but if you have to present more points for it, I'm happy to listen.

PR #525, opened 2019-01-04
fix typo

RomanKrasavtsev:


beorn7:
THanks. The tist failure is an unrelated problem, which I will fix ASAP.

PR #526, opened 2019-01-05
Update Makefile.common to current upstream version

beorn7:
This should fix the staticcheck test failures with Go 1.10


brian-brazil:
:+1:

PR #527, opened 2019-01-16
prometheus: reword comment to avoid cursing

mdlayher:
This wording isn't necessary, IMHO.


beorn7:
Yes, my bad…

mdlayher:
No worries! I was just grepping around and figured it might be worth changing.

Issue #531, opened 2019-01-27
DataRace detected with concurrent use of client


ldelossa:
I'm using v0.9.2 of the client. I just encountered the following data race:

```
==================
WARNING: DATA RACE
Write at 0x00c000350e10 by goroutine 80:
  github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api.(*httpClient).Do.func2()
      /Users/louis/git/go/src/github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api/client.go:116 +0xfa

Previous write at 0x00c000350e10 by goroutine 83:
  github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api.(*httpClient).Do()
      /Users/louis/git/go/src/github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api/client.go:122 +0x3e2
  github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api/prometheus/v1.apiClient.Do()
      /Users/louis/git/go/src/github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api/prometheus/v1/api.go:477 +0x80
  github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api/prometheus/v1.(*apiClient).Do()
      <autogenerated>:1 +0xb2
  github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api/prometheus/v1.(*httpAPI).Query()
      /Users/louis/git/go/src/github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api/prometheus/v1/api.go:335 +0x324
  github.com/Paperspace/metrics/prometheus.(*queryStreamer).query()
      /Users/louis/git/go/src/github.com/Paperspace/metrics/prometheus/querystreamer.go:114 +0x15b

Goroutine 80 (running) created at:
  github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api.(*httpClient).Do()
      /Users/louis/git/go/src/github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api/client.go:115 +0x22a
  github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api/prometheus/v1.apiClient.Do()
      /Users/louis/git/go/src/github.com/Paperspace/metrics/vendor/github.com/prometheus/client_golang/api/prometheus/v1/api.go:477 +0x80
[...]

PR #532, opened 2019-01-27
wait for done before writing to shared resp body

ldelossa:
This PR fixes what seems to be a data race condition. Previously if the caller of this Do() canceled the context, or the provided context has timed out the select statement would enter that branch and then close the response body. This could potentially occur concurrently with launched go routine on line 115. My solution simply moves the receive on the done channel before the close of the resp body. 

May want to consider whether we check context in the go routine launched on line 115 before reading the body as a good measure of "don't do work you don't need too". I'll wait for initial comments on this PR to make sure my logic is correct. 

fixes https://github.com/prometheus/client_golang/issues/531

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -119,8 +119,8 @@ func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response,
 
 	select {
 	case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
Thanks!

PR #533, opened 2019-01-27
Re-adjust to the old float formatting in prometheus/common

beorn7:
go.mod was already pointing to the new behavior, but it was never included in a release of client_golang.

Now we are using the brand new tag v0.2.0 in prometheus/common (which had been tagged in the meantime, so go mod would naturally pick the tagged version, including the new float formatting).


brian-brazil:
LGTM

Issue #194, opened 2016-03-09
documentation for API client


FPurchess:
It would be very useful to have a short example on how to perform query operations using the client (e.g. godoc examples) or, if a documentation of any form exists, reference it in the readme.

beorn7:
http://prometheus.io/docs/querying/api/ is the documentation, and it is linked from the README.md.

beorn7:
Please let us know in which way it isn't obvious enough.

FPurchess:
@beorn7 Thanks for the fast reply!

The link explains the API usage in general quite well, but does not explain how make queries using the client (https://godoc.org/github.com/prometheus/client_golang/api/prometheus). Do I miss something here?

beorn7:
Now I got it, sorry for not understanding earlier.

Yeah, this code is pretty much undocumented and arguably WIP.

I'll leave it to @fabxc to clarify.

FPurchess:
:+1: 

I assume it would look like that
``` go
import (
	"fmt"
	"golang.org/x/net/context"
	"github.com/prometheus/client_golang/api/prometheus"
)

ctx, cancel := context.WithCancel(context.Background())
defer cancel()

client, err = prometheus.New(prometheus.Config{Address:"<prometheus-url>"})
if err != nil {
	fmt.Errorf("%v", err)
}

q = prometheus.NewQueryAPI(client)

value, err := q.Query(ctx, "sum(metric)", time.Now())
if err != nil {
	fmt.Errorf("%v", err)
}

// FIXME how to get the real information out of "value" (model.Value)
```

RomainBelorgey:
Did someone know how to get the information from value in the example of @FPurchess ?
It gives me only a string and the metric's type. Is it possible to get the metric ?

carlpett:
@RomainBelorgey You need to inspect the result of `value.Type()`. I do something like this:

```go
            switch {
            case val.Type() == model.ValScalar:
                scalarVal := val.(*model.Scalar)
                // handle scalar stuff
            case val.Type() == model.ValVector:
                vectorVal := val.(model.Vector)
                for _, elem := range vectorVal {
                    // do something with each element in the vector
            // etc
```

`model` is this import: `"github.com/prometheus/common/model"`

yinchuan:
Thanks @carlpett

beorn7:
I'll re-assign this to @krasi-georgiev as @fabxc is unlikely to work in this.

[...]

PR #534, opened 2019-01-31
ADD example to query Prometheus HTTP API using client_golang

prakashmishra1598:
Added an example go file to demonstrate querying the Prometheus HTTP API using 'client_golang' (#194)

### DIFF
--- a/examples/prometheus_api_example/main.go
+++ b/examples/prometheus_api_example/main.go
@@ -36,6 +36,9 @@ func main()  {
 
 	config := prometheus.Config{Address: hostname, RoundTripper:prometheus.DefaultRoundTripper}
 	client, err := prometheus.NewClient(config)
+	if err != nil {
+		fmt.Println(err)
+	}
 	httpApi := v1.NewAPI(client)
 
 	//Query time series data(memory usage in bytes for containers) for 5m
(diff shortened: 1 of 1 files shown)


beorn7:
Thanks for your contribution.

Since this cannot run on its own (it needs a running Prometheus server and some code changes to make it work), perhaps it would be better to create an example as part of the package doc comment so that it shows up on https://godoc.org/github.com/prometheus/client_golang/api or https://godoc.org/github.com/prometheus/client_golang/api/prometheus/v1 ? See how it is done for https://godoc.org/github.com/prometheus/client_golang/prometheus . 

(As a counter example for stand-alone examples, the examples for the instrumentation client in the `/examples/` dir can just be `go run` and will then server metrics to look at or even point a Prometheus server at it.)

@krasi-georgiev What do you think?

prakashmishra1598:
@beorn7 @krasi-georgiev Yes, having a godoc example makes more sense. I'll add that instead ASAP :)

beorn7:
Thanks.

itaysk:
Just wanted to say this was helpful to me. Thanks @prakashmishra1598 for making the example, and I wish this was more discoverable then in a closed PR. (didn't find in godoc)

beorn7:
That's because it hasn't been added to godoc yet. The godoc beneath the `api` dir could use some love. PRs welcome.

AntonioBusillo:
+1

PR #536, opened 2019-02-07
Save memory on histograms: slightly faster with less code

pascaldekloe:
* one allocation for duo instantiations
* use index overflow instead of redundant last state field

name                         old time/op    new time/op    delta
HistogramWithLabelValues-12     204ns ± 0%     204ns ± 0%   ~     (all equal)
HistogramNoLabels-12           41.1ns ± 0%    40.5ns ± 0%   ~     (p=1.000 n=1+1)
HistogramObserve1-12           33.1ns ± 0%    33.9ns ± 0%   ~     (p=1.000 n=1+1)
HistogramObserve2-12            132ns ± 0%     122ns ± 0%   ~     (p=1.000 n=1+1)
HistogramObserve4-12            377ns ± 0%     364ns ± 0%   ~     (p=1.000 n=1+1)
HistogramObserve8-12           1.00µs ± 0%    0.96µs ± 0%   ~     (p=1.000 n=1+1)
HistogramWrite1-12             1.80µs ± 0%    1.76µs ± 0%   ~     (p=1.000 n=1+1)
HistogramWrite2-12             3.28µs ± 0%    3.25µs ± 0%   ~     (p=1.000 n=1+1)
HistogramWrite4-12             6.66µs ± 0%    6.66µs ± 0%   ~     (p=1.000 n=1+1)
HistogramWrite8-12             13.7µs ± 0%    13.6µs ± 0%   ~     (p=1.000 n=1+1)

name                         old alloc/op   new alloc/op   delta
HistogramWithLabelValues-12     48.0B ± 0%     48.0B ± 0%   ~     (all equal)
HistogramNoLabels-12            0.00B          0.00B        ~     (all equal)

name                         old allocs/op  new allocs/op  delta
HistogramWithLabelValues-12      1.00 ± 0%      1.00 ± 0%   ~     (all equal)
HistogramNoLabels-12             0.00           0.00        ~     (all equal)

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -183,11 +183,13 @@ func newHistogram(desc *Desc, opts HistogramOpts, labelValues ...string) Histogr
 		opts.Buckets = DefBuckets
 	}
 
+	// one memory alloctation
+	var bothCounts [2]histogramCounts
 	h := &histogram{
 		desc:        desc,
 		upperBounds: opts.Buckets,
 		labelPairs:  makeLabelPairs(desc, labelValues),
-		counts:      [2]*histogramCounts{&histogramCounts{}, &histogramCounts{}},
+		counts:      [2]*histogramCounts{&bothCounts[0], &bothCounts[1]},
 	}
 	for i, upperBound := range h.upperBounds {
 		if i < len(h.upperBounds)-1 {
@@ -205,9 +207,10 @@ func newHistogram(desc *Desc, opts HistogramOpts, labelValues ...string) Histogr
 		}
 	}
 	// Finally we know the final length of h.upperBounds and can make buckets
-	// for both counts:
-	h.counts[0].buckets = make([]uint64, len(h.upperBounds))
-	h.counts[1].buckets = make([]uint64, len(h.upperBounds))
+	// for both counts (with one memory allocation):
+	bothBounds := make([]uint64, 2*len(h.upperBounds))
+	h.counts[0].buckets = bothBounds[:len(bothBounds)/2]
+	h.counts[1].buckets = bothBounds[len(bothBounds)/2:]
 
 	h.init(h) // Init self-collection.
 	return h
@@ -227,7 +230,7 @@ type histogram struct {
 	// countAndHotIdx is a complicated one. For lock-free yet atomic
 	// observations, we need to save the total count of observations again,
 	// combined with the index of the currently-hot counts struct, so that
-	// we can perform the operation on both values atomically. The least
+	// we can perform the operation on both values atomically. The most
 	// significant bit defines the hot counts struct. The remaining 63 bits
 	// represent the total count of observations. This happens under the
 	// assumption that the 63bit count will never overflow. Rationale: An
@@ -250,7 +253,6 @@ type histogram struct {
 	// pointers to guarantee 64bit alignment of the histogramCounts, see
 	// http://golang.org/pkg/sync/atomic/#pkg-note-BUG.
 	counts [2]*histogramCounts
-	hotIdx int // Index of currently-hot counts. Only used within Write.
 
 	labelPairs []*dto.LabelPair
 }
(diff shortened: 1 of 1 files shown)


beorn7:
Note: The test failure is unrelated to this PR. It's caused by a Travis change, which I'll take care of separately.

prometheus/histogram.go
```diff
@@ -183,11 +183,13 @@ func newHistogram(desc *Desc, opts HistogramOpts, labelValues ...string) Histogr
 		opts.Buckets = DefBuckets
 	}
 
+	// one memory alloctation
+	var bothCounts [2]histogramCounts
```
beorn7: Not sure this is worth it. It is not in the hot path and thus most likely will never have a measurable impact. On the other hand, it requires an additional line of code, a line of explanatory comment, and makes the `counts` initialization less readable.

prometheus/histogram.go
```diff
@@ -205,9 +207,10 @@ func newHistogram(desc *Desc, opts HistogramOpts, labelValues ...string) Histogr
 		}
 	}
 	// Finally we know the final length of h.upperBounds and can make buckets
-	// for both counts:
-	h.counts[0].buckets = make([]uint64, len(h.upperBounds))
-	h.counts[1].buckets = make([]uint64, len(h.upperBounds))
+	// for both counts (with one memory allocation):
+	bothBounds := make([]uint64, 2*len(h.upperBounds))
+	h.counts[0].buckets = bothBounds[:len(bothBounds)/2]
+	h.counts[1].buckets = bothBounds[len(bothBounds)/2:]
```
beorn7: In the same spirit as above, this is not the hot path and this kind of micro optimization makes the code much less readable and any change thereof error-prone, while not providing a measurable gain in return.

prometheus/histogram.go
```diff
@@ -309,41 +311,30 @@ func (h *histogram) Write(out *dto.Metric) error {
...
 	// cold. To have a normal uint64 for the count, we bitshift by 1 and
-	// save the result in count. We also set h.hotIdx to 1 for the next
-	// Write call, and we will refer to counts #1 as hotCounts and to counts
-	// #0 as coldCounts.
+	// save the result in count.
 	//
 	// If the currently-hot counts struct is #1, we do the corresponding
 	// things the other way round. We have to _decrement_ h.countAndHotIdx
 	// (which is a bit arcane in itself, as we have to express -1 with an
 	// unsigned int...).
```
beorn7: This whole comment block can be simplified even more. It still contains parts that only are true for the old way and not the nice new way you found. As things are now less arcane, perhaps it's also better to have a line-by-line comment for explanation.

prometheus/histogram.go
```diff
@@ -309,41 +311,30 @@ func (h *histogram) Write(out *dto.Metric) error {
...
 
 	// Now we have to wait for the now-declared-cold counts to actually cool
 	// down, i.e. wait for all observations still using it to finish. That's
 	// the case once the count in the cold counts struct is the same as the
 	// one atomically retrieved from the upper 63bits of h.countAndHotIdx.
-	for {
-		if count == atomic.LoadUint64(&coldCounts.count) {
-			break
-		}
+	for count != atomic.LoadUint64(&coldCounts.count) {
```
beorn7: Oops, thanks for finding a way more obvious way for this `for` loop.

prometheus/histogram.go
pascaldekloe: Reasonable argument. I can revert.

prometheus/histogram.go
pascaldekloe: What do you think of my description at <https://github.com/pascaldekloe/metrics/blob/master/metrics.go#L144-L156>? Copy/paste over here?

beorn7: Looks good. I'd probably add more explanations to the lines below, but I can do that later, too.

beorn7:
If you rebase this on master, it should make the test green.

pascaldekloe:
I've borrowed some of your thinking so this is fair-trade. 😁

PR #538, opened 2019-02-11
Pull in new Makefile.common from prometheus/prometheus

beorn7:
This should fix the currently broken Travis builds for Go versions
before 1.11.


beorn7:
It seemed to have worked.

@simonpasquier FYI. I'll merge right ahead, as this is just pulling in the file from prometheus/prometheus.

simonpasquier:
Thanks!

PR #540, opened 2019-02-24
feat(push): add format builder option

FrankSpitulski:
Allow users to specify which format the pusher should push in.
Also added tests related to changing the format of the pusher.
This does not break existing functionality since the argument
is varadic.

@beorn7 

Signed-off-by: Frank Spitulski <fspituls@ucsd.edu>

### DIFF
--- a/prometheus/push/push.go
+++ b/prometheus/push/push.go
@@ -186,8 +186,10 @@ func (p *Pusher) BasicAuth(username, password string) *Pusher {
 }
 
 // Format configures the Pusher to use an encoding format given by the
-// provided expfmt.Format. For convenience, this method returns a
-// pointer to the Pusher itself.
+// provided expfmt.Format. The default format is expfmt.FmtProtoDelim and
+// should be used with the standard Prometheus Pushgateway. Custom
+// implementations may require different formats. For convenience, this
+// method returns a pointer to the Pusher itself.
 func (p *Pusher) Format(format expfmt.Format) *Pusher {
 	p.expfmt = format
 	return p
@@ -233,7 +235,7 @@ func (p *Pusher) push(method string) error {
 	if p.useBasicAuth {
 		req.SetBasicAuth(p.username, p.password)
 	}
-	req.Header.Set(contentTypeHeader, string(expfmt.FmtProtoDelim))
+	req.Header.Set(contentTypeHeader, string(p.expfmt))
 	resp, err := p.client.Do(req)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


beorn7:
What's your use case?

My first spontaneous reaction was: The PGW wants the protobuf format. Why add the complication?

My second reaction: You probably have another push target. It would be nice to know which. And then I would suggest to explore HTTP content negotiation.

My thirst reaction: This should be dealt with in the same way as the other options, i.e. with another of the "builder pattern" methods, and not with an additional parameter to `New`.

FrankSpitulski:
My use case is a custom push endpoint that only accepts text encoding. I saw content negotiation as a handler in the promhttp.HandleFor method, but it isn't a pusher and I can't find any examples of agent driven/reactive content negotiation in go so it doesn't appear to be the right path. Can you point me at the existing builder methods for other options? The pusher looks like it's pretty hard coded to just do protobuf.

FrankSpitulski:
I think this implementation fits better @beorn7.

beorn7:
Indeed, agent-driven content negotiation in HTTP seems to be quite hard and needed quite a bit of work on both sides. Thus, it's probably a good thing to let the user of the `push` package choose the format.

I'll review the code ASAP.

prometheus/push/push.go
```diff
@@ -197,7 +208,7 @@ func (p *Pusher) push(method string) error {
 		return err
 	}
 	buf := &bytes.Buffer{}
-	enc := expfmt.NewEncoder(buf, expfmt.FmtProtoDelim)
+	enc := expfmt.NewEncoder(buf, p.expfmt)
```
beorn7: `p.expfmt` has to be used, too, down in line 236, instead of the hardcoded `expfmt.FmtProtoDelim`.

prometheus/push/push.go
```diff
@@ -182,6 +185,14 @@ func (p *Pusher) BasicAuth(username, password string) *Pusher {
 	return p
 }
 
+// Format configures the Pusher to use an encoding format given by the
+// provided expfmt.Format. For convenience, this method returns a
+// pointer to the Pusher itself.
```
beorn7: Perhaps add what the default is (`expfmt.FmtProtoDelim`) and that that's the one most natural for the Prometheus Pushgateway, and that the Pushgateway will understand all the formats, but custom implementations of a push receiver might require a specific one.

prometheus/push/push.go
FrankSpitulski: good catch, updated

prometheus/push/push.go
FrankSpitulski: updated

beorn7:
Looks great. Thank you very much.

PR #541, opened 2019-03-04
Add Go 1.12 to Travis CI

simonpasquier:
Also update to the latest version of Makefile.common.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: go
 
 go:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,4 +1,3 @@
-sudo: false
```
simonpasquier: The `sudo:` key has been deprecated, see https://docs.travis-ci.com/user/reference/trusty/#container-based-infrastructure

beorn7:
Thanks.

PR #544, opened 2019-03-06
prometheus/promhttp: actually observe values in trace in example

s-urbaniak:
This is similar to
https://github.com/prometheus/client_golang/pull/535 fixing ExampleInstrumentRoundTripperDuration.

cc @brancz @mxinden @beorn7

### DIFF
--- a/prometheus/promhttp/instrument_client_1_8_test.go
+++ b/prometheus/promhttp/instrument_client_1_8_test.go
@@ -162,16 +162,16 @@ func ExampleInstrumentRoundTripperDuration() {
 	// functions that we want to instrument.
 	trace := &InstrumentTrace{
 		DNSStart: func(t float64) {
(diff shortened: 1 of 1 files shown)


brancz:
lgtm :+1: as a side note .. we should probably test in this test that after doing the request we actually get _some_ metrics populated, no?

s-urbaniak:
@brancz that's actually a good idea, let me see how I can add such a test.

beorn7:
It would be great to verify, but also note that this is just an example and mostly meant to document usage than actually testing anything. If the example gets confusing by adding the verification, it would defeat the main purpose of the example.

In different news: This code queries `http://google.com`. Since the example code is actually run, this actually performs the request, and if it fails, it will fail the test. That's another wart here.

In any case, merging this now to have a correct example in the documentation. Other concerns about this can be addressed separately.

PR #553, opened 2019-04-09
*: replace golang.org/x/net/context by context

simonpasquier:
The context package is available since Go 1.7 which is the minimal version supported by client_golang.

Found this after reading https://medium.com/@valyala/stripping-dependency-bloat-in-victoriametrics-docker-image-983fb5912b0d (mentioned by @brian-brazil on IRC). Though the complain isn't really convincing to me as dependency managers should be smart enough to vendor only the required packages and the library rely mostly on stdlib packages.


beorn7:
Thanks for spotting and fixing.

PR #555, opened 2019-04-14
Update dependencies

beorn7:
Mostly to use the new module support for prometheus/common and
beorn7/perks.

With :heart: for @SuperQ, who particularly likes those PRs. :smile:


beorn7:
Also down to five lines in `go.mod`. Really bloated, this library… :o)

PR #557, opened 2019-04-23
Implement POST with get fallback for Query/QueryRange

### LINKED ISSUE #428: API client doesn't support POST method for Query/QueryRange
POST support was added to query/query_range API endpoints in prometheus with https://github.com/prometheus/prometheus/pull/3322. This client currently *only* sends GET requests, which means that it isn't won't work for large query string values. As the API stands today:

``` go
Query(ctx context.Context, query string, ts time.Time) (model.Value, error)
QueryRange(ctx context.Context, query string, r Range) (model.Value, error)
```

There isn't really a place to put it, so I see a couple options:

1. add a `NewPOSTAPI` method to return an `httpAPI` instance with a flag to use POST for both query and query_range (this would be done by setting an option in the `httpAPI` struct-- to be added).
2. add QueryPost and QueryRangePost to the API interface
3. add `METHOD` argument to Query and QueryRange

I'm not sure what the compatibility guarantees are on this client interface, so I'm not sure how we want to proceed. I'm leaning towards 1 or 2 (probably in that order)

cc @juliusv

jacksontj:
Fixes #428


jacksontj:
cc @krasi-georgiev

jacksontj:
cc @krasi-georgiev

krasi-georgiev:
Thanks

There are few other places where get is used, should these be updated as well?

do we have a test that ensures this fallback?
I mean ensure that an endpoint that supports only GET uses the GET method and an endpoint that support POST uses the post method.

jacksontj:
> There are few other places where get is used, should these be updated as well?

This does a POST then a GET, so we only want to add this to endpoints we *expect* POST to work, do we have a list of the ones we want to support fallback on? The change is relatively small, so its easy to add in. As requested in the issue this does simply do a POST and fallback to a GET, there is no control on the client side which method you will use-- this also means if your server only supports GET you'll do a POST then a GET.

> do we have a test that ensures this fallback?

From reading the tests, it doesn't seem that *anything* that actually does this integration test. There are some hard-coded tests that check, but that assumes that the test defines it properly. So I have added a test that the fallback happens as we expect-- which should be sufficient.

api/client.go
```diff
@@ -60,6 +60,28 @@ type Client interface {
 	Do(context.Context, *http.Request) (*http.Response, []byte, error)
 }
 
+// DoGetFallback will attempt to do the request as-is, and on a 405 it will fallback to a GET request
```
krasi-georgiev: missing full stops here and few other places
```suggestion
// DoGetFallback will attempt to do the request as-is, and on a 405 it will fallback to a GET request.
```

api/client_test.go
```diff
@@ -113,3 +116,70 @@ func TestClientURL(t *testing.T) {
...
+
+	type testResponse struct {
+		Values string
+		Method string
+	}
+
+	acceptPost := true
+	// Start a local HTTP server
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		req.ParseForm()
```
krasi-georgiev: error not handled here and few other places.

Maybe use  `https://github.com/prometheus/tsdb/testutil` or `https://github.com/prometheus/prometheus/testutil` to facilitate error handling in tests

`testutil.OK(t,req.ParseForm())`

api/client_test.go
```diff
@@ -113,3 +116,70 @@ func TestClientURL(t *testing.T) {
...
+		r := &testResponse{
+			Values: req.Form.Encode(),
+			Method: req.Method,
+		}
+
+		body, _ := json.Marshal(r)
+
+		if req.Method == http.MethodPost {
+			if !acceptPost {
```
krasi-georgiev: hm , that global var is a bit confusing, how about using the url for this.Something like:

`if req.URL.Path=="../allowPost"...}`

api/client_test.go
jacksontj: This particular one is okay, since if this fails the result won't have the form values-- so there's no need to specifically check this error.

jacksontj:
@krasi-georgiev nits addressed :)

api/client_test.go
```diff
@@ -113,3 +118,70 @@ func TestClientURL(t *testing.T) {
...
+		if req.Method == http.MethodPost {
+			if req.URL.Path == "/blockPost" {
+				http.Error(w, string(body), http.StatusMethodNotAllowed)
+				return
+			}
+		}
+
+		w.Write(body)
+	}))
+	// Close the server when test finishes
```
krasi-georgiev: still missing full stop here and few other places :)

api/client_test.go
```diff
@@ -113,3 +118,70 @@ func TestClientURL(t *testing.T) {
...
+	defer server.Close()
+
+	u, err := url.Parse(server.URL)
+	testutil.Ok(t, err)
+	client := &httpClient{client: *(server.Client())}
+
+	// Do a post, and ensure that the post succeeds
+	_, b, err := DoGetFallback(client, context.TODO(), u, v)
+	if err != nil {
+		t.Fatalf("Error doing local request: %v", err)
```
krasi-georgiev: why not use `testutil` everywhere?

krasi-georgiev:
bummer, failing tests for the different golang versions

api/client_test.go
```diff
@@ -113,3 +118,70 @@ func TestClientURL(t *testing.T) {
...
+	}
+	resp := &testResponse{}
+	if err := json.Unmarshal(b, resp); err != nil {
+		t.Fatal(err)
```
krasi-georgiev: why not use testutil everywhere?

krasi-georgiev:
should also run `GO111MODULE=on go mod tidy` to update the `go.mod` file.

api/client_test.go
jacksontj: I've swapped the plain `t.Fatal(err)` to testutil, the rest provide more helpful error messages about the failure

api/client_test.go
```diff
@@ -113,3 +118,70 @@ func TestClientURL(t *testing.T) {
...
+	// Do a post, and ensure that the post succeeds.
+	_, b, err := DoGetFallback(client, context.TODO(), u, v)
+	if err != nil {
+		t.Fatalf("Error doing local request: %v", err)
+	}
+	resp := &testResponse{}
+	if err := json.Unmarshal(b, resp); err != nil {
+		testutil.Ok(t, err)
+	}
+	if resp.Method != http.MethodPost {
```
krasi-georgiev: you can use `testutil.Assert` with a message.

api/client_test.go
krasi-georgiev: you can use testutil.Assert with a message.

I think only .Ok doesn't support a message, but it shows the failing line so you can always go there and check.

krasi-georgiev:
still failing test because of the need to support older golang versions.

`api/client_test.go:154: server.Client undefined (type *httptest.Server has no field or method Client)`

beorn7:
httptest.Client has been added in 1.9. If it avoids undue trouble, we could raise the minimum requirement to Go1.9, as we are at Go1.12 already (supporting three versions ought to be enough for everyone). The 1.7 support we still have is essentially because we can.

jacksontj:
Personally I'd rather bump the required version. But I'll leave that decision up to you maintainers :)

krasi-georgiev:
yep I think what @beorn7 suggest is ok so can do it as part of this PR or separate one with some details why it was needed.

beorn7:
I'll prepare a PR. (We can weed out a bunch of code while we are on it.)

beorn7:
See #561 .

jacksontj:
@beorn7 @krasi-georgiev rebased with go version change, CI is passing now.

krasi-georgiev:
Thanks!

PR #561, opened 2019-04-28
Increase minimum required Go version to 1.9

beorn7:
This allows us to simplify a bunch of code while still supporting the
last four Go minor versions.

We have also run into minor annoyances a couple of times by now to
keep supporting 1.7 and 1.8.

It's time to pull the plug!

Signed-off-by: Bjoern Rabenstein <bjoern@rabenste.in>


krasi-georgiev:
LGTM

Issue #560, opened 2019-04-26
Add "storage.Warnings" to api client


jacksontj:
Prometheus has added warnings to the HTTP API -- as it stands today this client doesn't parse/return those warnings. This should simply be adding the warnings return to:

- Query()
- QueryRange()
- Series()

Assuming this is non-controversial I can definitely submit a PR with this feature.

krasi-georgiev:
I am not familiar what are these warning for and how should these be handled. Can you add a bit more context with links to the docs or code.

jacksontj:
These warning were added in https://github.com/prometheus/prometheus/pull/4832 so now the API response includes a warning section which is crucial to have, since it now is a way to annotate a failure to read a remote read endpoint. Personally it seems like a messy solution to the problem, since you now have to check error and warning and if you care about data completeness either is a failure... But unfortunately it was added upstream so the client (imo) should support it.

krasi-georgiev:
Yep it makes sense to add here so I am happy to review a PR. 

> Personally it seems like a messy solution to the problem,

I am just curious, what would you consider a better solution?

jacksontj:
> I am just curious, what would you consider a better solution?

These remote-reads are also errors, so IMO they should be treated as such. If prom wanted to make *some* errors be ignored that would have been a better option IMO. As it stands today if you have a warning in the response its unclear if the data you have is authoritative or not (if you asked for data only in the cache range vs outside of the cache range). If the data is still authoritative a warning isn't useful, if it isn't then its an error (as the result is *wrong* since it is missing data). IMO no need for warnings.

A more understandable use of warnings would be for prom to add warnings about specific matchers taking too long or matching too many series, but this add for remote_read seems wrong.

brian-brazil:
Only local storage is critical, we don't want to take out monitoring because a clustered storage system is down.

jacksontj:
> Only local storage is critical, we don't want to take out monitoring because a clustered storage system is down.

I don't disagree, what I'm saying is that the storage layers there should be more configurable. If the user wants remote_read errors to be fatal they should be-- and those errors would be errors (in the regular error return). If they don't want them to be fatal, they wouldn't be errors (meaning nothing returned to the client). It seems odd to add an entire new return type `warnings` which is actually just `[]error` just for this very specific use-case.

But, as I said before: 

> it was added upstream so the client (imo) should support it.

PR #562, opened 2019-04-29
Add storage.Warnings to client

jacksontj:
closes https://github.com/prometheus/client_golang/issues/560

This adds `storage.Warnings` to the error return of Query and QueryRange. Unfortunately the client.Do() method interface makes this a bit difficult to implement. 

So either we go down this path (where the specific error is a known type people should check for) *or* we need to change the Do() methods' interface to return a separate "warning" error.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -25,6 +25,42 @@ import (
 	"time"
 )
 
+func NewErrorAPI(err error, warnings []string) Error {
+	if err == nil && warnings == nil {
+		return nil
+	}
+	return &ErrorAPI{err, warnings}
+}
+
+type ErrorAPI struct {
+	err      error
+	warnings []string
+}
+
+func (w *ErrorAPI) Err() error {
+	return w.err
+}
+
+func (w *ErrorAPI) Error() string {
+	if w.err != nil {
+		return w.err.Error()
+	}
+	return "Warnings: " + strings.Join(w.warnings, " , ")
+}
+
+func (w *ErrorAPI) Warnings() []string {
+	return w.warnings
+}
+
+// Error encapsulates an error + warning
+type Error interface {
+	error
+	// Err returns the underlying error.
+	Err() error
+	// Warnings returns a list of warnings.
+	Warnings() []string
+}
+
 // DefaultRoundTripper is used if no RoundTripper is set in Config.
 var DefaultRoundTripper http.RoundTripper = &http.Transport{
 	Proxy: http.ProxyFromEnvironment,
@@ -55,14 +91,14 @@ func (cfg *Config) roundTripper() http.RoundTripper {
 // Client is the interface for an API client.
 type Client interface {
 	URL(ep string, args map[string]string) *url.URL
-	Do(context.Context, *http.Request) (*http.Response, []byte, error)
+	Do(context.Context, *http.Request) (*http.Response, []byte, Error)
 }
 
 // DoGetFallback will attempt to do the request as-is, and on a 405 it will fallback to a GET request.
-func DoGetFallback(c Client, ctx context.Context, u *url.URL, args url.Values) (*http.Response, []byte, error) {
+func DoGetFallback(c Client, ctx context.Context, u *url.URL, args url.Values) (*http.Response, []byte, Error) {
 	req, err := http.NewRequest(http.MethodPost, u.String(), strings.NewReader(args.Encode()))
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, NewErrorAPI(err, nil)
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
@@ -71,11 +107,14 @@ func DoGetFallback(c Client, ctx context.Context, u *url.URL, args url.Values) (
 		u.RawQuery = args.Encode()
 		req, err = http.NewRequest(http.MethodGet, u.String(), nil)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, NewErrorAPI(err, nil)
 		}
 
 	} else {
-		return resp, body, err
+		if err != nil {
+			return resp, body, NewErrorAPI(err, nil)
+		}
+		return resp, body, nil
 	}
 	return c.Do(ctx, req)
 }
@@ -115,7 +154,7 @@ func (c *httpClient) URL(ep string, args map[string]string) *url.URL {
 	return &u
 }
 
-func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte, error) {
+func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte, Error) {
 	if ctx != nil {
 		req = req.WithContext(ctx)
 	}
@@ -127,7 +166,7 @@ func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response,
 	}()
 
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, NewErrorAPI(err, nil)
 	}
 
 	var body []byte
@@ -147,5 +186,5 @@ func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response,
 	case <-done:
 	}
 
-	return resp, body, err
+	return resp, body, NewErrorAPI(err, nil)
 }
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/api"
@@ -95,13 +96,26 @@ const (
 
 // Error is an error returned by the API.
 type Error struct {
-	Type   ErrorType
-	Msg    string
-	Detail string
+	Type     ErrorType
+	Msg      string
+	Detail   string
+	warnings []string
 }
 
 func (e *Error) Error() string {
-	return fmt.Sprintf("%s: %s", e.Type, e.Msg)
+	if e.Type != "" || e.Msg != "" {
+		return fmt.Sprintf("%s: %s", e.Type, e.Msg)
+	}
+
+	return "Warnings: " + strings.Join(e.warnings, " , ")
+}
+
+func (w *Error) Err() error {
+	return w
+}
+
+func (w *Error) Warnings() []string {
+	return w.warnings
 }
 
 // Range represents a sliced time range.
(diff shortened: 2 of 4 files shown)


jacksontj:
Cc @krasi-georgiev

api/prometheus/v1/api.go
```diff
@@ -22,10 +22,12 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/prometheus/client_golang/api"
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/storage"
```
krasi-georgiev: Doesn't this add too much dependency with the whole Prometheus library?
I think the idea is to keep the go client as light as possible.
Can't you just duplicate the struct type?

jacksontj: I can duplicate the struct type, but this means even more code duplication. IMO copy/pasting code into this library doesn't really keep it "light", I'd even make the argument that it makes it heavier (since you now have to maintain all of that copy/pasted code). Regardless of my opinion on the matter, if you prefer I can copy/paste that here as well.

jacksontj:
@krasi-georgiev the main question is what the interface should be. Right now I'm wrapping the `Warnings` into a specific error type, which is different than the prom internals, but changes less of this client. Alternatively I can make the return of `Do()` add an additional return type of `Warnings` and have it be nil for everyone else -- either way works for me.

api/prometheus/v1/api.go
brian-brazil: Yes, we shouldn't depend on Prometheus itself. That'd be a circular dependency at least.

jacksontj: @krasi-georgiev do we want to copy/paste the type as is? or do we want this to be `[]string` (since we won't have specific error types to switch on)?

jacksontj:
@krasi-georgiev bump?

krasi-georgiev:
will review in 1-2 days.

krasi-georgiev:
I think it should be something like:
```
type Error struct {
	Type   ErrorType
	Msg    string
	Detail string
	Warnings []string
}
```

jacksontj:
@krasi-georgiev to clarify, you want to replace the `error` in the return with `*Error`? Or are you suggesting that `Warnings` be this new Error struct?.

From the upstream behavior warnings != error, so it seems incorrect for the client to be merging them. And if we decided that we wanted to then we'd definitely need a mechanism for the caller to determine what is an error vs a warning. In this proposed error struct thats a bit unclear. I'm also unclear what `detail` in the error would be, as the response struct (as taken from prometheus' api server) is: 

```
type response struct {
	Status    status      `json:"status"`
	Data      interface{} `json:"data,omitempty"`
	ErrorType errorType   `json:"errorType,omitempty"`
	Error     string      `json:"error,omitempty"`
	Warnings  []string    `json:"warnings,omitempty"`
}
```

krasi-georgiev:
maybe you are right lets try to keep these separately to be similiar to upstream.

in that case why not keep it seperately starting from the client?

```
func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte,Warnings,error) {
```

api/prometheus/v1/api.go
```diff
@@ -670,11 +680,20 @@ type apiClient struct {
 	api.Client
 }
 
+// TODO: create a wrapper for multiple errors, this way we can type switch on these
+// and do what we need to
```
krasi-georgiev: why not do it as part of this PR?

also the comment formating should be something like:
```
// Warnings is returned when .......
// TODO: create a wrapper for multiple errors, this way we can type switch on these
// and do what we need to.
```

jacksontj:
@krasi-georgiev It isn't split right now because I hadn't heard back on the plan -- https://github.com/prometheus/client_golang/pull/562#issuecomment-488042563

So it sounds like we want to keep warning/error separate all the way down? If so I can definitely do that.

api/prometheus/v1/api.go
jacksontj: WIP; so far I've been given some contradicting or unclear direction. This will get cleaned up before we merge.

jacksontj: This is an interface question. If we want to always return `returnValue, warnings, error` there is no need for this. If we want to instead keep the `returnValue, error` then we need to make that error switch-able.

So, do we want to *add* warnings to the return signature? Or try to type-switch it out?

jacksontj:
https://github.com/prometheus/client_golang/pull/562#issuecomment-489999962 

All the way down to the HTTP client doesn't quite work, at this client level all we do is return a resp, body, and error -- at this level we haven't unmarshaled the response enough to know if there are warnings.

api/prometheus/v1/api.go
```diff
@@ -734,5 +753,11 @@ func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, [
 		}
 	}
 
+	if err == nil {
+		if len(result.Warnings) != 0 {
+			err = Warnings(result.Warnings)
+		}
+	}
+
```
krasi-georgiev:
```
if err == nil {
		if len(result.Warnings) != 0 {
			err = Warnings(result.Warnings)
		}
	}
```
my idea was instead of this to just return 

`func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte,Warnings, error){


	return resp, []byte(result.Data),Warnings, err
}

jacksontj:
@krasi-georgiev here's it with warnings split all the way down

jacksontj:
For the error response, if we want to have a single return type, IMO we should go back to what I had before where we have a specific `Warnings` struct that implements error. This way we the interface remains the same and the client can decide if they want to treat warnings differently or not. IMO this is how upstream prom should have dealt with it -- but do realize that will make the client interface different from the internal interface.

krasi-georgiev:
> For the error response, if we want to have a single return type, IMO we should go back to what I had before where we have a specific Warnings struct that implements error. 

do you mean with typecasting? Do you think we can avoid typecasting by using an interface?

> but do realize that will make the client interface different from the internal interface.

I don't have a problem with that.

jacksontj:
Yes I mean with typecasting. Otherwise if they want an error or warning
separately the client would have to convert out custom interface or stuck
over. So what I had before was a `Warnings` type that implemented the error
interface, so the caller could check if it's a warning

krasi-georgiev:
I was thinking something like this:


```
type ErrorAPI struct {
	error    string
	warning  string
}

func (w *ErrorAPI) Error() string {
	return w.error
}

func (w *ErrorAPI) Warning() string {
	return w.warnings
}

type Error interface {
	error
	Warning() string
}
```

and than
```
func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte, Error) {
```

so that the caller can use `err.Error()` and `err.Warning()` as needed.

jacksontj:
@krasi-georgiev Updated with a single interface for the error, some small cleanup required; but since we are going back and forth a bunch I figured I'll let you do an initial review before I spend the time cleaning it up all the way.

krasi-georgiev:
yes, I think this is the best approach so far so go ahead with the cleanup and ping me when ready for a review.

jacksontj:
@krasi-georgiev Cleanup done, hopefully thats enough to get this in :)

api/client.go
```diff
@@ -55,14 +85,14 @@ func (cfg *Config) roundTripper() http.RoundTripper {
...
+	Do(context.Context, *http.Request) (*http.Response, []byte, Error)
 }
 
 // DoGetFallback will attempt to do the request as-is, and on a 405 it will fallback to a GET request.
-func DoGetFallback(c Client, ctx context.Context, u *url.URL, args url.Values) (*http.Response, []byte, error) {
+func DoGetFallback(c Client, ctx context.Context, u *url.URL, args url.Values) (*http.Response, []byte, Error) {
 	req, err := http.NewRequest(http.MethodPost, u.String(), strings.NewReader(args.Encode()))
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, &ErrorAPI{err: err}
```
krasi-georgiev: since we have a constructor best to use one or the other everywhere - either this or `NewErrorAPI()`

api/client.go
```diff
@@ -25,6 +25,36 @@ import (
...
+	return w.err.Error()
+}
+
+func (w *ErrorAPI) Warnings() []string {
+	return w.warnings
+}
+
+type Error interface {
+	error
+	Err() error
```
krasi-georgiev: Why the need for Err() and Error()?

api/client.go
```diff
@@ -25,6 +25,36 @@ import (
...
+	Warnings() []string
```
krasi-georgiev: All exported funcs should have a comment.

api/prometheus/v1/api_test.go
```diff
@@ -829,6 +833,21 @@ func TestAPIClientDo(t *testing.T) {
...
+		{
+			code: http.StatusOK,
+			response: &apiResponse{
+				Status:    "error",
+				Data:      json.RawMessage(`"test"`),
+				ErrorType: ErrTimeout,
+				Error:     "timed out",
+				Warnings:  []string{"a"},
+			},
+			expectedWarnings: []string{"a"},
```
krasi-georgiev: why do we need this separately.
can't we just have:
expectedErr: api.NewErrorAPI(nil, []string{"a"}),

and than the check can be just:
```
if test.expectedErr.Warning() != nil {
```

in the main block here :https://github.com/prometheus/client_golang/pull/562/files#diff-c03449ea3ae93306323741c88841a939R867

api/client.go
jacksontj: `Err()` returns the error itself. This is important if you want to check if the error is something in particular. This is especially important as we have been returning the err directly until now. So old code that did something like:

```
ret, err := client.DoSomething()
if err != nil {
    return nil, err
}
``

would now be able to do something like
```
ret, err := client.DoSomething()
if err != nil && err.Err() != nil {
    return nil, err.Err()
}
``

jacksontj:
@krasi-georgiev PR updated to address comments

api/client.go
krasi-georgiev: hm I still don't see the need for this.
Can you give an example for how and where it is needed and why it can't be achieved by just using err.Error() ?

jacksontj: err.Error() returns a string, if you need to check that is a specific error type or something then it is not sufficient.

For example, when we do a client get we could get back an error from the HTTP library. In the client it looks like so:
```
	_, body, apiErr := h.client.Do(ctx, req)
	if apiErr != nil {
		return AlertsResult{}, apiErr
	}
```

So if I wanted to handle the specific error type of `http.ProtocolError` I'd need to do something like:

```
if err != nil {
    if typedErr, ok := err.Err().(*http.ProtocolError); ok {
        // do something else
    }
    return nil, err
}
```

If we *only* have the string representation the people are forced to resort to string matching, which is a bad plan :)

krasi-georgiev: Yes I agree. Lets keep just 

```
// Error .....
type Error interface {
	// Error ......
	Error() error
	// Warnings returns a list of warnings
	Warnings() []string
}
```

There are still few places without comments the go-lint complains about it.

jacksontj: So you want to only have those 2 methods? This would mean the return type doesn't adhere to the error interface anymore meaning this is a completely breaking change.

 If so I'd call the errror one err (error is for string, err is for an actual error type).

jacksontj: As I'm thinking about this more, I think we want to stick with:

```
type Error interface{
    error
    Err() error
    Warnings() []string
}
```

and change the implementations of `Error()` (the string return) to check the warnings. This way its more-or-less backwards compatible -- although it would mean that warnings are errors (if you wanted to skip them you'd have to check the Err() method). This would make it more compatible, and a little less clunky.

If we want the default behavior to be that people have to check if it was an error or warning then I think going to some interface that doesn't adhere to the `error` interface would be okay -- it will just be completely incompatible with all client usage.

api/client.go
```diff
@@ -25,6 +25,38 @@ import (
...
+
+func (w *ErrorAPI) Error() string {
+	return w.err.Error()
+}
+
+func (w *ErrorAPI) Warnings() []string {
+	return w.warnings
+}
+
+type Error interface {
```
krasi-georgiev: missing comment.

api/client.go
```diff
@@ -25,6 +25,38 @@ import (
...
+	error
+	// Err returns the underlying error
```
krasi-georgiev: missing full stop here and below.

api/client.go
krasi-georgiev: Yes you are right lets keep as is for now and will refactor if needed.

api/prometheus/v1/api_test.go
```diff
@@ -849,15 +867,14 @@ func TestAPIClientDo(t *testing.T) {
 				if err == nil {
 					t.Fatalf("expected error %q but got none", test.expectedErr)
```
krasi-georgiev: this can be simplified to:
`testutil.NotOk(t,err)`


also further down can use the testutil as well for:

```
if err != nil {
		t.Fatalf("unexpeceted error %s", err)
}
want, got := test.expectedBody, string(body)
if want != got {
	t.Errorf("unexpected body: want %q, got %q", want, got)
}
```

api/prometheus/v1/api_test.go
```diff
@@ -849,15 +867,14 @@ func TestAPIClientDo(t *testing.T) {
...
 				}
-				if test.expectedErr.Error() != err.Error() {
-					t.Errorf("unexpected error: want %q, got %q", test.expectedErr, err)
-				}
+				testutil.Equals(t, test.expectedErr.Error(), err.Error(), "unexpected error: want %q, got %q", test.expectedErr, err)
```
krasi-georgiev:
```suggestion
				testutil.Equals(t, test.expectedErr.Error(), err.Error())
```
no need of a message it is more less the default message from `testutil.Equals` anyway

api/prometheus/v1/api_test.go
```diff
@@ -849,15 +867,14 @@ func TestAPIClientDo(t *testing.T) {
...
+
 				if test.expectedErr.Detail != "" {
 					apiErr := err.(*Error)
-					if apiErr.Detail != test.expectedErr.Detail {
-						t.Errorf("unexpected error details: want %q, got %q", test.expectedErr.Detail, apiErr.Detail)
-					}
+					testutil.Equals(t, apiErr.Detail, test.expectedErr.Detail, "unexpected error details: want %q, got %q", test.expectedErr.Detail, apiErr.Detail)
```
krasi-georgiev:
```suggestion
					testutil.Equals(t, apiErr.Detail, test.expectedErr.Detail)
```

no need of a message it is more less the default message from `testutil.Equals` anyway

api/prometheus/v1/api_test.go
```diff
@@ -849,15 +867,14 @@ func TestAPIClientDo(t *testing.T) {
...
 				}
+
+				testutil.Equals(t, test.expectedErr.Warnings(), err.Warnings(), "expected warnings %v, actual=%v", test.expectedErr.Warnings(), err.Warnings())
```
krasi-georgiev:
```suggestion
				testutil.Equals(t, test.expectedErr.Warnings(), err.Warnings())
```

no need of a message it is more less the default message from `testutil.Equals` anyway

api/client.go
jacksontj: @krasi-georgiev to clarify, do you want me to change the `Error()` implementation to return something if there was a warning (e.g. if no error but warnings, return the warnings concated somehow)?

jacksontj: @krasi-georgiev bump :)

krasi-georgiev: Sorry, I am at kubecon so will be slow on replies.
I meant to keep it as
```
type Error interface{
    error
    Err() error
    Warnings() []string
}
```

krasi-georgiev:
LGTM
@beorn7 any comment from your side?

jacksontj:
@krasi-georgiev squashed, should be ready for merge.

beorn7:
Didn't check all the details, but no general remarks from my side. I trust you to do the right thing in detail.

PR #563, opened 2019-04-30
Remove encode of params in query before DoGetFallback

jacksontj:

cc @krasi-georgiev this was missed, so right now it sends it as query params and post body for the post request.


krasi-georgiev:
Thanks

PR #567, opened 2019-05-03
Make TestCounterAddLarge more robust

### LINKED ISSUE #530: TestCounterAddLarge fails on some architectures
```
s390x:
=== RUN   TestCounterAddLarge
--- FAIL: TestCounterAddLarge (0.00s)
	counter_test.go:178: valBits expected 18446744073709551616.000000, got 0.000000.
	counter_test.go:181: valInts expected 0, got 18446744073709551615.

ppc64el:
=== RUN   TestCounterAddLarge
--- FAIL: TestCounterAddLarge (0.00s)
	counter_test.go:178: valBits expected 18446744073709551616.000000, got 0.000000.
	counter_test.go:181: valInts expected 0, got 18446744073709551615.

arm64:
=== RUN   TestCounterAddLarge
--- FAIL: TestCounterAddLarge (0.00s)
	counter_test.go:178: valBits expected 18446744073709551616.000000, got 0.000000.
	counter_test.go:181: valInts expected 0, got 18446744073709551615.
```

Otherwise passes on `amd64 i386 armhf`. Not sure if this is indicative of a 64-bit issue on the niche arches.

Version 0.9.0-1

beorn7:
The previous `float64(math.MaxUint64 + 1)` is too close to
`float64(math.MaxUint64)` to actually overflow as indended.

The counter code is actually converting forward and backward and
compare the original and twice-converted value. On most platform, this
will create a deviation and thus trigger the expected behavior. By
sheer "luck", one might end up with the same value and thus still use
the uint64 representation. Which is OK within the precision we can
expect. But it breaks the test. With this change, the next
representable floating point value greater than the floating point
value used to represent math.MaxUint64 is used.

@xnox this should fix #530. Could you verify?
@TheTincho you might have seen this issue, too, for the Debian builds on the affected platforms.

### DIFF
--- a/prometheus/counter_test.go
+++ b/prometheus/counter_test.go
@@ -172,7 +172,7 @@ func TestCounterAddLarge(t *testing.T) {
 	}).(*counter)
 
 	// large overflows the underlying type and should therefore be stored in valBits.
-	large := float64(math.MaxUint64 + 1)
+	large := math.Nextafter(float64(math.MaxUint64), 1e20)
 	counter.Add(large)
 	if expected, got := large, math.Float64frombits(counter.valBits); expected != got {
 		t.Errorf("valBits expected %f, got %f.", expected, got)


brian-brazil:
:+1:

beorn7:
Thanks @brian-brazil . This change makes things saner in any case. @xnox please follow up on this (or #530) if your tests are still failing.

xnox:
With this patch applied, build passed on arm64 ppc64el s390x. Cherrypicked into Ubuntu and uploaded.

PR #570, opened 2019-05-08
Switch from encoding/json -> jsoniter

### LINKED ISSUE #565: JSON Unmarshaling of large responses is excessively expensive
This is basically the same issue as https://github.com/prometheus/prometheus/issues/3601 and https://github.com/prometheus/prometheus/pull/3536

The current api client here uses the base `encoding/json` which is both slow and memory inefficient. For large responses it can generate garbage >10x the size of the response data. So, ideally we'd use a better (read: faster) json library. Upstream prom uses json-iter -- so presumably we'd do the same here?

Thoughts @krasi-georgiev

### LINKED ISSUE #656: Please add link to godocs in the github repo description
Can you add a link to https://godoc.org/github.com/prometheus/client_golang in the repo description?

For example, see how https://github.com/pallets/flask does it

Just makes it super quick to find the godocs from the repo

jacksontj:

Fixes #565


krasi-georgiev:
do you think you can add some benchmarking so we can see what is the actual performance gains?

jacksontj:
sure, and for easy reading here are the result:

```
goos: linux
goarch: amd64
pkg: github.com/prometheus/client_golang/api/prometheus/v1
BenchmarkSerialization/10/10/unmarshal/encoding/json-4         	  200000	     10272 ns/op	     232 B/op	       5 allocs/op
BenchmarkSerialization/10/10/unmarshal/jsoniter-4              	10000000	       154 ns/op	      16 B/op	       1 allocs/op
BenchmarkSerialization/10/100/unmarshal/encoding/json-4        	   20000	     87834 ns/op	     232 B/op	       5 allocs/op
BenchmarkSerialization/10/100/unmarshal/jsoniter-4             	10000000	       178 ns/op	      16 B/op	       1 allocs/op
BenchmarkSerialization/10/1000/unmarshal/encoding/json-4       	    2000	    966855 ns/op	     232 B/op	       5 allocs/op
BenchmarkSerialization/10/1000/unmarshal/jsoniter-4            	 5000000	       211 ns/op	      16 B/op	       1 allocs/op
BenchmarkSerialization/100/10/unmarshal/encoding/json-4        	   10000	    124487 ns/op	     232 B/op	       5 allocs/op
BenchmarkSerialization/100/10/unmarshal/jsoniter-4             	10000000	       217 ns/op	      16 B/op	       1 allocs/op
BenchmarkSerialization/100/100/unmarshal/encoding/json-4       	    1000	   1335330 ns/op	     232 B/op	       5 allocs/op
BenchmarkSerialization/100/100/unmarshal/jsoniter-4            	10000000	       184 ns/op	      16 B/op	       1 allocs/op
BenchmarkSerialization/100/1000/unmarshal/encoding/json-4      	     100	  11345940 ns/op	     232 B/op	       5 allocs/op
BenchmarkSerialization/100/1000/unmarshal/jsoniter-4           	 5000000	       219 ns/op	      16 B/op	       1 allocs/op
BenchmarkSerialization/1000/10/unmarshal/encoding/json-4       	    1000	   1093244 ns/op	     232 B/op	       5 allocs/op
BenchmarkSerialization/1000/10/unmarshal/jsoniter-4            	10000000	       170 ns/op	      16 B/op	       1 allocs/op
BenchmarkSerialization/1000/100/unmarshal/encoding/json-4      	     200	   9597895 ns/op	     232 B/op	       5 allocs/op
BenchmarkSerialization/1000/100/unmarshal/jsoniter-4           	10000000	       180 ns/op	      16 B/op	       1 allocs/op
BenchmarkSerialization/1000/1000/unmarshal/encoding/json-4     	      10	 104403258 ns/op	     232 B/op	       5 allocs/op
BenchmarkSerialization/1000/1000/unmarshal/jsoniter-4          	10000000	       176 ns/op	      16 B/op	       1 allocs/op
PASS
ok  	github.com/prometheus/client_golang/api/prometheus/v1	35.976s
```

jacksontj:
Fixed bug in benchmarks, for some reason the difference isn't as large as I expected:

```
$ go test -run=x -bench=. -benchmem
goos: linux
goarch: amd64
pkg: github.com/prometheus/client_golang/api/prometheus/v1
BenchmarkSerialization/10/10/unmarshal/encoding/json-8         	    5000	    258376 ns/op	   25970 B/op	     564 allocs/op
BenchmarkSerialization/10/10/unmarshal/jsoniter-8              	   10000	    231181 ns/op	   32649 B/op	     850 allocs/op
BenchmarkSerialization/10/100/unmarshal/encoding/json-8        	     500	   2320385 ns/op	  256520 B/op	    5964 allocs/op
BenchmarkSerialization/10/100/unmarshal/jsoniter-8             	    1000	   2187491 ns/op	  327974 B/op	    8950 allocs/op
BenchmarkSerialization/10/1000/unmarshal/encoding/json-8       	      50	  23265065 ns/op	 2571482 B/op	   59970 allocs/op
BenchmarkSerialization/10/1000/unmarshal/jsoniter-8            	      50	  22021770 ns/op	 3287473 B/op	   89955 allocs/op
BenchmarkSerialization/100/10/unmarshal/encoding/json-8        	     500	   2568609 ns/op	  258018 B/op	    5606 allocs/op
BenchmarkSerialization/100/10/unmarshal/jsoniter-8             	    1000	   2239641 ns/op	  326580 B/op	    8501 allocs/op
BenchmarkSerialization/100/100/unmarshal/encoding/json-8       	      50	  23747338 ns/op	 2576962 B/op	   59650 allocs/op
BenchmarkSerialization/100/100/unmarshal/jsoniter-8            	      50	  22705622 ns/op	 3288373 B/op	   89526 allocs/op
BenchmarkSerialization/100/1000/unmarshal/encoding/json-8      	       5	 233949527 ns/op	26711548 B/op	  600269 allocs/op
BenchmarkSerialization/100/1000/unmarshal/jsoniter-8           	       5	 228011137 ns/op	33462769 B/op	  899805 allocs/op
BenchmarkSerialization/1000/10/unmarshal/encoding/json-8       	      50	  26338162 ns/op	 2596940 B/op	   56224 allocs/op
BenchmarkSerialization/1000/10/unmarshal/jsoniter-8            	      50	  23486559 ns/op	 3282834 B/op	   85182 allocs/op
BenchmarkSerialization/1000/100/unmarshal/encoding/json-8      	       5	 241223130 ns/op	27157049 B/op	  678011 allocs/op
BenchmarkSerialization/1000/100/unmarshal/jsoniter-8           	       5	 227933347 ns/op	33707764 B/op	  974806 allocs/op
BenchmarkSerialization/1000/1000/unmarshal/encoding/json-8     	       1	2324619810 ns/op	311494832 B/op	 6319035 allocs/op
BenchmarkSerialization/1000/1000/unmarshal/jsoniter-8          	       1	2155111994 ns/op	361141288 B/op	 9300019 allocs/op
PASS
ok  	github.com/prometheus/client_golang/api/prometheus/v1	34.515s
```


I'll need to spend some more time on this later :/

krasi-georgiev:
any progress with this?

jacksontj:
Some, at this point I can confirm that the benchmarks are accurate (meaning the perf improvement isn't what I expected). So this will require more investigation, but its possible changes will be required elsewhere as well (e.g. common package). I'll try to get some more time to investigate this week.

jacksontj:
I've also added some marshal benchmarks (for anyone else interested):

```
$ go test -run=x -bench=. -benchmem
goos: linux
goarch: amd64
pkg: github.com/prometheus/client_golang/api/prometheus/v1
BenchmarkSerialization/10/10/marshal/encoding/json-8         	    3000	    398908 ns/op	   55704 B/op	    1972 allocs/op
BenchmarkSerialization/10/10/marshal/jsoniter-8              	    3000	    432072 ns/op	   56143 B/op	    1932 allocs/op
BenchmarkSerialization/10/10/unmarshal/encoding/json-8       	    5000	    274671 ns/op	   26698 B/op	     764 allocs/op
BenchmarkSerialization/10/10/unmarshal/jsoniter-8            	    5000	    237565 ns/op	   33371 B/op	    1050 allocs/op
BenchmarkSerialization/10/100/marshal/encoding/json-8        	     300	   3930002 ns/op	  549498 B/op	   19975 allocs/op
BenchmarkSerialization/10/100/marshal/jsoniter-8             	     300	   4278254 ns/op	  554065 B/op	   19934 allocs/op
BenchmarkSerialization/10/100/unmarshal/encoding/json-8      	    1000	   2325803 ns/op	  256444 B/op	    5964 allocs/op
BenchmarkSerialization/10/100/unmarshal/jsoniter-8           	     500	   2332108 ns/op	  328020 B/op	    8950 allocs/op
BenchmarkSerialization/10/1000/marshal/encoding/json-8       	      30	  41352309 ns/op	 6008460 B/op	  200016 allocs/op
BenchmarkSerialization/10/1000/marshal/jsoniter-8            	      30	  46626714 ns/op	 5561751 B/op	  199967 allocs/op
BenchmarkSerialization/10/1000/unmarshal/encoding/json-8     	      50	  24766228 ns/op	 2571484 B/op	   59970 allocs/op
BenchmarkSerialization/10/1000/unmarshal/jsoniter-8          	      50	  23100204 ns/op	 3287492 B/op	   89955 allocs/op
BenchmarkSerialization/100/10/marshal/encoding/json-8        	     300	   4334550 ns/op	  570430 B/op	   19706 allocs/op
BenchmarkSerialization/100/10/marshal/jsoniter-8             	     300	   4557554 ns/op	  561875 B/op	   19305 allocs/op
BenchmarkSerialization/100/10/unmarshal/encoding/json-8      	     500	   2861914 ns/op	  265226 B/op	    7606 allocs/op
BenchmarkSerialization/100/10/unmarshal/jsoniter-8           	     500	   2473652 ns/op	  333875 B/op	   10502 allocs/op
BenchmarkSerialization/100/100/marshal/encoding/json-8       	      30	  43987262 ns/op	 6039315 B/op	  199747 allocs/op
BenchmarkSerialization/100/100/marshal/jsoniter-8            	      30	  46285685 ns/op	 5558747 B/op	  199339 allocs/op
BenchmarkSerialization/100/100/unmarshal/encoding/json-8     	      50	  26428125 ns/op	 2576977 B/op	   59650 allocs/op
BenchmarkSerialization/100/100/unmarshal/jsoniter-8          	     100	  23110238 ns/op	 3283907 B/op	   89514 allocs/op
BenchmarkSerialization/100/1000/marshal/encoding/json-8      	       3	 409665279 ns/op	59081770 B/op	 1999900 allocs/op
BenchmarkSerialization/100/1000/marshal/jsoniter-8           	       3	 430603410 ns/op	58996882 B/op	 1999499 allocs/op
BenchmarkSerialization/100/1000/unmarshal/encoding/json-8    	       5	 233062181 ns/op	26713731 B/op	  658270 allocs/op
BenchmarkSerialization/100/1000/unmarshal/jsoniter-8         	       5	 229206061 ns/op	33464878 B/op	  957805 allocs/op
BenchmarkSerialization/1000/10/marshal/encoding/json-8       	      30	  44881487 ns/op	 6123904 B/op	  197052 allocs/op
BenchmarkSerialization/1000/10/marshal/jsoniter-8            	      30	  45518786 ns/op	 5661321 B/op	  193044 allocs/op
BenchmarkSerialization/1000/10/unmarshal/encoding/json-8     	      50	  26742119 ns/op	 2596950 B/op	   56224 allocs/op
BenchmarkSerialization/1000/10/unmarshal/jsoniter-8          	      50	  23756123 ns/op	 3282841 B/op	   85182 allocs/op
BenchmarkSerialization/1000/100/marshal/encoding/json-8      	       3	 388438656 ns/op	57343797 B/op	 1997168 allocs/op
BenchmarkSerialization/1000/100/marshal/jsoniter-8           	       3	 434228624 ns/op	59020432 B/op	 1993193 allocs/op
BenchmarkSerialization/1000/100/unmarshal/encoding/json-8    	       5	 234770845 ns/op	27129120 B/op	  600611 allocs/op
BenchmarkSerialization/1000/100/unmarshal/jsoniter-8         	       5	 220862249 ns/op	33679800 B/op	  897405 allocs/op
BenchmarkSerialization/1000/1000/marshal/encoding/json-8     	       1	3830350089 ns/op	623496024 B/op	19997237 allocs/op
BenchmarkSerialization/1000/1000/marshal/jsoniter-8          	       1	4229907983 ns/op	689501776 B/op	19993300 allocs/op
BenchmarkSerialization/1000/1000/unmarshal/encoding/json-8   	       1	2277582499 ns/op	311490592 B/op	 6199036 allocs/op
BenchmarkSerialization/1000/1000/unmarshal/jsoniter-8        	       1	2082645407 ns/op	361136920 B/op	 9180019 allocs/op
PASS
ok  	github.com/prometheus/client_golang/api/prometheus/v1	69.265s
```

Which seems to indicate that the improvements claims from https://github.com/prometheus/prometheus/pull/3536 aren't accurate.

api/prometheus/v1/api_bench_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+		m = append(m, ss)
+	}
+	return m
+}
+
+func BenchmarkSerialization(b *testing.B) {
+	for _, timeseriesCount := range []int{10, 100, 1000} {
+		for _, datapointCount := range []int{10, 100, 1000} {
+			b.Run(fmt.Sprintf("%d/%d", timeseriesCount, datapointCount), func(b *testing.B) {
```
brian-brazil: Nested Run calls will do this for you.

brian-brazil:
SamplePair is likely the issue here, I did a custom marshalling in Prometheus for Point. As a general note we're trying to get away from common/model.

api/prometheus/v1/api_bench_test.go
jacksontj: I'm not sure what you mean by this, are you suggesting to have 2 `b.Run` calls instead of one?

jacksontj:
I added Type Encoder/Decoder funcs and am now seeing the perf improvement I was expecting. This seems like a pretty ugly way to do this (since we have to copy/paste this marshal logic all over). So maybe it would be helpful to change the marshaling/unmarshaling in common to be fast?

```
benchmark                                                 old ns/op      new ns/op     delta
BenchmarkSerialization/1000/1000/marshal/jsoniter-8       4125222362     201316013     -95.12%
BenchmarkSerialization/1000/1000/unmarshal/jsoniter-8     2060620113     218847915     -89.38%

benchmark                                                 old allocs     new allocs     delta
BenchmarkSerialization/1000/1000/marshal/jsoniter-8       19993541       2030           -99.99%
BenchmarkSerialization/1000/1000/unmarshal/jsoniter-8     9240110        998003         -89.20%

benchmark                                                 old bytes     new bytes     delta
BenchmarkSerialization/1000/1000/marshal/jsoniter-8       689480872     106189225     -84.60%
BenchmarkSerialization/1000/1000/unmarshal/jsoniter-8     361144008     9749326       -97.30%
```

api/prometheus/v1/api_bench_test.go
brian-brazil: Yes. It'll be a little simpler.

brian-brazil:
It wouldn't make sense to add those to common (especially as we're trying to do away with model). What I'd suggest is more appropriate structures be defined here.

jacksontj:
If thats the case, then I'd just add these here for now and leave the model move for a separate PR; this is a *huge* perf improvement that I'd like to get in sooner.

api/prometheus/v1/api_bench_test.go
jacksontj: I've split it, IMO its not simpler (since it requires more indentation and another function wrap).

krasi-georgiev:
now with this custom Marshaling/Unmarshaling, do you think it is a good idea to also add a test to ensure the output is what is expected given some generated input data.

api/prometheus/v1/api.go
```diff
@@ -17,17 +17,83 @@ package v1
...
+
+	f, err := strconv.ParseFloat(iter.ReadString(), 64)
+	if err != nil {
+		iter.ReportError("unmarshal model.SamplePair", err.Error())
+		return
+	}
+	p.Value = model.SampleValue(f)
+
+	if iter.ReadArray() {
+		iter.ReportError("unmarshal model.SamplePair", "array must be of len 2")
```
krasi-georgiev: I don't understand what you mean by this comment.
Maybe something like
```suggestion
		iter.ReportError("unmarshal model.SamplePair", "array has more elements than expected")
```

api/prometheus/v1/api.go
```diff
@@ -17,17 +17,83 @@ package v1
...
+
+func unMarshalPointJSON(ptr unsafe.Pointer, iter *json.Iterator) {
+	p := *((*model.SamplePair)(ptr))
+	if !iter.ReadArray() {
+		iter.ReportError("unmarshal model.SamplePair", "value must be an array")
+		return
+	}
+	p.Timestamp = model.Time(iter.ReadFloat64())
+	if !iter.ReadArray() {
+		iter.ReportError("unmarshal model.SamplePair", "array must be of len 2")
```
krasi-georgiev: The comment is a bit unclear, but do we even need this check?

Shouldn't `f, err := strconv.ParseFloat(iter.ReadString(), 64)` just fail if there is nothing else to read?

api/prometheus/v1/api_bench_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"strconv"
+	"testing"
+	"time"
+
+	jsoniter "github.com/json-iterator/go"
+	"github.com/prometheus/common/model"
+)
+
+func generateData(timeseries, datapoints int) model.Matrix {
+	// Create the top-level matrix
```
krasi-georgiev:
```suggestion
```
Why do we need this comment? Isn't it obvious?

api/prometheus/v1/api.go
jacksontj: The iter requires calling readarray to proceed to the next item in the array, once you call it you have to call it until it returns false or you set an error.

api/prometheus/v1/api.go
jacksontj: The value array must have exactly 2 elements (1) timestamp and (2) value

api/prometheus/v1/api_bench_test.go
jacksontj: I can remove this, this benchmark is lifted from the issue I made against prom

api/prometheus/v1/api_bench_test.go
```diff
@@ -0,0 +1,110 @@
...
+	for i := 0; i < timeseries; i++ {
+		lset := map[model.LabelName]model.LabelValue{
+			model.MetricNameLabel: model.LabelValue("timeseries_" + strconv.Itoa(i)),
+		}
+		now := model.Now()
+		values := make([]model.SamplePair, datapoints)
+
+		for x := datapoints; x > 0; x-- {
+			values[x-1] = model.SamplePair{
+				Timestamp: now.Add(time.Second * -15 * time.Duration(x)), // Set the time back assuming a 15s interval
```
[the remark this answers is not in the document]
jacksontj: I used 14s at the time because that was my scrape interval, for benchmarking it only matters how many points we are marshaling (the spacing doesn't actually matter)

api/prometheus/v1/api.go
krasi-georgiev: yes I got that part, but in this particular case if `iter.ReadArray()` is true it means that that the parsing found more elements so the correct error should be:
"array has more elements than expected" , right?

jacksontj:
> Do you think it would make sense to add some benchmarking for `model.Vector` as well?

We could, but since the marshaling methods are on `model.SamplePair` we get more of them from Matrix than Vector. If we wanted to improve overall performance more we'd need to create these marshal methods for each of the types, which I'd prefer to leave for another PR (since this seemingly simple one is already over a week in ;) )

api/prometheus/v1/api.go
krasi-georgiev: yes I missed that part, thanks.
Maybe something like this would be a bit more clear for someone like me who is not that familiar with this package:

```
// Move the iterator to the the sample value.
if !iter.ReadArray() {
		iter.ReportError("unmarshal model.SamplePair", "array doesn't include a sample value")
```
can be even without the extra comment.

api/prometheus/v1/api_bench_test.go
krasi-georgiev: yep I don't see how it is usefull here.

api/prometheus/v1/api_bench_test.go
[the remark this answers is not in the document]
krasi-georgiev: Aah ok, Thanks. 
In that case let's just increment it by 1s or something similar  to avoid any confusion that it might actually have some special meaning.

krasi-georgiev:
> We could, but since the marshaling methods are on model.SamplePair we get more of them from Matrix than Vector. If we wanted to improve overall performance more we'd need to create these marshal methods for each of the types, which I'd prefer to leave for another PR (since this seemingly simple one is already over a week in ;) )

Yep It makes sense. Let's leave for another PR.

> now with this custom Marshaling/Unmarshaling for model.SamplePair, do you think it is a good idea to also add a test to ensure the output is what is expected given some generated input data.

how about that one?

jacksontj:
I copied over the tests (basically) from prom's web api. When I did so I found that `model.Time` doesn't properly unmarshal negative times (which I didn't think were even allowed?). I have opened https://github.com/prometheus/common/pull/193 to fix the unmarshaling, we can either (1) get that merged or (2) remove the test case with the negative time.

cc @krasi-georgiev

api/prometheus/v1/api.go
jacksontj: I changed the ReadArray errors to all be the same error `"SamplePair must be [timestamp, value]"` which is probably lots more descriptive anyways.

brian-brazil:
> (which I didn't think were even allowed?)

Times are int64s, so it is permitted though I've never come across it in practice.

api/prometheus/v1/api_bench_test.go
```diff
@@ -0,0 +1,111 @@
...
+			Metric: model.Metric(lset),
+			Values: values,
+		}
+
+		m = append(m, ss)
+	}
+	return m
+}
+
+func BenchmarkSerialization(b *testing.B) {
```
krasi-georgiev: I would prefer a more descriptive name like: `BenchmarkSamplesSerialization` or even `BenchmarkSamplesJsonSerialization` or `BenchmarkSamplesJsonMarshaling`

api/prometheus/v1/api_test.go
```diff
@@ -872,3 +876,90 @@ func TestAPIClientDo(t *testing.T) {
 
 	}
 }
+
+func TestSamplePair(t *testing.T) {
```
krasi-georgiev: need a comment to explain why do we need this test or what is it testing or a more descriptive name like: `TestSamplesJsonMarshalingCorrectness`

api/prometheus/v1/api_test.go
```diff
@@ -872,3 +876,90 @@ func TestAPIClientDo(t *testing.T) {
...
+			expected: `[0,"0.0000012345678"]`,
+		},
+		{
+			point:    model.SamplePair{0, 1.2345678e-67},
+			expected: `[0,"1.2345678e-67"]`,
+		},
+	}
+
+	for i, test := range tests {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
```
krasi-georgiev:
```suggestion
		t.Run(test.expected, func(t *testing.T) {
```

would be easyer to know which test is running or failed.

krasi-georgiev:
@beorn7 any idea why is this test failing https://travis-ci.org/prometheus/client_golang/jobs/533629401 
the PR seems unrelated to this failing test.

beorn7:
It's just a flaky test. I restarted it, and it worked. See #573 for tracking. This particular one might actually be fixed if you rebase.

jacksontj:
@krasi-georgiev comments addressed

@beorn7 I checked and I'm already based on master -- so it seems that it isn't fixed :(

beorn7:
Yeah, still the same kind of flakiness. Wasn't related to the data race.
I ran it again, and it worked. But you still have to add the DCO.

jacksontj:
I've squashed the commits, so they should be ready to merge now.

krasi-georgiev:
LGTM
@beorn7 any comments from you?

api/prometheus/v1/api_bench_test.go
```diff
@@ -0,0 +1,111 @@
+// Copyright 2017 The Prometheus Authors
```
beorn7: Not that it matters a lot, but for new files, we usually put the current year here.

api/prometheus/v1/api_bench_test.go
```diff
@@ -0,0 +1,111 @@
...
+package v1
+
+import (
+	"encoding/json"
+	"strconv"
+	"testing"
+	"time"
+
+	jsoniter "github.com/json-iterator/go"
+	"github.com/prometheus/common/model"
```
beorn7: Nit: In the other files, named imports are usually separated by a blank.

Not that I feel strongly, but I think we should be consistent.

api/prometheus/v1/api_test.go
```diff
@@ -15,17 +15,20 @@ package v1
...
 	"net/url"
 	"reflect"
 	"strings"
 	"testing"
 	"time"
 
+	json "github.com/json-iterator/go"
+
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/tsdb/testutil"
```
beorn7: Not part of this PR (as it happens in api/client_test.go already), but now that I'm seeing it here:

I find testutil packages problematic. The usual Go mantra is to not introduce custom test utilities and frameworks. Saving a few repetitive keystrokes comes at the cost of more dependencies and cognitive overhead for the reader (who has to adjust to the many different testutil packages out there). Using it in `prometheus/tsdb` might be justified as it is quite self-contained code. `prometheus/client_golang`, however, is deliberately meant to be used by many users with different stylistic ideas. The cognitive overhead for reading the code and the added dependency has more of an impact here. Note that the `prometheus/client_golang` has a dependency on `prometheus/tsdb` only because of the usage of `testutil`. Also note that this `testutil` package is not in `prometheus/common`. It was probably never meant to be used outside of `prometheus/tsdb`.

I'd like to see that dependency gone, but I'll open a separate issue for this.

go.mod
```diff
@@ -4,8 +4,12 @@ require (
 	github.com/beorn7/perks v1.0.0
 	github.com/go-logfmt/logfmt v0.4.0 // indirect
 	github.com/golang/protobuf v1.3.1
+	github.com/json-iterator/go v1.1.6
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.1 // indirect
 	github.com/prometheus/client_model v0.0.0-20190129233127-fd36f4220a90
-	github.com/prometheus/common v0.4.0
-	github.com/prometheus/procfs v0.0.0-20190507164030-5867b95ac084
+	github.com/prometheus/common v0.4.1-0.20190516160040-17f5ca174818
```
beorn7: This is an untagged version of `prometheus/common`.
- Is that really required?
- If yes, could we just ask @brian-brazil to tag a version in `prometheus/common`? It is very confusing for users of `client_golang` if they have an untagged dependency in an otherwise tagged repository.

api/prometheus/v1/api_test.go
jacksontj: Thats unfortunate because thats what I had originally (no testutil). If this is something we want going forward we should be clear about it up-front :)

go.mod
jacksontj: For the tests to pass this requires https://github.com/prometheus/common/pull/193 which is on master (otherwise untagged).

beorn7: I see. @brian-brazil I guess we could just tag a 0.4.1 in prometheus/common?

jacksontj:
@beorn7 @krasi-georgiev sounds like we just need to decide on the `common` package (don't update and have a broken test case, or get upstream to tag a new package).

api/prometheus/v1/api_test.go
beorn7: Yes, we definitely should.

If we want to deviate from common Go idioms and mantras, we should have an informed consensus about it among the Prometheus developers.

jacksontj: I've re-re-re written the tests, this time (again) to not use testutil :)

krasi-georgiev: >  The cognitive overhead for reading the code and the added dependency has more of an impact here. Note that the prometheus/client_golang has a dependency on prometheus/tsdb only because of the usage of testutil. 

In general I agree, but in this case I think it is exactly the opposite. It simplifies the code without any cognitive overhead.

In general it would be ideal to have in common and use across projects.

beorn7:
prometheus/common is now tagged as v0.4.1

krasi-georgiev:
>  The cognitive overhead for reading the code and the added dependency has more of an impact here. Note that the prometheus/client_golang has a dependency on prometheus/tsdb only because of the usage of testutil. 

In general I agree, but in this case I think it is exactly the opposite. It simplifies the code without any cognitive overhead.

What do you think about adding the testutils in the common repo?

jacksontj:
Would it be possible to tackle this testutil question outside of this PR? This PR is a huge performance improvement and its already been blocked up in review for over a week. I have rebased this and it should be ready for merge as-is.

IMO we don't *need* a testutil, it doesn't change the tests dramatically enough IMO to bother with the dep. If you guys want to maintain a shared testutils package somewhere though we could use it, but IIRC the plan was to remove `common`, so maybe adding there is not a great plan?

beorn7:
Yes, the testutil discussion is definitely out of scope of this PR. We can have it in #586 .

api/prometheus/v1/api_test.go
```diff
@@ -26,6 +26,8 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/api"
+	json "github.com/json-iterator/go"
```
beorn7: This line has to be before L28 to satisfy the CI style test.

However, for consistency, the named import should be in its own block (separated by a blank line), and L28 should be in the same block as L31 and L32.

beorn7:
See my comment for the last thing needed to satisfy the CI.

api/prometheus/v1/api_test.go
jacksontj: If we're going to be this picky (which I think is fine) I'd suggest adding a `make imports` function similar to https://github.com/jacksontj/promxy/blob/master/Makefile#L17

jacksontj:
@beorn7 Fixed the imports, although it looks like the unmarshaling doesn't work for nan/inf in <1.11 of go; do we need to support those versions or are we okay to drop them?

beorn7:
We have a commitment to support the last three Go minor releases at least, more if possible without undue hassle. Thus, dropping 1.9 should be fine, but for Go 1.10, we should wait for 1.13 to be out for a while.

Not sure what's a good way out of this.

jacksontj:
@beorn7 I have checked through the release notes and I don't see why it would be fixed in that version. I have also downloaded 1.10 and run the tests locally using it and they pass, so I'm not sure why its failing in CI

```
$ GO111MODULE= /usr/local/go.1.10/bin/go test -short  ./...
ok  	github.com/prometheus/client_golang/api	0.006s
ok  	github.com/prometheus/client_golang/api/prometheus/v1	0.004s
```

beorn7:
Semi-educated guess: This has to do with Go Modules support in Go 1.11+, but not in Go 1.10 and earlier.

With Go Modules, the tests use Prometheus 2.9. Without Go Modules, they use the tip of master, which has changed the encoding of the alert values. We need https://github.com/prometheus/client_golang/pull/585 merged to be compatible again, but we will only do that once Prometheus 2.10 is released and tagged.

jacksontj:
@beorn7 I don't think so? The error is marshaling a `model.SamplePair` which should have nothing to do with the alert value. From some more looking it seems that this is actually due to jsoniter upstream (https://github.com/json-iterator/go/issues/365) -- apparently that marshaling output is against RFC, but from what I understand that is what prometheus is expecting?

beorn7:
Thanks for the investigation. Indeed, my semi-educated guess was way too semi…

My guess was right, however, about Go Modules support being the issue. Go 1.11+ take json-iterator/go at the tag v1.1.6, while the earlier Go versions take the head of master and thus include the fix.

The encoding of Inf and NaN is indeed required by Prometheus. How could that have ever worked?

beorn7:
Early result: The Prometheus server is _not_ using `model.SamplePair` to marshal sample pairs into JSON. The sample value is represented as a string.

See https://prometheus.io/docs/prometheus/latest/querying/api/#expression-query-result-formats: “JSON does not support special float values such as NaN, Inf, and -Inf, so sample values are transferred as quoted JSON strings rather than raw numbers.”

jacksontj:
So should we change model SamplePair to a string instead of a float64?

beorn7:
`model.SamplePair` is correct, a value in Prometheus is a float64.
However, it's not suitable for JSON marshaling, as JSON floats are too restricted. (AFAIK, it is used nowhere for JSON marshalling.)

I'm not an expert when it comes to JSON APIs and testing thereof. Perhaps just provide the actual JSON code as reference for the test? You could also check out the prometheus/prometheus code to see how the JSON is created in the actual Prometheus API server code (which I'm not very familiar with). Just random ideas… Perhaps @krasi-georgiev has some advice.

krasi-georgiev:
will check in few days.

jacksontj:
@beorn7 @krasi-georgiev I have updated the PR to support this. Since we are already implementing our own marshal, I'm just doing the write as bytes instead of as a float64 -- this version works on jsoniter master.

beorn7:
Is there trailing whitespace in ./api/prometheus/v1/api.go L94?

beorn7:
@krasi-georgiev besides the formatting nit, does this look good to you?

jacksontj:
ran go fmt (fixed the whitespace), the CI run should pass now.

jacksontj:
And that test failure I've seen before that test is flaky for some reason :/ I don't see how I can re-trigger the tests -- I assume a maintainer has to do that?

beorn7:
Restarted the flaky test. (I have #573 in my pipeline to unflake.)

jacksontj:
@beorn7 tests have passed, so we should be good for a merge

beorn7:
I'll just wait for :+1: from @krasi-georgiev as he is the maintainer of this code.

krasi-georgiev:
back to full speed work mode in 1-2 days so will approve soon.

PR #572, opened 2019-05-08
Clarify deprecation of `DefObjectives`

beorn7:
Previously, the whole `Objectives` field was marked as deprecated by
linters.

@aixeshunter


aixeshunter:
Good @beorn7

PR #576, opened 2019-05-14
Fix regression issue 574

cviecco:
On commit 1173d734050af725504a9b2fbe93bff0b16f1b53 the push function was removed. Causing issue with wrappers that implement that function. If the wrapper implemented that function the delegator would fall in the range [16-31] and with NewDelegator returning a nil value. This then caused a panic.

This patch pulls back part of the removed items on 173d734050af725504a9b2fbe93bff0b16f1b53.


beorn7:
Ouch. Thanks for catching this.

PR #579, opened 2019-05-15
Cut v0.9.3 and announce new versioning plan

beorn7:
@simonpasquier to check sanity with Go Modules.
@krasi-georgiev for API client parts.


krasi-georgiev:
LGTM

beorn7:
Thanks. I'll just fix a race condition in a test that I have just spotted and get thin into the release.

beorn7:
OK, all done. Will merge on green (and then tag and rename the branch and such).

Issue #573, opened 2019-05-08
Unflake flaky tests


beorn7:
Umbrella issue for flaky tests, check those that have been unflaked:
- [x] `TestGoCollectorMemStats` is timing dependent and fails occasionally on Travis.
- [x] `TestGoCollectorGC` occasionally fails on Travis.
- [x] `TestClientMiddlewareAPI` contacts http://google.com (which might fail, and is also bad because it needs an internet connection).

Example failures of `TestGoCollectorMemStats`:
```
--- FAIL: TestGoCollectorMemStats (0.02s)
    go_collector_test.go:183: unexpected value of go_memstats_alloc_bytes, want 4, got 3
    go_collector_test.go:235: unexpected of msLast.Alloc, want 4, got 3
```

```
--- FAIL: TestGoCollectorMemStats (0.02s)
    go_collector_test.go:183: unexpected value of go_memstats_alloc_bytes, want 2, got 3
```

Example failure of `TestGoCollectorGC`:
```
--- FAIL: TestGoCollectorGC (0.00s)
    go_collector_test.go:142: want 1 new garbage collection run, got 2
```

beorn7:
`TestGoCollectorMemStats` goes crazy if run with race detector on my laptop, but works on Travis. Hmm…

beorn7:
Got it. Travis is configured to run `test-short`. The race condition is real (and non-flaky),  but it is only in the test, not in the code being tested.

beorn7:
Race condition is fixed. Might be that that also solves the other flakiness. Will see.

beorn7:
It doesn't.

beorn7:
I assume it is fixed now. Not completely sure, but I'll re-open if flaky tests happen again.

PR #581, opened 2019-05-16
Fix race in TestGoCollectorMemStats

beorn7:
Related to #573 (but only partially fixes it).

@simonpasquier @krasi-georgiev This is the fix I want to get in before cutting v0.9..3 in #579 .


krasi-georgiev:
sorry I am not familiar with these modules to give a review.

brian-brazil:
:+1:

PR #582, opened 2019-05-16
Make use of pre-existing context in InstrumentRoundTripperTrace

### LINKED ISSUE #580: Request context is overridden in instrumented http clients
It is quite common for golang http clients to make use of the `context` package. Example use cases are to halt stalled requests or shutdown appropriately for example. When an application uses the `github.com/prometheus/client_golang/prometheus` package to instrument an http client for metrics, it expects the transport wrappers to preserve the behavior and state of the wrapped client.   

However, what happens with a request context set before the instrumentation of the client is that it gets [replaced by the default root context](https://github.com/prometheus/client_golang/blob/master/prometheus/promhttp/instrument_client.go#L216)  and that makes quite of a difference:
```
r = r.WithContext(httptrace.WithClientTrace(context.Background(), trace))
```
From the [golang docs](https://golang.org/pkg/net/http/#Request.WithContext): 
"WithContext returns a shallow copy of r **with its context changed to ctx.**"

The effect is that if you created a context with, for example, `context.WithTimeout` and set it to a `http.Client` (`client.WithContext(ctx)`), and then instrument the client, for example as suggested [here](https://github.com/prometheus/client_golang/blob/master/prometheus/promhttp/instrument_client_test.go#L107-L186), and then send requests to a server that is slow enough to trigger timeout, the timeout will never occur.

What is expected instead is that the [`InstrumentRoundTripperTrace`](https://github.com/prometheus/client_golang/blob/master/prometheus/promhttp/instrument_client.go#L136) function respects existing request context, if any.

All findings are wrt [`v0.9.2`](https://github.com/prometheus/client_golang/blob/v0.9.2/prometheus/promhttp/instrument_client.go) and [`master`](https://github.com/prometheus/client_golang/blob/master/prometheus/promhttp/instrument_client.go).

beorn7:
Fixes #580

@shturec does this do the trick?

### DIFF
--- a/prometheus/promhttp/instrument_client.go
+++ b/prometheus/promhttp/instrument_client.go
@@ -14,7 +14,6 @@
 package promhttp
 
 import (
-	"context"
 	"crypto/tls"
 	"net/http"
 	"net/http/httptrace"
@@ -213,7 +212,7 @@ func InstrumentRoundTripperTrace(it *InstrumentTrace, next http.RoundTripper) Ro
 				}
 			},
 		}
-		r = r.WithContext(httptrace.WithClientTrace(context.Background(), trace))
+		r = r.WithContext(httptrace.WithClientTrace(r.Context(), trace))
 
 		return next.RoundTrip(r)
 	})
--- a/prometheus/promhttp/instrument_client_test.go
+++ b/prometheus/promhttp/instrument_client_test.go
@@ -14,15 +14,18 @@
 package promhttp
 
 import (
+	"context"
+	"fmt"
 	"log"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


shturec:
Yep, seems ok to me. 
I didn't see you have made a PR already and shot my own (#584) and I've contributed two tests inside that used to verify the behavior with context (using timeout context for the experiment). They work also with your solution so I think it's fine. 
Feel free to make use of them if you find them ok.

beorn7:
Thanks. Having tests for this is great. However, we don't want to pull in a dependency just for assertion. If you don't mind, I will just incorporate the essence of your tests into this PR.

beorn7:
I think I got it all nicely packaged up.

beorn7:
Incidentally, this also fixes the 1st flaky test mentioned in #573

PR #583, opened 2019-05-16
Remove outdated use-case description from doc comment

beorn7:
We stopped advertising binary-wide setting of a label quite a while
ago. This doc comment was missed in the cleanup.

@brian-brazil @Ashish-Bansal follow-up to the discussion in https://github.com/prometheus/client_python/issues/406


brian-brazil:
:+1:

PR #585, opened 2019-05-20
/api/v1/rules: alert value as string

holycheater:
a follow-up on https://github.com/prometheus/prometheus/pull/5582

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -226,7 +226,7 @@ type Alert struct {
 	Annotations model.LabelSet
 	Labels      model.LabelSet
 	State       AlertState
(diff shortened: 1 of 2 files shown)


beorn7:
The Prometheus change will be in 2.10. We should get a release of client_golang done at about the same time as the final 2.10.0 release.

daenney:
2.10-rc0 got tagged today

beorn7:
The final release will happen in three days at the earliest.

beorn7:
@krasi-georgiev Does this look good to you?

beorn7:
Once we have @krasi-georgiev 's LGTM, we'll merge this.

PR #588, opened 2019-05-26
Attempt to fix flakiness of TestGoCollectorMemStats

beorn7:
This is really lame as it essentially just uses longer times to
wait. The test is still timing-dependent and thus could still
theoretically fail with unlucky scheduling. However, we are testing
something that _is_ about timing. Turning this all into something not
timing-dependent would be first quite involved and second might defeat
the purpose of testing code that is inherently about timing.

Let's see how this works out in practice.

Signed-off-by: beorn7 <bjoern@rabenste.in>

### DIFF
--- a/prometheus/go_collector_test.go
+++ b/prometheus/go_collector_test.go
@@ -184,14 +184,14 @@ func TestGoCollectorMemStats(t *testing.T) {
 		}
 	}
 
-	// Speed up the timing to make the tast faster.
-	c.msMaxWait = time.Millisecond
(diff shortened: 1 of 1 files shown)


beorn7:
Together with #589 , this should fix #573.

brian-brazil:
Lgtm

We might need more, at least on some older versions of Go we ended up with
100ms (plus allowing for kernel process scheduling).

beorn7:
Let's see how it works out.

PR #589, opened 2019-05-26
Unflake TestGoCollectorGC

beorn7:
It is perfectly possible that a normal GC happens just before the
forced one. Thus seeing 2 GCs is fine.

Whenever this test failed, it was because two GCs were seen.

Signed-off-by: beorn7 <bjoern@rabenste.in>


beorn7:
Together with #588 , this should fix #573.

brian-brazil:
Lgtm

PR #593, opened 2019-06-04
 Remove use of deprecated procfs functions

beorn7:
@pgier

### DIFF
--- a/prometheus/process_collector.go
+++ b/prometheus/process_collector.go
@@ -126,7 +126,7 @@ func NewProcessCollector(opts ProcessCollectorOpts) Collector {
 	}
 
 	// Set up process metric collection if supported by the runtime.
-	if _, err := procfs.NewStat(); err == nil {
+	if _, err := procfs.NewDefaultFS(); err == nil {
 		c.collectFn = c.processCollect
 	} else {
 		c.collectFn = func(ch chan<- Metric) {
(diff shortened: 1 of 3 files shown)


prometheus/process_collector.go
```diff
@@ -126,7 +126,7 @@ func NewProcessCollector(opts ProcessCollectorOpts) Collector {
 	}
 
 	// Set up process metric collection if supported by the runtime.
-	if _, err := procfs.NewStat(); err == nil {
+	if _, err := procfs.NewDefaultFS(); err == nil {
```
gouthamve: I don't know the internals, but previously we are checking if we can ready the data correctly, but with this change we will be checking if the procfs mount exists. Not sure if that is enough.

beorn7: You are right, previously we _also_ tested actually reading the proc stats.

However, that was not a guarantee that it would always work. For example, if somebody removed the read permissions on the proc files, it would fail at the next collection. Thus, we always had error handling in place to deal with that. 

The initial test was mostly there to weed out platforms that have no procfs support in general so that we avoid trying it out on each collect cycle again.

Arguably, the way this change changes it is even better: If there is no procfs at all, it will prevent collection for good. If there is a procfs but it is not readable, that might be a temporary problem, so it makes sense to try on each collect cycle again. (Needless to say that all of this are very rare corner cases.)

gouthamve: Agreed, :+1: This PR is good to go!

Issue #542, opened 2019-03-05
Late encoding errors cannot create HTTP error


beorn7:
https://github.com/prometheus/client_golang/blob/2daed26f633ce23fab5efc02fb3f2e803b3526c3/prometheus/promhttp/http.go#L155

This call will fail with `multiple response.WriteHeader calls` if something has already been sent down the wire. This needs to be treated in a better way.

PR #594, opened 2019-06-05
Add an error counter for internal errors in the HTTP handler

### LINKED ISSUE #543: promhttp: Consider a metric for HTTP exposition errors
There are cases where an error won't or cannot be sent as an HTTP response with the appropriate status code (examples: `ContinueOnError` as configured error handling, #542 ). While those problem should be caught be logs, it is better if alerts can be set up for those problems. Thus, `promhttp` should add an error counter as a metric.

beorn7:
The doc comments explain the rationale in a quite detailed way.

Fixes #543 and #542

I'll add a test once we have decided this is the way to go.

### DIFF
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -84,9 +84,31 @@ func Handler() http.Handler {
 // instrumentation. Use the InstrumentMetricHandler function to apply the same
 // kind of instrumentation as it is used by the Handler function.
 func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
-	var inFlightSem chan struct{}
+	var (
+		inFlightSem chan struct{}
+		errCnt      = prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "promhttp_metric_handler_errors_total",
+				Help: "Total number of internal errors encountered by the promhttp metric handler.",
+			},
+			[]string{"cause"},
+		)
+	)
+
 	if opts.MaxRequestsInFlight > 0 {
 		inFlightSem = make(chan struct{}, opts.MaxRequestsInFlight)
+	}
+	if opts.Registry != nil {
+		// Initialize all possibilites that can occur below.
+		errCnt.WithLabelValues("gathering")
+		errCnt.WithLabelValues("encoding")
+		if err := opts.Registry.Register(errCnt); err != nil {
+			if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
+				errCnt = are.ExistingCollector.(*prometheus.CounterVec)
+			} else {
+				panic(err)
+			}
+		}
 	}
 
 	h := http.HandlerFunc(func(rsp http.ResponseWriter, req *http.Request) {
@@ -106,6 +128,7 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 			if opts.ErrorLog != nil {
 				opts.ErrorLog.Println("error gathering metrics:", err)
 			}
+			errCnt.WithLabelValues("gathering").Inc()
 			switch opts.ErrorHandling {
 			case PanicOnError:
 				panic(err)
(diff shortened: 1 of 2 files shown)


prometheus/promhttp/http.go
```diff
@@ -236,9 +264,12 @@ const (
...
 	// if no metrics can be served, serve an HTTP status code 500 and the
 	// last error message in the body. Only use this in deliberate "best
-	// effort" metrics collection scenarios. It is recommended to at least
-	// log errors (by providing an ErrorLog in HandlerOpts) to not mask
-	// errors completely.
+	// effort" metrics collection scenarios. In this case, it is highly
+	// recommended to provide other means of detecting errors: By setting an
+	// ErrorLog in HandlerOpts, the errors are logged. By providing a
+	// Registry in HandlerOpts, the exposed metrics include an error counter
+	// "promhttp_metric_handler_errors_total", which can easily be used for
```
brian-brazil: I'd `s/easily//` here, as it's not automatically easy for everyone.

prometheus/promhttp/http.go
```diff
@@ -84,10 +88,32 @@ func Handler() http.Handler {
...
 	if opts.MaxRequestsInFlight > 0 {
 		inFlightSem = make(chan struct{}, opts.MaxRequestsInFlight)
 	}
+	if opts.Registry != nil {
+		// Initialize all possibilites that can occur below.
+		errCnt.WithLabelValues("gathering")
+		errCnt.WithLabelValues("encoding")
+		if err := opts.Registry.Register(errCnt); err != nil {
+			if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
+				errCnt = are.ExistingCollector.(*prometheus.CounterVec)
```
brian-brazil: Using a global would seem cleaner here, rather than poking around error types to the same effect

prometheus/promhttp/http.go
```diff
@@ -62,18 +62,22 @@ var gzipPool = sync.Pool{
...
 // This function is meant to cover the bulk of basic use cases. If you are doing
 // anything that requires more customization (including using a non-default
 // Gatherer, different instrumentation, and non-default HandlerOpts), use the
 // HandlerFor function. See there for details.
 func Handler() http.Handler {
 	return InstrumentMetricHandler(
-		prometheus.DefaultRegisterer, HandlerFor(prometheus.DefaultGatherer, HandlerOpts{}),
+		prometheus.DefaultRegisterer, HandlerFor(
+			prometheus.DefaultGatherer,
+			HandlerOpts{Registry: prometheus.DefaultRegisterer},
```
brian-brazil: I'm not sure this makes sense for the default handler, as gathering will fail the whole scrape anyway and once we've OpenMetrics it has an explicit EOF that'll catch encoding issues. Given that a little wary of adding two more series for ~every target, for cases that don't/won't happen.

prometheus/promhttp/http.go
beorn7: Removed.

prometheus/promhttp/http.go
beorn7: A global CounterVec wouldn't help here. We would still need the same error type exercise. Even if you re-register the same collector, you'll get an error.

However, if we used a global here, you couldn't use separate counters when passing different registries.

brian-brazil: Yeah, that had just occurred to me.

prometheus/promhttp/http.go
beorn7: That's a good point. 99% of the usage for the new metric will be with `ContinueOnError`. However, the incident that triggered #542 would happen even with the default handler.

But yeah, adding metrics for the default case might be confusing. I'll remove it here.

brian-brazil:
:+1:

beorn7:
Thanks. Working on a test now.

beorn7:
Test added, will merge on green.

PR #596, opened 2019-06-06
Implement process collector for Windows

### LINKED ISSUE #376: Make `process_…` metrics work on MS Windows

carlpett:
This PR implements a process collector for Windows, resolves #376.

Some discussion points:
- The `process_cpu_seconds_total` uses an underlying data source that is, from my understanding, incremented with a resolution of ~16 ms. There might be better APIs to use? The ones I've found work with cycles, though, which isn't that easy to convert to seconds.
- There aren't really "file descriptors" on Windows. I did an interpretation of it to be handles (which covers lots of different things in addition to files) instead. And there is a hard-coded max of 16M handles per process.
- I'm not fully sure I did the mapping for Linux memory concepts to Windows correctly. Would love a sanity check there.

### DIFF
--- a/prometheus/process_collector_other.go
+++ b/prometheus/process_collector_other.go
@@ -1,0 +1,65 @@
+// Copyright 2015 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !windows
+
+package prometheus
+
+import (
+	"github.com/prometheus/procfs"
+)
+
+func canCollectProcess() bool {
+	_, err := procfs.NewDefaultFS()
+	return err == nil
+}
+
+func (c *processCollector) processCollect(ch chan<- Metric) {
+	pid, err := c.pidFn()
+	if err != nil {
+		c.reportError(ch, nil, err)
+		return
+	}
+
+	p, err := procfs.NewProc(pid)
+	if err != nil {
+		c.reportError(ch, nil, err)
+		return
+	}
+
+	if stat, err := p.Stat(); err == nil {
+		ch <- MustNewConstMetric(c.cpuTotal, CounterValue, stat.CPUTime())
+		ch <- MustNewConstMetric(c.vsize, GaugeValue, float64(stat.VirtualMemory()))
+		ch <- MustNewConstMetric(c.rss, GaugeValue, float64(stat.ResidentMemory()))
+		if startTime, err := stat.StartTime(); err == nil {
+			ch <- MustNewConstMetric(c.startTime, GaugeValue, startTime)
+		} else {
+			c.reportError(ch, c.startTime, err)
+		}
+	} else {
+		c.reportError(ch, nil, err)
+	}
+
+	if fds, err := p.FileDescriptorsLen(); err == nil {
+		ch <- MustNewConstMetric(c.openFDs, GaugeValue, float64(fds))
+	} else {
+		c.reportError(ch, c.openFDs, err)
+	}
+
+	if limits, err := p.Limits(); err == nil {
+		ch <- MustNewConstMetric(c.maxFDs, GaugeValue, float64(limits.OpenFiles))
+		ch <- MustNewConstMetric(c.maxVsize, GaugeValue, float64(limits.AddressSpace))
+	} else {
+		c.reportError(ch, nil, err)
+	}
+}
--- a/prometheus/process_collector_other.go
+++ b/prometheus/process_collector_other.go
@@ -1,4 +1,4 @@
-// Copyright 2015 The Prometheus Authors
+// Copyright 2019 The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/prometheus/process_collector_windows.go
+++ b/prometheus/process_collector_windows.go
@@ -95,9 +95,9 @@ func (c *processCollector) processCollect(ch chan<- Metric) {
 		c.reportError(ch, nil, err)
 		return
 	}
-	ch <- MustNewConstMetric(c.vsize, GaugeValue, float64(mem.WorkingSetSize))
+	ch <- MustNewConstMetric(c.vsize, GaugeValue, float64(mem.PrivateUsage))
 	ch <- MustNewConstMetric(c.maxVsize, GaugeValue, float64(mem.PeakWorkingSetSize))
-	ch <- MustNewConstMetric(c.rss, GaugeValue, float64(mem.PrivateUsage))
+	ch <- MustNewConstMetric(c.rss, GaugeValue, float64(mem.WorkingSetSize))
 
 	handles, err := getProcessHandleCount(h)
 	if err != nil {
@@ -105,7 +105,7 @@ func (c *processCollector) processCollect(ch chan<- Metric) {
 		return
 	}
 	ch <- MustNewConstMetric(c.openFDs, GaugeValue, float64(handles))
-	ch <- MustNewConstMetric(c.maxFDs, GaugeValue, float64(16*1024*1024)) // Windows has a hard-coded max limit, not per-process
+	ch <- MustNewConstMetric(c.maxFDs, GaugeValue, float64(16*1024*1024)) // Windows has a hard-coded max limit, not per-process.
 }
 
 func fileTimeToSeconds(ft windows.Filetime) float64 {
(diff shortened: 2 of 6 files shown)


carlpett:
ping @beorn7, what do you think? (Apart from my apparently having missed something with go modules, I'll fix that)

beorn7:
The test failures are because you haven't updated the go.mod file. (Let me know if you need help with the Go modules part.)

prometheus/process_collector_windows.go
```diff
@@ -0,0 +1,113 @@
...
+	ch <- MustNewConstMetric(c.cpuTotal, CounterValue, fileTimeToSeconds(kernelTime)+fileTimeToSeconds(userTime))
+
+	mem, err := getProcessMemoryInfo(h)
+	if err != nil {
+		c.reportError(ch, nil, err)
+		return
+	}
+	ch <- MustNewConstMetric(c.vsize, GaugeValue, float64(mem.WorkingSetSize))
+	ch <- MustNewConstMetric(c.maxVsize, GaugeValue, float64(mem.PeakWorkingSetSize))
+	ch <- MustNewConstMetric(c.rss, GaugeValue, float64(mem.PrivateUsage))
```
brian-brazil: I think RSS is WorkingSetSize

prometheus/process_collector_other.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2015 The Prometheus Authors
```
beorn7: Not that it matters a lot, but in a new file, we usually update the year to the current one.

prometheus/process_collector_windows.go
```diff
@@ -0,0 +1,113 @@
...
+		return
+	}
+
+	var startTime, exitTime, kernelTime, userTime windows.Filetime
+	err = windows.GetProcessTimes(h, &startTime, &exitTime, &kernelTime, &userTime)
+	if err != nil {
+		c.reportError(ch, nil, err)
+		return
+	}
+	ch <- MustNewConstMetric(c.startTime, GaugeValue, float64(startTime.Nanoseconds()/1e9))
```
beorn7: Is there a reason not to use `fileTimeToSeconds` here, too?

prometheus/process_collector_windows.go
```diff
@@ -0,0 +1,113 @@
...
+	ch <- MustNewConstMetric(c.cpuTotal, CounterValue, fileTimeToSeconds(kernelTime)+fileTimeToSeconds(userTime))
+
+	mem, err := getProcessMemoryInfo(h)
+	if err != nil {
+		c.reportError(ch, nil, err)
+		return
+	}
+	ch <- MustNewConstMetric(c.vsize, GaugeValue, float64(mem.WorkingSetSize))
+	ch <- MustNewConstMetric(c.maxVsize, GaugeValue, float64(mem.PeakWorkingSetSize))
```
beorn7: maxVsize is the maximum possible virtual memory size, not the observed peak. Not sure if there is a way to get this from somewhere on MS Windows. Perhaps it has to be hardcoded depending on whether it's 64bit or 32bit.

prometheus/process_collector_windows.go
beorn7: Yeah, the Windows terminology is kind of weird. WSS is something else than RSS, but apparently, when Windows says "WSS", it means "RSS".

prometheus/process_collector_windows.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	ch <- MustNewConstMetric(c.startTime, GaugeValue, float64(startTime.Nanoseconds()/1e9))
+	ch <- MustNewConstMetric(c.cpuTotal, CounterValue, fileTimeToSeconds(kernelTime)+fileTimeToSeconds(userTime))
+
+	mem, err := getProcessMemoryInfo(h)
+	if err != nil {
+		c.reportError(ch, nil, err)
+		return
+	}
+	ch <- MustNewConstMetric(c.vsize, GaugeValue, float64(mem.WorkingSetSize))
```
beorn7: Concluding from the comments below, the `WorkingSetSize` is more like the RSS. I have no clue how to get something like the vsize on Windows.

prometheus/process_collector_windows.go
```diff
@@ -0,0 +1,113 @@
...
+	ch <- MustNewConstMetric(c.maxVsize, GaugeValue, float64(mem.PeakWorkingSetSize))
+	ch <- MustNewConstMetric(c.rss, GaugeValue, float64(mem.PrivateUsage))
+
+	handles, err := getProcessHandleCount(h)
+	if err != nil {
+		c.reportError(ch, nil, err)
+		return
+	}
+	ch <- MustNewConstMetric(c.openFDs, GaugeValue, float64(handles))
+	ch <- MustNewConstMetric(c.maxFDs, GaugeValue, float64(16*1024*1024)) // Windows has a hard-coded max limit, not per-process
```
beorn7: Nit: End sentence with a period.

beorn7:
Wild guess: For vsize, we have to check PagefileUsage. If it is zero, we have to use PrivateUsage. 

But it would be really good if somebody with a good understanding of Windows's memory management could help out.

prometheus/process_collector_windows.go
carlpett: Yes. For whatever reason, `Filetime.Nanoseconds` [converts to UNIX epoch](https://godoc.org/golang.org/x/sys/windows#Filetime.Nanoseconds), while the actual time stored within is number of 100 ns intervals passed 🤷‍♂

prometheus/process_collector_windows.go
carlpett: It seems it might depend on OS version as well. I found this source https://blogs.technet.microsoft.com/markrussinovich/2008/11/17/pushing-the-limits-of-windows-virtual-memory/ where they claim 8TB on 64 bit. They have a reference table which goes to Windows 2012. On my Windows lab host, I could reserve 128TB, though.

I'll see if this is available through some API, but otherwise I don't know. Not super keen on the idea of maintaining a lookup table which is hard to know when it needs to be updated.

prometheus/process_collector_windows.go
carlpett: Yes, that seems correct. 
>The "working set" of a process is the set of memory pages currently visible to the process in physical RAM memory. These pages are resident and available for an application to use without triggering a page fault

prometheus/process_collector_windows.go
carlpett: Alternatively, on Linux it seems we represent `unlimited` with -1. Might make sense here too?

brian-brazil: We could not export it. I doubt there's many 32bit systems out there that would care (and it's a hardcoded value if they need it).

carlpett:
@beorn7 From my understanding of [PROCESS_MEMORY_COUNTERS_EX](https://docs.microsoft.com/en-us/windows/desktop/api/psapi/ns-psapi-_process_memory_counters_ex), `PrivateUsage` and `PagefileUsage` are the same, but the latter is deprecated and always zero (as some sort of backwards compatibility with the non-EX `PROCESS_MEMORY_COUNTERS`, I suppose)?

beorn7:
It seemed to me that `PagefileUsage` could be non-zero on older Windows version (which then would not have `PrivateUsage`?). The MS documentation is not really clear to me. Feels like a blast from the past...

prometheus/process_collector_windows.go
beorn7: Ugh…

carlpett:
Yeah, it is weirdly formulated. My interpretation is that PagefileUsage=PrivateUsage, except on older versions where PagefileUsage=0 (but PrivateUsage is still set). But I may be wrong.

prometheus/process_collector_windows.go
carlpett: That's also an option, of course. But 32-bit isn't actually something we can hardcode, if I understand it correctly. The value [will depend](https://docs.microsoft.com/en-us/windows/desktop/Memory/memory-limits-for-windows-releases#memory-and-address-space-limits) on if the OS is 32 bit or not, and if booted with a 2/2 GB or 3/1 GB split between OS and application. On a 64 bit OS, it depends on if the `IMAGE_FILE_LARGE_ADDRESS_AWARE` flag is set on the binary by the go compiler, which seems to be the case: https://github.com/golang/go/blob/e883d000f4ce0c47711c3a7c59df8bb2f0ec557f/src/cmd/link/internal/ld/pe.go#L785-L788

carlpett: But again, it is not clear if those hoops are worth jumping through.

beorn7: In doubt, let's just not export this metric, as Brian suggested.

carlpett: Sounds good to me. In that case, should we drop the "max fds" too? That value is pretty uninteresting (although "correct"), and I added it to try to keep parity.

beorn7: Yes, let's keep it as it is the technically correct value and doesn't imply any maintenance overhead to keep it correct.

beorn7:
In lack of better information, let's do it as discussed. I think the only remaining change is to remove the `c.maxVsize` metric. Then we can merge this.

carlpett:
Done!

beorn7:
Thanks 1M!

PR #598, opened 2019-06-07
Cut v0.9.4

beorn7:
@krasi-georgiev could you please check the API client parts in particular?

Also, is there anything in the pipeline for the API client we should delay the release for?

This is planned to be the last release before removing all deprecated code and make it v1.0.0 (see first section of the README.md for details).


krasi-georgiev:
> @krasi-georgiev could you please check the API client parts in particular?

all looks good.



> Also, is there anything in the pipeline for the API client we should delay the release for?

nope, can't think of anything.

beorn7:
Ok, I'll cut the release then. Thanks.

PR #599, opened 2019-06-08
Pass warnings through on non-error responses

jacksontj:

Without this we have 2 problems:

(1) we cannot differentiate an error from a warning from the API response body (as the `Err()` method is returning self always

(2) we can inadvertently layer these errors together when using `NewErrorAPI`

This PR addresses both.

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -211,8 +211,11 @@ func (e *Error) Error() string {
 	return "Warnings: " + strings.Join(e.warnings, " , ")
 }
 
-func (w *Error) Err() error {
-	return w
+func (e *Error) Err() error {
+	if e.Type != "" || e.Msg != "" {
+		return errors.New(fmt.Sprintf("%s: %s", e.Type, e.Msg))
+	}
+	return nil
 }
 
 func (w *Error) Warnings() []string {
@@ -847,8 +850,6 @@ func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, [
 
 	code := resp.StatusCode
 
-	var err api.Error
-
 	if code/100 != 2 && !apiError(code) {
 		errorType, errorMsg := errorTypeAndMsgFor(resp)
 		return resp, body, &Error{
@@ -869,6 +870,8 @@ func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, [
 		}
 	}
 
+	var err *Error
+
 	if apiError(code) != (result.Status == "error") {
 		err = &Error{
 			Type:     ErrBadResponse,
@@ -883,6 +886,13 @@ func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, [
 			Msg:      result.Error,
 			warnings: result.Warnings,
 		}
+	}
+
+	if len(result.Warnings) > 0 {
+		if err == nil {
+			err = &Error{}
+		}
+		err.warnings = result.Warnings
 	}
 
 	return resp, []byte(result.Data), err
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -22,7 +22,6 @@ import (
 	"math"
 	"net/http"
 	"strconv"
-	"strings"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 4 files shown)


jacksontj:
On a related note, I've been trying to integrate this warnings behavior as a client and I have to say the behavior/interface is infuriating to use. 

According to the upstream prom docs the warnings are meant to be non-fatal. If the response has a warning and no error we currently return an `Error` type which implements the `error` interface. This is a bit clunky for direct usage, but manageable. If you are layering methods this is completely untenable. In golang if the err is not nil methods generally don't pass the value back up. A basic case would look like:

```
func PassThrough() (interface{}, error) {
    v, err := DownstreamFunc()
    if err != nil {
        return nil, err
    }
    return v, nil
```

Since this client is returning a non-nil thing that *is* a warning (and implements `erro`), if you want to still get the value when there is a warning but not an error it requires changing all code you have written before -- and (worse) to change your coding habits moving forward (as this is very not idiomatic go). To show how messy this is an example of this passing-through would look like:

```
func PassThrough() (interface{}, error) {
    v, err := DownstreamFunc()
    if apiErr, ok := err.(api.Error); ok  {
        if apiErr.Err() != nil {
            return nil, apiErr
        }
    }
    return v, api.NewErrorAPI(err)
```

Assuming upstream plans on keeping this behavior (warnings are non-fatal) IMO we should change the signature to pass warning separate from error. As it stands you basically have to rewrite everything once *and* change passthrough funcs and future coding habits. If we were to change the return to `(value, warnings, error)` It would require the one rewrite of the caller of the client but avoid problems with layering/passthrough.

Thoughts?


cc @krasi-georgiev

krasi-georgiev:
@jacksontj yep looks like this implementation make the use of error quite bad so yeah lets try with your original suggestion.

jacksontj:
Ok, I'll try and get a pr together some time this week.

jacksontj:
@krasi-georgiev here is the change, for now I only wired up warnings to query/queryRange -- as they are the only ones that pass that information back. If these get added to others in the future it would be easy to add -- but I don't think thats planned.

api/prometheus/v1/api_test.go
```diff
@@ -705,7 +714,11 @@ func TestAPIs(t *testing.T) {
 		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
 			client.curTest = test
 
-			res, err := test.do()
+			res, warnings, err := test.do()
+
+			if (test.inWarnings == nil) != (warnings == nil) {
```
krasi-georgiev: I am a bit confused by this. The `queryTests` don't have the `inWarnings` set anywhere so this will always be true `test.inWarnings == nil`

jacksontj: @krasi-georgiev true, I've added a couple cases that do so.

krasi-georgiev: but than you should also check the content right? `testutil.Equals(t,test.inWarnings,warnings)`

krasi-georgiev: and don't forget the full stop for the comments to follow the golang standards. I know it is annoying until you get used to it :smile_cat:

jacksontj: Updated again :)

krasi-georgiev:
Thanks!

api/prometheus/v1/api.go
```diff
@@ -230,34 +216,34 @@ type Range struct {
...
+	DeleteSeries(ctx context.Context, matches []string, startTime time.Time, endTime time.Time) error
 	// Flags returns the flag values that Prometheus was launched with.
-	Flags(ctx context.Context) (FlagsResult, api.Error)
+	Flags(ctx context.Context) (FlagsResult, error)
 	// LabelValues performs a query for the values of the given label.
-	LabelValues(ctx context.Context, label string) (model.LabelValues, api.Error)
+	LabelValues(ctx context.Context, label string) (model.LabelValues, error)
 	// Query performs a query for the given time.
-	Query(ctx context.Context, query string, ts time.Time) (model.Value, api.Error)
+	Query(ctx context.Context, query string, ts time.Time) (model.Value, api.Warnings, error)
```
verdverm: This breaks a lot of downstream users of this library who have been relying on go get without a lock file

jacksontj: Unfortunately this was expected, upstream prometheus added these warnings and due to the semantics of those warnings we end up basically requiring this (more discussion in #562).

verdverm:
Legit, we are not using go ver mgmt, on us in my mind

api/prometheus/v1/api.go
jacksontj: Personally I'm finding this `Warnings` thing a major pain to integrate as it complicates things quite a bit more (and it doesn't really help IMO that its not really a "warning" from prometheus).

PR #600, opened 2019-06-11
Remove all deprecated features

beorn7:
This is in preparation of the upcoming v1 release.


brian-brazil:
:+1: 

Not sure what that test failure is.

beorn7:
Seems to have been in internal hiccup in golangci lint. Couldn't even find that error number anywhere. However, rerunning the test worked.

Let's brace for impact from all those that build from head and have ignored the deprecation warnings so far.

PR #603, opened 2019-06-14
Add warnings to series

jacksontj:

cc @krasi-georgiev

### DIFF
--- a/api/prometheus/v1/api_test.go
+++ b/api/prometheus/v1/api_test.go
@@ -163,8 +163,7 @@ func TestAPIs(t *testing.T) {
 
 	doSeries := func(matcher string, startTime time.Time, endTime time.Time) func() (interface{}, api.Warnings, error) {
 		return func() (interface{}, api.Warnings, error) {
-			v, err := promAPI.Series(context.Background(), []string{matcher}, startTime, endTime)
-			return v, nil, err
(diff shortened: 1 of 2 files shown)


jacksontj:
FYI I've also opened https://github.com/prometheus/prometheus/issues/5661 as this warnings return seems like it would need to be included in `LabelValues` and `LabelNames` as well.

krasi-georgiev:
Is this already in prometheus? can you put a link to the code.

jacksontj:
Yes, this is already in prometheus (https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L492). I have opened up the other issue because logically this same `Warnings` should exist for those other 2 APIs as well (but don't today).

krasi-georgiev:
can you add a test for this one as well?

jacksontj:
@krasi-georgiev sure, I thought I had already -- but that was for the other PR, here we go :)

api/prometheus/v1/api_test.go
```diff
@@ -376,6 +401,21 @@ func TestAPIs(t *testing.T) {
 			},
 			err: fmt.Errorf("some error"),
 		},
+		// series with error and warning
```
krasi-georgiev: capital letter and full stop for all comments :)

jacksontj:
@krasi-georgiev that should do it.

krasi-georgiev:
Thanks!

PR #604, opened 2019-06-14
Add /labels API to client

jacksontj:
API ref https://prometheus.io/docs/prometheus/latest/querying/api/#getting-label-names

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -124,6 +124,7 @@ const (
 	epAlertManagers   = apiPrefix + "/alertmanagers"
 	epQuery           = apiPrefix + "/query"
 	epQueryRange      = apiPrefix + "/query_range"
+	epLabels          = apiPrefix + "/labels"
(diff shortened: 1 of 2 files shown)


beorn7:
Note: I plan to cut v1 today. Would be cool to get this in if it is ready for it.

krasi-georgiev:
Thanks!

jacksontj:
@beorn7 as long as we aren't moving this client to GA yet. IMO https://github.com/prometheus/prometheus/issues/5661 should get done first -- so we don't have to do another interface break.

beorn7:
@jacksontj What is GA?
I would like to do v1 of client_golang very soon and not wait for Prometheus v2.11.
The API client will be explicitly excluded from semver guarantees.

jacksontj:
@beorn7 I was referring to making this API client "GA" (meaning it follows the semver stuff) -- I wasn't sure what all was included in a v1 release. So if this API client remains as "experimental" (so we can break the API as needed), I'm fine.

PR #606, opened 2019-06-14
Update doc comment of NewProcessCollector

beorn7:
- Now also works on MS Windows. Thanks @carlpett.
- The hints for updating from older versions is obsolete by now.

@carlpett we totally forgot to update the doc comment.


carlpett:
Whoops. Good catch!

brian-brazil:
:+1:

Issue #605, opened 2019-06-14
Wrapped Register does not play well with AlreadyRegisteredError


lukedirtwalker:
We use abad2d1bd44235a26707c172eab6bca5bf2dbad3 but I checked the code and the problem is still in 1335ef46bddd1cb120c21240851f292904aa91f6

*Problem:* If you register something in a wrapped register it can't be recovered from an AlreadyRegisteredError

```go
        // set default registry
        labels := prometheus.Labels{"foo": "bar"}
	reg := prometheus.NewRegistry()
	prometheus.DefaultRegisterer = prometheus.WrapRegistererWith(labels, reg)
	prometheus.DefaultGatherer = reg

	g := prometheus.NewGauge(prometheus.GaugeOpts{
		Name:        "mygague",
		Help:        "Help for the gauge",
	})
```
Now check for already registered error according to example in GoDoc
```go
	if err := prometheus.Register(g); err != nil {
		if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
			g = are.ExistingCollector.(prometheus.Gauge)
		} else {
			panic(err)
		}
	}
```
Result: `Panic msg="interface conversion: *prometheus.wrappingCollector is not prometheus.Gauge: `

The user of the API has no possibility to cast to wrappingCollector and extract the underlying collector.

So either the doc should discourage the usage of `AlreadyRegisteredError` when using a wrapped registry or we should make the underlying collector somehow accessible.

beorn7:
Thanks for the catch. I'll think about it if we can unwrap the original collector.

PR #607, opened 2019-06-14
Make the AlreadyRegisteredError useful for wrapped registries

beorn7:
@lukedirtwalker please have a look. This should fix your use case.

### DIFF
--- a/prometheus/registry_test.go
+++ b/prometheus/registry_test.go
@@ -746,37 +746,120 @@ func BenchmarkHandler(b *testing.B) {
 }
 
 func TestAlreadyRegistered(t *testing.T) {
-	reg := prometheus.NewRegistry()
 	original := prometheus.NewCounterVec(
 		prometheus.CounterOpts{
-			Name: "test",
-			Help: "help",
+			Name:        "test",
+			Help:        "help",
+			ConstLabels: prometheus.Labels{"const": "label"},
 		},
 		[]string{"foo", "bar"},
 	)
 	equalButNotSame := prometheus.NewCounterVec(
 		prometheus.CounterOpts{
+			Name:        "test",
+			Help:        "help",
+			ConstLabels: prometheus.Labels{"const": "label"},
+		},
+		[]string{"foo", "bar"},
+	)
+	originalWithoutConstLabel := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
 			Name: "test",
 			Help: "help",
 		},
 		[]string{"foo", "bar"},
 	)
-	var err error
-	if err = reg.Register(original); err != nil {
-		t.Fatal(err)
+	equalButNotSameWithoutConstLabel := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "test",
+			Help: "help",
+		},
+		[]string{"foo", "bar"},
+	)
+
+	scenarios := []struct {
+		name              string
+		originalCollector prometheus.Collector
+		registerWith      func(prometheus.Registerer) prometheus.Registerer
+		newCollector      prometheus.Collector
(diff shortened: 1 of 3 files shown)


beorn7:
This fixes #605

prometheus/registry_test.go
```diff
@@ -746,38 +746,153 @@ func BenchmarkHandler(b *testing.B) {
...
-	}
+		if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
+			if are.ExistingCollector != original {
+				t.Error("expected original collector but got something else")
+			}
+			if are.ExistingCollector == equalButNotSame {
+				t.Error("expected original collector but got new one")
+			}
+		} else {
+			t.Error("unexpected error:", err)
```
lukedirtwalker: Why are you using t.Error here. I think this is never expected so we could just use fatal?

prometheus/registry_test.go
```diff
@@ -746,38 +746,153 @@ func BenchmarkHandler(b *testing.B) {
...
+			t.Fatal("expected error when registering equal collector")
 		}
-	} else {
-		t.Error("unexpected error:", err)
-	}
+		if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
+			if are.ExistingCollector != original {
+				t.Error("expected original collector but got something else")
+			}
+			if are.ExistingCollector == equalButNotSame {
```
lukedirtwalker: I think I would just have that included in the error above
```go
t.Errorf("expected original collector but got %v", are.ExiststingCollector)
```
(or even use Fatalf)

prometheus/registry_test.go
```diff
@@ -746,38 +746,153 @@ func BenchmarkHandler(b *testing.B) {
...
-		if are.ExistingCollector == equalButNotSame {
-			t.Error("expected original callector but got new one")
+
+		if err = reg.Register(equalButNotSame); err == nil {
+			t.Fatal("expected error when registering equal collector")
 		}
-	} else {
-		t.Error("unexpected error:", err)
-	}
+		if are, ok := err.(prometheus.AlreadyRegisteredError); ok {
```
lukedirtwalker: This if/else block is almost the same for all test maybe add a function to check it, something like `func checkAlreadyRegisteredError(t *testing.T, err error, original, equalButNotSame Collector) {...}`

prometheus/registry_test.go
beorn7: It's definitely fatal as in "the test is broken". But I only use t.Fatal in tests if the rest of the test wouldn't make sense to run anymore. However, if the rest of the test can still in principle work and tell me about other broken things, I use t.Error so that it continue to runs.

In this case, the Error is the last thing in the (sub-)test, so the difference doesn't really matter. It's more that I personally default to t.Error for the reasons above.

lukedirtwalker: Makes sense yeah. :+1:

prometheus/registry_test.go
beorn7: The %v representation of a collector can look a bit daunting. As this is an `==` test, you'll see from the line where the test fails what precise collector was returned. My take is that this is slightly clearer than dumping a possibly complex text representation to the test output (which might have the right hint for the qualified reader, but will probably require to look at the test code anyway).

prometheus/registry_test.go
beorn7: The whole test could (or should) be transformed into a table based one, filling in the registries used for registering, re-registering, and the CounterVecs to register and reregister.

In general, I don't mind code duplication in tests if it makes things very explicit. But perhaps at this point, the table approach is even easier to grok.

I'll work on it...

beorn7: The test is now table-driven.

PR #613, opened 2019-06-28
Add Delete method to Pusher

### LINKED ISSUE #611: Support for DELETE calls to PushGateway
Is there any plan to support the DELETE API for the PushGateway?
With the package as-is, we are able to push metrics to the PushGateway, but not able to clean them up if they are not intended to be overwritten.

beorn7:
Fixes #611.

@nnoto what do you think?

This allows
```Go
push.New("http://localhost:9091", "some_job").Grouping("foo", "bar").Delete(),
```


nnoto:
Looks good to me, much appreciated!

beorn7:
OK, I'll merge. You can use it by building from head. I'll see what else is in the pipeline to decide when to cut a release.

Issue #614, opened 2019-07-02
API client Series() call can't handle minTime/maxTime set by prometheus


jacksontj:
This is a bit of a convoluted error so I'll do my best to explain.

If an API call to `/series` is made without a start/end time prometheus sets minTime/maxTime as the value (https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L462). If I then turn around and use this time with the API client (to presumably send the same request) I get an error:

```
{
"status": "error",
"errorType": "execution",
"error": "bad_data: cannot parse \"292277025-08-17T23:12:54.999-08:00\" to a valid timestamp"
}
```

This is because the times used for minTime/maxTime can't be formatted to an RFC3339Nano timestamp. So for my use case in [promxy](https://github.com/jacksontj/promxy) I get a call from grafana (in this example) with no start/end time -- so prometheus sets the time to these min/max times but then I cannot pass that time down to another prometheus to *fetch* the data, as its not a valid timestamp. 

Go playground showing the behavior: https://play.golang.org/p/hM_DWyv0JMK

So I'm not sure what we want to do exactly to handle this. Locally I have copied the min/max times into the client package and are optionally *not* setting the start/end times if they match. I'm not a major fan of the copy/paste of vars, but unfortunately those times are also private in the prometheus api/v1 package (https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L441). Alternatively we could make it so the client doesn't set the start/end if start/end are `time.IsZero()` which would allow the client/caller to pick when they don't send the timestamp (instead of us transparently doing it in the client).

brian-brazil:
That sounds like a bug on the Prometheus end, however I also recall that Go has some limits in terms of what it can handle time wise.

jacksontj:
@brian-brazil Yea, I could go either way on this TBH (client or server issue). If we were to tackle this server-side then IMO we'd want to pass some other time down the stack (presumably min/max that can be represented in what the API requires) -- right now the time we set is not expressible in the format the API requires.

beorn7:
@krasi-georgiev assigning to you as the API expert. Perhaps you have an opinion/idea?

brian-brazil:
Something else that occurs to me is that if this is something that could also affect other languages, then we should probably fix it on the server.

krasi-georgiev:
+1 on fixing server side

jacksontj:
What fix do we want server side? Should I just change the min/max time to
[...]

PR #617, opened 2019-07-07
Change all time formatting to UTC and off of time.RFC3339Nano

jacksontj:
Prometheus has issues parsing RFC3339Nano timestamps if the year has more than 4 digits, in addition it is the second-pass parse attempt. Since this is a client library and the interface is a `time.Time` it makes sense that we pick the clearest simplest format-- so I propose we use the `model.Time` representation of time in our communications to prometheus. This (1) removes the issues with timezones in those queries going downstream and (2) completely works around this #614 issue as the parsing mechanism in prometheus can handle those times in this format.

Related to #614

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -678,7 +678,7 @@ func (h *httpAPI) QueryRange(ctx context.Context, query string, r Range) (model.
 	q.Set("query", query)
 	q.Set("start", formatTime(r.Start))
 	q.Set("end", formatTime(r.End))
-	q.Set("step", strconv.FormatFloat(r.Step.Seconds(), 'f', 3, 64))
+	q.Set("step", strconv.FormatFloat(r.Step.Seconds(), 'f', -1, 64))
 
 	_, body, warnings, err := api.DoGetFallback(h.client, ctx, u, q)
 	if err != nil {
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -594,8 +594,8 @@ func (h *httpAPI) DeleteSeries(ctx context.Context, matches []string, startTime 
 		q.Add("match[]", m)
 	}
 
-	q.Set("start", startTime.Format(time.RFC3339Nano))
-	q.Set("end", endTime.Format(time.RFC3339Nano))
+	q.Set("start", formatTime(startTime))
+	q.Set("end", formatTime(endTime))
 
 	u.RawQuery = q.Encode()
 
@@ -659,7 +659,7 @@ func (h *httpAPI) Query(ctx context.Context, query string, ts time.Time) (model.
 
 	q.Set("query", query)
 	if !ts.IsZero() {
-		q.Set("time", ts.Format(time.RFC3339Nano))
+		q.Set("time", formatTime(ts))
 	}
 
 	_, body, warnings, err := api.DoGetFallback(h.client, ctx, u, q)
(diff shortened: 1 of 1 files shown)


krasi-georgiev:
As discussed this would be fixed on the server side so that clients in other languages don't hit the same problem.

brian-brazil:
I think using unixtime makes more sense, and greatly reduces the chances we'll hit issues compared to a more human readable format.

We're trying to kill off model, so please avoid increasing usage of it.

jacksontj:
I understand the want to reduce model usage bit this clients API is already based on it, so at this point removal of common from this specific client is a significantly larger project.

brian-brazil:
I'm not saying to remove existing uses right now, I'm saying let's not add more.

jacksontj:
That change seems entirely pointless to me. At this point there are already 88 occurances of `model.*` in this package-- why would I take the time to copy/paste that single struct over (which BTW is also used in all the `model.Value` etc.) to remove 1 occurance? At this point that seems like wasted effort, not to mention duplicating the code (since we'd use both `model.TIme` and a copy of it in the same package).

beorn7:
@brian-brazil:
> I think using unixtime makes more sense,

Do I understand this correctly that you are in general in favor of this change?

@jacksontj: `return model.TimeFromUnixNano(t.UTC().UnixNano()).String()` is just a very elaborate way of turning `t` into a string-formatted float with the unixtime as a value. I think what @brian-brazil wants is to simply replace that line with
```Go
return strconv.FormatFloat(float64(t.UnixNano())/1e9, 'f', -1, 64)
```
There is really no reason to involve any logic from any other Prometheus packages.

brian-brazil:
> Do I understand this correctly that you are in general in favor of this change?

Yes.

beorn7:
OK, cool. I'd say, let's just use the line I suggested above instead of the `model.TimeFromUnixNano` call, and then this is good to go. (Modulo one tiny note added below.)

api/prometheus/v1/api.go
```diff
@@ -675,16 +675,10 @@ func (h *httpAPI) QueryRange(ctx context.Context, query string, r Range) (model.
...
-		step  = strconv.FormatFloat(r.Step.Seconds(), 'f', 3, 64)
-	)
-
 	q.Set("query", query)
-	q.Set("start", start)
-	q.Set("end", end)
-	q.Set("step", step)
+	q.Set("start", formatTime(r.Start))
+	q.Set("end", formatTime(r.End))
+	q.Set("step", strconv.FormatFloat(r.Step.Seconds(), 'f', 3, 64))
```
beorn7: Not really related to this change, but I'm wondering if there was a reason to use `3` here instead of `-1`. We do the same for the timestamp formatting now, so perhaps adjust it to `-1`, now that we are on it?

jacksontj: No reason, it was just that already (I just changed the format to match the other methods). Updated :)

jacksontj:
@beorn7 PR updated with comments, should be ready for merge

krasi-georgiev:
yep all good.

beorn7:
Thanks everyone!

jacksontj:
@beorn7 seems that the formatTime function you have me doesn't quite work

Doing some local testing this is the output of:
```
	fmt.Println(startTime, formatTime(startTime))
	fmt.Println(endTime, formatTime(endTime))
```
```
-292273086-05-16 16:47:06 +0000 UTC 6795364580.679345
292277025-08-18 07:12:54.999 +0000 UTC -6795364579.680346
```

jacksontj:
Here's the same on go playground: https://play.golang.org/p/qMLQcs_WG2J

jacksontj:
#619 has the fix :)

beorn7:
Ooof, yeah, of course, because UnixNano is rather limited in its time range.
(Would have been the same problem if using the `model` package.)
Thanks for the fix.

PR #624, opened 2019-07-22
Support new base64 encoding for pushing to the Pushgateway

beorn7:
This should only be released after PGW 0.9 is released.

Interestingly, the code had a bug so far as it would neither properly URL-encode label values with special characters nor reject them (as long as none of the special characters was a `/`). Nobody noticed so far…

### DIFF
--- a/prometheus/push/push.go
+++ b/prometheus/push/push.go
@@ -36,6 +36,7 @@ package push
 
 import (
 	"bytes"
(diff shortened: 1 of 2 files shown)


beorn7:
Also @Sinjo and @dmagliola (as inspiration for the Ruby client ;o).

brian-brazil:
:+1:

PR #627, opened 2019-07-30
Update .gitignore

beorn7:
Inspired by
https://github.com/prometheus/common/pull/202

A lot of entries in the old file were from times where we used cgo and had a completely different build system.

@tariq1890


tariq1890:
LGTM

beorn7:
Thanks @tariq1890 .

PR #628, opened 2019-07-31
add minor Dockerfile improvements

tariq1890:


Dockerfile
```diff
@@ -4,7 +4,7 @@
 #    docker build -f examples/$name/Dockerfile -t prometheus/golang-example-$name .
 
 # Builder image, where we build the example.
-FROM golang:1 AS builder
+FROM golang:1.12 AS builder
```
tariq1890: I don't think it's a good idea to pin to golang 1 as breaking changes could happen once golang 1.13 is out.

Dockerfile
```diff
@@ -15,9 +15,9 @@ WORKDIR /go/src/github.com/prometheus/client_golang/examples/simple
 RUN CGO_ENABLED=0 GOOS=linux go build -a -tags netgo -ldflags '-w'
 
 # Final image.
-FROM prom/busybox
+FROM quay.io/prometheus/busybox:latest
```
tariq1890: Looking at prometheus/busybox repo, quay seems to be current hosted registry for the busybox image.

Dockerfile
simonpasquier: at the same time, client_golang should always work with the latest stable version of Go.

Dockerfile
simonpasquier: correct

Dockerfile
beorn7: And the dockerfile is only for the examples. I don't expect any problems with any Go 1.x version, given the Go stability guarantees.

beorn7: You could even turn the argument around: We deliberately want to build the examples with the latest Go version. If something doesn't work, that's a bug.

tariq1890: Makes sense. I have fixed that.

tariq1890:
Np :)

PR #630, opened 2019-08-08
Provide godoc examples for using the api

### LINKED ISSUE #194: documentation for API client

csmarchbanks:
Closes #194

cc: @krasi-georgiev

### DIFF
--- a/api/prometheus/v1/example_test.go
+++ b/api/prometheus/v1/example_test.go
@@ -1,0 +1,106 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package v1_test provides examples making requests to Prometheus using the
+// Golang client.
+package v1_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/prometheus/client_golang/api"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+)
+
+func ExampleAPI_Query() {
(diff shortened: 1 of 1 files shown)


api/prometheus/v1/example_test.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	result, warnings, err := api.QueryRange(ctx, "rate(prometheus_tsdb_head_samples_appended_total[5m])", r)
+	if err != nil {
+		fmt.Printf("Error querying Prometheus: %v\n", err)
+		os.Exit(1)
+	}
+	if len(warnings) > 0 {
+		fmt.Printf("Warnings: %v\n", warnings)
+	}
+	fmt.Printf("Result:\n%v\n", result)
```
krasi-georgiev: wondering whether it is a good idea to add a comment for the exepcted output like in the examples:

https://blog.golang.org/examples
// Output: golly

api/prometheus/v1/example_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"os"
+	"time"
+
+	"github.com/prometheus/client_golang/api"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+)
+
+func ExampleAPI_Query() {
+	client, err := api.NewClient(api.Config{
+		Address: "http://demo.robustperception.io:9090",
```
krasi-georgiev: Just tried to open it and it is down. 
A mock client would be ok, but would make the example a bit strange.

api/prometheus/v1/example_test.go
csmarchbanks: I considered having an Output section, but we don't know exactly what the Output would be, and if it is wrong the tests will fail.

We could make a mock Prometheus, where we know the response, but I agree with your comment below that the example would be a bit strange in that case.

api/prometheus/v1/example_test.go
csmarchbanks: As I said in the other comment, I agree that a mock client is weird for an example. Since these are just being compiled and tests don't depend on them passing perhaps it is ok for robustperception to be down occasionally?

krasi-georgiev: Yeap, agree , it is a good start anyway!

krasi-georgiev:
LGTM

PR #634, opened 2019-08-19
Add WriteHeader call to Flush

beorn7:
Flush is another of the methods that will call WriteHeader if it
hasn't happened yet. Since we want to call observeWriteHeader (if
set), we need to do the WriteHeader call already here, similar to what
we have done in Write and ReadFrom.

This commit also adds comments explaining the above to not tempt
developers to remove the WriteHeader call.


beorn7:
It really depends on the implementation. The implementation I looked at in the standard library (https://golang.org/src/net/http/server.go) do the same thing, i.e. they have an unexported tracking variable `wroteHeader`. Since it is unexported, the delegator we are implementing here cannot access it and we have to do our own tracking.

PR #636, opened 2019-09-01
Truncate time to avoid rounding issues on prometheus

jacksontj:
Upstream issue: https://github.com/prometheus/prometheus/issues/5972

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -873,5 +873,6 @@ func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, [
 }
 
 func formatTime(t time.Time) string {
+	t = t.Truncate(time.Nanosecond*1e6)
 	return strconv.FormatFloat(float64(t.Unix())+float64(t.Nanosecond())/1e9, 'f', -1, 64)
 }


api/prometheus/v1/api.go
```diff
@@ -873,5 +873,6 @@ func (c apiClient) Do(ctx context.Context, req *http.Request) (*http.Response, [
 }
 
 func formatTime(t time.Time) string {
+	t = t.Truncate(time.Nanosecond*1e6)
```
beorn7: `go fmt` (at least the most recent version) wants blanks around the `*`.

brian-brazil:
I don't think we should be changing the user request like this in the API library, and as noted in the upstream issue there's no bug here - and this likely counts as a breaking change. If you try to integrate two systems that aren't designed to be integrated you can't be surprised when there's oddities.

beorn7:
Ball seems to have been dropped here. @krasi-georgiev @brian-brazil any chance your combined wisdom can converge on how to solve this (what's in this PR plus prometheus/prometheus#5972)?

krasi-georgiev:
I would go with what @brian-brazil  decides as I don't have any experience with how is this api used.

brian-brazil:
Overall I think this is a GIGO situation. If a use case needs specific rounding, it should probably do it on their end. Erroring out on our end would likely break a lot of people (and there's cases like InfluxDB which may have more than ms precision).

beorn7:
@jacksontj It looks there is no real appetite for changing the current behavior. I hope you'll understand.

I'll close this and the related https://github.com/prometheus/prometheus/issues/5972 . I guess it cannot harm to document the rounding behavior of the Prometheus API (for which a PR will be welcome).

PR #637, opened 2019-09-03
Add support for HTTP authenticated Prometheus endpoints

sjug:
If we secure the Prometheus endpoint we'd still like to be able to access it using the API.


brian-brazil:
Why not pass in your own round tripper?

sjug:
@brian-brazil Fair point, removed the InsecureRoundTripper. Will rename PR as well.

brian-brazil:
The RoundTripper can handle auth too.

PR #644, opened 2019-09-25
process_collector: Account for recent Windows API changes

zx2c4:
Recent necessary changes mean that GetCurrentProcess() no longer
returns an error. And actually, it now returns a simple typed constant
(-1), which means that by using it directly, the Go compiler will inline
all uses of it into essentially a no-op. This commit updates the code
for this API change.

cc @scudette


beorn7:
Thanks for this PR. However, we already have #643 for that (and concluded over there that we need build tags to make sure the library works on both older Go versions and the current one). I'll close this in lieu of #643.

PR #647, opened 2019-10-01
remove duplicated assignment in v1/api.go

beautytiger:


beorn7:
@krasi-georgiev should that assignment have been anything else?

For now, I'm merging this as it doesn't change behavior.

krasi-georgiev:
Yeah I don't know, but it looks like a copy and paste error.

PR #655, opened 2019-10-14
Unflake TestGoCollectorGoroutines

### LINKED ISSUE #653: Flaky test again…
```
--- FAIL: TestGoCollectorGoroutines (0.00s)
    go_collector_test.go:78: want 1 new goroutine, got -1
```

Seems to affect Go1.13 in particular.

beorn7:
This is not a great solution, but it's also hard to test for this
moving target.

Fixes #653 .


beorn7:
Merging this as the flakiness for Go 1.13 is so bad that CI fails quite often.
It's only a test. Concerns can be addressed later, if there are any.

Issue #633, opened 2019-08-13
Potential for false AlreadyRegistered errors


awilliams:
This issue describes the potential for false [`AlreadyRegisteredError`](https://godoc.org/github.com/prometheus/client_golang/prometheus#AlreadyRegisteredError) errors when registering collectors.

The below test case is a simplified version taken from a real-world application where registration of collectors incorrectly failed with `duplicate metrics collector registration attempted` errors, even though they were all unique. The test case attempts to shows the likelihood of these false errors and also the underlying uniqueness algorithm.

A potential fix would be to change how [`collectorID`](https://github.com/prometheus/client_golang/blob/4efc3ccc7a660996803fff4721a2643b33dc3f77/prometheus/registry.go#L301) is calculated. Instead of summing the descriptor hashes, the fnv hashing function (same as used by the descriptors) could be used to create a hash of the descriptor hashes. I'm not sure why summing the hashes leads to collisions, but experimentally have seen that hashing the hashes does not have the same weakness.

### [Go Playground Test Cases](https://play.golang.org/p/HpV7YE6LI_4)

Related issues: #222

Version of Prometheus: `master` at the time of writing ([`4efc3ccc7a660996803fff4721a2643b33dc3f77`](https://github.com/prometheus/client_golang/tree/4efc3ccc7a660996803fff4721a2643b33dc3f77))

beorn7:
Yeah, I was naively hoping this would never became a problem in practice. Thanks for proving me wrong. (Meant seriously, not sarcastically.)

In v2 of this library, I'm pretty sure we'll simply drop hashing here at all. It's probably not really worth it as this code path is unlikely to be performance critical. As hinted at in #222.

Hashing the hashes would be an easy improvement, but the fnv hash function is not very strong anyway (and our original sin was to ever use it without hash collision detection in place).

I'm more inclined to use a stronger hash function (like the one used in prometheus/prometheus now, which is even faster than fnv if the hardware supports it). Or somehow kick the hashing out in v1 already.

awilliams:
Thanks @beorn7 for the insight.

We have come up with a workaround for our particular case. I'll mention it below in case it's useful for anyone. 

Using names from the included [test case](https://play.golang.org/p/HpV7YE6LI_4), instead of having `myCollector` implement `prometheus.Collector`, we were instead added the following methods:

```go
type myCollector struct {
    a, b, c, d prometheus.Collector
}

func (m *myCollector) Register(r prometheus.Registerer) error {
[...]

PR #657, opened 2019-10-14
Make hash collisions in the registry much less likely

beorn7:
Sadly, that's not yet a complete solution (like the one planned for v2, see last item in #222). However, using xxhash for the registry improves things a lot. It also moves from summing to XOR'ing for the collectorID. (Hashing the hashes would be an option, too, but it would require sorting of the hashes. I didn't want to go down that rabbit hole and rather stick with a commutative and associative operation.)

(Interestingly, just moving from sum to XOR solved #633 already, but of course, other hash collisions might happen in practice despite the XOR. In fact, hash collisions might still occur. This PR is mostly there to make this sufficiently unlikely so that the world will survive until v2 is released.)


beorn7:
@awilliams this will be of interest for you.

beorn7:
Id' consider this a fix for #633. The "proper" solution is contained in #222.

awilliams:
Thanks @beorn7, this looks like a great bridge between now and v2. Nice to see the collision test case as well!

beorn7:
Thanks for having a look.

prometheus/metric.go
```diff
@@ -24,6 +24,8 @@ import (
 
 const separatorByte byte = 255
 
+var separatorByteSlice = []byte{255} // For convenient use with xxhash.
```
cstyan: not that it's likely to change, but should this not be `[]byte{separatorByte}`

beorn7: Yeah, that's better. (Originally, I wanted to remove separatorByte altogether, but then realized I want to keep the old hashing for vectors.) Will sent you a PR.

PR #663, opened 2019-10-17
Fix collectorID calculation for Unregister()

### LINKED ISSUE #662: Regression: unregistering "builtin" collectors stopped working with 1.2.0
This simple test case showcases the problem:
```go
func TestRegisterUnregisterCollector(t *testing.T) {
	col := prometheus.NewGoCollector()

	reg.MustRegister(col)
	reg.Unregister(col)
	if metrics, err := reg.Gather(); err != nil {
		t.Error("error gathering sample metric")
	} else if len(metrics) != 0 {
		t.Error("should have unregistered metric")
	}
}
```
This passed up to v1.1.0, but fails on v1.2.0 (and master).

The calculation of collectorID got changed in 1.2.0 and the same change was not applied to unregistering. PR follows shortly.

costela:
This PR fixes the regression from 1.2.0 where the calculation of `collectorID` changed during `Register()`, but not during `Unregister()`

~It also brings the vec code closer to the rest by addopting `xxhash` everywhere.~ (reverted, see https://github.com/prometheus/client_golang/pull/663#issuecomment-543145728)

Finally, we also add a simple test case to ensure this doesn't happen again! :wink:  

fixes #662

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -361,7 +361,7 @@ func (r *Registry) Unregister(c Collector) bool {
 	var (
 		descChan    = make(chan *Desc, capDescChan)
 		descIDs     = map[uint64]struct{}{}
-		collectorID uint64 // Just a sum of the desc IDs.
+		collectorID uint64 // All desc IDs XOR'd together.
 	)
 	go func() {
 		c.Describe(descChan)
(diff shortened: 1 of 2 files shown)


beorn7:
Thanks for the fix of unregister.

However, the vec part should stay with fnv64a. See commit description of https://github.com/prometheus/client_golang/pull/657/commits/ee1078a03c073762e2660d305738bfdafbf69ba6

costela:
@beorn7 got it. Will revert that part.

costela:
Done! :heavy_check_mark:

Issue #661, opened 2019-10-17
go get fails for xxhash/v2


sepastian:
I am following the guide for `client_golang`: https://prometheus.io/docs/guides/go-application/

Installing `promhttp` with `go get` fails because of an invalid dependency on `xxhash/v2`.

```
$ go get github.com/prometheus/client_golang/prometheus/promhttp
package github.com/cespare/xxhash/v2: cannot find package "github.com/cespare/xxhash/v2" in any of:
	/home/sebastian/.asdf/installs/golang/1.10/go/src/github.com/cespare/xxhash/v2 (from $GOROOT)
	/home/sebastian/Code/sipts/worker-base/src/github.com/cespare/xxhash/v2 (from $GOPATH)

$ go version
go version go1.10 linux/amd64
```

The URL https://github.com/cespare/xxhash/v2 is invalid; should this read https://github.com/cespare/xxhash/tree/v2 instead?

beorn7:
This is a change introduced by go modules. To make it work with older Go versions, those got patched. You need to run go go 1.10.3+.

I'll add that to the README.md.

beorn7:
Now addressed in README.md.

sepastian:
I tried with go 1.13.1 yesterday, it works there. Thanks!

PR #664, opened 2019-10-17
Add more precise Go version requirements.

beorn7:
@sepastian

“Fixes” #661


beorn7:
Merging as it is just documentation, and this should make it into v1.2.1.

PR #665, opened 2019-10-17
Cut v1.2.1

beorn7:
@costela here we go.


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.2.1 / 2019-10-17
+
+* [BUGFIX] Fix regression in the implementation of `Registerer.Unregister`.
```
costela: shouldn't we mention the issue ID?
```suggestion
* [BUGFIX] Fix regression in the implementation of `Registerer.Unregister`. #662
```

beorn7: Yes, indeed. Or to be more precise: The PR. (I prefer linking the PR as the more interesting place to look at. Both PR and issue are linked to each other anyway.)

PR #666, opened 2019-10-18
Remove old reference to basicMetricVec

jeffwidman:
Contrary to the code comment, I see no `basicMetricVec` implementation. Grep'ing this project shows this is the only reference. So I suspect it's an outdated comment and can be removed to minimize confusion.

I'm unclear whether other parts of that comment are also incorrect and need updating.


beorn7:
Yes, that's just a remnant. Thanks for spotting it.
Could you add the [DCO](https://github.com/prometheus/client_golang/pull/666/checks?check_run_id=264971059)?

beorn7:
The remaining comment is up to date, AFAICS.

jeffwidman:
Sounds good, added the DCO

PR #668, opened 2019-10-22
Send Graphite metrics with tags

kamaev:
Hello! I have added some changes to the Bridge struct 
and related methods to make it possible to use Graphite tags


kamaev:
ooops, forgot to remove one thing..

beorn7:
Thanks. Will review ASAP. @stuartnelson3 as the author of the bridge, perhaps you can have a look, too.

PR #687, opened 2019-11-29
Provide a mechanism to reset all metrics for a Gatherer

stristr:
@beorn7 
This PR adds the ability for consumers to reset all metrics—e.g. in a test or test suite where we're testing the values captured by a GaugeVec, we can call `prometheus.DefaultGatherer.ResetMetrics()`.

I wasn't sure if this was a worthy enough use case for adding to the `Gatherer` interface, but it seemed to fit well semantically. For the most common `prometheus.DefaultGatherer.ResetMetrics()` use case, we do need access to the `Registry`'s mutex to ensure correctness.

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -158,6 +158,10 @@ type Gatherer interface {
 	// expose an incomplete result and instead disregard the returned
 	// MetricFamily protobufs in case the returned error is non-nil.
 	Gather() ([]*dto.MetricFamily, error)
+
+	// ResetMetrics restores all registered metric Collectors to their initial
+	// state.
+	ResetMetrics()
 }
 
 // Register registers the provided Collector with the DefaultRegisterer.
@@ -192,6 +196,12 @@ type GathererFunc func() ([]*dto.MetricFamily, error)
 // Gather implements Gatherer.
 func (gf GathererFunc) Gather() ([]*dto.MetricFamily, error) {
 	return gf()
+}
+
+// ResetMetrics implements Gatherer.
+func (gf GathererFunc) ResetMetrics() {
+	// GathererFunc's state is determined by its value, so resetting its
+	// metrics has no effect.
 }
 
 // AlreadyRegisteredError is returned by the Register method if the Collector to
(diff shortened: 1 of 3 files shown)


beorn7:
Thanks for your PR and your explanation.

I don't think we should add anything to the `Gatherer` interface purely for test purposes. This library already has way more and fatter types than I'm comfortable with.

The other problem is that the reset only works for collectors that happen to implement the `Reset` method, most likely only the provided `...Vec` types. However, in real-world scenarios, you'll often have custom `Collector` implementations that might wrap `...Vec` types or that simply collect metrics equivalent to a `...Vec` type, but don't have a `Reset` method. That's just getting very confusing.

In tests, I would strongly prefer to stay in full control. If I really just want to reset the `...Vec` instances I have created, it's easy to do so. In general cases, I would probably let each test work on a newly created registry to start with a ensured clean slate.

stristr:
Thanks for the quick response! This makes sense to me. In my use case, I'm mainly using built-in `*Vec` types registered with `promauto` in `init()` functions. I essentially am looking for a way to integration test the metrics without (a) reinitializing the entire app or (b) caring which order tests execute in.

It's reasonable enough to move the `init()` code into a public method I can call from tests, but I would like to keep using the convenience functions without delving into the implementation details of the client library if possible…what would you think about an API for restoring the client library default registry to its initial state?

Something like this:

```diff
diff --git a/prometheus/registry.go b/prometheus/registry.go
index b412cf7..d2c8716 100644
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -52,12 +52,18 @@ const (
 // consequences. Users who want to avoid global state altogether should not use
 // the convenience functions and act on custom instances instead.
 var (
-	defaultRegistry              = NewRegistry()
-	DefaultRegisterer Registerer = defaultRegistry
-	DefaultGatherer   Gatherer   = defaultRegistry
+	DefaultRegisterer Registerer
+	DefaultGatherer   Gatherer
 )
 
 func init() {
+	InitializeDefaults()
+}
+
+func InitializeDefaults() {
+	r := NewRegistry()
+	DefaultRegisterer = r
+	DefaultGatherer = r
 	MustRegister(NewProcessCollector(ProcessCollectorOpts{}))
 	MustRegister(NewGoCollector())
 }
```

beorn7:
One of the reason to avoid global state is precisely that it is inconvenient for tests.

In general, I recommend to avoid the global registry and the `promauto` package in any non-trivial program. It has its place in very simply programs, but if you really need to reset vectors and such in your test cases, that's a clear sign that your program is non-trivial and you should avoid the global registry.

In particular, it would defeat the purpose of the simplicity of `promauto` and the global registry if we added non-trivial tooling to handle global state in non-trivial programs.

stristr:
Okay, that's a fair position to take. Thanks so much for talking through it with me!

PR #689, opened 2019-12-03
add default go collector and process collector var so that we can unregister them 

bj-wangjia:
Signed-off-by: bj-wangjia <jia.wang@mintegral.com>

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -52,14 +52,16 @@ const (
 // consequences. Users who want to avoid global state altogether should not use
 // the convenience functions and act on custom instances instead.
 var (
-	defaultRegistry              = NewRegistry()
-	DefaultRegisterer Registerer = defaultRegistry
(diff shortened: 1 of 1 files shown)


beorn7:
If you need a registry without or with custom process collector and go collector, I'd recommend to use a local registry.

The global registry is really only meant for simple cases where you are fine with the default behavior.

bj-wangjia:
This  patch just add two const var,  do not change the default behavior.
Developer can unregister the default  collector by the two var if they  need not them

beorn7:
Global state in Go programs is considered highly problematic. The reason we have some global state in this library is mostly pragmatic: For very simple programs, using the global state simplifies things, and you also don't run into the typical problems of global state, precisely because your program is simple.

As said, the recommendation for non-trivial programs is to not use global state at all, because of the huge potential of problems. Thus, I don't want to include tooling that mostly exist to support non-trivial use cases with the global registry.

Also, I don't want to add even more global state, which your PR does.

PR #692, opened 2019-12-06
ExampleInstrumentHandlerDuration: use “code” label

stapelberg:
The label name must be code, not handler. Without this change, InstrumentHandlerDuration panics.


beorn7:
The `handler` label and the missing of the `code` label is intentional. The `handler` label is curried away in 384 and 391 and should therefore not create a problem. If this code is really panicking, we need to find out what's wrong.

stapelberg:
Ah, nevermind then. I selectively copied from the example and entirely missed the currying bit. Sorry for the noise!

PR #698, opened 2019-12-11
remove "v2" in import from desc.go

wqc200:
xxhash has not package "v2"


beorn7:
This repository is using v2.x of xxhash. With the introduction of Go modules, this requires the `v2` in the import path. It works if you use a recent enough Go version. Even older Go versions were patched for that. The minimum required patch releases for older Go versions are Go1.9.7 and Go1.10.3, as stated in the README.md.

PR #699, opened 2019-12-11
Return Prometheus Warnings

joe-elliott:
This PR replaces #648.

- Moved warnings from the `api` package to `api/prometheus/v1`
- Split the `api.Client` interface.  Previously it was being used in both `api` and `v1`.  Now a separate interface (`v1.apiClient`) better defines this internal http client.
- Moved DoGetFallback into method into `v1`
- Moved tests as appropriate.
- Correctly returned warnings from `Do` and `DoGetFallback`.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -25,8 +25,6 @@ import (
 	"time"
 )
 
-type Warnings []string
-
 // DefaultRoundTripper is used if no RoundTripper is set in Config.
 var DefaultRoundTripper http.RoundTripper = &http.Transport{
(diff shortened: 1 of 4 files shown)


beorn7:
Many thanks. I'll have a look ASAP.

@krasi-georgiev in case you find time: You might have some context here. If you see anything we haven't seen, please let us know.

krasi-georgiev:
sorry I wouldn't be able to have a look anytime soon.

PR #703, opened 2020-01-09
Added `testutil.CollectAndCount`

bwplotka:
:wave: 

I would like to propose additional function to `testutil.go` which will count metrics in a given Collector. The example usage can be seen here: https://github.com/thanos-io/thanos/blob/d758432436718ff6cd0d9136bc53653f4ec5cf47/pkg/compact/compact_e2e_test.go#L179

I also read the package documentation and I think I can agree that mock would be preferred for all of those test assertions IF such mock would be ready to use somewhere AND it would be easy to inject those mocked counter, gauges etc. The latter might be even impossible given the `*Vec` metrics are not interfaces e.g:
```
func NewCounterVec(opts CounterOpts, labelNames []string) *CounterVec {
```

Overall I think, with this small function `testutil` gives all you need to unit test your instrumentation without mocks, except maybe registering part, which is easier to mock.

### DIFF
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -108,6 +108,33 @@ func ToFloat64(c prometheus.Collector) float64 {
 	panic(fmt.Errorf("collected a non-gauge/counter/untyped metric: %s", pb))
 }
 
+// CollectAndCount collects all Metrics from the provided Collector and returns their number.
+//
+// This can be used to assert the number of metrics collected by a given collector after certain operations.
+//
+// This function is only for testing purposes, and even for testing, other approaches
+// are often more appropriate (see this package's documentation).
+func CollectAndCount(c prometheus.Collector) int {
+	var (
+		mCount int
+		mChan  = make(chan prometheus.Metric)
+		done   = make(chan struct{})
+	)
+
+	go func() {
+		for range mChan {
+			mCount++
+		}
+		close(done)
+	}()
+
+	c.Collect(mChan)
+	close(mChan)
+	<-done
+
+	return mCount
+}
+
 // CollectAndCompare registers the provided Collector with a newly created
 // pedantic Registry. It then does the same as GatherAndCompare, gathering the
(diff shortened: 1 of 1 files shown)


prometheus/testutil/testutil.go
```diff
@@ -108,6 +108,34 @@ func ToFloat64(c prometheus.Collector) float64 {
 	panic(fmt.Errorf("collected a non-gauge/counter/untyped metric: %s", pb))
 }
 
+// MetricCount collects all Metrics from the provided Collector and counts them.
```
beorn7: Suggestion: MetricCount collects all Metrics from the provided Collector and returns their number.

prometheus/testutil/testutil.go
```diff
@@ -108,6 +108,34 @@ func ToFloat64(c prometheus.Collector) float64 {
...
+//
+// Typical use case is when you want to assert certain number of series of given collector after
+// some operation.
```
beorn7: Suggestion: This can be used to assert the number of metrics collected by a given collector after certain operations.

prometheus/testutil/testutil.go
```diff
@@ -108,6 +108,34 @@ func ToFloat64(c prometheus.Collector) float64 {
...
+//
+// This function is only for testing purposes, and even for testing, other approaches
+// are often more appropriate (see this package's documentation).
+func MetricCount(c prometheus.Collector) int {
```
beorn7: Perhaps call it `CollectAndCount` so that it corresponds to `CollectAndCompare`?

bwplotka:
Thanks for speedy review @beorn7 - all applied. 

Side note: With interfaces, mock might be doable indeed. Do you think we should host such a prepared mock as well in v2 then? I think it would be useful.

beorn7:
> Do you think we should host such a prepared mock as well in v2 then? I think it would be useful.

That's the plan, at least vaguely.

PR #706, opened 2020-01-20
Add exemplars to counter and histogram

beorn7:
This still needs tests. I'd like to show it to you already, though, to find out if this is the right approach.

If it works for you, I'll add tests before merging this.


prometheus/counter.go
```diff
@@ -112,7 +128,23 @@ func (c *counter) Write(out *dto.Metric) error {
...
+	var exemplar *dto.Exemplar
+	if e := c.exemplar.Load(); e != nil {
+		exemplar = e.(*dto.Exemplar)
+	}
+
+	return populateMetric(CounterValue, val, c.labelPairs, exemplar, out)
+}
+
+func (c *counter) updateExemplar(v float64, l Labels) {
+	if len(l) == 0 {
```
brian-brazil: Empty exemplar labels are valid. You probably want to check for nil here instead.

prometheus/histogram.go
```diff
@@ -205,9 +215,10 @@ func newHistogram(desc *Desc, opts HistogramOpts, labelValues ...string) Histogr
 		}
 	}
 	// Finally we know the final length of h.upperBounds and can make buckets
-	// for both counts:
+	// for both counts as well es exemplars:
```
brian-brazil: as

prometheus/counter.go
beorn7: Good to know. Will adjust.

prometheus/histogram.go
beorn7: fixed.

beorn7:
Thanks, addressed both of the comments, but not yet the 64 character limit.

Is that just the length of the label names and values added up, or does it include the curly braces and quotes and commas and perhaps even escaping? Is it in runes or in bytes?

brian-brazil:
> Is that just the length of the label names and values added up, or does it include the curly braces and quotes and commas and perhaps even escaping? 

Just the names and values, before escaping.

> Is it in runes or in bytes?

Runes. Everything else in OM is about UTF-8 characters, so it'd seem odd to use bytes here.

beorn7:
Thanks. Working on implementing the limit…

beorn7:
Limit implemented. I assume the general direction is fine and will add tests now.

beorn7:
This is now ready for review.

prometheus/counter.go
```diff
@@ -40,6 +41,16 @@ type Counter interface {
 	// Add adds the given value to the counter. It panics if the value is <
 	// 0.
 	Add(float64)
+	// AddWithExemplar works like Add but also replaces the currently saved
+	// exemplar (possibly none) with a new one, created from the provided
```
brian-brazil: "(if any)" would be the usual way to word that

prometheus/counter.go
```diff
@@ -40,6 +41,16 @@ type Counter interface {
...
+	// value, the current time as timestamp, and the provided labels. Empty
+	// Labels will lead to a valid (label-less) exemplar. But if Labels is
+	// nil, the current exemplar is left in place. This method panics if the
+	// value is < 0, if any of the provided labels are invalid, or if the
+	// provided labels contain more than 64 runes in total. AddWithExemplar
```
brian-brazil: Mightn't hurt to put numbers on this. What you and I consider costly for instrumentation isn't the same as everyone else, given that e.g. sending a udp packet on every event or using client-side quantiles is common.

prometheus/counter.go
```diff
@@ -78,6 +89,9 @@ type counter struct {
 	desc *Desc
 
 	labelPairs []*dto.LabelPair
+	exemplar   atomic.Value // *dto.Exemplar
```
brian-brazil: full stop

prometheus/promhttp/http.go
```diff
@@ -163,22 +168,38 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
...
+			lastErr = err
+			if opts.ErrorLog != nil {
+				opts.ErrorLog.Println("error encoding and sending metric family:", err)
+			}
+			errCnt.WithLabelValues("encoding").Inc()
+			switch opts.ErrorHandling {
+			case PanicOnError:
+				panic(err)
+			case ContinueOnError:
+				// Handled later.
```
brian-brazil: Returning here would make more sense now with the refactoring.

prometheus/counter.go
beorn7: Yes, much better. Changed.

prometheus/counter.go
beorn7: Hmm, those numbers would be quite involved. It's not just runtime, but also that the creation of the Exemplar needs a number of allocations.

Re-reading the current wording of the doc comment, I find it now a bit confusing myself. It would probably best to explain the usual strategy for really hot paths (stochastically do some observations with and most without exemplar or even do some more sophisticated adaptive rate limiting), but then it would get way too involved for a doc comment.

I removed the sentence about the performance impact for now. This is after all still an experiment. (I'll mark it as such in the changelog.)

prometheus/counter.go
beorn7: Hmm, that looks weird as this comment is not in English but in Go.

To resolve the ambiguity here, I changed it to actual English: `Always a *dto.Exemplar.`

beorn7: Or more precisely: `Containing nil or a *dto.Exemplar.`

prometheus/promhttp/http.go
beorn7: I thought so, too, at first. But it would be redundant with the return in L191. I had it first there, and also had a `default` clause to explicitly handle all possibilities (which then doesn't require the final `return` in the function anymore), but then thought that's also redundant.

On the other hand, to really do the shortest possible code, the `ContinueOnError` could be completely omitted. I did that now, with a short comment.

prometheus/counter.go
brian-brazil: My thought was mainly to avoid scaring people away from using this for reasons that they don't actually care about. If they really care about a hot path, they'll probably have their own micro-benchmarks anyway.

PR #707, opened 2020-01-25
Cut v1.4.0

beorn7:
Note that adding a method to the `Histogram` and `Counter` interfaces is technically a breaking change. However, I cannot really think about any option that wouldn't be terribly convoluted. (It would be much easier if the constructor returned a concrete type that implements certain interfaces, but this mistake was made long ago.)

I would expect that very few will have implemented their own `Histogram` and `Counter`, and if they did, it will be easy to fix. So the impact will (hopefully) be quite limited.


brian-brazil:
:+1:

beorn7:
@bwplotka : Yes, your thoughts are precisely what I considered as a strictly compatible solution, but then I thought it's too clunky, at least if the issue we are avoiding by it is not very relevant. But I'm on the fence about it. Or in other words: I could easily be convince to do it that way, for example if there is a lot of backlash with this release. But perhaps we shouldn't even risk it.

A related point is that the methods of a `HistogramVec` and a `SummaryVec` to retrieve a child return an `Observer`, which is the interface shared by `Summary` and `Histogram`. But now the `Histogram` has an `ObserveWithExemplar`, but the `Summary` does not. So the moment a vector is in the game, you have to do the interface upgrade anyway (and that has already occurred [in practice](https://github.com/beorn7/common/blob/15f8ff97bd5206b72529a1d449e762ddc4669760/middleware/instrument.go#L58-L64) at Grafana).

I think I'll create a proposal PR with an `ObserverWithExemplar` and an `AdderWithExemplar` interface so that you can all look at it and decide what's the better trade-off.

bwplotka:
Perfect thanks :+1: I guess this means we wait with the release then.

beorn7:
I have rebased with #708 included.

The only `CHANGE` is now for the metric doc string, which is benign.

Please give final approvement or raise objections.

go.mod
```diff
@@ -4,11 +4,16 @@ require (
 	github.com/beorn7/perks v1.0.1
 	github.com/cespare/xxhash/v2 v2.1.1
 	github.com/golang/protobuf v1.3.2
-	github.com/json-iterator/go v1.1.8
+	github.com/google/go-cmp v0.4.0 // indirect
```
simonpasquier: I'm not sure why about the indirect dependencies. I think they shouldn't be there.

beorn7: They got introduced by the update of dependencies. `go mod tidy` doesn't remove them.

It seems that's the way it has to be.

Do you know a way I could try to get rid of them?

PR #708, opened 2020-01-25
Pull out ...WithExemplar methods into separate interfaces

beorn7:
This is, sadly, the only way to avoid a breaking change. The cost is
that anyone using exemplars has to perform a type assertion. This is,
however, a common pattern where interfaces turn out to need additional
methods in a stable library or only some implementations can provide
the additional methods (AKA "interface upgrade").

Needless to say that in v2 of this library, we'll do things in a more
straight forward way.

By now, I'm fairly confident we need to go down this way. It would be extremely confusing and damaging if enough users got bitten by the original approach so that we had to revert to the approach in this PR anyway.

The trade-off in this PR is the following: I'd expect relatively few users to use exemplars. And those who do are most likely advanced users who are not shocked by a type assertion. Thus, the collective cost is relatively low, but it buys as the benefit of not risking breakage.

On top of all that, a user of a `HistogramVec` has to do the type assertion anyway.

In different news, I'm still torn about the naming. The interface name should follow the pattern with the `...er` suffix, but `ObserveWithExemplarer` is not really an option. `ExemplarObserver` is not too bad, but perhaps we should rename the method to `ExemplarObserve` then? Not quite as descriptive, but the upside is that it is shorter.

WDYT?

### DIFF
--- a/examples/random/main.go
+++ b/examples/random/main.go
@@ -90,9 +90,13 @@ func main() {
 		for {
 			v := (rand.NormFloat64() * *normDomain) + *normMean
 			rpcDurations.WithLabelValues("normal").Observe(v)
-			rpcDurationsHistogram.ObserveWithExemplar(
-				// Demonstrate exemplar support with a dummy ID. This would be
-				// something like a trace ID in a real application.
+			// Demonstrate exemplar support with a dummy ID. This
+			// would be something like a trace ID in a real
+			// application.  Note the necessary type assertion. We
+			// already know that rpcDurationsHistogram implements
+			// the ExemplarObserver interface and thus don't need to
+			// check the outcome of the tipe assertion.
+			rpcDurationsHistogram.(prometheus.ExemplarObserver).ObserveWithExemplar(
 				v, prometheus.Labels{"dummyID": fmt.Sprint(rand.Intn(100000))},
 			)
 			time.Sleep(time.Duration(75*oscillationFactor()) * time.Millisecond)
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -41,13 +41,18 @@ type Counter interface {
 	// Add adds the given value to the counter. It panics if the value is <
 	// 0.
 	Add(float64)
-	// AddWithExemplar works like Add but also replaces the currently saved
-	// exemplar (if any) with a new one, created from the provided value,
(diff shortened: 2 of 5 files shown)


examples/random/main.go
```diff
@@ -90,9 +90,10 @@ func main() {
 		for {
 			v := (rand.NormFloat64() * *normDomain) + *normMean
 			rpcDurations.WithLabelValues("normal").Observe(v)
-			rpcDurationsHistogram.ObserveWithExemplar(
-				// Demonstrate exemplar support with a dummy ID. This would be
-				// something like a trace ID in a real application.
+			// Demonstrate exemplar support with a dummy ID. This would be
+			// something like a trace ID in a real application.
+			// Note the necessary type assertion.
+			rpcDurationsHistogram.(prometheus.ExemplarObserver).ObserveWithExemplar(
```
bwplotka: I would maybe in this example check if it's not implementing this? So:

```
if e, ok := rpcDurationsHistogram.(prometheus.ExemplarObserver); ok {
  e..ObserveWithExemplar( ...
}
```

bwplotka: This is because encouraging avoiding panics in your code is strongly suggested... But one can argue that we directly call our library constructor, so up to you here, not a blocker (:

beorn7: Yeah, it's all in the same file. I have added a comment now that explains why we don't check the type assertion.

beorn7:
> Given the proper support for exemplars will take for backends months, we will probably have already a v2 where we can change things.

I wouldn't be that optimistic. I expect v2 to take a while to reach a usable state. And even after that has happened, we'll need some time where the v2 branch can be used experimentally but will still need some time to stabilize for a real v2 release. And ever after that, I'd expect v1 to be used by a significant fraction of the current ~12k known importers for a long time…

Anyway, I assume that the above won't swing your otherwise quite clear verdict here. I'm fairly confident this is the way to go, too. I'll merge this and then adjust the release PR for final approval.

examples/random/main.go
beorn7: And I'll also fix the "tipo". :o(

bwplotka: Thanks!

bwplotka:
Awesome, yup same opinion on my side.

PR #713, opened 2020-02-12
Add a "factory" to make promauto work for custom registries

beorn7:
Besides the scary naming (happy to hear better ideas than the dreaded "factory"), the whole package documentation needs to updated, should we go down this path. (We don't need to talk about how promauto only works with the global registry anymore.)

@bwplotka I think this comes close to your request for a `promauto` extension, but it is done in a way that reads more nicely than an additional `Registerer` parameter in each constructor function.

### DIFF
--- a/prometheus/doc.go
+++ b/prometheus/doc.go
@@ -84,25 +84,21 @@
 // of those four metric types can be found in the Prometheus docs:
 // https://prometheus.io/docs/concepts/metric_types/
 //
-// A fifth "type" of metric is Untyped. It behaves like a Gauge, but signals the
-// Prometheus server not to assume anything about its type.
-//
-// In addition to the fundamental metric types Gauge, Counter, Summary,
-// Histogram, and Untyped, a very important part of the Prometheus data model is
-// the partitioning of samples along dimensions called labels, which results in
+// In addition to the fundamental metric types Gauge, Counter, Summary, and
+// Histogram, a very important part of the Prometheus data model is the
+// partitioning of samples along dimensions called labels, which results in
 // metric vectors. The fundamental types are GaugeVec, CounterVec, SummaryVec,
-// HistogramVec, and UntypedVec.
+// and HistogramVec.
 //
 // While only the fundamental metric types implement the Metric interface, both
 // the metrics and their vector versions implement the Collector interface. A
 // Collector manages the collection of a number of Metrics, but for convenience,
-// a Metric can also “collect itself”. Note that Gauge, Counter, Summary,
-// Histogram, and Untyped are interfaces themselves while GaugeVec, CounterVec,
-// SummaryVec, HistogramVec, and UntypedVec are not.
+// a Metric can also “collect itself”. Note that Gauge, Counter, Summary, and
+// Histogram are interfaces themselves while GaugeVec, CounterVec, SummaryVec,
+// and HistogramVec are not.
 //
 // To create instances of Metrics and their vector versions, you need a suitable
-// …Opts struct, i.e. GaugeOpts, CounterOpts, SummaryOpts, HistogramOpts, or
-// UntypedOpts.
+// …Opts struct, i.e. GaugeOpts, CounterOpts, SummaryOpts, or HistogramOpts.
 //
 // Custom Collectors and constant Metrics
 //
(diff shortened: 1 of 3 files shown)


prometheus/promauto/auto.go
```diff
@@ -221,3 +221,155 @@ func NewHistogramVec(opts prometheus.HistogramOpts, labelNames []string) *promet
 	prometheus.MustRegister(h)
 	return h
 }
+
+// Factory provides factory methods to create Collectors that are automatically
+// registered with a Registerer. Create a Factory with the With function,
+// providing a Registerer to register created Collectors with. The zero value of
+// a Factory creates Collectors that are not registered with any
+// Registerer. Each method of the Factory panics if the registration fails.
+type Factory struct {
```
bwplotka: Why not just embedding Registerer? (:

bwplotka: Ignore me, we don't want Register method to be exposed, no point.

beorn7:
With @bwplotka and @kakkoyun endorsing this approach, I'll rework the doc comments (and think about naming again) to make this ready for merging.

beorn7:
This is now ready for the final review.

As a byproduct, I realized that `NewUntypedFunc` got never added to the promauto package, which I have done now. An the same time, I realized a few obsolete remnants of "untyped", which I also fixed. (They are now all in this PR, but at least in separate commits.)

Context for the above: Untyped metrics cannot really happen in direct instrumentation, that's why `NewUntypedVec` and stuff was removed long ago. However, `NewUntypedFunc` can be used to mirror 3rd party metrics, and those might actually be of unknown type.

prometheus/doc.go
```diff
@@ -118,13 +114,16 @@
...
-// the Collect method. The Describe method has to return separate Desc
-// instances, representative of the “throw-away” metrics to be created later.
-// NewDesc comes in handy to create those Desc instances. Alternatively, you
-// could return no Desc at all, which will mark the Collector “unchecked”.  No
-// checks are performed at registration time, but metric consistency will still
-// be ensured at scrape time, i.e. any inconsistencies will lead to scrape
+// NewConstSummary (and their respective Must… versions). NewConstMetric is used
+// for all metric types with just a float64 as their value: Counter, Gauge, and
+// a special “type” called Untyped. Use the latter if you are not sure if the
+// mirrored metric is a Counter or a Gauge. Creation of the Metric instance
```
bwplotka: Just curious. Why in this case:

>  mirrored metric is a Counter or a Gauge. 

We cannot just use Gauge?

prometheus/promauto/auto.go
```diff
@@ -11,11 +11,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package promauto provides constructors for the usual Prometheus metrics that
-// return them already registered with the global registry
-// (prometheus.DefaultRegisterer). This allows very compact code, avoiding any
-// references to the registry altogether, but all the constructors in this
-// package will panic if the registration fails.
+// Package promauto provides alternative constructors for the fundamental
```
bwplotka: :+1:

prometheus/promauto/auto.go
```diff
@@ -79,51 +84,78 @@
...
+// registration of the previously existing metric). This may come as an even
+// bigger surprise with the global registry, where simply importing another
+// package can trigger a panic (if the newly imported package registers metrics
+// in its init function). At least, in the prometheus package, creation of
+// metrics and other collectors is separate from registration. You first create
+// the metric, and then you decide explicitly if you want to register it with a
+// local or the global registry, and if you want to handle the error or risk a
+// panic. With the constructors in the promauto package, registration is
+// automatic, and if it fails, it will always panic. Furthermore, the
+// constructors will often be called in the var section of a file, which means
```
bwplotka: That should be really not recommended. I would remove the word `often` .. (:

prometheus/promauto/auto.go
```diff
@@ -79,51 +84,78 @@
...
+// that panicking will happen as a side effect of merely importing a package.
+//
+// A separate package allows conservative users to entirely ignore it. And
+// whoever wants to use it, will do so explicitly, with an opportunity to read
+// this warning.
+//
+// Enjoy promauto responsibly!
```
bwplotka: haha, thanks!

prometheus/promauto/auto.go
```diff
@@ -132,92 +164,211 @@ import "github.com/prometheus/client_golang/prometheus"
...
+// Factory provides factory methods to create Collectors that are automatically
+// registered with a Registerer. Create a Factory with the With function,
+// providing a Registerer to auto-register created Collectors with. The zero
+// value of a Factory creates Collectors that are not registered with any
+// Registerer. All methods of the Factory panic if the registration fails.
+type Factory struct {
+	r prometheus.Registerer
+}
+
+// With creates a Factory using the provided Registerer for registration of the
```
bwplotka: What if registerer is nil? (We often do that for tests)

bwplotka: Oh actually it is just not registered... Awesome! :heart:

prometheus/doc.go
beorn7: Because you couldn't/shouldn't use `rate` on a gauge. It's really a difference if something is a Gauge for sure or if you don't know if it is a Gauge or a Counter.

`Untyped` is really mostly for mirroring some 3rd party metrics where you don't know, or the kind of metrics even changes between different scenarios (happens typically for hardware metrics exposed in generic ways).

prometheus/promauto/auto.go
beorn7: A lot of people do use the global registry, and for them it's normal to create metrics in the `var` section. Even important players in the Prometheus community recommend that. We have to leave it open here what way to pick. If I remove the "often", it will sounds like you would _always_ create metrics in the `var` section.

prometheus/promauto/auto.go
beorn7: That's also documented in the `Factory` doc comment.

PR #714, opened 2020-02-13
Explicitly forward metricVec methods Collect, Describe, Reset

beorn7:
Interestingly, methods implicitly forwarded from embedded types are
detected by GoDoc if they are just one level deep. Methods of embedded types in
the embedded type are not recognized. This commit therefore adds
explicit forwarding methods for Collect, Describe, and Reset.

@bwplotka I discovered that while working on the promauto changes. Please have a look, easy review. (o:

### DIFF
--- a/prometheus/vec.go
+++ b/prometheus/vec.go
@@ -91,6 +91,18 @@ func (m *metricVec) Delete(labels Labels) bool {
 	return m.metricMap.deleteByHashWithLabels(h, labels, m.curry)
 }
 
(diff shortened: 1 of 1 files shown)


beorn7:
There are plenty of related bugs: https://github.com/golang/go/issues/2971 https://github.com/golang/go/issues/6127 https://github.com/golang/go/issues/6600 … it seems to be a complicated and hairy topic.

roidelapluie:
Thanks for digging, at leas we have some references.

PR #718, opened 2020-02-26
API Client: Support new metadata endpoint in v1

### LINKED ISSUE #705: API client: Support for new metadata endpoints (in v1)
There should be support for the new Prometheus metadata endpoints described in https://github.com/prometheus/prometheus/issues/6395 .

cc @gotjosh

gotjosh:
Introduces support for the new metadata endpoint from Prometheus. The new endpoint provides information independent of targets and collapses the unique combinations of HELP, TYPE and UNIT.

Fixes #705

Signed-off-by: gotjosh <josue@grafana.com>

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -130,6 +130,7 @@ const (
 	epSeries          = apiPrefix + "/series"
 	epTargets         = apiPrefix + "/targets"
 	epTargetsMetadata = apiPrefix + "/targets/metadata"
+	epMetricsMetadata = apiPrefix + "/metadata"
 	epRules           = apiPrefix + "/rules"
 	epSnapshot        = apiPrefix + "/admin/tsdb/snapshot"
 	epDeleteSeries    = apiPrefix + "/admin/tsdb/delete_series"
@@ -248,6 +249,8 @@ type API interface {
 	Targets(ctx context.Context) (TargetsResult, error)
 	// TargetsMetadata returns metadata about metrics currently scraped by the target.
 	TargetsMetadata(ctx context.Context, matchTarget string, metric string, limit string) ([]MetricMetadata, error)
+	// MetricMetadata returns metadata about metrics currently scraped by the metric name.
+	MetricsMetadata(ctx context.Context, metric string, limit string) (map[string][]Metadata, error)
 }
 
 // AlertsResult contains the result from querying the alerts endpoint.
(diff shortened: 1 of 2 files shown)


api/prometheus/v1/api.go
```diff
@@ -366,6 +369,13 @@ type MetricMetadata struct {
 	Unit   string            `json:"unit"`
 }
 
+// Metadata models the metadata of a metric.
+type Metadata struct {
```
gotjosh: Initially, I tried to re-use the `MetricMetadata` struct above. However, it felt misleading to return a type that will _always_ have two empty fields (target, metric).

beorn7:
On my review queue. Help would be appreciated. @slrtbtfs or @joe-elliott perhaps?

api/prometheus/v1/api_test.go
```diff
@@ -202,6 +202,13 @@ func TestAPIs(t *testing.T) {
 		}
 	}
 
+	doMetricsMetadata := func(metring string, limit string) func() (interface{}, Warnings, error) {
```
slrtbtfs:
```suggestion
	doMetricsMetadata := func(metric string, limit string) func() (interface{}, Warnings, error) {
```

api/prometheus/v1/api_test.go
```diff
@@ -202,6 +202,13 @@ func TestAPIs(t *testing.T) {
...
+		return func() (interface{}, Warnings, error) {
+			v, err := promAPI.MetricsMetadata(context.Background(), metring, limit)
```
slrtbtfs:
```suggestion
			v, err := promAPI.MetricsMetadata(context.Background(), metric, limit)
```

api/prometheus/v1/api.go
```diff
@@ -802,6 +812,29 @@ func (h *httpAPI) TargetsMetadata(ctx context.Context, matchTarget string, metri
 	return res, json.Unmarshal(body, &res)
 }
 
+func (h *httpAPI) MetricsMetadata(ctx context.Context, metric string, limit string) (map[string][]Metadata, error) {
```
slrtbtfs: It might make sense to name the method just `Metadata` so it has the same name as the endpoint.

(nitpick)

joe-elliott: Agree.  This would also align with the struct name as well.

gotjosh: Sounds good to me!

gotjosh: Done!

gotjosh:
One last 👀 @joe-elliott / @slrtbtfs ?

PR #723, opened 2020-03-11
Export gatherers and registerer of pusher

nghialt:
Hi beorn@soundcloud.com,

I have a need to export `gatherers` and `registerer` of Pusher struct. Please review and comment

Regards,
NghiaLT.

### DIFF
--- a/prometheus/push/push.go
+++ b/prometheus/push/push.go
@@ -298,6 +298,16 @@ func (p *Pusher) fullURL() string {
 	return fmt.Sprintf("%s/metrics/%s", p.url, strings.Join(urlComponents, "/"))
 }
 
+// Registerer returns registerer object of pusher
+func (p *Pusher) Registerer() prometheus.Registerer {
+	return p.registerer
+}
+
+// Gatherers returns gatherers object of pusher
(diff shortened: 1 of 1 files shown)


bwplotka:
Can we learn more about the need? Use cases? (: @nghialt

beorn7:
I have fought very hard to keep the still quite large number of exported types and methods at bay in client_golang to make it less confusing to use. This PR would defeat that purpose. There needed to be a _really_ good reason to do so.

nghialt:
@bwplotka @beorn7 my case is to use `Handler` together with `Pusher`, but `registerer` and `gatherers` are not exported to use in `HandlerFor`

beorn7:
The suggested flow here would be to maintain your registry normally and then expose it with `promhttp.HandlerFor` and push it with `push.Pusher.Gatherer`.

beorn7:
I assume this is clarified sufficiently now.

If you still have questions, I recommend to use the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users), where more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

Issue #672, opened 2019-10-30
httpError (http.go:306)


AlanQin1024:
{"duration":9708394250,"level":"info","msg":"Metrics updated","time":"2019-10-30T11:57:42+08:00"}
2019/10/30 11:57:42 http: superfluous response.WriteHeader call from github.com/prometheus/client_golang/prometheus/promhttp.httpError (http.go:306)

beorn7:
Could you provide a bit more context? I can only guess from what you have provided. In particular, which version of this repo did you use? (https://github.com/prometheus/client_golang/pull/634 might be of relevance. If that's not in the code you were running, it would explain the message.)

beorn7:
Closing for lack of follow-up. I'll re-open once more information is provided.

cameronkerrnz:
I have been encountering this today using snmp_exporter 

```
snmp_exporter[26979]: level=info ts=2019-12-18T11:14:56.238Z caller=main.go:141 msg="Starting snmp_exporter" version="(version=0.16.1, branch=HEAD, revision=4b2bb401378cf40c409aba91ab25189be5670183)"
snmp_exporter[26979]: level=info ts=2019-12-18T11:14:56.238Z caller=main.go:142 msg="Build context" (gogo1.13.5,userroot@89bd31baeb1f,date20191210-15:23:21)=(MISSING)
snmp_exporter[26979]: level=info ts=2019-12-18T11:14:56.245Z caller=main.go:235 msg="Listening on address" address=:9116

snmp_exporter[26979]: 2019/12/19 00:24:05 http: superfluous response.WriteHeader call from github.com/prometheus/client_golang/prometheus/promhttp.httpError (http.go:344)
```

The version of snmp_exporter is v0.16.1, the current 'latest' release (https://github.com/prometheus/snmp_exporter/releases/tag/v0.16.1)

The WriteHeader is logged not infrequently, and I think it correlates with context exception deadlines being reported by Prometheus

I'm not aware of this behaviour with its previous version (v0.15.0), but I could have easily missed it.

It's possible that it may be the result of a timeout? I think some of the scrapes (eg. a little over 2s) may be taking a bit longer than permitted by default, but I haven't reviewed the timeout configuration yet.

cameronkerrnz:
I can further confirm that I didn't see this behaviour with v0.15.0 of snmp_exporter.

In terms of client_golang, this would be a difference of v1.0.0 to v1.2.1

[...]

PR #724, opened 2020-03-11
fix: superfluous response.WriteHeader call

lidel:
This makes `WriteHeader` do nothing if `r.wroteHeader` is already true.

It is a more robust way of avoiding `superfluous response.WriteHeader call` error in the console. Old version fixed redundant calls made by client_golang, this one closes the gap on calls made by third party code, removing error from stdout:

```
2020/03/11 14:57:06 http: superfluous response.WriteHeader call from github.com/prometheus/client_golang/prometheus/promhttp.(*responseWriterDelegator).WriteHeader (delegator.go:58)
```

It also makes delegator code simpler and less error prone during refactors.

### DIFF
--- a/prometheus/promhttp/delegator.go
+++ b/prometheus/promhttp/delegator.go
@@ -53,6 +53,10 @@ func (r *responseWriterDelegator) Written() int64 {
 }
 
 func (r *responseWriterDelegator) WriteHeader(code int) {
+	// Avoid superfluous response.WriteHeader call
+	if r.wroteHeader {
+		return
+	}
 	r.status = code
 	r.wroteHeader = true
 	r.ResponseWriter.WriteHeader(code)
@@ -64,9 +68,7 @@ func (r *responseWriterDelegator) WriteHeader(code int) {
 func (r *responseWriterDelegator) Write(b []byte) (int, error) {
 	// If applicable, call WriteHeader here so that observeWriteHeader is
 	// handled appropriately.
-	if !r.wroteHeader {
-		r.WriteHeader(http.StatusOK)
-	}
+	r.WriteHeader(http.StatusOK)
 	n, err := r.ResponseWriter.Write(b)
 	r.written += int64(n)
 	return n, err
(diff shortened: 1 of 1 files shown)


prometheus/promhttp/delegator.go
```diff
@@ -53,6 +53,10 @@ func (r *responseWriterDelegator) Written() int64 {
 }
 
 func (r *responseWriterDelegator) WriteHeader(code int) {
+	// Avoid superfluous response.WriteHeader call
```
brian-brazil: Full stop

prometheus/promhttp/delegator.go
```diff
@@ -86,9 +88,7 @@ func (d closeNotifierDelegator) CloseNotify() <-chan bool {
 func (d flusherDelegator) Flush() {
 	// If applicable, call WriteHeader here so that observeWriteHeader is
```
beorn7: The "If applicable" is now a bit confusing (as there is no `if` statement anymore). How about changing this and the other comments to "Call WriteHeader here so that observeWriteHeader is handled appropriately if still needed."

beorn7:
Thinking about it, perhaps "more robust" is actually not the right way here. I mean, the console message is not a problem on its own. It is there to expose the bug that `WriteHeader` is called to often. This change will actually mask this error rather than fixing it.

Perhaps we should hear opinions from people experienced with Go HTTP stuff.

beorn7:
Current state of my thinking process: I tend towards the delegator not changing anything it is not in charge of, i.e. it should not mask/fix superfluous calls of `WriteHeader`. However, it should be robust against those, so I guess the `r.observeWriteHeader(code)` should only happen if `r.wroteHeader` is still false, but the delegation to the original `WriteHeader` should happen in any case.

beorn7:
Breaking news: Now we have pretty solid evidence for the root cause, see #672. It's `promhttp` after all.

lidel:
@beorn7 cool, I see you mentioned working on a fix – should we close this PR then?

beorn7:
Yes, I'll close this now.

I mostly kept it open to remind myself of https://github.com/prometheus/client_golang/pull/724#issuecomment-597677662 . I'll work that into my upcoming fix.

Thanks everyone.

PR #731, opened 2020-04-02
Improve API error handling

huanggze:
Signed-off-by: huanggze <loganhuang@yunify.com>

The current way we handle error drops some info from `errorType` and `error` for some scenarios. 

Consider a real response example below, we expect the error object carry more info than `inconsistent body for response code`. 

```json
HTTP OK
---
{
    "status":"error",
    "errorType":"internal",
    "error":"multiple matches for labels: many-to-one matching must be explicit (group_left/group_right)"
}
```

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -905,14 +905,14 @@ func (h *apiClientImpl) Do(ctx context.Context, req *http.Request) (*http.Respon
 		}
 	}
 
-	if apiError(code) != (result.Status == "error") {
+	if apiError(code) && result.Status == "success" {
 		err = &Error{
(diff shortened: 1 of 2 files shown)


huanggze:
@beorn7

beorn7:
@bboreham @joe-elliott you might have most context here. Does this make sense?

joe-elliott:
Without knowing the details of prometheus's API response code it's hard to say definitively, but I do prefer the changes @huanggze has made to the existing code.

With these changes the error is only hardcoded to "inconsistent body for response code" if the body says "success" but the http status is error.  Otherwise you always get the error as returned from prometheus which I like.

PR #732, opened 2020-04-07
Rewrite CurryWith wrappers in a more idomatic way

lebaptiste:
Hi @beorn7, I was diving into the prometheus code and I found  the CurryWith wrappers a bit confusing so here is a small PR I hope would help. Thanks


lebaptiste:
I've checked the implementation of metricVec.curryWith and since it is not exported, I could get the guarantee that the "idomatic" version I proposed is _currently_ exactly the same.
However I see you point, the semantic has changed and if you anticipate you could return in future a combination of `err!=nil` and `vec != nil`, it's probably better not having this PR.
This gives me a chance to thank you for a such great work!

Issue #728, opened 2020-03-19
process_resident_memory_bytes detected is wrong in Win7 32bit


cobolbaby:
### Build Version: 

- golang: 1.13.8
- client_golang: 1.4.1

### Running OS:

- Windows7 32bit

### Details:

The memory usage I saw through resource management was about 30M. But the return value of the metrics interface is 300T, as shown below:

![2020-03-19 11-23-55屏幕截图](https://user-images.githubusercontent.com/11460848/77028726-a4609e00-69d4-11ea-9da0-a65573b7d67c.png)

In addition, process_virtual_memory_bytes metrics is 0.

beorn7:
The memory definitions on Microsoft Windows are certainly very different than on Unix-like OSs. While that might explain the zero for `process_virtual_memory_bytes`, 300TiB for `process_resident_memory_bytes` is certainly not what this metric is intended for.

Sadly, I don't have access to Microsoft Windows machines. Help welcome!

cobolbaby:
Help welcome!

PR #734, opened 2020-04-10
fix: process_resident_memory_bytes detected is incorrect in Win7 32bit

cobolbaby:
Ref: https://github.com/prometheus/client_golang/issues/728

### DIFF
--- a/prometheus/process_collector_windows.go
+++ b/prometheus/process_collector_windows.go
@@ -33,18 +33,22 @@ var (
 )
 
 type processMemoryCounters struct {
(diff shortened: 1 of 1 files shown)


beorn7:
Cool. Thank you. 

I have seen https://golang.org/src/internal/syscall/windows/psapi_windows.go now, where they use `uintptr` rather than `uint` I guess it doesn't really make a difference in practice, but is there perhaps a reason for that, and would `uintptr` be even better here?

cobolbaby:
@beorn7 No problem after verification．

PR #736, opened 2020-04-16
Add GaugeFuncVec collecting vectors of GaugeFuncs

colega:
`GaugeFunc` is great for datasources that can provide instant collection of metrics, like `sql.DB.Stats()`, however they don't allow multiple gauges on the same metric name with different label values.

A `GaugeFuncVec` works similarly to a `GaugeVec`, but instead of creating `Gauge` metrics, it accepts a function as the datasource (just like `GaugeFunc`) and registers it for a given set of labels.

Most of the code was already present in the `metricVec`, we just need to skip the metric creation part and provide method for adding metrics directly.

I did't open a previous discussion on Google Groups because this is a port of my own piece of code I did in https://github.com/colega/gaugefuncvec however I considered that it might be interesting to have that in the client.

- Pros: people can find this easily and it reuses most of the code from `metricsVec`. It's also easier to keep the API updated with most of the changes.
- Cons: makes the API of this client wider which isn't necessary since the implementation in my library provides same contract and (and performance in most cases). I can also understand the concern of letting people expose too many gauge funcs affecting the scraping performance, however I'd also expect people to understand what they're doing (and maybe add a warning in the comments?) 

Anyway, we can move the discussion about this to Google Groups if needed.

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -315,5 +315,5 @@ func NewCounterFunc(opts CounterOpts, function func() float64) CounterFunc {
 		opts.Help,
 		nil,
 		opts.ConstLabels,
-	), CounterValue, function)
+	), CounterValue, nil, function)
 }
--- a/prometheus/gauge.go
+++ b/prometheus/gauge.go
@@ -285,5 +285,118 @@ func NewGaugeFunc(opts GaugeOpts, function func() float64) GaugeFunc {
 		opts.Help,
 		nil,
 		opts.ConstLabels,
-	), GaugeValue, function)
+	), GaugeValue, nil, function)
+}
+
+// GaugeFuncVec is a Collector that bundles a set of GaugeFuncs that all share the same
+// Desc, but expose different funcs on different labels.
+// This is used, for example, if you want to report the number of open SQL connections for each of the sql.DB instance
+// you have by setting a gauge as func() float64 { return float64(db.Stats().OpenConnections) }
+type GaugeFuncVec struct {
+	*metricVec
+}
+
+// NewGaugeFuncVec creates a new GaugeFuncVec based on the provided GaugeOpts and
+// partitioned by the given label names.
+// GaugeFuncs can be added using the Add* methods
+func NewGaugeFuncVec(opts GaugeOpts, variableLabelNames []string) *GaugeFuncVec {
+	desc := NewDesc(
+		BuildFQName(opts.Namespace, opts.Subsystem, opts.Name),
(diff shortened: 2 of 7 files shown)


colega:
I guess that if we go forward with this, it should be added to `promauto` package.

beorn7:
I have contemplated such a feature for quite some time. So far, I arrived at the conclusion that the feature would be too heavy weight for the limited use case. I found it easier to use a bunch of `NewGaugeFunc` with const labels for the easy cases or a custom `Collector` for the complex cases.

Your code adds a quite heavy-weight type `GaugeFuncVec` with six methods and a constructor function. The same needed to be done for `CounterFunc`, and, as you already said, all needs to be mirrored to `promauto`. That's a _lot_ of surface to digest. This package is already quite fat with its many types (I worked hard to reduce them over the years).

Could you share example code that uses your proposed `GaugeFuncVec`? Then I could demonstrate how I'd implement it with the given primitives, and then we can better assess if it is worth the effort.

colega:
D'oh! I didn't realize (or I forgot, since I knew that at some point) that multiple metric collectors can be registered on the same metric name as long as their desc's are compatible. So, this is definitely useless.

My usecase was this one fwiw:
```go

	prometheus.MustRegister(
		prometheus.NewGaugeFunc(
			prometheus.GaugeOpts{
				Namespace:   "mysql",
				Name:        "connections_open",
				Help:        "Number of mysql connections open",
				ConstLabels: prometheus.Labels{"connection": "master"},
			},
			func() float64() { master.Stats().OpenConnections },
		),
		prometheus.NewGaugeFunc(
			prometheus.GaugeOpts{
				Namespace:   "mysql",
				Name:        "connections_open",
				Help:        "Number of mysql connections open",
				ConstLabels: prometheus.Labels{"connection": "slave"},
			},
			func() float64() { slave.Stats().OpenConnections },
		),
	)
``` 

We can close this. Maybe we can clarify the possiblity of this approach on the `NewGaugeFunc` docs? 

Thanks!

beorn7:
Yeah, that's actually a really good example for the use of `ConstLabels`. (The [Collector example](https://pkg.go.dev/github.com/prometheus/client_golang@v1.5.0/prometheus?tab=doc#example-Collector) used to use `ConstLabels`, but then `WrapRegistererWith` got added and proved to be the better technique.

Would you like to add a godoc example along the lines you have above? (On the non-technical side, I'd rather avoid the usage of the loaded words master and slave, though.)

PR #739, opened 2020-04-23
Porting promlint from prometheus/prometheus

RainbowMango:
Refer https://github.com/prometheus/prometheus/issues/5317.

I split this work into two commits:
- Porting : Just simply copy `promlint` in.
- Minor Improvement: Fix static check warning by some IDE, such as Goland. [More Deatils](https://github.com/prometheus/prometheus/pull/6445#issuecomment-564468169)

I wish it will be easier for the reviewer to catch what has been changed during porting.

### DIFF
--- a/prometheus/testutil/promlint/promlint.go
+++ b/prometheus/testutil/promlint/promlint.go
@@ -1,4 +1,4 @@
-// Copyright 2017 The Prometheus Authors
+// Copyright 2020 The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -21,8 +21,9 @@ import (
 	"sort"
 	"strings"
 
-	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
+
+	dto "github.com/prometheus/client_model/go"
 )
 
 // A Linter is a Prometheus metrics linter.  It identifies issues with metric
@@ -40,21 +41,16 @@ type Problem struct {
 	Text string
 }
 
-// problems is a slice of Problems with a helper method to easily append
-// additional Problems to the slice.
-type problems []Problem
-
-// Add appends a new Problem to the slice for the specified metric, with
-// the specified issue text.
-func (p *problems) Add(mf dto.MetricFamily, text string) {
-	*p = append(*p, Problem{
+// newProblem is helper function to create a Problem.
+func newProblem(mf dto.MetricFamily, text string) Problem {
+	return Problem{
 		Metric: mf.GetName(),
 		Text:   text,
-	})
+	}
 }
 
 // New creates a new Linter that reads an input stream of Prometheus metrics.
-// Only the text exposition format is supported.
+// Only the Prometheus text exposition format is supported.
 func New(r io.Reader) *Linter {
 	return &Linter{
 		r: r,
@@ -118,11 +114,11 @@ func lint(mf dto.MetricFamily) []Problem {
 
 // lintHelp detects issues related to the help text for a metric.
 func lintHelp(mf dto.MetricFamily) []Problem {
-	var problems problems
+	var problems []Problem
 
 	// Expect all metrics to have help text available.
 	if mf.Help == nil {
-		problems.Add(mf, "no help text")
+		problems = append(problems, newProblem(mf, "no help text"))
 	}
 
 	return problems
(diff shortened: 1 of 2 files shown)


RainbowMango:
cc @beorn7 @brancz

By the way, I'm curious why Travis not run?

RainbowMango:
I can't see the Travis build log, `travis-ci.org` always sys `Under Maintenance` when access the `Details` link. I will check it later.

brancz:
lgtm but leaving last call to @beorn7

prometheus/testutil/promlint/promlint.go
```diff
@@ -0,0 +1,367 @@
+// Copyright 2017 The Prometheus Authors
```
beorn7, line 1: It doesn't matter a lot, but we use to fill in the current year if we create a new file.

prometheus/testutil/promlint/promlint.go
```diff
@@ -0,0 +1,367 @@
...
+
+import (
+	"fmt"
+	"io"
+	"regexp"
+	"sort"
+	"strings"
+
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
```
beorn7, line 25: Nit: The super strict Prometheus convention is to put named imports in their own block after the unnamed ones. So line 24 should be after line 25, separated by a blank line.

prometheus/testutil/promlint/promlint_test.go
```diff
@@ -0,0 +1,788 @@
+// Copyright 2017 The Prometheus Authors
```
beorn7, line 1: As above, let's put 2020 here.

prometheus/testutil/promlint/promlint.go
```diff
@@ -0,0 +1,367 @@
...
+// A Problem is an issue detected by a Linter.
+type Problem struct {
+	// The name of the metric indicated by this Problem.
+	Metric string
+
+	// A description of the issue for this Problem.
+	Text string
+}
+
+// newProblem is helper function to create a Problem.
```
beorn7, line 43: :rofl:

prometheus/testutil/promlint/promlint.go
```diff
@@ -0,0 +1,367 @@
...
+func newProblem(mf dto.MetricFamily, text string) Problem {
+	return Problem{
+		Metric: mf.GetName(),
+		Text:   text,
+	}
+}
+
+// New creates a new Linter that reads an input stream of Prometheus metrics.
+// Only the text exposition format is supported.
```
beorn7, line 52: Perhaps "Only the Prometheus text exposition format is supported." so that people don't get confused with OpenMetrics?

beorn7:
Many thanks :rainbow::mango:, :st_barthelemy::fish:, and :peace_symbol:.

I have a question: Do you feel this should in its own `promlint` package? I'm a bit concerned the package structure gets a bit nested. On the other hand, it's a nice and clean separation. If we keep it in a sub-package, I'd like to add `testutil.CollectAndLint` and `testutil.GatherAndLint` to make it easier to include the linting in metrics testing. Do you think that's a good idea?

RainbowMango:
> I have a question: Do you feel this should in its own promlint package? I'm a bit concerned the package structure gets a bit nested.

I'm glad you ask this.

Yes, I think if we put it in `client_golang/prometheus/promlint` would be much better.
But I'm not worried about the nest, I think `promlint` not just test util. It could be more significant in the future.

Indeed, we can take promlint as test util for now, because it can only check the gathered metrics. 
I have a thought to extend promlint, making it possible to check a metric before register to a registry. So, we can take `promlint` as a tool for metric conformance. From this point of view, even we put `promlint` to a separated repo could be acceptable.

RainbowMango:
Actually, I [started a program](https://github.com/promlint/promlint) to run my thought months ago.

I don't want to disrupt the Prometheus community, I just want the `best practice` cloud spread to more projects.

Would you mind I continue working on my program? If it really works, I will be happy to donate it to the Prometheus community.

beorn7:
We already arrived at the conclusion that having the promlint package here in client_golang is a good idea. A stand-alone program would than simply import the package and do the boilerplate around it to turn it into e.g. a command line tool or whatever else.

My question was more about having the moving parts directly in `client_golang/prometheus/testutil` or in its own sub-package `client_golang/prometheus/testutil/promlint` as it is now in this PR. I'd find `client_golang/prometheus/promlint` confusing as I see the linting as part of testing from the perspective of an instrumentation library.

Let's go with `client_golang/prometheus/testutil/promlint` as in this PR for now. I'll sand a PR for those helpers in `testutil` I mentioned above.

Thanks everyone and in particular @RainbowMango .

PR #743, opened 2020-04-24
Add linter helpers to testutil

beorn7:
This for one adds direct `MetricFamily` support to `promlint`.

It then adds helpers `CollectAndLint` and `GatherAndLint` akin to `CollectAndCompare` and `GatherAndCompare` to the `testutil` package, which simplifies to integrate linting into testing of instrumented code.

@RainbowMango

### DIFF
--- a/prometheus/testutil/lint.go
+++ b/prometheus/testutil/lint.go
@@ -20,9 +20,9 @@ import (
 	"github.com/prometheus/client_golang/prometheus/testutil/promlint"
 )
 
-// CollectAndLint registers the provided Collector with a newly created
-// pedantic Registry. It then does the same as GatherAndLint, gathering the
-// metrics from the pedantic Registry.
+// CollectAndLint registers the provided Collector with a newly created pedantic
+// Registry. It then calls GatherAndLint with that Registry and with the
+// provided metricNames.
 func CollectAndLint(c prometheus.Collector, metricNames ...string) ([]promlint.Problem, error) {
 	reg := prometheus.NewPedanticRegistry()
 	if err := reg.Register(c); err != nil {
--- a/prometheus/testutil/lint.go
+++ b/prometheus/testutil/lint.go
@@ -1,0 +1,46 @@
+// Copyright 2020 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testutil
+
+import (
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil/promlint"
+)
+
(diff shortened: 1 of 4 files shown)


prometheus/testutil/lint.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil/promlint"
+)
+
+// CollectAndLint registers the provided Collector with a newly created
+// pedantic Registry. It then does the same as GatherAndLint, gathering the
```
bwplotka, line 24:
```suggestion
// pedantic Registry. It then does the same as GatherAndLint, gathers the
```
Kind of scared to suggest this to you, but .... am I right here? :smile: 

Previous version was reading oddly

prometheus/testutil/promlint/promlint.go
```diff
@@ -29,7 +29,8 @@ import (
 // A Linter is a Prometheus metrics linter.  It identifies issues with metric
 // names, types, and metadata, and reports them to the caller.
 type Linter struct {
-	r io.Reader
+	r   io.Reader
```
bwplotka, line 32: I would comment this is `oneof` between two. I was bit confused and I can imagine package devs might be as well..

Not relevant to package users

prometheus/testutil/lint.go
```diff
@@ -0,0 +1,46 @@
...
+
+// GatherAndLint gathers all metrics from the provided Gatherer and checks them
+// with the linter in the promlint package. If any metricNames are provided,
+// only metrics with those names are checked.
+func GatherAndLint(g prometheus.Gatherer, metricNames ...string) ([]promlint.Problem, error) {
+	got, err := g.Gather()
+	if err != nil {
+		return nil, fmt.Errorf("gathering metrics failed: %s", err)
+	}
+	if metricNames != nil {
```
bwplotka, line 42: I  think I would prefer `len(metricNames) > 0`, less surprising.

prometheus/testutil/lint.go
beorn7: I could argue why my version is correct. But I agree it's convoluted. I'll reword to avoid the complication.

prometheus/testutil/promlint/promlint.go
beorn7: It would actually work with both having values, just that you don't have a constructor doing so (but we might have in the future). I'll add a comment explaining that.

prometheus/testutil/lint.go
beorn7: My thinking was that the linter will never recognize fewer problems in future versions, but it might recognize more. I changed the comment to clarify that.

RainbowMango:
Awesome!

RainbowMango:
@beorn7 
By the way, are you gonna publish a new release after this PR? 

I can help clean up `promlint` package in `prometheus/prometheus` repo.
- promlint in  `prometheus/prometheus` should import `prometheus/client_golang` as you mentioned on https://github.com/prometheus/prometheus/issues/5317#issuecomment-616607117.

I was wondering if we can deprecate promlint in  `prometheus/prometheus`.

beorn7:
@RainbowMango Yes, I'll cut v1.6.0 ASAP (hopefully today).

Once it is released, it would be very much appreciated if you could help switching over the prometheus/prometheus code.

RainbowMango:
> Once it is released, it would be very much appreciated if you could help switching over the prometheus/prometheus code.

OK. Will do.

prometheus/testutil/lint.go
bwplotka: Thx!

PR #752, opened 2020-05-13
Properly handle empty job and label values

### LINKED ISSUE #750: push: properly handle empty label values in grouping key
See https://github.com/prometheus/pushgateway/pull/346 for a documentation update for the PGW about empty label values in the grouping key. (Note: It's really just documenting the current state of the PGW. No change of behavior of the PGW involved.)

The `push` package currently accepts empty label values for the grouping key, as demonstrated in https://github.com/prometheus/pushgateway/issues/344 . However, it then attempts to push in an invalid way, see again https://github.com/prometheus/pushgateway/issues/344 for details.

While the use case for empty label values in the grouping key is somewhat esoteric, a case can be made for it. Thus, we should probably support it in the way now [documented](https://github.com/prometheus/pushgateway/pull/346). Alternative is to disallow it in the `push` package. Just the current state of pretending it works but then fail to push is clearly a bug.

beorn7:
An empty job name was always an error, but it was previously only
detected when pushing to the PGW and receiving an error. With this
commit, the error is detected before pushing.

An empty label value should have been OK but was encoded in a way that
couldn't be pushed to the
PGW. Cf. https://github.com/prometheus/pushgateway/issues/344 . This
commit changes the creation of the path in the URL so that it works
with empty label values.

@matevz
Fixes #750 .

### DIFF
--- a/prometheus/push/push.go
+++ b/prometheus/push/push.go
@@ -37,6 +37,7 @@ package push
 import (
 	"bytes"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


beorn7:
Thanks @roidelapluie

PR #753, opened 2020-05-13
Improve CollectAndCount

beorn7:
Now that we have also added CollectAndLint and GatherAndLint, I
thought we should bring CollectAndCount in line. So:

- Add GatherAndCount.
- Add filtering by metric name.
- Add tests.

Minor wart: CollectAndCount should now return `(int, error)`, but that
would be a breaking change as the current version just returns
`int`. I decided to let the new version panic when it should return an
error. An error is anyway very unlikely, so the biggest annoyance here
is really just the inconsistency.

@bwplotka you might like this.

@cstyan you needed this recently (but as already discussed, with `promauto` you won't, but your request still inspired this PR)

### DIFF
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -112,31 +112,43 @@ func ToFloat64(c prometheus.Collector) float64 {
 	panic(fmt.Errorf("collected a non-gauge/counter/untyped metric: %s", pb))
 }
 
-// CollectAndCount collects all Metrics from the provided Collector and returns their number.
-//
-// This can be used to assert the number of metrics collected by a given collector after certain operations.
-//
-// This function is only for testing purposes, and even for testing, other approaches
-// are often more appropriate (see this package's documentation).
-func CollectAndCount(c prometheus.Collector) int {
-	var (
-		mCount int
-		mChan  = make(chan prometheus.Metric)
-		done   = make(chan struct{})
-	)
+// CollectAndCount registers the provided Collector with a newly created
+// pedantic Registry. It then calls GatherAndCount with that Registry and with
+// the provided metricNames. In the unlikely case that the registration or the
+// gathering fails, this function panics. (This is inconsistent with the other
+// CollectAnd… functions in this package and has historical reasons. Changing
+// the function signature would be a breaking change and will therefore only
+// happen with the next major version bump.)
+func CollectAndCount(c prometheus.Collector, metricNames ...string) int {
+	reg := prometheus.NewPedanticRegistry()
+	if err := reg.Register(c); err != nil {
+		panic(fmt.Errorf("registering collector failed: %s", err))
+	}
+	result, err := GatherAndCount(reg, metricNames...)
+	if err != nil {
+		panic(err)
+	}
+	return result
+}
 
-	go func() {
-		for range mChan {
-			mCount++
-		}
-		close(done)
-	}()
+// GatherAndCount gathers all metrics from the provided Gatherer and counts
+// them. It returns the number of metric children in all gathered metric
+// families together. If any metricNames are provided, only metrics with those
+// names are counted.
+func GatherAndCount(g prometheus.Gatherer, metricNames ...string) (int, error) {
+	got, err := g.Gather()
+	if err != nil {
+		return 0, fmt.Errorf("gathering metrics failed: %s", err)
+	}
+	if metricNames != nil {
+		got = filterMetrics(got, metricNames)
+	}
 
-	c.Collect(mChan)
-	close(mChan)
-	<-done
-
-	return mCount
+	result := 0
+	for _, mf := range got {
+		result += len(mf.GetMetric())
+	}
+	return result, nil
 }
 
 // CollectAndCompare registers the provided Collector with a newly created
--- a/prometheus/testutil/testutil_test.go
+++ b/prometheus/testutil/testutil_test.go
@@ -306,3 +306,30 @@ some_total{label1="value1"} 1
 		t.Errorf("Expected\n%#+v\nGot:\n%#+v", expectedError, err.Error())
 	}
 }
+
+func TestCollectAndCount(t *testing.T) {
+	c := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "some_total",
+			Help: "A value that represents a counter.",
+		},
+		[]string{"foo"},
+	)
+	if got, want := CollectAndCount(c), 0; got != want {
+		t.Errorf("unexpected metric count, got %d, want %d", got, want)
+	}
+	c.WithLabelValues("bar")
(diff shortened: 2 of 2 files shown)


beorn7:
You wanted to write a test if a certain metric shows up in your exposition. That's possible with the new tooling here because you can just put that metric name into the filter and count if there is at least one in the exposition.

cstyan:
Isn't the pedantic registry kind of messing with that? I wanted to check if a list of metric names had been registered with an existing `prometheus.Registry`.

beorn7:
That's where the `GatherAndCount` comes to your rescue (also added in this PR).

prometheus/testutil/testutil.go
```diff
@@ -112,31 +112,43 @@ func ToFloat64(c prometheus.Collector) float64 {
...
-		close(done)
-	}()
+// CollectAndCount registers the provided Collector with a newly created
+// pedantic Registry. It then calls GatherAndCount with that Registry and with
+// the provided metricNames. In the unlikely case that the registration or the
+// gathering fails, this function panics. (This is inconsistent with the other
+// CollectAnd… functions in this package and has historical reasons. Changing
+// the function signature would be a breaking change and will therefore only
+// happen with the next major version bump.)
+func CollectAndCount(c prometheus.Collector, metricNames ...string) int {
```
bwplotka, line 122: I totally don't see `CollectAndCount(c prometheus.Collector, metricNames ...string)` signature. It does not make sense to me - too many ways of doing the same thing.

prometheus/testutil/testutil.go
```diff
@@ -112,31 +112,43 @@ func ToFloat64(c prometheus.Collector) float64 {
...
-		for range mChan {
-			mCount++
-		}
-		close(done)
-	}()
+// CollectAndCount registers the provided Collector with a newly created
+// pedantic Registry. It then calls GatherAndCount with that Registry and with
+// the provided metricNames. In the unlikely case that the registration or the
+// gathering fails, this function panics. (This is inconsistent with the other
+// CollectAnd… functions in this package and has historical reasons. Changing
```
bwplotka, line 119: Do we really need to have error? Previous version was error-free

beorn7:
@bwplotka You must have misunderstood. `CollectAndCount` will continue to work in exactly the same way as before. Just that it has now optional filtering.

And as I tried to explain, we don't need to change it to return an error. This PR does _not_ change it to return an error.

The registration is needed to enable the filtering. Since the registration will never fail in practice, I wouldn't introduce a different code path for the case of no filtering.

And yes, there is a need for `GatherAndCount`. @cstyan did not find another solution, he just did something different that happened to not test for presence of metrics anymore.

And there are always people who ask for the option to handle an error instead of panicking. That's why all the other functions return an error. But as you personally do not seem to need `GatherAndCount`, there is no reason for you to dislike this because `CollectAndCount` still does not return an error.

prometheus/testutil/testutil.go
beorn7: And the current version is still error-free.

prometheus/testutil/testutil.go
beorn7: I don't understand your comment. What's the other way of doing the same thing?

bwplotka:
> CollectAndCount will continue to work in exactly the same way as before. Just that it has now optional filtering.

Hm but the single collector has just one metric name, so what is this filtering :thinking: I guess historgrams / summaries which have more than one? Ok, then fine (: 

> And as I tried to explain, we don't need to change it to return an error. This PR does not change it to return an error.

I know @beorn7 but you say it's broken that it does not return error... so I am arguing with that. (:

bwplotka:
> And there are always people who ask for the option to handle an error instead of panicking. That's why all the other functions return an error. But as you personally do not seem to need GatherAndCount, there is no reason for you to dislike this because CollectAndCount still does not return an error.

Cool, looks like what we need is `Must...` version then. (:

bwplotka:
How brutal would be to introduce breaking change? and mention this in comment?

beorn7:
> Hm but the single collector has just one metric name, so what is this filtering?

That's not true. A `Collector` can return any number of differently named metrics. A prominent example used by almost everyone is https://github.com/prometheus/client_golang/blob/master/prometheus/go_collector.go .

With the filtering, you can easily test for presence of a particular metric in your home-grow `Collector`.

> I guess historgrams / summaries which have more than one?

That's not true, either. (o: Histograms and summaries only have one metric name. (The "magic" metric names you see in Prometheus and in the text format are not "real" names.)

> I know @beorn7 but you say it's broken that it does not return error... so I am arguing with that. (:

I'm not saying it's broken. It's just inconsistent with the four or five other functions we have in this package. I'd prefer to have them _all_ panic or _all_ return an error. But either change would be breaking.

> Cool, looks like what we need is `Must...` version then. (: 

Strictly speaking, yes, but I think in this case, it would be overkill. Personally, I'd be fine with all the functions panicking, but the error-return version is less controversial and more flexible (even if it makes the handling less elegant – but why are you using Go if you want elegant code? :exploding_head: ). We got the version with errors from the 1st time contributor, and when you contributed `CollectAndCount`, I didn't realize that we could align it better with the existing functions.

What we have in this PR now has just a tiny consistency wart.

> How brutal would be to introduce breaking change?

Given the gigantic adoption of this module, I'd really bump the major version for any breaking change in a part that is not explicitly marked as experimental. But then I want to bundle a bunch of batched up breaking changes to not inflict the cost of a major version bump to change a single function signature.

cstyan:
Hmm okay, yes I can see how `GatherAndCount` could be used for what I wanted to try. Thanks @beorn7 !

prometheus/testutil/testutil.go
bwplotka: Sure, but you plan to make it error full (:

prometheus/testutil/testutil.go
bwplotka: You answered in main thread, I did not know there exists implementations of collector with many names.

beorn7:
> But still, I don't fully agree with the comment that is "predicting" that panicking is wrong here (:

There will certainly be a few people patronizing me at the next meetup (as it happened before, and of course if we will have proper meetups again any time soon…). :nerd_face: 

> Unless I am missing something, the only kind of errors that can happen in Collector is someone creating an invalid metric name

Which can plausibly happen given that the helper function will often be used to test custom collectors. The collector could also collect a set of metrics that is not self-consistent.

> Plus this is only testing package.

Some people feel especially strong about not panicking in test code.

In general, instead of trying to convince everyone to do the one right thing, I try to accommodate both schools of thought, at least if that's possible without a lot of overhead.

bwplotka:
> Some people feel especially strong about not panicking in test code.

Kind of I am. But I am sometimes also not decided xd

:+1:

PR #754, opened 2020-05-14
Add lint:ignore for protobuf deprecation

beorn7:
`github.com/golang/protobuf/proto` is deprecated in lieu of
`google.golang.org/protobuf/proto`. However, we cannot simply
migrate. Types from the proto package are exposed to users of packages
in this repo. If we migrate here, users have to migrate to. Thus, we
could only migrate with a major version bump.

In different news, with all the inline lint:ignore comments, including
the existing ones, there is no need to repeat the exception in the
Makefile.

A current version of `staticcheck` is happy with the code after this
commit. golangci-lint is broken at the moment, however, and ignores
the lint:ignore comments in the code as well as those via envvar.

@bentucker this does not fix the golangci-lint problem, but at least it addresses the legitimate protobuf deprecation warnings that even the proper staticcheck would catch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,13 +13,6 @@
 
 include Makefile.common
 
-# http.CloseNotifier is deprecated but we don't want to remove support
-# from client_golang to not break anybody still using it.
-STATICCHECK_IGNORE = \
-  github.com/prometheus/client_golang/prometheus/promhttp/delegator*.go:SA1019 \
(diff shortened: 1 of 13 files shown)


beorn7:
@simonpasquier gentle ping.

beorn7:
Thanks everyone.

PR #755, opened 2020-05-18
api/prometheus/v1/api.go: Add support for status/runtimeinfo endpoint

### LINKED ISSUE #746: Add support for /status/runtimeinfo endpoint to the api client
Seems like there is `status/config` and `status/flags` but not `status/runtimeinfo`, any particular reason why this endpoint is missing or just simply was not added yet? 
(Happy to add it, if it's the latter)

https://github.com/prometheus/client_golang/blob/6edbbd9e560190e318cdc5b4d3e630b442858380/api/prometheus/v1/api.go#L138-L139

lilic:
Adds support for the status/runtimeinfo API endpoint as per the API docs https://prometheus.io/docs/prometheus/latest/querying/api/#runtime-information

Closes #746

cc @beorn7

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -137,6 +137,7 @@ const (
 	epCleanTombstones = apiPrefix + "/admin/tsdb/clean_tombstones"
 	epConfig          = apiPrefix + "/status/config"
 	epFlags           = apiPrefix + "/status/flags"
+	epRuntimeinfo     = apiPrefix + "/status/runtimeinfo"
 )
 
 // AlertState models the state of an alert.
@@ -238,6 +239,8 @@ type API interface {
 	Query(ctx context.Context, query string, ts time.Time) (model.Value, Warnings, error)
 	// QueryRange performs a query for the given range.
 	QueryRange(ctx context.Context, query string, r Range) (model.Value, Warnings, error)
+	// Runtimeinfo returns the various runtime information properties about the Prometheus server.
+	Runtimeinfo(ctx context.Context) (RuntimeinfoResult, error)
 	// Series finds series by label matchers.
 	Series(ctx context.Context, matches []string, startTime time.Time, endTime time.Time) ([]model.LabelSet, Warnings, error)
 	// Snapshot creates a snapshot of all current data into snapshots/<datetime>-<rand>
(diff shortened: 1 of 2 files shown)


api/prometheus/v1/api.go
```diff
@@ -238,6 +239,8 @@ type API interface {
 	Query(ctx context.Context, query string, ts time.Time) (model.Value, Warnings, error)
 	// QueryRange performs a query for the given range.
 	QueryRange(ctx context.Context, query string, r Range) (model.Value, Warnings, error)
+	// Runtimeinfo returns the various runtime information properties about the Prometheus server.
```
bwplotka, line 242:
```suggestion
	// RuntimeInfo returns the various runtime information properties about the Prometheus server.
```

api/prometheus/v1/api.go
```diff
@@ -238,6 +239,8 @@ type API interface {
...
+	Runtimeinfo(ctx context.Context) (RuntimeinfoResult, error)
```
bwplotka, line 243:
```suggestion
	RuntimeInfo(ctx context.Context) (RuntimeinfoResult, error)
```
```suggestion
	Runtimeinfo(ctx context.Context) (RuntimeInfoResult, error)
```

api/prometheus/v1/api.go
```diff
@@ -277,6 +280,22 @@ type ConfigResult struct {
 // FlagsResult contains the result from querying the flag endpoint.
 type FlagsResult map[string]string
 
+// RuntimeinfoResult contains the result from querying the runtimeinfo endpoint.
+type RuntimeinfoResult struct {
```
bwplotka, line 284: ditto

api/prometheus/v1/api.go
```diff
@@ -640,6 +659,23 @@ func (h *httpAPI) Flags(ctx context.Context) (FlagsResult, error) {
...
 
+func (h *httpAPI) Runtimeinfo(ctx context.Context) (RuntimeinfoResult, error) {
+	u := h.client.URL(epRuntimeinfo, nil)
+
+	req, err := http.NewRequest(http.MethodGet, u.String(), nil)
+	if err != nil {
+		return RuntimeinfoResult{}, err
+	}
+
+	_, body, _, err := h.client.Do(ctx, req)
```
bwplotka, line 670: It looks like matches the other methods, but why we discard the status code completely? No point in marshaling if we don't check 2** status. :thinking:

lilic: Like you mentioned this just uses what the other functions use, but the custom `.Do` does check for status :) https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go#L880

api/prometheus/v1/api.go
lilic: I disagree with this, as the endpoint is /runtimeinfo not /runtime_info, at least that was the logic I used here. :)

lilic:
@beorn7 @bwplotka PTAL, thanks! :)

beorn7:
Sorry for the waiting time, @lilic . I'm not really an expert of this HTTP client code. So I was just waiting for @bwplotka to signal if his nits have been addressed. And then I missed for how long nothing happened here. I now checked myself, and from my somewhat limited perspective, it looks fine to me. As @brancz seems happy, too, I'll merge this.

Thank you very much!

lilic:
@beorn7 no problem, if @bwplotka is not happy I can fix it in a follow up as well! :)

api/prometheus/v1/api.go
bwplotka: Interesting, not a blocker, but this looks weird (:

api/prometheus/v1/api.go
bwplotka: I think this is something to fix in follow up PRs if any, just it feels wrong, will add issue to follow up (:

lilic: Agreed, we should fix it for all the functions!

PR #758, opened 2020-05-24
Implement noop process collector for js/wasm

slugalisk:
Adds minimal support for building for WebAssembly.

Builds targeting js/wasm currently fail trying to compile procfs. Browsers don't provide APIs for retrieving memory or CPU usage so metric collection is disabled.

### DIFF
--- a/prometheus/process_collector_js.go
+++ b/prometheus/process_collector_js.go
@@ -1,4 +1,4 @@
-// Copyright 2019 The Prometheus Authors
+// Copyright 2020 The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build js,wasm
+// +build js
 
 package prometheus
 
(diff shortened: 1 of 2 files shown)


prometheus/process_collector_js.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2019 The Prometheus Authors
```
beorn7, line 1: Nit: This should be updated to the current year for new files.

beorn7:
A different thought: Does the code as is compile on js/wasm? The idea so far is that the code in `process_collector_other.go` would gracefully just not collect any metrics on un-supported platforms, and that you only need to provide specific code for platforms that _do_ support process metric collection, but in other ways.

slugalisk:
No, the current `process_collector_other.go` does not compile due to the procfs dependency. It seems unlikely that procfs would work on any wasm platform since one of the big selling point of wasm is sandboxing. It also seems like updating procfs to support future wasm APIs would be out of scope for that project. FWIW I think it makes sense to use the noop collector for both js and wasm but supporting just js suits my needs... I've updated the diff accordingly.

beorn7:
I agree that procfs wouldn't yield anything useful on js/wasm, but the approach here is that the procfs package just err's out in those cases, and we handle that here appropriately, see https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector_other.go#L22-L25

So I think we should fix procfs to behave like that on js/wasm.

beorn7:
I'll file a bug against procfs.

beorn7:
Or even better, fix it right ahead: https://github.com/prometheus/procfs/pull/303

Closing this one. Thanks for catching.

slugalisk:
👍 Thanks for fixing this.

PR #761, opened 2020-05-28
testutil/promlint: allow Kelvin as a base unit for color temperature

mdlayher:
Signed-off-by: Matt Layher <mdlayher@gmail.com>

I'm exporting metrics from my new Elgato Key Light and the device exposes the color temperature of its light in Kelvin, e.g. 4200K. I don't think this measurement would make any sense in Celsius and propose allowing Kelvin as a "base unit", but still recommending that "kelvins" be changed to "kelvin".

### DIFF
--- a/prometheus/testutil/promlint/promlint.go
+++ b/prometheus/testutil/promlint/promlint.go
@@ -313,9 +313,10 @@ var (
 		// Base units.
 		"amperes": "amperes",
 		"bytes":   "bytes",
-		"celsius": "celsius", // Celsius is more common in practice than Kelvin.
+		"celsius": "celsius", // Also allow Celsius because it is common in typical Prometheus use cases.
 		"grams":   "grams",
 		"joules":  "joules",
+		"kelvin":  "kelvin", // SI base unit, used in special cases (e.g. color temperature, scientific measurements).
 		"meters":  "meters", // Both American and international spelling permitted.
 		"metres":  "metres",
 		"seconds": "seconds",
@@ -328,8 +329,7 @@ var (
 		"days":    "seconds",
 		"weeks":   "seconds",
 		// Temperature.
-		"kelvin":     "celsius",
-		"kelvins":    "celsius",
+		"kelvins":    "kelvin",
 		"fahrenheit": "celsius",
 		"rankine":    "celsius",
 		// Length.
--- a/prometheus/testutil/promlint/promlint_test.go
+++ b/prometheus/testutil/promlint/promlint_test.go
@@ -166,6 +166,14 @@ x_seconds 10
 x_joules 10
 `,
 		},
+		{
+			name: "kelvin",
+			in: `
+# HELP x_kelvin Test metric.
+# TYPE x_kelvin untyped
+x_kelvin 10
+`,
+		},
 		// bad cases.
 		{
 			name: "milliamperes",
@@ -288,18 +296,6 @@ x_days 10
 			}},
 		},
 		{
-			name: "kelvin",
-			in: `
-# HELP x_kelvin Test metric.
-# TYPE x_kelvin untyped
-x_kelvin 10
-`,
-			problems: []promlint.Problem{{
-				Metric: "x_kelvin",
-				Text:   `use base unit "celsius" instead of "kelvin"`,
-			}},
-		},
-		{
 			name: "kelvins",
 			in: `
 # HELP x_kelvins Test metric.
@@ -308,7 +304,7 @@ x_kelvins 10
 `,
 			problems: []promlint.Problem{{
 				Metric: "x_kelvins",
-				Text:   `use base unit "celsius" instead of "kelvins"`,
+				Text:   `use base unit "kelvin" instead of "kelvins"`,
 			}},
 		},
 		{


prometheus/testutil/promlint/promlint.go
```diff
@@ -313,9 +313,10 @@ var (
 		// Base units.
 		"amperes": "amperes",
 		"bytes":   "bytes",
-		"celsius": "celsius", // Celsius is more common in practice than Kelvin.
+		"celsius": "celsius", // Celsius is preferred in computing.
```
beorn7, line 316: How about "Also allow Celsius because it is common in typical Prometheus use cases."

prometheus/testutil/promlint/promlint.go
```diff
@@ -313,9 +313,10 @@ var (
...
 		"grams":   "grams",
 		"joules":  "joules",
+		"kelvin":  "kelvin", // Kelvin is used for color temperatures, not Celsius.
```
beorn7, line 319: How about: "SI base units, used in special cases (e.g. color temperature, scientific measurements)."

mdlayher:
Done and done, thanks!

brian-brazil:
For Colour temperature I can see the argument, but the whole point of having one chosen base unit so that we don't end up with a mix of Kelvin and Celsius. Accordingly I don't see a reason to make an exception here, as with any other temperature unit it is easy to convert at the display stage.

The wording as proposed here also indicates that Kelvin is in fact preferred, which is not the case.

beorn7:
I actually prefer Kelvin. But I acknowledge that this is not the case for everyone.

I think temperature is special, not only because it involves a number not everyone knows for the top of their head (273.15) but because the conversion involves an addition rather than a multiplication. Common mathematical operations with temperature, like exp(–E/kT) don't make sense with Celsius.

I think the preferred unit depends on the context, and I think that's well expressed by this PR. If you want to improve the wording, please make a suggestion.

brian-brazil:
I don't think temperate is special, and the same sort of  argument could be made that we should allow Fahrenheit and milliseconds and minutes and bits because that's how the implementer thinks of it - which defeats the whole purpose of having exactly one base unit.

If someone would prefer Kelvin displayed in Grafana they're free to do so, but that doesn't mean that that is what also goes in the exposition.

Both the Prometheus docs and the Openmetrics draft only allow for Celsius, and both also happen to explicitly mention Kelvin. Allowing this would be a significant change in best practices, as it'd basically be abandoning the notion of base units.

beorn7:
> I don't think temperate is special, and the same sort of argument could be made that we should allow Fahrenheit and milliseconds and minutes and bits because that's how the implementer thinks of it - which defeats the whole purpose of having exactly one base unit.

I completely disagree with this statement, but I don't want to sink an enormous amount of time into this fairly low-significance discussion.

> Both the Prometheus docs and the Openmetrics draft only allow for Celsius, and both also happen to explicitly mention Kelvin.

Openmetrics isn't finalized, and Prometheus uses the word "should" and "preferred" not "allow".

The code comments here make pretty clear that Celsius is more common than Kelvin.

Still I was never entirely happy with the statement on prometheus.io, but I could tolerate it as merely a recommendation. Now this PR brought up another case that is less far fetched than the low-temperature experiments I remember from my time in the lab.

I think the linter should not complain about Kelvin used as a unit. If the only issue you have here is that the comment in the code (which the user of the linter doesn't even see), then please make a suggestion to improve.

If you want the linter to complain about Kelvin, then we have a problem.

brian-brazil:
> If you want the linter to complain about Kelvin, then we have a problem.

I want the linter to continue to complain about Kelvin, for the same reasons we complain about minutes, bits, and Fahrenheit.

beorn7:
> for the same reasons we complain about minutes, bits, and Fahrenheit.

I have explained why the reasons are different for Kelvin.

I guess we have sufficiently discussed this fairly lightweight issue. Time to call a vote.

prometheus/testutil/promlint/promlint.go
```diff
@@ -328,8 +329,7 @@ var (
 		"days":    "seconds",
 		"weeks":   "seconds",
 		// Temperature.
-		"kelvin":     "celsius",
-		"kelvins":    "celsius",
+		"kelvins":    "kelvin",
 		"fahrenheit": "celsius",
 		"rankine":    "celsius",
```
csmarchbanks, line 334: Just a random thought, should rankine be suggested as kelvin as they are both absolute zero based?

beorn7: That makes sense, but I'd be careful here. We don't really want to push people towards Kelvin. It's only that there are cases where you really cannot _not_ use Kelvin, in which case the lint warnings are really annoying, given there's no "lint-ignore" pragma possible in the Prometheus metrics format.

If you get the lint warning "Use `celsius` instead of `rankine`", and you have strong reasons for using absolute temperature, you are probably deep enough into it that you'll recognize that switching to `kelvin` will still fix the warning.

beorn7:
By voting, we passed the following proposal (11 for, 1 against): “Allow Kelvin as a base unit in certain cases and update our documented recommendation and the linter code accordingly.”

I therefore assume that there are no disagreements left no this PR. (Despite prompting twice, no suggestions to change the code comments were made.)

PR #762, opened 2020-06-03
Improve doc comments in promauto

beorn7:
In particular, point out that `With(nil)` is valid.


bwplotka:
Thanks @beorn7 :+1:

PR #764, opened 2020-06-04
Ensure that nil registers are treat as a no-op, even when wrapping.

tomwilkie:
Passing a `nil` is useful in tests, but our treatment of `nil` is inconsistent.  

This code does not panic:

```go
promauto.With(nil).NewCounter(...)
```

Where as the following does panic.

```go
promauto.With(prometheus.WrapRegistererWith(prometheus.Labels{"foo": "bar"}, nil)).NewCounter(...)
```

### DIFF
--- a/prometheus/promauto/auto_test.go
+++ b/prometheus/promauto/auto_test.go
@@ -1,0 +1,29 @@
+// Copyright 2020 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package promauto
+
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func TestWrapNil(t *testing.T) {
+	// A nil registerer should be treated as a no-op by promauto, even when wrapped.
+	registerer := prometheus.WrapRegistererWith(prometheus.Labels{"foo": "bar"}, nil)
+	c := With(registerer).NewCounter(prometheus.CounterOpts{
+		Name: "test",
+	})
+	c.Inc()
+}
--- a/prometheus/promauto/auto_test.go
+++ b/prometheus/promauto/auto_test.go
@@ -22,8 +22,6 @@ import (
 func TestWrapNil(t *testing.T) {
 	// A nil registerer should be treated as a no-op by promauto, even when wrapped.
 	registerer := prometheus.WrapRegistererWith(prometheus.Labels{"foo": "bar"}, nil)
-	c := With(registerer).NewCounter(prometheus.CounterOpts{
-		Name: "test",
-	})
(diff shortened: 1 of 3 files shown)


prometheus/promauto/auto_test.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func TestWrapNil(t *testing.T) {
+	// A nil registerer should be treated as a no-op by promauto, even when wrapped.
+	registerer := prometheus.WrapRegistererWith(prometheus.Labels{"foo": "bar"}, nil)
+	c := With(registerer).NewCounter(prometheus.CounterOpts{
```
bwplotka, line 25: We can do all in one line as well here

prometheus/promauto/auto_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func TestWrapNil(t *testing.T) {
+	// A nil registerer should be treated as a no-op by promauto, even when wrapped.
+	registerer := prometheus.WrapRegistererWith(prometheus.Labels{"foo": "bar"}, nil)
```
bwplotka, line 24: I think test for Wrap should be in `prometheus` package and there is no need for using promauto there.

In the same time same test might be here for just `With` (:

tomwilkie: I wanted to use `promauto` as that is where the assumption that a `nil` registerer comes from, and I wanted to show that the `Wrap..` functions should also work.  I can add another test in `prometheus/` if you like?

tomwilkie:
I'm a fan of the NoopRegisterer approach too, but I spoke with @beorn7 and he preferred this approach, and I kinda agree with him now.  We already assume nil is safe in promauto, and therefore adding a NoopRegisterer would not allow us to remove these checks, in effect doing it both ways.  Plus this way the zero value for Registerer is a valid, which is mode idiomatic, no?

prometheus/promauto/auto_test.go
beorn7: Yes, would be good to quickly test in the `prometheus` package that wrapping `nil` works.

beorn7:
`Registerer` is an interface type, so I'm not sure if it is correct to talk about a zero value for it. But I do like to encourage that functions receiving a `Registerer` will treat a `nil` value benignly.

It's trivial to implement a no-op implementation of a `Registerer` if you need it. However, if we provided one in the `prometheus` package, it would signal that that's the usual way to go. Then we would inevitably have some people implementing functions that want a `NoopRegisterer` to signal "no instrumentation", while others will implement functions that are also happy with `nil`. I like the latter because users of that function don't even have to import the `prometheus` package.

beorn7:
Will merge on green!

beorn7:
`TestSummaryDecay` has started to become flaky recently. See https://github.com/prometheus/client_golang/issues/765

bwplotka:
Thanks for this discussion, overall I agree with the way to go for now, but I would be curious what would be the best approach for future (e.g `v2`)

> Then we would inevitably have some people implementing functions that want a NoopRegisterer to signal "no instrumentation", while others will implement functions that are also happy with nil. I like the latter because users of that function don't even have to import the prometheus package if they don't want instrumentation.

I don't get why import is the problem. Anyway, I think I agree with this direction going forward. While I definitely like explicitly marking no instrumentation with `Noop`, however, I tend to agree that people will just create wrappers that accept nils... Actually we did that too: https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/extprom/extprom.go#L10

PR #773, opened 2020-06-22
Add support for tsdb endpoint

HimaVarsha94:
Two primary updates:
 - api/prometheus/v1/api.go: Add support for tsdb endpoint
 - api/prometheus/v1/api_test.go: Add test case for tsdb endpoint

Signed-off-by: Hima Varsha <hdureddy@apple.com>

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -138,6 +138,7 @@ const (
 	epConfig          = apiPrefix + "/status/config"
 	epFlags           = apiPrefix + "/status/flags"
 	epRuntimeinfo     = apiPrefix + "/status/runtimeinfo"
+	epTSDB            = apiPrefix + "/status/tsdb"
 )
 
 // AlertState models the state of an alert.
@@ -254,6 +255,8 @@ type API interface {
 	TargetsMetadata(ctx context.Context, matchTarget string, metric string, limit string) ([]MetricMetadata, error)
 	// Metadata returns metadata about metrics currently scraped by the metric name.
 	Metadata(ctx context.Context, metric string, limit string) (map[string][]Metadata, error)
+	// TSDB returns the cardinality statistics.
+	TSDB(ctx context.Context) (TSDBResult, error)
 }
 
 // AlertsResult contains the result from querying the alerts endpoint.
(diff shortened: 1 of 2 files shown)


HimaVarsha94:
cc @beorn7

beorn7:
Thanks @HimaVarsha94 .

@lilic @bwplotka could you have a look? You know this stuff better than me.

api/prometheus/v1/api.go
```diff
@@ -404,6 +407,20 @@ type queryResult struct {
 	v model.Value
 }
 
+// TSDBResult contains the result from querying the tsdb endpoint
+type TSDBResult struct {
```
bwplotka, line 411: I am maybe bias, but I wish we could just import https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L1172

This is fine though I guess - to avoid importing prometheus.

api/prometheus/v1/api.go
```diff
@@ -404,6 +407,20 @@ type queryResult struct {
...
+	SeriesCountByMetricName     []MetricCountPair `json:"seriesCountByMetricName"`
+	LabelValueCountByLabelName  []MetricCountPair `json:"labelValueCountByLabelName"`
+	MemoryInBytesByLabelName    []MetricCountPair `json:"memoryInBytesByLabelName"`
+	SeriesCountByLabelValuePair []MetricCountPair `json:"seriesCountByLabelValuePair"`
+}
+
+// MetricCountPair models information about metric name and count
```
bwplotka, line 418:
```suggestion
// Stat models information about statistic value.
```

api/prometheus/v1/api.go
```diff
@@ -404,6 +407,20 @@ type queryResult struct {
...
+type MetricCountPair struct {
```
bwplotka, line 419: I would vote for renaming this to `Stat` (:

Reasons:

* It is called `stat` in Prometheus
* It's not technically a "metric"
* It's not always a "count" 

What do you think? :hugs:

HimaVarsha94: Makes sense to me. I was focusing on a single use case but this is better.

api/prometheus/v1/api.go
lilic: It might to avoid cyclic dependencies? In any case, I would stick to what already is standard practice here but we could open issue for this to discuss?

PR #778, opened 2020-07-01
Add Filtering Functionality to Registry and HTTP Handler

### LINKED ISSUE #135: Allow filtering via query parameters
For federation we can provide `match[]` query parameters with metric selectors to ask for specific metrics. By default none are shown.

Regular scrape targets should provide the same options but show all metrics by default.

nathanjsweet:
Add a filter function to the Gather interface.
Add a filter generator function to the http options for
the http handler.

Signed-off-by: Nate Sweet <nathanjsweet@pm.me>

This fixes #135

### DIFF
--- a/prometheus/examples_test.go
+++ b/prometheus/examples_test.go
@@ -642,7 +642,7 @@ humidity_percent{location="inside"} 33.2
 temperature_kelvin{location="somewhere else"} 4.5
 `
 
-	parseText := func() ([]*dto.MetricFamily, error) {
+	parseText := func(filter prometheus.GatherFilter) ([]*dto.MetricFamily, error) {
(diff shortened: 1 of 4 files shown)


nathanjsweet:
We are talking about merging this into cilium [here](https://github.com/cilium/cilium/pull/12369)

beorn7:
This is unfortunately a breaking change and can only be implemented in v2. That's why #135 is in the v2 milestone.

This also has to be coordinated with other client libraries. We don't want every instrumented binary to require a different ad-hoc way of filtering metrics.

I'm very sorry that we don't even have a plan yet how this will look like. I'll try to push this up on our agenda so that we something sketched out in the not too far future.

In the meantime, I'm afraid, you have to work with your own fork if you need the feature right now.

PR #779, opened 2020-07-06
fix tests warning about string(int) type conversions

johejo:
See https://github.com/golang/go/issues/32479


johejo:
Oh, that's right, my fix is wrong.
I'll fix it soon.

johejo:
Fixed.

PR #784, opened 2020-07-22
Remove reference to Metric after reslicing

hummerd:
If we remove metric from slice, last value in old `metrics` still has value of `metricWithLabelValues` and holds reference to `Metric` (see Note in https://github.com/golang/go/wiki/SliceTricks#delete-without-preserving-order).
```
m.metrics[h] = append(metrics[:i], metrics[i+1:]...)
```
So we must zero last value in old metrics slice after we ressliced `metrics`.
```
old := metrics
m.metrics[h] = append(metrics[:i], metrics[i+1:]...)
old[len(old)-1] = metricWithLabelValues{}
```

Signed-off-by: Dima Kozlov <hummerd@mail.ru>

### DIFF
--- a/prometheus/vec.go
+++ b/prometheus/vec.go
@@ -276,7 +276,9 @@ func (m *metricMap) deleteByHashWithLabelValues(
 	}
 
 	if len(metrics) > 1 {
+		old := metrics
 		m.metrics[h] = append(metrics[:i], metrics[i+1:]...)
+		old[len(old)-1] = metricWithLabelValues{}
 	} else {
 		delete(m.metrics, h)
 	}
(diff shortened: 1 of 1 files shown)


hummerd:
Thanks for accepting PR!

We had actual issue, but i'm not sure that prometheus client was the cause of it. We had memory leak problem that happened from time to time with our service and one of last modification to this service was adding metrics. We create a huge GaugeVec (150k+ different labels) with a code like this:
```
		for ch, rec := range newRecords {
			var timeout, _ = strconv.ParseFloat(rec.Timeout, 64)
			if timeout > 0 {
				chTTLGauge.WithLabelValues(ch).Set(timeout)
			} else {
				chTTLGauge.DeleteLabelValues(ch)
			}
		}
```
So I started to look if it can cause problems and found code with deleting metrics. 

But! Now we made some performance improvements to our code and it seems like problem is gone (even with old version of client). So i think it problem was in our code, not in prometheus client!

beorn7:
Thanks for the context.

PR #786, opened 2020-07-30
Report user and system cpu times

steinz:
Hello @beorn7!

This PR adds user and system CPU timing counters and was written to be consistent with the node.js client library ([equivalent code](https://github.com/siimon/prom-client/blob/master/lib/metrics/processCpuTotal.js)).

### DIFF
--- a/prometheus/process_collector.go
+++ b/prometheus/process_collector.go
@@ -23,6 +23,8 @@ type processCollector struct {
 	pidFn           func() (int, error)
 	reportErrors    bool
 	cpuTotal        *Desc
+	cpuUser         *Desc
+	cpuSys          *Desc
 	openFDs, maxFDs *Desc
 	vsize, maxVsize *Desc
 	rss             *Desc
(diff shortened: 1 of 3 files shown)


brian-brazil:
The node.js client should be consistent with this library, rather than the other way around. `process_` is discussed at https://prometheus.io/docs/instrumenting/writing_clientlibs/#process-metrics, which doesn't include these metrics.

steinz:
Thanks for the link and feedback!
I'm investigating using `process_cpu_seconds_total` for the dashboards I care about, but I'll be back if I find a compelling use case for exposing system and user cpu usage separately 🙂

beorn7:
Thanks for your contribution @steinz . To change the `process_...` metrics, the first step is indeed to add it to https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/writing_clientlibs.md#process-metrics . I recommend to open a PR there, which your reasoning, or – usually better for more involved discussions – go to #prometheus-dev on irc.freenode.net or post it on the [prometheus-developers mailing list](https://groups.google.com/forum/#!forum/prometheus-developers).

beorn7:
I'll close this PR for now as it is not actionable in the context of this repo. Feel free to reference this PR in any upcoming discussion, and we can re-open or re-create should the common spec for `process_...` metrics change.

PR #793, opened 2020-08-22
Replace with the standard library constant.

johncming:
The Standard Library later added a 422 status code. 

https://github.com/golang/go/commit/b9ec0024fbc18dd94eff7240afd82fac6b4d8fdc


beorn7 (approved):
Thank you.

PR #795, opened 2020-08-31
Update go.mod

shakertanim:


lilic:
@shakertanim can you clarify the reason why this update is needed? Seems like CI is not a fan of it. :)

beorn7:
Thanks @lilic for chiming in.

In my understanding of Go Modules, we _have_ to add the `v2` to the path in the go.mod file. See https://github.com/golang/go/wiki/Modules#semantic-import-versioning .

I'm closing this, assuming I'm right, but if you have more explanations, @shakertanim , please follow up on this issue.

Issue #801, opened 2020-09-08
Error querying Prometheus: bad_data: cannot parse "" to a valid timestamp


kayrosik:
This is part of my function QueryRange:

```
	start := time.Now()
	end := now.Add(time.Duration(-config.Prometheus.MetricsTimeRange) * time.Minute)
        resolution := math.Max(math.Floor(end.Sub(start).Seconds()/250), 1)
	step := time.Duration(resolution * 1e9)

	v1api := v1.NewAPI(client)
	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
	defer cancel()
	r := v1.Range{
		Start: start,
		End:   end,
		Step:  step,
	}
	result, warnings, err := v1api.QueryRange(ctx, metric, r)
```

I use trickster + prometheus(Thanos). 
When I try to call the function, I get:
`Error querying Prometheus: bad_data: cannot parse "" to a valid timestamp`

And I can't understand why it can be (

I change DoGetFallback function (always HTTP-Get):
```
	resp, body, warnings, err := h.Do(ctx, req)
	u.RawQuery = args.Encode()
	req, err = http.NewRequest(http.MethodGet, u.String(), nil)
	if err != nil {
		return nil, nil, warnings, err
	}
	return h.Do(ctx, req)
```

And it works! The problem is POST-request.

beorn7:
It's is actually recommended to use POST for the query and query-range endpoints because it avoids complications with length restrictions of the URL. (PromQL queries can become quite long.)

Maybe Trickster or Thanos don't support POST requests for those endpoints?

But even then, the HTTP client should transparently fallback to GET. Perhaps something doesn't work as intended with the fallback.

My understanding is limited here. This was implemented back in #557 by @jacksontj . Or perhaps @lilic @bwplotka or @joe-elliott would have an idea?

joe-elliott:
`DoGetFallback` relies on the server returning a 405.  @kayrosik can you confirm what status code the POST is returning?  My guess is that `Error querying Prometheus: bad_data: cannot parse "" to a valid timestamp` is a 400.

Also guessing that error is returned by Trickster based on the text, but it would be nice to confirm.

kayrosik:
Status-code is 400

beorn7:
OK, that explains it then. Thanks @joe-elliott for the catch.

@kayrosik I recommend to file an issue with Trickster to either support POST (as it is recommended anyway) or return a 405 so that the client will properly retry.

Issue #794, opened 2020-08-24
Allowed HTTP methods shall be limited on metrics 


4ndris:
### Host operating system: output of `uname -a`

```Linux ca543372ca59 4.19.76-linuxkit prometheus/node_exporter#1 SMP Tue May 26 11:42:35 UTC 2020 x86_64 GNU/Linux```

### node_exporter version: output of `node_exporter --version`

```node_exporter, version 1.0.1 (branch: HEAD, revision: 3715be6ae899f2a9b9dbfd9c39f3e09a7bd4559f)```

### node_exporter command line flags
```/bin/node_exporter --path.rootfs=/host```

### Are you running node_exporter in Docker?
yes

It seems that node_exporter exposes metrics endpoints without limiting allowed HTTP methods, this way allowing to be invoked with GET, POST, PUT, DELETE. It is recommended to limit all the allowed methods to GET, HEAD (and POST in case its necessary) otherwise it shall return 405.

roidelapluie:
Yes, I have moved the issue. Thanks @4ndris.

beorn7:
Currently, the `http.Handler` implementations provided by `promhttp` are indeed completely ignorant of methods (you could even use made-up method names). The handler always behaves exactly the same. While I wouldn't claim that was a deliberate decision, what's exactly the impact or problem with that?

> It is recommended to limit all the allowed methods to GET, HEAD (and POST in case its necessary) otherwise it shall return 405.

While I could see people arguing that way, what is the source of this recommendation? A quick web search only reveals overeager "security" scanners complaining about certain methods being allowed. But I'm totally willing to learn about a recommendation of some weight and/or following a convincing line of argument.

In principle, I think it would not do any harm to only serve metrics on GET, but at this point, it will be hardly possible to restrict the method in the handler without a major version bump. I can see legitimate cases for using other methods than GET and HEAD (if users have wrapped the stock metrics handler with something else and read parameters from a POST form, for example), but more commonly I'd expect accidental usage of other methods that would suddenly stop to work. Since the current `promhttp` documentation nowhere requires the GET method, one might argue the currently (implicitly) documented behavior is to allow all methods, even if only GET makes sense semantically. 

[...]

PR #802, opened 2020-09-10
API client: Enable fallback on status code 501, too

beorn7:
When discussing #801, I remembered #794. While dealing with the
latter, I read the HTTP RFC, stumbling upon the following:

    When a request method is received
    that is unrecognized or not implemented by an origin server, the
    origin server SHOULD respond with the 501 (Not Implemented) status
    code.  When a request method is received that is known by an origin
    server but not allowed for the target resource, the origin server
    SHOULD respond with the 405 (Method Not Allowed) status code.

Concluding from that, it is possible that a server desiring a fallback
to GET will send a status code of 501. It is even preferred if that
server does not offer any resource to be used with the POST method.

Therefore, I think we should fallback to GET on a 501, too.

@joe-elliott it would be great if you could have a look here, too.

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -1004,7 +1004,8 @@ func (h *apiClientImpl) Do(ctx context.Context, req *http.Request) (*http.Respon
 
 }
 
-// DoGetFallback will attempt to do the request as-is, and on a 405 it will fallback to a GET request.
+// DoGetFallback will attempt to do the request as-is, and on a 405 or 501 it
+// will fallback to a GET request.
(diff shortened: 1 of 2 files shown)


joe-elliott:
That's an annoying definition :).  501 is a superset of 405.  If the server returns 501 you don't know if the method was unrecognized or not implemented.

Will take a look at this when I get a chance.

joe-elliott (approved):
Looks good to me.  Good job digging through the bowels of the HTTP RFC to find this one.

beorn7:
Thanks @joe-elliott .

Issue #796, opened 2020-08-31
Feature Request: Permit histogram's Observe() method to take an observation count


lavalamp:
We had a case where we in theory update a histogram at some frequency. In actuality, we only update it when things are happening, and when nothing is happening, we just count the amount of time that has passed and then update it with that many ~zeros~ [edit: observations] when the next thing happens.

Right now adding a value to a historgram N times takes N calls of Observe(). It would be very nice if we could pass N as a parameter instead, and if I follow the code, it could be done in constant time.

lavalamp:
cc @MikeSpreitzer @logicalhan

MikeSpreitzer:
@lavalamp you don't mean "update it with that many zeros", "zero" here is not operative.  We call `Observe` the desired number of times.

MikeSpreitzer:
Thanks for opening this.

lavalamp:
Sorry, you're right, that was a confusing level of detail, and it looks like I misunderstood the code, too :)

brancz:
cc @beorn7

beorn7:
Apologies for late (post vacation) reply.

I don't quite get the context here. If nothing is happening, why do you need to perform _n_ observations when the next thing is happening? Would that next thing just require that one observations? Nothingness should require no observations, no? (o:

MikeSpreitzer:
We want the effect of making Histogram Observations of a certain variable at a highish rate.  The implementation is that whenever the variable being observed changes, we call the histogram's Observe method N times, where N is the number of observation period boundaries that have been crossed since the variable last changed.

beorn7:
OK, to rephrase: You want to sample a relatively rarely changing measured value faster than you could reasonably scrape it with Prometheus, right?

I mean, if you want to call `Observe(X)` for example every second, I'd recommend to simply expose X directly as a Gauge and let Prometheus scrape it every second. You then have the more fundamental time series data in Prometheus and can act in any way you want on it.

The Summary and Histogram are more meant for "actual" observations that can happen at any time rather than "artificially generated" observations that are just regular samples (or "mini scrapes") of what's essentially a Gauge.

But let's assume that your use case is really one where you need to do that, there is still a problem: A Prometheus endpoint can be scraped at any time. Let's imagine the following timeline:

- t=0s: X changes to 42. `Observe(42)` gets called _n_ times according to your formula above.
- t=5s: Endpoint gets scraped by a Prometheus server.
[...]

PR #803, opened 2020-09-10
Export MetricVec (again)

beorn7:
MetricVec was already exported in early versions of this library, but
nobody really used it to implement vectors of custom Metric
implementations. Now #796 has shown up with a fairly special yet valid use case
for which I'd prefer a custom implementation of a special
"auto-sampling histogram" outside of this library. Therefore, I'd like
to reinstate support for creating vectors of custom Metric
implementations.

@MikeSpreitzer @lavalamp is that helpful? If yes, I'd be grateful for a review of this PR.

I played around for quite some while with the option of a separate
package providing the tools one would need to create vectors of custom
Metric implementations. However, with the current structure of the
prometheus/client_golang/prometheus package, this leads to a lot of
complications with circular dependencies. (The new package would need
the primitives from the prometheus package, while the existing metric
vectors like GaugeVec need to import the new vector package to not
duplicate the entire MetricVec implementation. Separating vector types from the main
prometheus package is out of the question at this point because that
would be a breaking change.)

### DIFF
--- a/prometheus/vec.go
+++ b/prometheus/vec.go
@@ -20,12 +20,20 @@ import (
 	"github.com/prometheus/common/model"
 )
 
-// metricVec is a Collector to bundle metrics of the same name that differ in
-// their label values. metricVec is not used directly (and therefore
-// unexported). It is used as a building block for implementations of vectors of
-// a given metric type, like GaugeVec, CounterVec, SummaryVec, and HistogramVec.
-// It also handles label currying.
-type metricVec struct {
+// MetricVec is a Collector to bundle metrics of the same name that differ in
+// their label values. MetricVec is not used directly but as a building block
+// for implementations of vectors of a given metric type, like GaugeVec,
+// CounterVec, SummaryVec, and HistogramVec. It is exported so that it can be
+// used for custom Metric implementations.
+//
+// To create a FooVec for custom Metric Foo, embed a pointer to MetricVec in
+// FooVec and initialize it with NewMetricVec. Implement wrappers for
+// GetMetricWithLabelValues and GetMetricWith that return (Foo, error) rather
+// than (Metric, error). Similarly, create a wrapper for CurryWith that returns
+// (*FooVec, error) rather than (*MetricVec, error). It is recommended to also
+// add the convenience methods WithLabelValues, With, and MustCurryWith, which
+// panic instead of returning errors. See also the MetricVec example.
+type MetricVec struct {
 	*metricMap
 
 	curry []curriedLabelValue
@@ -35,9 +43,9 @@ type metricVec struct {
 	hashAddByte func(h uint64, b byte) uint64
 }
 
-// newMetricVec returns an initialized metricVec.
-func newMetricVec(desc *Desc, newMetric func(lvs ...string) Metric) *metricVec {
-	return &metricVec{
+// NewMetricVec returns an initialized metricVec.
+func NewMetricVec(desc *Desc, newMetric func(lvs ...string) Metric) *MetricVec {
+	return &MetricVec{
 		metricMap: &metricMap{
 			metrics:   map[uint64][]metricWithLabelValues{},
 			desc:      desc,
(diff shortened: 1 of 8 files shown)


MikeSpreitzer:
I would like to try developing against this.  Is there some `go.mod` tweak I can make in https://github.com/MikeSpreitzer/kubernetes/tree/add-sampling-histogram that will get me this?

beorn7:
If you run `go get github.com/prometheus/client_golang@d1387ee3d99a5d16c8640663b6ae63663f6ae3b0` in the root of your repo (where the `go.mod` file is), your `go.mod` file should be updated accordingly. The relevant line in the `require` section should look like `github.com/prometheus/client_golang v1.7.2-0.20200910160544-d1387ee3d99a` then.

MikeSpreitzer (approved):
I was able to develop a client using this, see https://github.com/MikeSpreitzer/kubernetes/blob/add-sampling-histogram-vec/staging/src/k8s.io/component-base/metrics/prometheusextension/sampling-histogram.go (based off https://github.com/kubernetes/kubernetes/pull/94672).

beorn7:
@MikeSpreitzer Glad to see that. Your approach looks good in general. I'll comment in #796 in more detail.

I'll get a second opinion from someone of prometheus-team before I merge this. E.g. @bwplotka as a heavy user of client_golang, do you think this makes sense to add this "construction kit for metric vectors" to client_golang?

beorn7:
@bwplotka :bellhop_bell: Could you have a look and let me know if you think this is useful to have in client_golang?

beorn7:
@bwplotka another attempt to get your attention… :pray:

lavalamp:
cc @logicalhan

prometheus/vec.go
```diff
@@ -20,12 +20,19 @@ import (
...
-// metricVec is a Collector to bundle metrics of the same name that differ in
-// their label values. metricVec is not used directly (and therefore
-// unexported). It is used as a building block for implementations of vectors of
-// a given metric type, like GaugeVec, CounterVec, SummaryVec, and HistogramVec.
-// It also handles label currying.
-type metricVec struct {
+// MetricVec is a Collector to bundle metrics of the same name that differ in
+// their label values. MetricVec is not used directly but as a building block
+// for implementations of vectors of a given metric type, like GaugeVec,
+// CounterVec, SummaryVec, and HistogramVec. Is is exported so that it can be
```
bwplotka, line 26:
```suggestion
// CounterVec, SummaryVec, and HistogramVec. It is exported so that it can be
```
(:

prometheus/vec.go
```diff
@@ -20,12 +20,19 @@ import (
...
+// used for custom Metric implementations. To create a FooVec for custom Metric
```
bwplotka, line 27: Super useful comment. I would suggesting adding new line to separate this tip for `Vec` implementation for custom type with upper summary?

prometheus/vec.go
```diff
@@ -95,15 +102,32 @@ func (m *metricVec) Delete(labels Labels) bool {
...
+// (GetMetric)WithLabelValues methods). It is possible to curry a curried
+// vector, but only with labels not yet used for currying before.
+//
+// The metrics contained in the MetricVec are shared between the curried and
+// uncurried vectors. They are just accessed differently. Curried and uncurried
+// vectors behave identically in terms of collection. Only one must be
+// registered with a given registry (usually the uncurried version). The Reset
+// method deletes all metrics, even if called on a curried vector.
+//
+// Note that CurryWith is usually not called directly but through a wrapper
```
bwplotka, line 127: :+1:

bwplotka (approved):
Love it, just a tiny suggestion and typo to fix and LGTM :tada: 

Sorry for the lag, I am behind all GH notifications, accidentally seen this ping... 

 

Will do my best next time, if I'm not responding after 3-4d feel free to ping me on Slack (:

prometheus/vec.go
beorn7: Sure. Done.

prometheus/vec.go
beorn7: Fixed.

beorn7:
Thanks @bwplotka . I have addressed the comments and will merge on green.

PR #806, opened 2020-10-15
Cut v1.8.0

beorn7:
@lilic @lavalamp @MikeSpreitzer


beorn7:
Hmm, different Go versions create a different `go.mod` in the corresponding test.

@roidelapluie @simonpasquier I vaguely remember that one of you had a solution for that. Do you?

beorn7:
Was it that we just removed the `go mod` test from older versions?

roidelapluie:
They are forward compatible, so you can run go mod tidy on an old version IIRC

bwplotka (approved):
Awesome, thanks! :+1: 

Not sure about the go.mod version incompatibility. The go.mod change visible on CI fail seems not important. It might be that some dependency is having some additional file build only for pre Go 1.12  versions with this indirect dep :thinking:

beorn7:
So, as expected, if I generate go.mod with an older Go version, the tests for the newer versions will complain.

Next try: Remove the module test from older versions.

beorn7:
OK, that worked.

PR #808, opened 2020-10-19
Add prototype Go runtime mixin

sdboyer:
First pass at a monitoring mixin based on the runtime metrics emitted by the default Go collector. Quite bare bones, but it's just intended as a start, without any glaring flaws. It doesn't cover every emitted metric, but it gets the biggies - heap and stack use, number of goroutines, gc durations, etc.

The panels in the dashboard apply no aggregations. I waffled on this a bit, but opted for that approach on the premise that the dashboard is basic enough, and the panels show few enough individual metrics, that the dashboard itself is most likely to be helpful for folks really just getting started. (That, and, they can use the job/instance template variables to drill into a particular instance if they really need to see something that's stuck.) Very open to persuasion on changing this to an aggregated approach.

There's basic CI included to lint the jsonnet (which is almost entirely a passthrough to exported Grafana dashboard JSON). No rules or alerts as there's no way of specifying them when we're abstracted so far away in a client library.

cc @beorn7


beorn7:
Will have a look ASAP.

README.md
```diff
@@ -43,6 +43,8 @@ The
 [`examples` directory](https://github.com/prometheus/client_golang/tree/master/examples)
 contains simple examples of instrumented code.
 
+The [`mixin` directory](https://github.com/prometheus/client_golang/tree/master/mixin) contains a monitoring mixin - a a basic dashboard for visualizing Go runtime metrics emitted by processes that use this library.
```
kakkoyun, line 46:
```suggestion
The [`mixin` directory](https://github.com/prometheus/client_golang/tree/master/mixin) contains a monitoring mixin - a basic dashboard for visualizing Go runtime metrics emitted by processes that use this library.
```

mixin/README.md
```diff
@@ -0,0 +1,30 @@
+# Go Runtime Mixin
+
+_This mixin is a work in progress. We aim for it to become a good role model for
+dashboards eventually, but it's not there yet._
+
+Mixins are a collection of configurable, reusable Prometheus rules, alerts
+and/or Grafana dashboards for a particular system, usually created by experts
+in that system. By [applying them]() to Prometheus and Grafana, you can quickly
```
beorn7, line 8: Link missing.

mixin/dashboards/go-runtime.json
```diff
@@ -0,0 +1,1153 @@
...
+      "2d",
+      "7d",
+      "30d"
+    ]
+  },
+  "timezone": "",
+  "title": "Go runtime metrics",
+  "uid": "CgCw8jKZz3",
+  "version": 3
+}
```
beorn7, line 1153: No newline at end of file.

beorn7 (changes requested):
Some minor nits. Will follow up with general questions as separate comments.

mixin/README.md
```diff
@@ -0,0 +1,30 @@
...
+set up appropriate monitoring for your systems.
+
+This mixin is for Go applications, and contains a dashboard for visualzing the
```
beorn7, line 11: visualzing → visualizing

beorn7:
General question before doing a detailed review of the dashboard panels: In the other mixins, we use the `grafana-builder` or the `grafonnet` libraries to build dashboards. Here we have just the bare dashboard JSON. In that just a WIP step or is the intention to ship dashboards as raw dashboard JSON? I'm not an expert for the Grafana part, but I'd assume in the raw dashboard JSON form, it's much harder to override, amend, or modify dashboards than in the Jsonnet form used for grafana-builder or grafonnet.

beorn7:
Minor directory layout request: I'd prefer to not create another top-level directory. In other mixins, we have them underneath `docs` or `documentation`. The `example` directory is perhaps quite close to that. Could we move the `mixin` directory into it, adjusting its description accordingly? Perhaps also call it `go-mixin` or `go-runtime-mixin` as this more specific naming is what we have done in other mixins so far.

sdboyer:
Thanks for the review!

A bit more clarity, before you look at the dashboard itself - i elected not to try to create panels for all the metrics (particularly the more abstruse ones), and more could be done to explain them. The bar we've been aiming for with the first pass is "not obviously wrong."

> In the other mixins, we use the `grafana-builder` or the `grafonnet` libraries to build dashboards.

We've been experimenting with just providing raw JSON, rather than generated jsonnet, as a way of trying to make creating these more accessible - it's generally held that jsonnet is one of the barriers to entry with mixins. i do feel like this form is a bit clunkier, but as this is really intended to be a first pass that we can iteratively improve, i figured i'd try out this pattern.

> I'd assume in the raw dashboard JSON form, it's much harder to override, amend, or modify dashboards than in the Jsonnet form used for grafana-builder or grafonnet.

i have heard some concerns expressed in a similar vein. I'm a bit circumspect about such amendments and overrides in general - i think they're clunky, error-prone, and don't compose particularly well. That doesn't make it not a real concern, of course.

For now, i'm inclined to keep with the "just use JSON" experiment, and see how it pans out for a bit. However, if you have a strong preference, i'm happy to put them in the traditional form.

> The `example` directory is perhaps quite close to that. Could we move the mixin directory into it, adjusting its description accordingly?

i haven't noticed that in the rash of recent mixin work, but i've no particularly strong opinion on this. Will make the move.

> Perhaps also call it `go-mixin` or `go-runtime-mixin` as this more specific naming is what we have done in other mixins so far.

This was bothering me, too. I'll do `go-runtime-mixin`, as that's really what this is.

mixin/README.md
sdboyer: whoops!

sdboyer: i think i'm actually just gonna remove this for now - eventually i want that to be a link, but the docs to which i would link aren't really written yet

beorn7:
> it's generally held that jsonnet is one of the barriers to entry with mixins

I can totally believe that. But what's the plans with recording and alerting rules, then? If we still provide them as jsonnet, then we still expose our users to jsonnet (but without making use of it in the dashboarding part of the mixin).

Providing a raw dashboard JSON file also raises the question why we have the Makefile/mixtool/jsonnetfmt contraption wrapped around it if in the end the dashboard JSON that it creates is exactly the same as the dashboard JSON provided. (I understand that it will make sense once we have more than just that one dashboard JSON file, but before we have that, it's really WIP.)

In any case, I'll have a look at the dashboard queries now, as that review makes sense, no matter in which form we'll later distribute the dashboard.

beorn7 (changes requested):
Here just notes about the individual dashboard expressions.

Possibly stupid question by someone not terribly familiar with Grafana: Why are there two `"yaxes"` each time? It appears to me we are only every using the first.

examples/go-runtime-mixin/dashboards/go-runtime.json
```diff
@@ -0,0 +1,1153 @@
...
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "rate(go_memstats_mallocs_total{job=\"$job\", instance=~\"$instance\"}[$__interval])",
```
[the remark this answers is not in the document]
sdboyer: i was wondering about this, and figured i probably got it wrong with just `$__interval` and not setting a min step.

AFAIK we don't have any restrictions, but i'll double-check. (Seems to me that providing a nudge to update oughtn't be a bad thing.)

sdboyer:
> Possibly stupid question by someone not terribly familiar with Grafana: Why are there two `"yaxes"` each time? It appears to me we are only every using the first.

Not sure why it did that. i didn't actually mess with the axes at all. Might've been a relic of the upstream dashboard i initially pulled from.

Will make these changes!

sdboyer:
> I can totally believe that. But what's the plans with recording and alerting rules, then? If we still provide them as jsonnet, then we still expose our users to jsonnet (but without making use of it in the dashboarding part of the mixin).

YAML for those, is the idea. See prometheus/mysqld_exporter#503 for an example. It just didn't seem like i really had any to do, here.

> Providing a raw dashboard JSON file also raises the question why we have the Makefile/mixtool/jsonnetfmt contraption wrapped around it if in the end the dashboard JSON that it creates is exactly the same as the dashboard JSON provided. (I understand that it will make sense once we have more than just that one dashboard JSON file, but before we have that, it's really WIP.)

Another good question, to which i have no satisfying answer. Thoughts, @tomwilkie ?

i should note that @malcolmholmes has a (local) experiment in [grizzly](https://github.com/grafana/grizzly) that can do a bidirectional json<>jsonnet transform (though the completeness of the json->jsonnet transform is a function of the completeness of the [dashboard spec](https://github.com/grafana/dashboard-spec)). Having that will, at least for me, address a lot of the concern about asking people to express dashboards in jsonnet.

malcolmholmes:
> i should note that @malcolmholmes has a (local) experiment in grizzly that can do a bidirectional json<>jsonnet transform (though the completeness of the json->jsonnet transform is a function of the completeness of the dashboard spec). Having that will, at least for me, address a lot of the concern about asking people to express dashboards in jsonnet.

Yes - with this we could render the JSON as Jsonnet, but just use it as a readable rendering of a dashboard, not as something that people necessarily actually code. I could see this both in Grizzly (`grizzly get --jsonnet`) but also as a tab in the JSON Model page. Simply because it will be easier to digest that JSON.

beorn7:
I might missing the point of the whole exercise here.

It was always possible to hand out dashboard JSON files or example rules in plain Yaml files to tell people about ways to setup their dashboards and alerts.

However, as it turned out, that wasn't terribly useful beyond inspirations or starting points because people had to adjust dashboards and rules to their local setup. Some templating mechanism had to be created.

Which was the starting point for the monitoring mixins effort, which picked Jsonnet as the templating mechanism of choice.

Now Jsonnet is obviously not a silver bullet and comes with issues and tradeoffs, might that be the adoption barrier or more fundamental concerns about the best way of modifying/combining/extending configurations.

But what's happening in this PR or https://github.com/prometheus/mysqld_exporter/pull/503 looks like going back to square one and provide just dashboard JSON and rules Yaml as examples once more, which the added overhead of `jsonnetfmt` (with no Jsonnet in the game?!?).

If the idea is that Jsonnet isn't the best choice for templating, then sure, let's discuss another solution. But this looks like reverting the existing approach _without_ presenting a different solution to the original problem.

Perhaps this is all just an iteration towards a real solution. That's also fine, but then I'd ask you to keep the changes of this PR in a branch until we have something that has at least the potential of being a working solution. This repository is one of the most used Go repositories at all, and anything we release here, even if marked as just an intermediate step of an experiment, will inevitably be taken at face value by many, which could be quite confusing.

sdboyer:
i'd call that a good summary of the history. And an apt characterization of the oddity of our current position, though i would make one note:

> looks like going back to square one and provide just dashboard JSON and rules Yaml as examples once more

Sort of square one, but things are different today because there exists this de-facto standard form for the `mixin.libsonnet` file, it has seen use (even if not as widespread as we might like), and there is tooling built around it, and - most crucially - it still does meet that goal of allowing a (presumably) trusted expert to provide a baseline, in contrast to the wild and wooly marketplace.

This is significant to the extent that it gives us an interface to operate behind - we _can_ do these experiments, and as long as we're not taking away any templating capability that is present in any existing mixin, we haven't actually made anything worse.

As to Jsonnet:

> If the idea is that Jsonnet isn't the best choice for templating, then sure, let's discuss another solution. But this looks like reverting the existing approach without presenting a different solution to the original problem.

When @brancz @metalmatze @tomwilkie et al. met a couple weeks ago to discuss the state of affairs, we generally agreed that even if jsonnet is part of the problem, it's not clearly enough of the problem that changing it would obviously produce a better outcome.

Increasingly, my belief is that the file formats stuff is a red herring because templating is the wrong way to think about what's we're actually doing. When we offer a template file from a repo, then expect that file to be templated and applied to running Grafana/Prom instances as fully-realized sets of rules and alerts, all context about what degrees of freedom were intended by the original template have been lost, but users are still confronted with the same editable dashboard interface as always.

The implicit premise is that the user has made all the relevant choices during templating, because they're doing infrastructure-as-code. But people - and i cannot emphasize this enough - _do not write whole dashboards as code_. At most, they create the dashboards in the UI, then export them to code, then _maybe_ make small changes to them directly in code. But even that is quite onerous, and if a dashboard needs to be significantly modified, it's back to the UI, then re-export to code. That's all hard enough when it's just people maintaining their own orgs' dashboards; trying to incorporate a mixin from a third-party compounds that further.

IMO, for the goals of mixins to be properly realized, choice about how to consume and fix free variables from within a mixin must be deferred, until it can be presented to the user in the place where they actually make the rest of their choices: the UI. That's not incompatible with infra-as-code, though it does take some reimagning. i'm working on this. 

All that said...

> This repository is one of the most used Go repositories at all, and anything we release here, even if marked as just an intermediate step of an experiment, will inevitably be taken at face value by many, which could be quite confusing.

This is an excellent point, and i can't make any serious counterargument. So, i'm going to close this PR. We'll put this mixin in [jsonnet-libs](https://github.com/grafana/jsonnet-libs) until we have better answers to these questions, and/or i have a chance to convert it to jsonnet 😄

beorn7:
Thanks @sdboyer . This all makes sense to me.

Perhaps a note about my role in the various Prometheus related mixins: I got penciled in as the "maintainer" of those, but mostly because I took some care (and still have to do way more of that) for getting the Prometheus and Grafana "content" (alerts, recording rules, dashboards, PromQL, …) right (or at least not blatantly wrong), not so much because I would have a lot of context on the mixin framework part. I took the mixin framework as a given and just filled in PromQL expressions and such. This also makes it hard for me to review PRs like https://github.com/prometheus/node_exporter/pull/1875 , which doesn't really touch the Prometheus content, but requires context of how the mixins should be used.

PR #812, opened 2020-11-05
Add unique label lint

ipstatic:
Prometheus will fail scrapes if a metric has more than one label with
the same name. However promtool does not report this failure currently.
This change will produce an error if a metric does not have unique label
names.

Signed-off-by: Jarod Watkins <jarod@42lines.net>

### DIFF
--- a/prometheus/testutil/promlint/promlint.go
+++ b/prometheus/testutil/promlint/promlint.go
@@ -119,6 +119,7 @@ func lint(mf *dto.MetricFamily) []Problem {
 		lintReservedChars,
 		lintCamelCase,
 		lintUnitAbbreviations,
+		lintUniqueLabels,
 	}
 
 	var problems []Problem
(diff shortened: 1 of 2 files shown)


brian-brazil:
We should be catching this as a parse failure over in the parser in common (which is the reference implementation), rather than as a lint failure here.

ipstatic:
> We should be catching this as a parse failure over in the parser in common (which is the reference implementation), rather than as a lint failure here.

If I move this to there, will `promtool check metrics` return an error if there is a metric with duplicate labels?

bwplotka:
Yes as long as we will fix same decoer: https://github.com/prometheus/client_golang/blob/405bb9b05f1a8235b98de85fd5cb0a0d597a3120/prometheus/testutil/promlint/promlint.go#L24 (:

ipstatic:
I have created a PR in common [here](https://github.com/prometheus/common/pull/263), however now when I run `promtool check metrics` with my updated common library, I do not get a failure with metrics that have duplicate labels.

beorn7:
Are you perhaps compiling from the vendored files in prometheus/prometheus? Or in Go module mode, where it takes the released version of prometheus/common?

In any case, since this will be solved in prometheus/common, I'm closing this PR.

ipstatic:
Can you recommend how I could go about testing that? I have added overrides in the `replace()` section of go.mod for the various modules I touched and attempt to run `promtool` with those overrides but I don't know if I am doing it correctly. When I run `promtool` with metrics that do have duplicate labels, it still returns with no error.

ipstatic:
It must have been how I was overriding, I updated `go.mod` to use the newest released version of common and it works as expected.

PR #813, opened 2020-11-13
removes the unnecessary switch statements from both the sanitizeMetho…

Schweppesale:
removes the unnecessary switch statements from both the sanitizeMethod + sanitizeCode functions

### DIFF
--- a/prometheus/promhttp/instrument_server.go
+++ b/prometheus/promhttp/instrument_server.go
@@ -320,128 +320,15 @@ func computeApproximateRequestSize(r *http.Request) int {
 }
 
 func sanitizeMethod(m string) string {
-	switch m {
-	case "GET", "get":
-		return "get"
-	case "PUT", "put":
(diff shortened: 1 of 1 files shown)


bwplotka (commented):
Looks good generally, however not sure why we need those methods at the end after this refactor. Seems extremely shallow to me (: 

cc @beorn7

brian-brazil:
I have a very vague recollection that this was a performance thing.

beorn7:
These methods might be in a very hot path, so we want to avoid allocating new strings.

bwplotka:
Can we at least add the comment that this is for allocs? Otherwise we will have the same PR every month (:

beorn7:
We can do that. (But I predict that it will provoke discussions if it's really better to do it this way… we had those PRs as well in the past. Avoiding allocations was certainly more important when this code was written.)

PR #814, opened 2020-11-15
Don't use newlines in error messages

SuperQ:
Don't use newlines for error formatting, avoids newlines in logs.

Related to https://github.com/prometheus/node_exporter/issues/1886

Signed-off-by: Ben Kochie <superq@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
+## unreleased / unreleased
+
+* [BUGFIX] Don't use newlines in error messages
+
 ## 1.8.0 / 2020-10-15
 
 * [CHANGE] API client: Use `time.Time` rather than `string` for timestamps in `RuntimeinfoResult`. #777
 * [FEATURE] Export `MetricVec` to facilitate implementation of vectors of custom `Metric` types. #803
-* [FEATURE API client: Support `/status/tsdb` endpoint. #773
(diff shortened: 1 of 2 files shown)


beorn7:
The current behavior is not a bug, it's a feature.

We might discuss if multi-line error messages are really so bad that it is worth changing the behavior (risking to break users – if someone depends on the current behavior, they will certainly claim it's a breaking change).

However, if we really want to touch the behavior of the multi-error type used here, we might as well adjust it to new stuff in https://golang.org/pkg/errors/ and perhaps make the behavior customizable. See https://github.com/hashicorp/go-multierror and https://github.com/uber-go/multierr/ for more or less popular other implementations. Note that the former also creates a multi-line string by default.

beorn7:
I thought about this. Sadly, we cannot sneak in a formatter because the current `MultiError` is just a `[]error` slice with methods. Changing it to a `struct` would again be a breaking change.

All those breakages are kind of soft, but this is also one of the most used Go packages in the known universe, so I'm afraid it will affect some users in surprising ways. I don't really want to break anyone if the benefit is merely adhering to a convention that is not even universally accepted.

Note that I have created #821 to document the behavior more clearly and describe a work around. I have also filed #820 to fix the problem when we can, i.e. in the next major release.

PR #785, opened 2020-07-24
Add headers and basic authentication

rsevilla87:
Provide basic authentication and a mechanism to add custom headers to the API requests.

Signed-off-by: Raul Sevilla <rsevilla@redhat.com>

PR #816, opened 2020-11-19
Add configuration option to include headers

jasonmartens:
Currently the go http client doesn't have any way of configuring authentication for protected endpoints. This allows setting the Authorization header as well as any other headers the client might need. 
@beorn7

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -40,6 +40,9 @@ type Config struct {
 	// The address of the Prometheus to connect to.
 	Address string
 
+	// Extra headers to add to each request (ex: authorization)
+	Headers map[string]string
+
 	// RoundTripper is used by the Client to drive HTTP requests. If not
 	// provided, DefaultRoundTripper will be used.
 	RoundTripper http.RoundTripper
@@ -68,14 +71,23 @@ func NewClient(cfg Config) (Client, error) {
 	}
 	u.Path = strings.TrimRight(u.Path, "/")
 
+	var h map[string]string
+	if cfg.Headers != nil {
+		h = cfg.Headers
+	} else {
+		h = map[string]string {}
+	}
+
 	return &httpClient{
 		endpoint: u,
+		headers: h,
 		client:   http.Client{Transport: cfg.roundTripper()},
 	}, nil
 }
 
 type httpClient struct {
 	endpoint *url.URL
+	headers map[string]string
 	client   http.Client
 }
 
@@ -96,6 +108,9 @@ func (c *httpClient) URL(ep string, args map[string]string) *url.URL {
 func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte, error) {
 	if ctx != nil {
 		req = req.WithContext(ctx)
+	}
+	for name, contents := range c.headers {
+		req.Header.Add(name, contents)
 	}
 	resp, err := c.client.Do(req)
 	defer func() {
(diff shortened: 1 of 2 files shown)


api/client.go
```diff
@@ -40,6 +40,9 @@ type Config struct {
 	// The address of the Prometheus to connect to.
 	Address string
 
+	// Extra headers to add to each request (ex: authorization)
```
beorn7, line 43: Consistency nits: Please end with a period and use `e.g.` instead of `ex:`.

api/client.go
```diff
@@ -68,14 +71,23 @@ func NewClient(cfg Config) (Client, error) {
 	}
 	u.Path = strings.TrimRight(u.Path, "/")
 
+	var h map[string]string
+	if cfg.Headers != nil {
+		h = cfg.Headers
+	} else {
+		h = map[string]string{}
+	}
```
beorn7, line 79: Iterating over a `nil` map (in line 112) is just fine, so you don't need this. Also see comment about line 83 below.

api/client.go
```diff
@@ -68,14 +71,23 @@ func NewClient(cfg Config) (Client, error) {
...
+
 	return &httpClient{
 		endpoint: u,
+		headers:  h,
```
beorn7, line 83: I guess you can just write `headers: cfg.Headers,` here and don't need the indirection via `h`.

beorn7 (changes requested):
Thank you.

I left a few minor comments, but let's also clarify if this is the way to go in general. I guess you could also provide an `http.RoundTripper` that injects the headers. That would definitely be the way for things like TLS config. But perhaps it would be too cumbersome if you only want to add a simple header entry? We had #785 before, which @lilic chimed in on. @lilic what do you think about this one?

beorn7:
To extend on the above, after a quick web search, this is probably how header injection via the RoundTripper would look like:

```Go

// headerAdder is an http.RoundTripper that adds additional headers to the request
type headerAdder struct {
    headers map[string][]string

    rt http.RoundTripper
}

func (h headerAdder) RoundTrip(req *http.Request) (*http.Response, error) {
    for k, vv := range h.headers {
        for _, v := range vv {
            req.Header.Add(k, v)
        }
    }
    return h.rt.RoundTrip(req)
}

cfg := api.Config {
    Address: "http://demo.robustperception.io:9090",
    RoundTripper: headerAdder {
        rt: http.DefaultTransport,
        headers: map[string][]string{"foo": []string{"bar", "baz"}},
    }
}
```

api/client.go
```diff
@@ -40,6 +40,9 @@ type Config struct {
 	// The address of the Prometheus to connect to.
 	Address string
 
+	// Extra headers to add to each request (ex: authorization)
+	Headers map[string]string
```
beorn7, line 44: HTTP headers can have multiple values, so this should probably be `map[string][]string`.

beorn7:
@bwplotka has just now proposed #817. Those examples make a lot of sense to me.

bwplotka:
Yes, you don't need any of this. Our code already makes it possible. In fact common lib has amazing helpers as well to make it easy to use (:

bwplotka (changes requested):
As mentioned above, I would vote for having one way of doing the same thing (e.g injecting auth token)

jasonmartens:
Hello all.
I did consider using the RoundTripper, but then saw the note that the request shouldn't be modified, so attempted this approach instead. However, if it's possible to use the RoundTripper approach that certainly works for me. I'm new to Go so I trust your judgement here.

Issue #822, opened 2020-12-04
promhttp instrumented handler hangs with the invalid metric name


AlekSi:
https://github.com/prometheus/client_golang/blob/47cfdc9bb8ceaf18b12a4b7ba778b06e12d1a97a/prometheus/promhttp/instrument_server.go#L228-L232
This code does not check what error is returned. If an invalid metric name is passed, the process consumes all memory and is killed by the OOM killer.

I found it when I passed sub-test's `t.Name()` as a part of the metric name: `/` can't a part of the metric name.

PR #823, opened 2020-12-09
promhttp: Correctly detect invalid metric and label names

beorn7:
Without this fix, the `InstrumentHandler...` middlewares get locked in
an endless loop in case of an invalid Collector, eating all the memory.

@AlekSi


beorn7:
Fixes #822.

beorn7:
@stuartnelson3 in case you feel bored and you like the blast from the past, you could have a look here.

AlekSi (approved):
:shipit:

beorn7:
Thanks @AlekSi . I'll check if there is anything else to wrap up before I cut a new release in the next couple of days.

PR #828, opened 2020-12-29
Support matchers in labels API

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

As pr https://github.com/prometheus/prometheus/pull/8301 was merged, this pr adds `matches` to `LabelNames` and `LabelValues` APIs. 

One thing I am not very sure is the testing, seems we are just mocking the HTTP response in the unit tests, it cannot really ensure correctness. But I already test this manually and it works.

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -230,9 +230,9 @@ type API interface {
 	DeleteSeries(ctx context.Context, matches []string, startTime time.Time, endTime time.Time) error
 	// Flags returns the flag values that Prometheus was launched with.
 	Flags(ctx context.Context) (FlagsResult, error)
-	// LabelNames returns all the unique label names present in the block in sorted order.
+	// LabelNames returns the unique label names present in the block in sorted order by given time range and matchers.
 	LabelNames(ctx context.Context, matches []string, startTime time.Time, endTime time.Time) ([]string, Warnings, error)
-	// LabelValues performs a query for the values of the given label.
+	// LabelValues performs a query for the values of the given label, time range and matchers.
 	LabelValues(ctx context.Context, label string, matches []string, startTime time.Time, endTime time.Time) (model.LabelValues, Warnings, error)
 	// Query performs a query for the given time.
 	Query(ctx context.Context, query string, ts time.Time) (model.Value, Warnings, error)
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -231,9 +231,9 @@ type API interface {
 	// Flags returns the flag values that Prometheus was launched with.
 	Flags(ctx context.Context) (FlagsResult, error)
 	// LabelNames returns all the unique label names present in the block in sorted order.
-	LabelNames(ctx context.Context, startTime time.Time, endTime time.Time) ([]string, Warnings, error)
+	LabelNames(ctx context.Context, matches []string, startTime time.Time, endTime time.Time) ([]string, Warnings, error)
 	// LabelValues performs a query for the values of the given label.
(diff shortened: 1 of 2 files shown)


api/prometheus/v1/api.go
```diff
@@ -231,9 +231,9 @@ type API interface {
 	// Flags returns the flag values that Prometheus was launched with.
 	Flags(ctx context.Context) (FlagsResult, error)
 	// LabelNames returns all the unique label names present in the block in sorted order.
-	LabelNames(ctx context.Context, startTime time.Time, endTime time.Time) ([]string, Warnings, error)
+	LabelNames(ctx context.Context, matches []string, startTime time.Time, endTime time.Time) ([]string, Warnings, error)
```
bwplotka, line 234: This is technically a breaking change. I am not sure what are the rules for this project cc @beorn7 

Do we allow this or not in single major version?

bwplotka (commented):
Hmm I think overall LGTM, except the compatibility break. Not sure what is the way forward here.

In terms of tests we test what we have to test, no? We check if parameters are passed correctly to the underlying URL query.

api/prometheus/v1/api.go
yeya24: > In terms of tests we test what we have to test, no? We check if parameters are passed correctly to the underlying URL query.

Now only HTTP method and path are tested https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api_test.go#L71-L76. We don't check parameters.

yeya24:
> In terms of tests we test what we have to test, no? We check if parameters are passed correctly to the underlying URL query.

Now only HTTP method and path are tested https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api_test.go#L71-L76. We don't check parameters.

api/prometheus/v1/api.go
beorn7: The whole API client is marked as experimental and not covered by the semantic versioning.

(In fact, I'd much prefer to have the API client on a separate release schedule. But as long as the API client is effectively pre-1.0, it doesn't really matter. Once we are ready to declare the API client as stable, we can revisit the decision if we want to keep it bundled with the rest of client_golang or separate it out (into a new repo or a separate Go module in a sub-directory).)

api/prometheus/v1/api.go
```diff
@@ -231,9 +231,9 @@ type API interface {
 	// Flags returns the flag values that Prometheus was launched with.
 	Flags(ctx context.Context) (FlagsResult, error)
 	// LabelNames returns all the unique label names present in the block in sorted order.
```
beorn7, line 233: I think this doc comment should be adjusted (it apparently didn't get adjusted when the time range was added, and now we have even matchers).

beorn7 (commented):
No worries about the breaking change. That's still OK with the API client.

@bwplotka I leave it to you to approve and merge this once you are satisfied.

api/prometheus/v1/api.go
yeya24: Got it. I will update this.

jotak:
Hello @yeya24  @bwplotka , thank you for this
Do you think it can be merged before the next client release?

yeya24:
Updated. Please take a look @beorn7 @bwplotka if you have time.

beorn7:
Thanks @yeya24 . I have no other comments, but I'd much prefer if @bwplotka could give his final approval.

beorn7 (approved):
No word from @bwplotka . I'll merge this now.

zhamdoctor:
> and path are tested

image
image
image
image
max(delta(go_memstats_alloc_bytes[5m])),this promql was transferred to unicode pattern,and causes an error.

PR #833, opened 2021-01-22
Bump prometheus/procfs to 0.3.0 to fix building on riscv64

zhsj:
This picks up https://github.com/prometheus/procfs/pull/325


beorn7 (approved):
Sure, we can do this now.

Note that each release usually includes an update of the dependencies anyway.

PR #839, opened 2021-02-23
Make InstrumentHandlerCounter return a http.Handler.

tomwilkie:
This is backwards incompatible I think, so lets not merge - if we do want to do this I'll introduce a new function.


beorn7:
Do you have an example where you need this?

Anything accepting an `http.Handler` should alse accept `http.HandlerFunc` because the latter implements the former. (But the opposite is not true.)

The existing code follows the "accept interfaces, return concrete types" paradigm.

tomwilkie:
Yes you're right, I shouldn't need this.  Let me try again.

PR #841, opened 2021-03-06
Add buildinfo method

ntk148v:


beorn7:
@bwplotka or @lilic , as my resident API reviewers, could I bother you with having a look here?

Don't worry about the go1.9 and go1.10 test failures. I'll remove the support prior to the next release.

api/prometheus/v1/api.go
```diff
@@ -281,6 +284,16 @@ type ConfigResult struct {
 // FlagsResult contains the result from querying the flag endpoint.
 type FlagsResult map[string]string
 
+// BuildinfoResult contains the results from queyring the buildinfo endpoint
```
lilic, line 287: Nit:
```suggestion
// BuildinfoResult contains the results from querying the buildinfo endpoint.
```

api/prometheus/v1/api.go
```diff
@@ -281,6 +284,16 @@ type ConfigResult struct {
...
+type BuildinfoResult struct {
+	Version   string `json:"version"`
+	Revision  string `json:"revision"`
+	Branch    string `json:"branch"`
+	BuildUser string `json:"buildUser"`
+	BuildDate string `json:"buildDate"`
```
lilic, line 293: This should be type `time.Time` instead.

lilic (commented):
Nice, thank you! Few comments only.

api/prometheus/v1/api.go
ntk148v: At first, I've though the same :smile: it is string actually.

https://github.com/prometheus/prometheus/blob/main/web/api/v1/api.go#L109

lilic: Interesting, wonder why the difference. 🤔 Thanks!

lilic (approved):
looks good to me, thank you!

api/prometheus/v1/api.go
beorn7: It's because it's a string in https://github.com/prometheus/common/blob/main/version/info.go , and I guess it's a string there because it's populated from command line flags.

beorn7:
@lilic thank you very much for the review.

@ntk148v we need the DCO on your second commit, too, see https://github.com/prometheus/client_golang/runs/2083213476 .

Or you squash both commits into one and force-push.

api/prometheus/v1/api.go
lilic: > and I guess it's a string there because it's populated from command line flags.

That makes sense, thanks!

beorn7 (approved):
Thank you everyoe.

PR #843, opened 2021-03-11
Update dependencies

beorn7:
In preparation for the next release.

This also bumps the minimum required Go version to 1.13, which is required by the prometheus/procfs package using the new
`%w` printf directives. However, it also allows us to remove some special casing about build info.

This also adds 1.16 to the testing. Thus, we still support the last four minor releases, which should be plenty.


beorn7:
Not sure why the old test still show up. They have been removed from the circleci config.

beorn7:
Perhaps because only the 2nd commit removed them?

beorn7:
Ping?

SuperQ (approved):
LGTM

SuperQ:
I updated the required CI tests.

.circleci/config.yml
```diff
@@ -78,6 +62,10 @@ workflows:
         name: go-1-15
         go_version: "1.15"
         run_lint: true
```
roidelapluie, line 64:
```suggestion
        run_lint: false
```

roidelapluie (approved):

beorn7:
@SuperQ :
> I updated the required CI tests.

What did you do?

roidelapluie:
In github you can mark tests as `required`, in which case they will always be present in the UI and must pass. They would appear even if they disappear in circleci.

beorn7:
I had a vague recollection of that setting, but I could not find it anywhere. Am I too stupid for GH's level of UX?

beorn7:
OK, found it. It's a branch protection rule...

roidelapluie:
Settings > Branches > Branch Protection rules > master - Edit > Require status checks to pass before merging

PR #845, opened 2021-03-17
Cut v1.10.0

beorn7:
Mostly because of API client features. But it has been a while, and the dependency bump alone is worth it.


beorn7:
@SuperQ usually we don't call out changes that happen due to the dependency update, unless of course it is significant enough to call it out. I don't think we need to mentioned the improved and more consistent error formatting in procfs (which triggered the Go version bump), but if there is anything else you remember, let me know.

SuperQ (approved):
LGTM. Agreed, I don't think there's a need to call out the `%w` change.

beorn7:
OK, thanks.

I have updated the date to today and will merge on green.

PR #855, opened 2021-04-11
Add newer fields to Rules API

gouthamve:
https://github.com/prometheus/prometheus/pull/6503/files added new fields to the API that the client wasn't using. I've updated it now.


beorn7 (approved):
Thanks.

(Obligatory note that I only "best-effort-reviewed" this because I'm not the best person to maintain HTTP API code. If anyone spots concerns here, please let me know.)

PR #856, opened 2021-04-11
Add missing fields to targets API

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

Just noted that the Targets response is not up-to-date with the latest Prometheus targets API response. https://github.com/prometheus/prometheus/blob/main/web/api/v1/api.go#L755


beorn7 (approved):
Thank you.

PR #870, opened 2021-05-26
Simplify sanitisation

### LINKED ISSUE #869: Simplify sanitisation code in instrument_server.go
Not sure if I'm missing the big picture, but some sanitization methods in `instrument_server.go` look redundant:

* https://github.com/prometheus/client_golang/blob/a7515ca7c9c6388a5ab84ea336faef795bbf866f/prometheus/promhttp/instrument_server.go#L333-L354
* https://github.com/prometheus/client_golang/blob/a7515ca7c9c6388a5ab84ea336faef795bbf866f/prometheus/promhttp/instrument_server.go#L359-L458

Maybe these could be simplified as:

```go
func sanitizeMethod(m string) string {
	return strings.ToLower(m)
}

func sanitizeCode(s int) string {
	if s == 0 {
		return "200"
	}
	return strconv.Itoa(s)
}
```

I doubt that switch/case over particular string constants would bring any noticeable performance benefits than calling `ToLower` on any input string (same about `Itoa`).

zserge:
Closes #869

### DIFF
--- a/prometheus/promhttp/instrument_server.go
+++ b/prometheus/promhttp/instrument_server.go
@@ -331,128 +331,15 @@ func computeApproximateRequestSize(r *http.Request) int {
 }
 
 func sanitizeMethod(m string) string {
-	switch m {
-	case "GET", "get":
-		return "get"
-	case "PUT", "put":
-		return "put"
-	case "HEAD", "head":
-		return "head"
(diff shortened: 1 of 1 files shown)


beorn7:
Duplicate of #813, closing for the same reason.

PR #871, opened 2021-05-26
Document implications of negative observations

### LINKED ISSUE #850: Document implications of negative observations for Summaries and Histograms
tl;dr: If the `Observe` method is ever called with a negative value, the sum of observations goes down. With the right (un)lucky timing of scrapes, this will lead to a `..._sum` time series in Prometheus that doesn't behave like a counter anymore. 

Historically, there have been different ideas among the Prometheus developers how to deal with that. See https://github.com/prometheus/prometheus/issues/6669 for a detailed discussion. The conclusion there is that we keep the status quo, which means that in the case above, you cannot meaningfully apply a `rate` or `increase` to the `..._sum` time series. Had this outcome been anticipated, we would have disallowed negative observations in the first place, but we cannot change that in v1 now (and hopefully, in the future, Prometheus will solve this better). What we have to do now is to document implications. Broadly along the following lines:

- Negative observations are extremely rare in practice.
- In the unlikely case you need them anyway for some reason, that's fine, but note that the `..._sum` time series ending up in Prometheus might go down then and thus doesn't behave like a counter anymore. So you cannot use `rate` or `increase` on them.
- If you really need _both_, i.e. negative observations _and_ applying `rate` or `increase` to the `..._sum` series, there is the work around of creating _two_ Summaries/Histograms, one for the positive observations and one for the negative observations (with inverted sign). You have to combine the results from both with appropriate PromQL expressions.

(And if really anyone needs the latter and has questions about it, then I hereby commit to write a blog post about it.)

beorn7:
Fixes #850.

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -47,7 +47,12 @@ type Histogram interface {
 	Metric
 	Collector
 
-	// Observe adds a single observation to the histogram.
+	// Observe adds a single observation to the histogram. Observations are
+	// usually positive or zero. Negative observations are accepted but
+	// prevent current versions of Prometheus from properly detecting
+	// counter resets in the sum of observations. See
+	// https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations
+	// for details.
 	Observe(float64)
 }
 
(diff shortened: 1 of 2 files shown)


beorn7:
Gentle :bellhop_bell: @bwplotka .
(Or @kakkoyun – as introduction to new duties… ;)

kakkoyun (approved):
LGTM :100:

PR #877, opened 2021-06-07
Cut v1.11.0

### LINKED ISSUE #876: Release v1.11.0
Since the changes to drop github.com/go-kit/kit have been merged, it would be great to tag a new release to encourage downstream importers to pick up the new version. Without doing this, the change is not effective since *all* dependencies need to be updated to now import the large dependency to have any impact.

See https://github.com/prometheus/common/issues/255 for context.

Note: the PR to drop the dependency is already merged, I was just hoping to tag the release as tagged releases are *much* more likely to be adopted.

kakkoyun:

Fixes #876 

cc @beorn7 Could you have look at it in case I've missed anything? It's my first release 😅


kakkoyun:
cc @lilic

lilic:
Thank you! 🎉

bwplotka (approved):
LGTM! 💪🏽

beorn7 (commented):
Looks good.

Just my nits, feel free to ignore or take into account for the future as you see fit:
* So far, the lines end with periods.
* We usually put changes first, then features, then enhancements, then bugfixes.
* Not sure if I would call the new collectors package a change. But then it includes a deprecation, so it might be best to call it out as a change.

Thinking about the last point: Best might be to call out the new dbstats collector as a feature, and mention that it is in a new collectors package. And then call out the deprecation as a change in a separate entry.

PR #881, opened 2021-06-17
Updating dependency versions + cleanup

sivabalan:
Please review @bwplotka @kakkoyun and merge.


go.sum
```diff
@@ -96,57 +186,284 @@ github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXf
...
+google.golang.org/genproto v0.0.0-20190911173649-1774047e7e51/go.mod h1:IbNlFCBrqXvoKpeg0TB2l7cyZUmoaFKYIwrEpbDKLA8=
+google.golang.org/genproto v0.0.0-20191108220845-16a3f7862a1a/go.mod h1:n3cpQtvxv34hfy77yVDNjmbRyujviMdxYliBSkLhpCc=
+google.golang.org/genproto v0.0.0-20191115194625-c23dd37a84c9/go.mod h1:n3cpQtvxv34hfy77yVDNjmbRyujviMdxYliBSkLhpCc=
+google.golang.org/genproto v0.0.0-20191216164720-4f79533eabd1/go.mod h1:n3cpQtvxv34hfy77yVDNjmbRyujviMdxYliBSkLhpCc=
+google.golang.org/genproto v0.0.0-20191230161307-f3c370f40bfb/go.mod h1:n3cpQtvxv34hfy77yVDNjmbRyujviMdxYliBSkLhpCc=
+google.golang.org/genproto v0.0.0-20200115191322-ca5a22157cba/go.mod h1:n3cpQtvxv34hfy77yVDNjmbRyujviMdxYliBSkLhpCc=
+google.golang.org/genproto v0.0.0-20200122232147-0452cf42e150/go.mod h1:n3cpQtvxv34hfy77yVDNjmbRyujviMdxYliBSkLhpCc=
+google.golang.org/genproto v0.0.0-20200204135345-fa8e72b47b90/go.mod h1:GmwEX6Z4W5gMy59cAlVYjN9JhxgbQH6Gn+gFDQe2lzA=
+google.golang.org/genproto v0.0.0-20200212174721-66ed5ce911ce/go.mod h1:55QSHmfGQM9UVYDPBsyGGes0y52j32PQ3BqQfXhyH3c=
+google.golang.org/genproto v0.0.0-20200224152610-e50cd9704f63/go.mod h1:55QSHmfGQM9UVYDPBsyGGes0y52j32PQ3BqQfXhyH3c=
```
bwplotka, line 410: This feels very unnecessary. Can you describe what you did? Can you try to remove `go.sum` and run `go mod tidy`?

bwplotka (commented):
Thank! One suggestion (:

go.sum
sivabalan: Thanks for the quick review @bwplotka. Dependency on prometheus/common 0.26.0 added a bunch of older dependencies (some of them with known vulnerabilities). Dependency tree generated from go-kit 0.10.0. 
So all I did was update the version of prometheus/common to 0.29.0 and run go mod tidy after removing go.sum. 

Did that again as per your suggestion and there is no change in the generated go.sum.

sagikazarmark: I think this is mostly because of the circular dependency between common and client_golang.

sagikazarmark:
Anything else to be done here?

go.sum
beorn7: Yeah, known issue with Go Modules dependency resolution. None of this code will be compiled

kakkoyun (approved):
Thanks for the contribution. It seems like there's nothing much more to be done here.
With the lazy module loading that's coming to Go, I hope our dependency problem will be less intrusive.

PR #886, opened 2021-06-23
Switch sparse histograms to base-2 buckets

beorn7:
This seem what OTel is converging towards, see
open-telemetry/oteps#149 .

I see pros and cons with base-10 vs base-2. They are discussed in
detail in that OTel PR, and the gist of the discussion is pretty much
in line with my design doc. Since the balance is easy to tip here, I
think we should go with base-2 if OTel picks base-2. This also seems
to be in agreement with several proprietary solution (see again the
discussion on that OTel PR.)

The idea to make the number of buckets per power of 2 (or formerly 10)
a power of 2 itself was also sketched out in the design doc
already. It guarantees mergeability of different resolutions. I was
undecided between making it a recommendation or mandatory. Now I think
it should be mandatory as it has the additional benefit of playing
well with OTel's plans.

This commit also addresses a number of outstanding TODOs.

This depends on https://github.com/prometheus/client_model/pull/47 (go.mod currently pinned to that commit).

@codesome note that this is not going to master, but just to the highly experimental sparsehistogram branch. I still wanted to create a PR so that you can chime in.

### DIFF
--- a/examples/random/main.go
+++ b/examples/random/main.go
@@ -54,10 +54,10 @@ var (
 	// normal distribution, with 20 buckets centered on the mean, each
 	// half-sigma wide.
 	rpcDurationsHistogram = prometheus.NewHistogram(prometheus.HistogramOpts{
-		Name:                    "rpc_durations_histogram_seconds",
-		Help:                    "RPC latency distributions.",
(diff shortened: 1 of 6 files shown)


codesome (commented):
:+1:

PR #893, opened 2021-07-22
Adding extension testutils.ToInt()

kuritka:
@bwplotka, @kakkoyun

During testing I need to read metrics which are very often (in our case mostly) integer values.
Prometheus offers only the `testutils.ToFloat64` function, so all results need to be continuously re-typed into integers.
This PR only extends the ToFloat functionality by adding the ToInt function.

Signed-off-by: kuritka <kuritka@gmail.com>

### DIFF
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -112,6 +112,11 @@ func ToFloat64(c prometheus.Collector) float64 {
 	panic(fmt.Errorf("collected a non-gauge/counter/untyped metric: %s", pb))
 }
 
+// ToInt returns the integer value of the collected Metric. For more information see ToFloat64.
+func ToInt(c prometheus.Collector) int {
+	return int(ToFloat64(c))
+}
+
 // CollectAndCount registers the provided Collector with a newly created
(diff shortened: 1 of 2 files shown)


kuritka:
@bwplotka, @kakkoyun , seems circleci is pending. Should I do some additional steps ? 
thx

bwplotka (changes requested):
Hi, thanks for your effort! But really it's just one line function - to me it's too shallow to have a dedicated function for this. Anyone can just do `int` anytime that want. (: 

I would prefer not merging it, there is little value here. What do you think? 🤗

kuritka:
@bwplotka, thanks for the review although it is rejected. In a way, you're right, but everyone will now have to implement this simple minor feature themselves. A bit of usefulness is the only reason to put it in the utilities, regardless of the small amount of code it contains. 

The change is not crucial for me, so I'll leave it as it is. 

Thx and good luck

bwplotka:
Thanks!

It's better to minimize the stuff we ship to thousand people to bare high quality minimum. Simpler = better.

You can always have quick local function in your project.

While we are here I would definitely change method name.. toFloat is not super helpful 🙈

PR #899, opened 2021-08-04
add ExponentialBucketsRange function

sbunce:
This function calculates exponential buckets with different arguments
than the existing ExponentialBuckets function. Instead of specifying the
start and factor, the user can specify the min and max bucket value. We
have been doing it this way internally at my company for some time.

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -116,6 +116,34 @@ func ExponentialBuckets(start, factor float64, count int) []float64 {
 	return buckets
 }
 
+// ExponentialBucketsRange creates 'count' buckets, where the lowest bucket is
+// 'min' and the highest bucket is 'max'. The final +Inf bucket is not counted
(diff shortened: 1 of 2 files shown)


dswarbrick:
For the record, tests which expect bit-identical floating points will probably fail on s390x and possibly some other less mainstream archs, due to subtle differences in Go's floating point implementation on those archs, i.e. https://go.dev/ref/spec#Floating_point_operators

I'm seeing the exponential bucket range test fail on Debian's s390x buildds, and I suspect that it's for the same reason that this fails https://github.com/gohugoio/hugo/issues/6387

sbunce:
@dswarbrick 
I did not know that! Thank you for educating me. I created a change which switches to a comparison with a tolerance "abs(a-b)<tolerance". The tolerance I picked "0.001" looks higher than the differences in your test failure output.

The PR is here.
https://github.com/prometheus/client_golang/pull/1133

PR #901, opened 2021-08-31
Implement strategy to limit the sparse bucket count

beorn7:
_This is merely a PR against the sparsehistogram branch._

@dieterbe @codesome Finally a more or less clean version of the bucket limitation strategy.


beorn7:
This is using go1.15 features, which explains why the tests for older go versions fail.

beorn7:
I'll merge this now to cut a Cortex gateway with this.

PR #940, opened 2021-11-25
registry: permit the collection of the same metric with multiple timestamps

mvisonneau:
When a timestamp if being associated to the metric, use it as part of the checksum. This will permit the multiple collection of a single metric with different timestamps. It can be useful when attempting to generate a dump. As an example:

```go
package main

import (
	"net/http"
	"time"

	"github.com/prometheus/client_golang/prometheus/promhttp"
)

type myCollector struct {
	metric *prometheus.Desc
}

func (c *myCollector) Describe(ch chan<- *prometheus.Desc) {
	ch <- c.metric
}

func (c *myCollector) Collect(ch chan<- prometheus.Metric) {
	t := time.Now().Add(-1 * time.Hour)
	ch <- prometheus.NewMetricWithTimestamp(t, prometheus.MustNewConstMetric(c.metric, prometheus.CounterValue, 123))

	t = time.Now().Add(-2 * time.Hour)
	ch <- prometheus.NewMetricWithTimestamp(t, prometheus.MustNewConstMetric(c.metric, prometheus.CounterValue, 456))
}

func main() {
	collector := &myCollector{
		metric: prometheus.NewDesc(
			"my_metric",
			"This is my metric with custom TS",
			nil,
			nil,
		),
	}

	r := prometheus.NewRegistry()
	r.MustRegister(collector)
	http.Handle("/metrics", promhttp.HandlerFor(r, promhttp.HandlerOpts{EnableOpenMetrics: true}))
	http.ListenAndServe(":8000", nil)
}
```

output:

```shell
~$ curl http://localhost:8000/metrics
my_metric 456 1637838943285
my_metric 123 1637842543285
```

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"unicode/utf8"
@@ -897,6 +898,14 @@ func checkMetricConsistency(
 		h.WriteString(lp.GetValue())
 		h.Write(separatorByteSlice)
 	}
+
+	// If a timestamp if being associated to the metric, use it as part of the checksum.
+	// This will permit the multiple collection of a single metric with different timestamps.
+	// Useful when attempting to generate a dump.
+	if dtoMetric.TimestampMs != nil {
+		h.WriteString(strconv.FormatInt(*dtoMetric.TimestampMs, 10))
+	}
+
 	hSum := h.Sum64()
 	if _, exists := metricHashes[hSum]; exists {
 		return fmt.Errorf(


beorn7:
This is unfortunately not valid Prometheus exposition format.

https://prometheus.io/docs/instrumenting/exposition_formats/#grouping-and-sorting reads: "Each line must have a unique combination of a metric name and labels. Otherwise, the ingestion behavior is undefined."

Prometheus itself tries to be tolerant, and it might even ingest this exposition as expected, but that might change at any time.

kakkoyun:
Closing this, as the reason stated above.

yiqinguo:
How to add a value to CounterValue, not Set.

PR #962, opened 2022-01-17
promhttp: Check validity of method and code label values

kakkoyun:

### DIFF
--- a/prometheus/promhttp/instrument_client.go
+++ b/prometheus/promhttp/instrument_client.go
@@ -50,7 +50,8 @@ func InstrumentRoundTripperInFlight(gauge prometheus.Gauge, next http.RoundTripp
 // The CounterVec must have zero, one, or two non-const non-curried labels. For
 // those, the only allowed label names are "code" and "method". The function
 // panics otherwise. For the "method" label a predefined default label value set
-// is used to filter given values. `additionalMethods` can be used to add more
+// is used to filter given values. Values besides predefined values will count
+// as `unknown` method.`WithExtraMethods` can be used to add more
 // methods to the set. Partitioning of the CounterVec happens by HTTP status code
 // and/or HTTP method if the respective instance label names are present in the
 // CounterVec. For unpartitioned counting, use a CounterVec with zero labels.
@@ -81,7 +82,8 @@ func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.Rou
 // ObserverVec.  The ObserverVec must have zero, one, or two non-const
 // non-curried labels. For those, the only allowed label names are "code" and
 // "method". The function panics otherwise. For the "method" label a predefined
-// default label value set is used to filter given values. `additionalMethods`
+// default label value set is used to filter given values. Values besides
+// predefined values will count as `unknown` method. `WithExtraMethods`
 // can be used to add more methods to the set. The Observe method of the Observer
 // in the ObserverVec is called with the request duration in
 // seconds. Partitioning happens by HTTP status code and/or HTTP method if the
--- a/prometheus/promhttp/instrument_client.go
+++ b/prometheus/promhttp/instrument_client.go
@@ -49,7 +49,9 @@ func InstrumentRoundTripperInFlight(gauge prometheus.Gauge, next http.RoundTripp
 // http.RoundTripper to observe the request result with the provided CounterVec.
 // The CounterVec must have zero, one, or two non-const non-curried labels. For
 // those, the only allowed label names are "code" and "method". The function
-// panics otherwise. Partitioning of the CounterVec happens by HTTP status code
+// panics otherwise. For the "method" label a predefined default label value set
+// is used to filter given values. `additionalMethods` can be used to add more
+// methods to the set. Partitioning of the CounterVec happens by HTTP status code
 // and/or HTTP method if the respective instance label names are present in the
 // CounterVec. For unpartitioned counting, use a CounterVec with zero labels.
 //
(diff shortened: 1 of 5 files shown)


prometheus/promhttp/instrument_client.go
```diff
@@ -49,21 +49,23 @@ func InstrumentRoundTripperInFlight(gauge prometheus.Gauge, next http.RoundTripp
...
+// methods to the set. Partitioning of the CounterVec happens by HTTP status code
 // and/or HTTP method if the respective instance label names are present in the
 // CounterVec. For unpartitioned counting, use a CounterVec with zero labels.
 //
 // If the wrapped RoundTripper panics or returns a non-nil error, the Counter
 // is not incremented.
 //
 // See the example for ExampleInstrumentRoundTripperDuration for example usage.
-func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.RoundTripper) RoundTripperFunc {
+func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.RoundTripper, additionalMethods ...string) RoundTripperFunc {
```
bwplotka, line 62: Should we start the options pattern at this point? This would be safe for future options. Otherwise we would need major version bump of function deprecations.

I mean something like:

```
type Option func(*option)

WithExtraMethods([]string) Option
```
and
 
```suggestion
func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.RoundTripper, o ...Option) RoundTripperFunc {
```

?

prometheus/promhttp/instrument_client.go
```diff
@@ -49,21 +49,23 @@ func InstrumentRoundTripperInFlight(gauge prometheus.Gauge, next http.RoundTripp
 // http.RoundTripper to observe the request result with the provided CounterVec.
 // The CounterVec must have zero, one, or two non-const non-curried labels. For
 // those, the only allowed label names are "code" and "method". The function
-// panics otherwise. Partitioning of the CounterVec happens by HTTP status code
+// panics otherwise. For the "method" label a predefined default label value set
+// is used to filter given values. `additionalMethods` can be used to add more
+// methods to the set. Partitioning of the CounterVec happens by HTTP status code
```
bwplotka, line 54:
```suggestion
// panics otherwise. For the "method" label a predefined default label value set
// is used to filter given values. Values outside of predefined values will count as `unknown` method. `WithExtraMethods` can be used to add more
// methods to the set. Partitioning of the CounterVec happens by HTTP status code
```

bwplotka (commented):
This is epic! Proposed a more flexible API, otherwise, it's perfect, thanks!

prometheus/promhttp/instrument_client.go
kakkoyun: Even though current changes wouldn't also break the existing method calls, I like using options pattern. It would be more flexible. I'll update the PR.

prometheus/promhttp/instrument_server.go
```diff
@@ -348,15 +394,27 @@ func sanitizeMethod(m string) string {
 		return "options"
 	case "NOTIFY", "notify":
 		return "notify"
+	case "TRACE", "trace":
+		return "trace"
+	case "PATCH", "patch":
+		return "patch"
 	default:
-		return strings.ToLower(m)
+		if len(additionalMethods) > 0 {
```
beorn7, line 402: Does this `if` really improve the performance? I would assume a `range` over an empty slice will exit very fast. But perhaps you know something I don't…

prometheus/promhttp/instrument_server.go
```diff
@@ -348,15 +394,27 @@ func sanitizeMethod(m string) string {
...
+			for _, method := range additionalMethods {
+				if strings.EqualFold(m, method) {
+					return strings.ToLower(m)
```
beorn7, line 405: Can't we just use `method` here instead of `m`? It wouldn't require `ToLower` if we documented that additional methods are used with the same capitalization as provided (which some users might even need).

beorn7 (commented):
LGTM in general, just a few thoughts.

prometheus/promhttp/instrument_server.go
kakkoyun: Noop, you're it doesn't have any functional or performance effect. Just it was easier to reason about the implicit else clause `return "unknown"`, happy to remove and simplify.

prometheus/promhttp/instrument_server.go
kakkoyun: Of course, it's an option. I just respected the previous design decisions here. Everything was the lower case before and just kept it. I'm inclined to keep the lower case behaviour and if someone requests it we can change it; it's a minor one anyways.

prometheus/promhttp/instrument_server.go
```diff
@@ -453,6 +509,9 @@ func sanitizeCode(s int) string {
 		return "511"
 
 	default:
-		return strconv.Itoa(s)
+		if s >= 100 && s <= 599 {
+			return strconv.Itoa(s)
+		}
+		return "unknown"
```
bwplotka, line 515: (:

prometheus/promhttp/option.go
```diff
@@ -0,0 +1,31 @@
...
+
+type option struct {
+	additionalMethods []string
+}
+
+// WithAdditionalMethods adds additional HTTP methods to the list of allowed methods.
+// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods for the default list.
+//
+// See the example for ExampleInstrumentHandlerWithAdditionalMethods for example usage.
+func WithAdditionalMethods(methods ...string) Option {
```
bwplotka, line 27: In comment on the top you have `WithExtraMethods` here you have `WithAdditionalMethods`. Let's agree on one. I vote for Extra since it's a lot shorter, but let's use whatever works - as long as we keep this consistent.

prometheus/promhttp/instrument_server.go
```diff
@@ -114,7 +130,10 @@ func InstrumentHandlerCounter(counter *prometheus.CounterVec, next http.Handler)
 // until the response headers are written. The ObserverVec must have valid
 // metric and label names and must have zero, one, or two non-const non-curried
 // labels. For those, the only allowed label names are "code" and "method". The
-// function panics otherwise. The Observe method of the Observer in the
+// function panics otherwise. For the "method" label a predefined default label
+// value set is used to filter given values. Values besides predefined values
+// will count as `unknown` method.`WithExtraMethods` can be used to add more
```
bwplotka, line 135: `WithExtraMethods` is not consistent with actual name, unless we change it.

bwplotka (approved):
LGTM, just a small inconsistency. I would pick shorter name, additional is quite long word to use everywhere and in those case we often say `extra`. (:

PR #974, opened 2022-01-24
Reduce granularity of histogram buckets for Go 1.17 collector

### LINKED ISSUE #967: v1.12.0 exposes some high cardinality metrics
It seems that v1.12.0 includes some extra high cardinality metrics, most likely autogenerated by Go itself thanks to #955.

Here's a full diff of metrics between 1.11 and 1.12 - https://gist.github.com/prymitive/3d16b79a8187448bf6e0d61db7336ca0

Having so many extra metrics exposed by default might cause problems for people with lots of Go services.

mknyszek:
The Go runtime/metrics package currently exports extremely granular
histograms. Exponentially bucket any histogram with unit "seconds"
or "bytes" instead to dramatically reduce the number of buckets, and
thus the number of metrics.

This change also adds a test to check for expected cardinality to
prevent cardinality surprises in the future.

Fixes #967 

Signed-off-by: Michael Anthony Knyszek <mknyszek@google.com>

### DIFF
--- a/prometheus/go_collector_go117.go
+++ b/prometheus/go_collector_go117.go
@@ -20,6 +20,7 @@ import (
 	"math"
 	"runtime"
 	"runtime/metrics"
+	"strings"
 	"sync"
 
 	//nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
@@ -53,9 +54,20 @@ type goCollector struct {
 // Deprecated: Use collectors.NewGoCollector instead.
 func NewGoCollector() Collector {
 	descriptions := metrics.All()
-	descMap := make(map[string]*metrics.Description)
-	for i := range descriptions {
-		descMap[descriptions[i].Name] = &descriptions[i]
+
+	// Collect all histogram samples so that we can get their buckets.
+	// The API guarantees that the buckets are always fixed for the lifetime
+	// of the process.
+	var histograms []metrics.Sample
+	for _, d := range descriptions {
+		if d.Kind == metrics.KindFloat64Histogram {
+			histograms = append(histograms, metrics.Sample{Name: d.Name})
+		}
+	}
+	metrics.Read(histograms)
+	bucketsMap := make(map[string][]float64)
+	for i := range histograms {
+		bucketsMap[histograms[i].Name] = histograms[i].Value.Float64Histogram().Buckets
 	}
 
 	// Generate a Desc and ValueType for each runtime/metrics metric.
@@ -80,6 +92,7 @@ func NewGoCollector() Collector {
 		var m collectorMetric
 		if d.Kind == metrics.KindFloat64Histogram {
 			_, hasSum := rmExactSumMap[d.Name]
+			unit := d.Name[strings.IndexRune(d.Name, ':')+1:]
 			m = newBatchHistogram(
 				NewDesc(
 					BuildFQName(namespace, subsystem, name),
@@ -87,6 +100,7 @@ func NewGoCollector() Collector {
 					nil,
 					nil,
 				),
+				internal.RuntimeMetricsBucketsForUnit(bucketsMap[d.Name], unit),
 				hasSum,
 			)
 		} else if d.Cumulative {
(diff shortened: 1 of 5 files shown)


prometheus/go_collector_metrics_go117_test.go
```diff
@@ -37,3 +37,5 @@ var expectedRuntimeMetrics = map[string]string{
 	"/sched/goroutines:goroutines":                "go_sched_goroutines_goroutines",
 	"/sched/latencies:seconds":                    "go_sched_latencies_seconds",
 }
+
+const expectedRuntimeMetricsCardinality = 79
```
bwplotka, line 41: 👍🏽

prometheus/internal/go_runtime_metrics.go
```diff
@@ -75,3 +76,49 @@ func RuntimeMetricsToProm(d *metrics.Description) (string, string, string, bool)
...
+	switch unit {
+	case "bytes":
+		// Rebucket as powers of 2.
+		return rebucketExp(buckets, 2)
+	case "seconds":
+		// Rebucket as powers of 10 and then truncate all buckets
+		// above 1 second and merge them into a single bucket.
+		b := rebucketExp(buckets, 10)
+		for i := range b {
+			if b[i] > 1 {
```
bwplotka, line 94: nit: We could eliminate <= 1 first to keep code more left indent

prometheus/internal/go_runtime_metrics.go
```diff
@@ -75,3 +76,49 @@ func RuntimeMetricsToProm(d *metrics.Description) (string, string, string, bool)
...
+// scheme.
+func rebucketExp(buckets []float64, base float64) []float64 {
+	bucket := buckets[0]
+	if bucket == math.Inf(1) {
+		return buckets
+	}
+	var newBuckets []float64
+	for i := 1; i < len(buckets); i++ {
+		if (bucket >= 0 && buckets[i] >= bucket*base) ||
+			(bucket < 0 && buckets[i] >= bucket/base) ||
```
bwplotka, line 117: ditto

prometheus/internal/go_runtime_metrics.go
```diff
@@ -75,3 +76,49 @@ func RuntimeMetricsToProm(d *metrics.Description) (string, string, string, bool)
...
+	return buckets
+}
+
+// rebucketExp takes a list of bucket boundaries (lower bound inclusive) and
+// downsamples the buckets to those a multiple of base apart. The end result
+// is a roughly exponential (in many cases, perfectly exponential) bucketing
+// scheme.
+func rebucketExp(buckets []float64, base float64) []float64 {
+	bucket := buckets[0]
+	if bucket == math.Inf(1) {
```
bwplotka, line 111: What this case represents? Is this even possible?

prometheus/internal/go_runtime_metrics.go
```diff
@@ -75,3 +76,49 @@ func RuntimeMetricsToProm(d *metrics.Description) (string, string, string, bool)
...
+// a reduced set of buckets. This function always removes any -Inf bucket as it's represented
+// as the bottom-most upper-bound inclusive bucket in Prometheus.
+func RuntimeMetricsBucketsForUnit(buckets []float64, unit string) []float64 {
+	switch unit {
+	case "bytes":
+		// Rebucket as powers of 2.
+		return rebucketExp(buckets, 2)
+	case "seconds":
+		// Rebucket as powers of 10 and then truncate all buckets
+		// above 1 second and merge them into a single bucket.
```
bwplotka, line 91: Can we add comment why this decision to treat 1+ second as the Inf+ bucket?

prometheus/go_collector_go117.go
```diff
@@ -303,37 +317,38 @@ type batchHistogram struct {
 	sum     float64 // Used if hasSum is true.
 }
 
-func newBatchHistogram(desc *Desc, hasSum bool) *batchHistogram {
-	h := &batchHistogram{desc: desc, hasSum: hasSum}
+func newBatchHistogram(desc *Desc, buckets []float64, hasSum bool) *batchHistogram {
+	h := &batchHistogram{
+		desc:    desc,
+		buckets: buckets,
+		counts:  make([]uint64, len(buckets)-1),
```
bwplotka, line 324: Maybe I missed something, but why -1?

bwplotka: Ah, ok found it, it similar to this in metrics:

> 	// Given N buckets, Count[n] is the weight of the range
>	// [bucket[n], bucket[n+1]), for 0 <= n < N.

Worth adding similar comment?

prometheus/go_collector_go117.go
```diff
@@ -303,37 +317,38 @@ type batchHistogram struct {
...
-		// it's going to get reused.
-		h.buckets = make([]float64, len(buckets))
-		copy(h.buckets, buckets)
-
-		h.counts = make([]uint64, len(counts))
+	// Reduce buckets.
+	var j int
+	var total uint64
+	for i, count := range counts {
+		total += count
```
bwplotka, line 345: nit: Hm, I feel if we get rid `total` variable and add directly in `h.buckets` it would be more clear? Fewer code lines, and to me `total` purpose was a bit misleading (it's not total, it's actually temp total between buckets). Not a blocker (:

bwplotka (approved):
Great job! Only readability nits, otherwise LGTM! Thanks 💪🏽

kakkoyun (commented):
LGTM. Thanks @mknyszek for taking care of this.

As @bwplotka, it would be nice add more explanation comments, it'd be easier for us to maintain in the long run.

Also we might consider handling https://github.com/prometheus/client_golang/issues/967#issuecomment-1021650447 as well before the next patch release. Not necessarily as part of this PR though.

prometheus/internal/go_runtime_metrics.go
mknyszek: it's not a possible case. I think I left that in by mistake. removed it.

prometheus/go_collector_go117.go
mknyszek: done, but I broke out the clearing separately. I think it's better this way.

prometheus/go_collector_go117.go
mknyszek: done

prometheus/internal/go_runtime_metrics.go
mknyszek: done

prometheus/internal/go_runtime_metrics.go
mknyszek: done, and I made some other parts of this clearer and more explicit

prometheus/internal/go_runtime_metrics.go
mknyszek: done

kakkoyun (approved):
LGTM. Thanks for the timely fix.

PR #975, opened 2022-01-24
Use simpler locking in the Go 1.17 collector

mknyszek:
A previous PR made it so that the Go 1.17 collector locked only around
uses of rmSampleBuf, but really that means that Metric values may be
sent over the channel containing some values from future metrics.Read
calls. While generally-speaking this isn't a problem, we lose any
consistency guarantees provided by the runtime/metrics package.

Also, that optimization to not just lock around all of Collect was
premature. Truthfully, Collect is called relatively infrequently, and
its critical path is fairly fast (10s of µs). To prove it, this test
also adds a benchmark.

Signed-off-by: Michael Anthony Knyszek <mknyszek@google.com>

### DIFF
--- a/prometheus/go_collector_go117.go
+++ b/prometheus/go_collector_go117.go
@@ -31,8 +31,11 @@ import (
 type goCollector struct {
 	base baseGoCollector
 
+	// mu protects updates to all fields ensuring a consistent
+	// snapshot is always produced by Collect.
+	mu sync.Mutex
+
 	// rm... fields all pertain to the runtime/metrics package.
-	rmSampleMu  sync.Mutex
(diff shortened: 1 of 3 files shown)


mknyszek:
CC @beorn7 @bwplotka @kakkoyun

kakkoyun:
@mknyszek Could you attach a benchmark comparison if it's possible? For documentation purposes.

kakkoyun (approved):
LGTM. I'll wait for @beorn7's comments since he has more context and elven eyes :)

beorn7 (approved):
Thank you very much.

mknyszek:
@kakkoyun Added it to the commit message.

PR #978, opened 2022-01-29
Cut v1.12.1

kakkoyun:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.12.1 / 2022-01-29
+
+* [BUGFIX] Make the Go 1.17 collector thread-safe #969
```
beorn7, line 3: Nit: I would call it "concurrency-safe". "Thread" is a rather specific term, and the Go runtime, while utilizing threads, pretty much decouples Goroutines from threads. (AFAIK, even on a single-CPU single-threaded machine you could still run into data races.)

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+  * Use simpler locking in the Go 1.17 collector #975
+* [BUGFIX] Reduce granularity of histogram buckets for Go 1.17 collector #974
+* [CHANGE] API client: make http reads more efficient #976
```
beorn7, line 6: Isn't that more of an [ENHANCEMENT]? If I see [CHANGE] in a changelog, I think it's an actual change of behavior that requires a user action to accommodate.

beorn7: And nit: I would capitalize HTTP.

beorn7 (changes requested):
Thank you very much. Some minor nits, see comments.

beorn7 (approved):

PR #979, opened 2022-01-30
Fix superfluous response.WriteHeader

S7evinK:
This fixes superfluous calls to `WriteHeader`

Related: https://github.com/matrix-org/dendrite/issues/2123


prometheus/promhttp/delegator.go
```diff
@@ -61,8 +61,10 @@ func (r *responseWriterDelegator) WriteHeader(code int) {
 		r.observeWriteHeader(code)
 	}
 	r.status = code
-	r.wroteHeader = true
-	r.ResponseWriter.WriteHeader(code)
```
beorn7, line 65: As described in the comment above (lines 57 to 60), we intentionally delegate the `WriteHeader` call to not mask the actual bug. Correct code wouldn't call `WriteHeader` more than once in the first place.

beorn7 (changes requested):
Thanks for the attempt, but I don't think this fixes the problem. It only masks it. With this change, the delegator detects that `WriteHeader` has been called twice and decides to not delegate that call. However, that's not the responsibility of the delegator.

kakkoyun:
@S7evinK Do you want to continue working on this?

S7evinK:
As I'm out of ideas, nope. Will close this.

PR #980, opened 2022-01-31
go.mod: Set minimal support version of go to 1.15

mrueg:
As requested in https://github.com/prometheus/common/pull/353#discussion_r793113705

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,14 +48,6 @@ workflows:
     jobs:
       # Refer to README.md for the currently supported versions.
       - test:
-          name: go-1-13
-          go_version: "1.13"
-          run_lint: true
-      - test:
-          name: go-1-14
-          go_version: "1.14"
-          run_lint: true
-      - test:
           name: go-1-15
           go_version: "1.15"
           run_lint: true
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This is the [Go](http://golang.org) client library for
 instrumenting application code, and one for creating clients that talk to the
 Prometheus HTTP API.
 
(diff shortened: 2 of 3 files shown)


beorn7:
Since this library is so widely used, we usually only bump the version requirement if we really have to (i.e. because there is code requiring the more recent version). Supporting the last two or three Go versions is a _minimal_ requirement (i.e. we would use build tags and such for features requiring newer Go versions). As long as older Go versions still work without additional hassle, we should IMHO keep the support.

mrueg:
Thanks for the feedback. Unfortunately I believe it needs to be raised due to the x/net dependency upgraded in https://github.com/prometheus/common/pull/348 to mitigate a potential vuln. With the upgraded x/net, prometheus/common doesn't seem to work with earlier versions of go than 1.15.

@roidelapluie and @SuperQ can chime in, in which order the change should enter as this repo also depends on prometheus/common.

beorn7:
OK, a security update of a dependency, thereby having dependencies requiring newer Go versions, sounds like a good reason.

That context wasn't clear to me.

mrueg:
Apologies again for not providing that context earlier. 
Still not sure how releasing both repos usually works because of the circular dependency. I guess client_golang starts first, then common depends on the new version and then client_golang has another release?

SuperQ (approved):
LGTM

SuperQ:
I'm not sure we need to cut a release for these changes.

SuperQ:
I've updated the branch protection rules to reflect this change.

kakkoyun (approved):
LGTM 

Thanks for taking care of it.

Issue #868, opened 2021-05-21
feature: Add exemplar without observation/incrementing?


LouisStAmour:
The use case I have in mind is for trace-derived metrics in something like an Otel Collector, where you want to output near real-time metrics but you want to add traces as exemplars (such as what Grafana supports) but you can't do both: sometimes you'll want to mark an interesting trace as an exemplar but you only know it's interesting because a later-processed inner span had an exception event and your exemplar would thus have an "exception" label on it to highlight it in red, possibly.

A similar idea I have is to automatically pick some exemplars based on what buckets they go into when observing for a histogram. This would I suppose be a new API call to observeWithExemplarCandidate say, that would take the same exemplar data but based on the histogram bucketing it could be picked as an exemplar or not and would return whether picked as an exemplar or not (maybe?)

The alternative to such a sampling API would be to allow observations to histograms like normal where the choice of observing with exemplar could be compared against the current value of the histogram, such as to know whether the value is an outlier or not. The difficulty here is that you might not have many requests and you might want to sample one request per bucket. At that point, you again need to keep your bucket data for up to a minute, then decide on which exemplars to add up to a minute after you've first observed them. That's again something that requires separating backdated exemplars from real-time observations, thus proving the need for this API call.

I'm new to the Prometheus client library community, so please let me know if this idea has merit. I'm thinking of suggesting the same in the Otel Collector community as I'm having a hard time adopting trace-derived metrics for Grafana's new support of exemplars without being able to separate real-time metrics from past-observed exemplar values.

beorn7:
I'm not sure I fully understand what you are proposing. Some parts of it seem to be easily possible with the current primitives. Others (like outlier detection) seem a bit too specific and too out-of-scope to be built into the instrumentation library. But in both cases, I don't understand why you would like to add an examplar without observation. An exemplar is _always_ linked to an observation, isn't it?

Perhaps you could provide code examples (imagining the method call you desire already exist)?

LouisStAmour:
The primary use case I'm supporting here is for tail-based sampling decisions while also preserving real-time metrics. The endpoint collecting Prometheus exemplars should not have to check and see if a trace was kept or not.

Right now, if we want to record every trace ID or a small sample of them as exemplars, we can. But that overwhelms the server collecting exemplars because it has to make sure every trace ID logged as an exemplar still exists. This can be done by choosing in advance what traces to keep but that doesn't allow for tail-based sampling.

[...]

PR #986, opened 2022-02-10
Add exemplar support for const histogram

wperron:
Relates to #868. We're looking to add support for exemplars on OpenTelemetry's collector contrib's `prometheusexporter` which uses the `constHistogram` internally. This change solves the case where metrics coming from another system (in our case the otel collector) need to be converted to the Prometheus format and retain their exemplars. Right now we've only implemented exemplars on the const histogram, I'm more than happy to do all the necessary adjustments.

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -582,7 +582,9 @@ func (h *constHistogram) Desc() *Desc {
 
 func (h *constHistogram) Write(out *dto.Metric) error {
 	his := &dto.Histogram{}
+	// h.buckets, buckets and bounds are all the same length
 	buckets := make([]*dto.Bucket, 0, len(h.buckets))
+	bounds := make([]float64, 0)
 
 	his.SampleCount = proto.Uint64(h.count)
 	his.SampleSum = proto.Float64(h.sum)
@@ -591,15 +593,24 @@ func (h *constHistogram) Write(out *dto.Metric) error {
 			CumulativeCount: proto.Uint64(count),
 			UpperBound:      proto.Float64(upperBound),
 		})
+		bounds = append(bounds, upperBound)
 	}
 
+	// make sure that both bounds and buckets have the same ordering
 	if len(buckets) > 0 {
 		sort.Sort(buckSort(buckets))
 	}
+	sort.Float64s(bounds)
 
 	if len(h.exemplars) > 0 {
-		for i := 0; i < len(buckets); i++ {
-			buckets[i].Exemplar = h.exemplars[i]
+		r := len(buckets)
+		for i := 0; i < r; i++ {
+			bound := sort.SearchFloat64s(bounds, *h.exemplars[i].Value)
+			// Only append the exemplar if it's within the bounds defined in the
+			// buckets.
+			if bound < r {
+				buckets[bound].Exemplar = h.exemplars[i]
+			}
 		}
 	}
 
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -573,6 +573,7 @@ type constHistogram struct {
 	sum        float64
 	buckets    map[float64]uint64
 	labelPairs []*dto.LabelPair
+	exemplars  []*dto.Exemplar
 }
 
 func (h *constHistogram) Desc() *Desc {
@@ -585,7 +586,6 @@ func (h *constHistogram) Write(out *dto.Metric) error {
 
 	his.SampleCount = proto.Uint64(h.count)
 	his.SampleSum = proto.Float64(h.sum)
-
 	for upperBound, count := range h.buckets {
 		buckets = append(buckets, &dto.Bucket{
 			CumulativeCount: proto.Uint64(count),
@@ -596,11 +596,25 @@ func (h *constHistogram) Write(out *dto.Metric) error {
 	if len(buckets) > 0 {
 		sort.Sort(buckSort(buckets))
 	}
+
+	if len(h.exemplars) > 0 {
+		for i := 0; i < len(buckets); i++ {
+			buckets[i].Exemplar = h.exemplars[i]
+		}
+	}
+
 	his.Bucket = buckets
 
 	out.Histogram = his
 	out.Label = h.labelPairs
 
+	return nil
+}
+
+func (h *constHistogram) GetExemplars() []*dto.Exemplar {
+	if h != nil {
+		return h.exemplars
+	}
 	return nil
 }
 
@@ -617,6 +631,7 @@ func (h *constHistogram) Write(out *dto.Metric) error {
 //
 // NewConstHistogram returns an error if the length of labelValues is not
 // consistent with the variable labels in Desc or if Desc is invalid.
+
 func NewConstHistogram(
 	desc *Desc,
 	count uint64,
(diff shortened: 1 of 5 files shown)


prometheus/histogram.go
```diff
@@ -655,6 +670,48 @@ func MustNewConstHistogram(
 	return m
 }
 
+func NewConstHistogramWithExemplar(
```
bwplotka, line 673: Can we use `NewConstHistogramWithExemplar` in `NewConstHistogram` for less code repetition?

bwplotka (approved):
Makes sense to me, thanks! Just one nit.

@beorn7 @kakkoyun  any objections?

beorn7:
Counters can also have exemplars. Instead of also adding `NewConstMetricWithExemplar`, how about using the `NewMetricWithExemplar` approach described in [this comment](https://github.com/prometheus/client_golang/issues/868#issuecomment-848830320)? This wrapper would detect if the metrics to be wrapped is a histogram or a counter and then act accordingly (and panic for other metric types).

beorn7:
It would also avoid having to add the `Must...` version separately. Instead of `NewConstHistogramWithExemplar`, `MustNewConstHistogramWithExemplar`, `NewConstMetricWithExemplar`, and `MustNewConstMetricWithExemplar`, we would just need the one `NewMetricWithExemplar` wrapper.

prometheus/histogram.go
```diff
@@ -604,6 +611,13 @@ func (h *constHistogram) Write(out *dto.Metric) error {
 	return nil
 }
 
+func (h *constHistogram) GetExemplars() []*dto.Exemplar {
```
beorn7, line 614: This library has avoided adding exporter Getters to metric types so far. I would suggest to keep it that way for consistency. (The reason behind that is a different story. Should we ever change the approach, it should be done in an informed and consistent way in a separate PR, not snug in as a byproduct.)

wperron: All good, I also raised this internally and it turns out we didn't need it anyway 👍

wperron:
> Counters can also have exemplars. Instead of also adding `NewConstMetricWithExemplar`, how about using the `NewMetricWithExemplar` approach described in [this comment](https://github.com/prometheus/client_golang/issues/868#issuecomment-848830320)? This wrapper would detect if the metrics to be wrapped is a histogram or a counter and then act accordingly (and panic for other metric types).
>
> It would also avoid having to add the `Must...` version separately. Instead of `NewConstHistogramWithExemplar`, `MustNewConstHistogramWithExemplar`, `NewConstMetricWithExemplar`, and `MustNewConstMetricWithExemplar`, we would just need the one `NewMetricWithExemplar` wrapper.

So I started looking into this last Friday and it's less straight forward than that comment makes it seem. `NewMetricWithTimestamp` is pretty straight forward because every `Metric` can have a timestamp associated with it, whereas not every `Metric` can have exemplars. Also, exemplars on Histograms are pretty different than on simple values. I agree it'd be the best outcome, I just thought I'd at least open the PR to start discussion and go from there. I'll get cracking on this and hopefully have something up soon that addresses those comments.

prometheus/histogram.go
wperron: Just to clarify, did you mean to use `NewConstHistogram` inside `NewConstHistogramWithExemplar`? My goal was to keep the existing API unchanged but whatever you folks prefer is fine by me.

wperron:
Hey folks, so I've been looking into a potential `func WithExemplar() (Metric, error)` function that could be a general case, sorta like `WithTimestamp` does but the more I look at it, the less I'm convinced that there's a good way to implement it.

From what I understand there’s 4 Metrics that can have exemplars: ExemplarAdder, ExemplarObserver, constMetric and constHistogram. For the first two, the layer of abstraction doesn’t add a lot, it ends up just doing what ppl are already doing right now, asserting that the metric implements the interface, then just call AddWithExemplar or ObserveWithExemplar. For the last two, the data they need is so different I’m struggling to find a way to make a function signature that would work with both of them that would also be clear enough. For reference, here are what the current signatures look like:

```
NewConstHistogram(desc *Desc, count uint64, sum float64, buckets map[float64]uint64, labelValues ...string)
NewConstMetric(desc *Desc, valueType ValueType, value float64, labelValues ...string) (Metric, error)
```

To add exemplars, the first one needs a []*dto.Exemplar because each bucket has it’s own exemplar, and the second one needs `*dto.Exemplar` because, well, there's only one, and I can't use `...*dto.Exemplar` because the existing functions already use a variadic parameter for the LabelValues.

I can add a `WithExemplar` variant to `NewConstMetric` and `MustNewConstMetric` to make this PR more complete, or just leave it as is too. Any thoughts?

beorn7:
Hmm, I increasingly believe there is no real good way of solving this in the current context.

I really don't like to expose `dto.Exemplar` to the user of the library directly. In `ObserveWithExemplar` and `AddWithExemplar`, we use `(value float64, exemplar Labels)` as the signature, not `(exemplar *dto.Exemplar)`. If we let users construct their own `dto.Exemplar`, we are not that far away from letting them create their own `dto.Metric` (which is not meant sarcastically, there are cases where that's just the right thing to do).

I think, from a pure interface perspective, the least bad solution would be a `func NewMetricWithExemplar(value float64, exemplar Labels, metric Metric) Metric`. It would work like `func NewMetricWithTimestamp(t time.Time, m Metric) Metric`, i.e. we would have a `metricWithExemplar` wrapper type corresponding to `timestampedMetric`, and we would put the exemplar into the protobuf at `Write` time. `NewMetricWithExemplar` would work on _any_ `Metric` implementation. You could use it repeatedly to add multiple exemplars. At `Write` time, the following happens:
- In case the written protobuf is a counter and the `value` is not negative, the exemplar is added to it, overwriting any existing exemplar, i.e. the outermost exemplar wins (in case `NewMetricWithExemplar` was used multiple times or `NewMetricWithExemplar` was used on an `ExemplarAdder` implementation that already had an examplar).
- In case the written protobuf is a histogram, the suitable bucket for the `value` is found and the exemplar is added to it, again overwriting existing exemplars.
- In all other cases, the exemplar is silently ignored.

The last point is the most "dirty" aspect of this, but maybe it's OK (think of the function as "add an exemplar to this metric if it makes sense"). It's slightly annoying that you have to call `NewMetricWithExemplar` multiple times to add many exemplars (most commonly to add an exemplar for each bucket of a multi-bucket histogram). But you can do that in a loop, so it might not be that bad. However, that leads to the serious concern I have, that the implementation cannot be very efficient. If you have a histogram with 20 buckets, you'll end up with a 20-fold wrapped histogram. _Maybe_ that's not that bad for most use cases (and for the few that need optimal performance, those can fall back to build their own `dto.Metric`).

So maybe, overall, the approach in this PR is already the least bad. I don't know. @bwplotka and @kakkoyun you should make the call here. :smile_cat:

wperron:
Hey folks, any updates on this PR?

beorn7:
@bwplotka and @kakkoyun, I think all the pros and cons of the various approaches have been laid out above. We need you to make a call what direction to take.

bwplotka:
Ack, we will sync on it with @kakkoyun on Monday. Thanks!

bwplotka:
Ack, will look on Tuesday, adding to my todo list.

bwplotka:
Looking now, sorry for lag

prometheus/histogram.go
bwplotka: no, opposite. I think you misunderstood, let me show in PR to your PR 🙃

bwplotka:
@wperron I proposed a PR to your PR. Tests are failing as I broke histogram upper bound check, fizzing now: https://github.com/Shopify/client_golang/pull/3 but shows the alternative usave (:

bwplotka (changes requested):
Blocking this to decide what's the best approach before merging 🙃

PR #989, opened 2022-02-23
Added Transactional Gatherer allowed cached solutions

bwplotka:
See https://github.com/prometheus/client_golang/pull/929 on how it can be used.

API proved to work well, but we have to experiment with different caching solutions before providing something here.


prometheus/value.go
```diff
@@ -110,18 +132,20 @@ func MustNewConstMetric(desc *Desc, valueType ValueType, value float64, labelVal
 }
 
 type constMetric struct {
-	desc       *Desc
-	valType    ValueType
-	val        float64
-	labelPairs []*dto.LabelPair
+	desc   *Desc
+	metric *dto.Metric
```
bwplotka, line 136: reusing some structs for readability as an effect of https://github.com/prometheus/client_golang/pull/929 experiements.

kakkoyun (approved):
LGTM :100:

PR #998, opened 2022-03-07
smart diff to testutil.GatherAndCompare

### LINKED ISSUE #996: Add smart diff to testutil.GatherAndCompare family of utilities.
Currently, it's much better to just use `require.Equal` from testify to compare outputs of scrape, since it gives nice diff:

```
0304 12:33:58.224133 4120671 server.go:406] "Adding debug handlers to kubelet server"
    server_test.go:1449: 
        	Error Trace:	server_test.go:1449
        	Error:      	Not equal: 
        	            	...
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -3648,3 +3648,3 @@
        	            	 container_processes{container="",id="/user.slice/user-1000.slice/user@1000.service/session.slice/pulseaudio.service",image="",name="",namespace="",pod=""} 0 1645535816846
        	            	-# HELP container_scrape_error 1 if there was an error while getting container metrics, 0 otherwise.
        	            	+# HELP container_scrape_error 1 if there was an error while getting container metrics, 0 otherwise
        	            	 # TYPE container_scrape_error gauge
        	            	@@ -3939,3 +3939,3 @@
        	            	 container_spec_cpu_period{container="",id="/user.slice/user-1000.slice/user@1000.service/session.slice/pulseaudio.service",image="",name="",namespace="",pod=""} 0
        	            	-# HELP container_spec_cpu_shares CPU share (limit) of the container.
        	            	+# HELP container_spec_cpu_shares CPU share of the container.
        	            	 # TYPE container_spec_cpu_shares gauge
        	Test:       	TestCadvisorMetrics

```

If I do this with our compare, it is huge unreadable blob without diff ): Just `got, want`. Let's improve experience here!

sourikghosh:
Signed-off-by: Sourik Ghosh <sourikghosh31@gmail.com>
closes #996

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.32.1
 	github.com/prometheus/procfs v0.7.3
+	github.com/stretchr/testify v1.4.0
 	golang.org/x/sys v0.0.0-20220114195835-da31bd327af9
 	google.golang.org/protobuf v1.26.0
 )
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -41,8 +41,10 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"testing"
 
 	"github.com/prometheus/common/expfmt"
+	"github.com/stretchr/testify/require"
 
 	dto "github.com/prometheus/client_model/go"
 
@@ -154,27 +156,27 @@ func GatherAndCount(g prometheus.Gatherer, metricNames ...string) (int, error) {
 // CollectAndCompare registers the provided Collector with a newly created
 // pedantic Registry. It then calls GatherAndCompare with that Registry and with
 // the provided metricNames.
-func CollectAndCompare(c prometheus.Collector, expected io.Reader, metricNames ...string) error {
+func CollectAndCompare(t *testing.T, c prometheus.Collector, expected io.Reader, shouldFail bool, metricNames ...string) error {
 	reg := prometheus.NewPedanticRegistry()
 	if err := reg.Register(c); err != nil {
 		return fmt.Errorf("registering collector failed: %s", err)
 	}
-	return GatherAndCompare(reg, expected, metricNames...)
+	return GatherAndCompare(t, reg, expected, shouldFail, metricNames...)
 }
 
 // GatherAndCompare gathers all metrics from the provided Gatherer and compares
 // it to an expected output read from the provided Reader in the Prometheus text
 // exposition format. If any metricNames are provided, only metrics with those
 // names are compared.
-func GatherAndCompare(g prometheus.Gatherer, expected io.Reader, metricNames ...string) error {
-	return TransactionalGatherAndCompare(prometheus.ToTransactionalGatherer(g), expected, metricNames...)
+func GatherAndCompare(t *testing.T, g prometheus.Gatherer, expected io.Reader, shouldFail bool, metricNames ...string) error {
+	return TransactionalGatherAndCompare(t, prometheus.ToTransactionalGatherer(g), expected, shouldFail, metricNames...)
 }
 
 // TransactionalGatherAndCompare gathers all metrics from the provided Gatherer and compares
 // it to an expected output read from the provided Reader in the Prometheus text
 // exposition format. If any metricNames are provided, only metrics with those
 // names are compared.
-func TransactionalGatherAndCompare(g prometheus.TransactionalGatherer, expected io.Reader, metricNames ...string) error {
+func TransactionalGatherAndCompare(t *testing.T, g prometheus.TransactionalGatherer, expected io.Reader, shouldFail bool, metricNames ...string) error {
 	got, done, err := g.Gather()
 	defer done()
 	if err != nil {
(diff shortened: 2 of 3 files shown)


sourikghosh:
review @bwplotka

kakkoyun (changes requested):
Looking good. 
I like to idea of using `require.Equals` but these are changing existing APIs. I'm reluctant to change existing public APIs.

How about adding equivalent new version of the APIs that have `t *testing.T` as a first parameter? and those can use require package under the hood.

kakkoyun (commented):
Rather than MethodNameV2, I'd prefer `MethodNameWithTesting` or `MethodNameWithT`

sourikghosh:
Ok replacing it with `MethodNameWithT`
Ehh  `ci/circleci: go-1-16`  for `TestSummaryDecay` failed but I run both `make test` and `test-short` locally before pushing.

kakkoyun (approved):

prometheus/testutil/testutil.go
```diff
@@ -41,8 +41,10 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"testing"
 
 	"github.com/prometheus/common/expfmt"
+	"github.com/stretchr/testify/require"
```
bwplotka, line 47: Hm.. is there a way to avoid this package? For a reason we wanted to make sure it's not used, so we are not leaking it.

Since this is not testing code, every project that imports `client_golang` needs to import `testify`....  I think we want to copy the diff function and maintain here.


See how we did that here https://github.com/efficientgo/tools/blob/main/core/pkg/testutil/testutil.go#L101

bwplotka (changes requested):
Hey, thanks for this work 💪🏽 

It is MUST have functionality, I miss it so much <3

But I think we need to be careful. I think we could improve this PR a bit:

* Ideally we don't force people to import extra (quite large) `testify` module. `client_golang` lib is well used library, so we need to be careful here. I think there is no strong need. Proposed alternative.
* I am not fan of extra interface (so many extra methods). I think we can remove that and add nice diff without `t *testing.T` - especially if we want to skip `testify` module import.


WDYT?

prometheus/testutil/testutil.go
kakkoyun: I totally agree. I think I was a little hasty to review this, and I haven't noticed testify is not there already :see_no_evil:

kakkoyun: I guess because it's already a transitive dependency.

sourikghosh: yeah, I agree too, this makes sense
but the above diff function has a dep on
https://github.com/efficientgo/tools/blob/fe763185946be83b20da626605319733bd7f97cb/core/pkg/testutil/testutil.go#L16 which is not maintained should I transfer the module implementation over here or just simply import it.

go.mod
```diff
@@ -3,8 +3,10 @@ module github.com/prometheus/client_golang
 require (
 	github.com/beorn7/perks v1.0.1
 	github.com/cespare/xxhash/v2 v2.1.2
+	github.com/davecgh/go-spew v1.1.1
 	github.com/golang/protobuf v1.5.2
 	github.com/json-iterator/go v1.1.12
+	github.com/pmezard/go-difflib v1.0.0
```
bwplotka, line 9: I don't want to be a pain but this module is `THIS PACKAGE IS NO LONGER MAINTAINED.` Not the best, since all users depending on our module will have to download it too.

I wonder if it makes sense to just copy the code for now with the required code (with test) over. Looks reasonably small. There is not benefit in depending on module that does not allow merging new changes. 

Or we could find different package.

bwplotka (changes requested):
Thanks! Sorry for complaining, but I think there is small risk with the module of choice 🤔 WDYT?

go.mod
sourikghosh: 🙃that was my question previously https://github.com/prometheus/client_golang/pull/998#discussion_r829694057

Ideally, I would prefer an alternative to the package but I couldn't find a good replacement. So maybe coping small piece of code and maintaining it over here ?!!!

bwplotka: Yea, I think copy might be better

sourikghosh:
@bwplotka 👀

kakkoyun:
@bwplotka Could you please have another look?

prometheus/testutil/difflib.go
```diff
@@ -0,0 +1,649 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// It provides tools to compare sequences of strings and generate textual diffs.
+//
+// Maintaining `GetUnifiedDiffString` here because original repository
+// is no loger maintained.(https://github.com/pmezard/go-difflib)
```
bwplotka, line 17:
```suggestion
// (https://github.com/pmezard/go-difflib) is no longer maintained.
```

prometheus/testutil/difflib.go
```diff
@@ -0,0 +1,649 @@
+// Copyright 2022 The Prometheus Authors
```
bwplotka, line 1: I believe we want to make sure this is under `internal` path so no one can import it. Otherwise we have to care about API, Go doc comments etc. 

And I don't client_golang have to be library which offers diffing 🤔

bwplotka (changes requested):
Thanks for this work, and sorry for lag. Again - please ping me on CNCF slack (: 

I found quite important thing to do before we merge, WDYT?

bwplotka (approved):
Finally!

Thanks for your hard work 💪🏽 

LGTM

PR #1001, opened 2022-03-16
Add statebot config

kakkoyun:

The thresholds are pretty generous. This shouldn't create any pressure on contributors.
On top of that, I have excluded issues with a milestone attached.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -24,7 +24,7 @@ exemptLabels:
 exemptProjects: true # default: false
 
 # Set to true to ignore issues in a milestone (defaults to false)
-exemptMilestones: false
+exemptMilestones: true # default: false
 
 # Set to true to ignore issues with an assignee (defaults to false)
 exemptAssignees: false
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,90 @@
...
+
+pull:
+  daysUntilClose: 14
+  daysUntilStale: 60
+  markComment: >
+    Hello 👋 Looks like there was no activity on this amazing PR for the last 60 days.
+    **Do you mind updating us on the status?** Is there anything we can help with? If you plan to still work on it, just comment on this PR or push a commit. Thanks! 🤗
+    
+    If there will be no activity in the next 2 weeks, this issue will be closed (we can always reopen a PR if you get back to this!).
+    Alternatively, use [`remind` command](https://probot.github.io/apps/reminders/) if you wish to be reminded at some point in future.
```
bwplotka, line 75: Great! But I think we don't have remind thingy. But we could mention stop staleness bot labels (:

kakkoyun: Good catch! I didn't realize it's a separate app. I'll remove those lines.

PR #1005, opened 2022-03-17
When prefix is empty, no more dots should be written

### LINKED ISSUE #1004: When prefix is empty, no more dots should be written
When prefix is empty, no more dots should be written
https://github.com/prometheus/client_golang/blob/868ec2137f6748fea08e345af668655c4c9e26ce/prometheus/graphite/bridge.go#L200-L207

zhijian-pro:
Close #1004


bwplotka (approved):
Great job, thanks!

PR #1010, opened 2022-03-18
Remove workaround for pre go1.15

johejo:
Related #980 #867 #866 


bwplotka:
👋🏽 Thanks for this!

I assume this means we are not supporting Go1.15 anymore? (We have `This library requires Go1.15 or later.` statement in our README).

bwplotka (approved):
Actually it is removing things required PRE 1.15, so all good. Thanks!

Issue #834, opened 2021-02-04
Enable deletion of metrics from a vector based on partially specified labels


beorn7:
In most cases, users of a metric vector are well aware of the label combinations for which metric children have been created in the vector. And they should be, because unpredictable label values often go along with unbounded cardinality. However, there are legitimate use cases where tracking the used label names is hard. (This could merely be a code maintenance problem, e.g. a HTTP counter partitioned by endpoints in a situation where other parts of the codebase can add endpoints.)

While those cases exist, it's also rarely _necessary_ to track the used label names. However, once you are in a situation where you want to delete metric children, you have to know them by their full label set. You cannot just say "delete all children with `tenant="42"` plus an unknown set of other labels. (Cortex instrumentation recently [ran into this problem when a tenant gets deleted entirely, and all the metrics related to it should disappear](https://github.com/cortexproject/cortex/pull/3782/).) This is essentially seeing the labeled metric children arranged in a tree in which you want to prune one whole branch.

Work-arounds include:
1. Separately track the used label names to allow deletion in the current way (with a full label set). (Which is kind of what Cortex is doing in the case above, but it is brittle: if another label is added somewhere in the code, it could be easily missed in the deletion code.)
2. Very hacky and weird: Call `Collect` on the vector and then `Write` on the collected `Metric` instances. Then look at the resulting protobuf message to find the metric children to delete.
3. Probably the most elegant with the current feature set: Have separate vectors for all the collectors for a single tenant, with the `tenant` label as a `ConstLabel`. Those collectors can then be separately registered and de-registered upon creation and removal of a tenant.

Note that it is not possible to use vector currying and then call `Reset` an a curried vector because `Reset` always acts on the underlying vector as a whole.

[...]

PR #1013, opened 2022-03-25
Implement deletion based on partially matching labels

stone-z:
Towards https://github.com/prometheus/client_golang/issues/834. This is an approach which I think feels user-friendly and accomplishes what folks were asking for. Open to suggestions of course, I'm new here.

This PR introduces a new deletion method:
- `DeletePartialMatch(labels Labels)` - deletes a metric if all of the given labels and associated values are present in that metric.

It returns the number of metrics which were deleted.

### DIFF
--- a/prometheus/vec.go
+++ b/prometheus/vec.go
@@ -80,6 +80,10 @@ func (m *MetricVec) DeleteLabelValues(lvs ...string) bool {
 	return m.metricMap.deleteByHashWithLabelValues(h, lvs, m.curry)
 }
 
+func (m *MetricVec) DeletePartialMatchLabelValues(lvs ...string) bool {
+	return m.metricMap.deleteByLabelValues(lvs, m.curry)
+}
+
 // Delete deletes the metric where the variable labels are the same as those
 // passed in as labels. It returns true if a metric was deleted.
 //
@@ -97,6 +101,10 @@ func (m *MetricVec) Delete(labels Labels) bool {
 	}
 
 	return m.metricMap.deleteByHashWithLabels(h, labels, m.curry)
+}
+
+func (m *MetricVec) DeletePartialMatch(labels Labels) bool {
+	return m.metricMap.deleteByLabels(labels, m.curry)
 }
 
 // Without explicit forwarding of Describe, Collect, Reset, those methods won't
@@ -353,6 +361,84 @@ func (m *metricMap) deleteByHashWithLabelValues(
 	return true
 }
 
+func (m *metricMap) deleteByLabelValues(
+	lvs []string, curry []curriedLabelValue,
+) bool {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	for h, metrics := range m.metrics {
+		i := findMetricWithPartialLabelValues(metrics, lvs, curry)
+		if i >= len(metrics) {
+			// Didn't find matching label values in this metric slice.
+			continue
+		}
+		delete(m.metrics, h)
+		return true
+	}
+
+	return false
+}
+
+// findMetricWithPartialLabelValues returns the index of the matching metric or
+// len(metrics) if not found.
+func findMetricWithPartialLabelValues(
+	metrics []metricWithLabelValues, lvs []string, curry []curriedLabelValue,
+) int {
+	for i, metric := range metrics {
+		if matchPartialLabelValues(metric.values, lvs, curry) {
+			return i
+		}
+	}
+	return len(metrics)
+}
+
+func matchPartialLabelValues(values []string, lvs []string, curry []curriedLabelValue) bool {
+	for _, v := range lvs {
+		// Check if the target value exists in our metrics and get the index.
+		if _, validValue := indexOfLabel(v, values); validValue {
+			continue
+		}
+		return false
+	}
+	return true
+}
+
+func xmatchPartialLabelValues(values []string, lvs []string, curry []curriedLabelValue) bool {
+	// if len(values) != len(lvs)+len(curry) {
+	// 	return false
+	// }
+
+	for i, v := range values {
+		fmt.Printf("%d, %v", i, v)
+		// See if the value was previously curried
+		for _, vCurry := range curry {
+			if vCurry.value == v {
+				continue
+			}
+		}
+		// for _, vMetric := range values {
+		// 	if vMetric
+		// }
+	}
+
+	var iLVs, iCurry int
+	for i, v := range values {
+		if iCurry < len(curry) && curry[iCurry].index == i {
+			if v != curry[iCurry].value {
+				return false
+			}
+			iCurry++
+			continue
+		}
+		if v != lvs[iLVs] {
+			return false
+		}
+		iLVs++
+	}
+	return true
+}
+
 // deleteByHashWithLabels removes the metric from the hash bucket h. If there
 // are multiple matches in the bucket, use lvs to select a metric and remove
 // only that metric.
@@ -379,6 +465,139 @@ func (m *metricMap) deleteByHashWithLabels(
 		delete(m.metrics, h)
 	}
 	return true
+}
+func (m *metricMap) deleteByLabels(
+	labels Labels, curry []curriedLabelValue,
+) bool {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	for h, metrics := range m.metrics {
+		i := findMetricWithPartialLabels(m.desc, metrics, labels, curry)
+		if i >= len(metrics) {
+			continue
+		}
+		delete(m.metrics, h)
+		return true
+	}
+
+	return false
+}
+
+func findMetricWithPartialLabels(
+	desc *Desc, metrics []metricWithLabelValues, labels Labels, curry []curriedLabelValue,
+) int {
+	for i, metric := range metrics {
+		if matchPartialLabels(desc, metric.values, labels, curry) {
+			return i
+		}
+	}
+	return len(metrics)
+}
+
+// TODO: This is untested and might not work
+func getCurriedLabelValue(desc *Desc, values []string, targetLabel string, curry []curriedLabelValue) (string, bool) {
+	iCurry := 0
+	for i, k := range desc.variableLabels {
+		if iCurry < len(curry) && curry[iCurry].index == i {
+			if k == targetLabel {
+				return curry[iCurry].value, true
+			}
+			iCurry++
+			continue
+		}
+	}
+	return "", false
+}
+
+// indexOfLabel searches the given list of labels for the target string and returns
+// the index or len(labels) as well as a boolean whether the search succeeded.
+func indexOfLabel(target string, labels []string) (int, bool) {
+	for i, l := range labels {
+		if l == target {
+			return i, true
+		}
+	}
+	return len(labels), false
+}
+
+func matchPartialLabels(desc *Desc, values []string, labels Labels, curry []curriedLabelValue) bool {
+	for l, v := range labels {
+		// Check if the target label exists in our metrics and get the index.
+		varLabelIndex, validLabel := indexOfLabel(l, desc.variableLabels)
+		if validLabel {
+			// Check the value of that label against the target value.
+			if values[varLabelIndex] == v {
+				continue
+
+			}
+		}
+		return false
+	}
+	return true
+}
+
+func xmatchPartialLabels(desc *Desc, values []string, labels Labels, curry []curriedLabelValue) bool {
+
+	// for label, value := range labels {
+	// 	// Check if the label was previously curried, and check the value if so.
+	// 	curriedValue, ok := getCurriedLabelValue(desc, values, label, curry)
+	// 	if ok {
+	// 		if curriedValue == value {
+	// 			continue
+	// 		}
+	// 	}
+
+	// 	// See if the target label is among the remaining labels in this metric, and get the index if so.
+
+	// }
+
+	iCurry := 0
+	for i, k := range desc.variableLabels {
+		// Check if the label was previously curried, and check the value if so.
+
+		if iCurry < len(curry) && curry[iCurry].index == i {
+			if values[i] != curry[iCurry].value {
+				return false
+			}
+			iCurry++
+			continue
+		}
+		if targetValue, ok := labels[k]; ok {
+			if values[i] != targetValue {
+				return false
+			}
+			continue
+		}
+
+		// }
+		// Check the value at the index. Abort if it doesn't match our value. Continue if it does.
+
+		// See if the target label is among the remaining labels in this metric, and get the index if so.
+
+		// Check the value at the index. Abort if it doesn't match our value. Continue if it does.
+
+		// If we haven't continued, the target label or the target value for a label did not exist. Abort.
+		return false
+	}
+
+	// If we get through all the labels, we've got a match.
+	return true
+
+	// iCurry := 0
+	// for i, k := range desc.variableLabels {
+	// 	if iCurry < len(curry) && curry[iCurry].index == i {
+	// 		if values[i] != curry[iCurry].value {
+	// 			return false
+	// 		}
+	// 		iCurry++
+	// 		continue
+	// 	}
+	// 	if values[i] != labels[k] {
+	// 		return false
+	// 	}
+	// }
+	// return true
 }
 
 // getOrCreateMetricWithLabelValues retrieves the metric by hash and label value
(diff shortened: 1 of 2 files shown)


prometheus/vec.go
```diff
@@ -80,6 +80,16 @@ func (m *MetricVec) DeleteLabelValues(lvs ...string) bool {
 	return m.metricMap.deleteByHashWithLabelValues(h, lvs, m.curry)
 }
 
+// DeletePartialMatchLabelValues deletes all metrics where the variable labels
+// contain all of the values passed. The order of the passed values does not matter.
+// It returns the number of metrics deleted.
+
+// This method deletes a metric if the given value is present for any label.
+// To delete a metric based on a partial match for a particular label, use DeletePartialMatch().
+func (m *MetricVec) DeletePartialMatchLabelValues(lvs ...string) int {
```
bwplotka, line 89: shouldn't we have label name passed for it too?

bwplotka: Do we have use case when we don't know label name? This seems tedious for implementation and prone to errors 🤔

bwplotka (commented):
Nice! It looks quite good to me however, I personally did not have a code that needs functionality like this. This is why it would be amazing to have opinions from potential users: @taiyang-li @pstibrany

I added a small comment though, I think the `DeletePartialMatchLabelValues(values..)` semantics is too open. I wonder if `DeletePartialMatchLabelValues(name, values ...)` make more sense (:

prometheus/vec.go
stone-z: You're right, I had in my head that deleting with unknown label names would also be needed but I don't have a use case for it personally. I'm fine to drop it and support only `DeletePartialMatch`. Alternatively, `DeletePartialMatchLabelValues` could be made to delete based on the presence of any of the given values for a particular label as you described, if that's desired behavior.

stone-z: I dropped the `DeletePartialMatchLabelValues` implementation -- if someone finds it to be valuable then we can re-introduce it

prometheus/vec_test.go
```diff
@@ -125,6 +125,86 @@ func testDeleteLabelValues(t *testing.T, vec *GaugeVec) {
...
+	c2.With(Labels{"l1": "11"}).Inc()
+
+	// Delete with valid curried pair l2: l2CurriedValue.
+	if got, want := c2.DeletePartialMatch(Labels{"l2": "l2CurriedValue"}), 1; got != want {
+		t.Errorf("got %v, want %v", got, want)
+	}
+	// Try to delete with a label value from before currying.
+	if got, want := c2.DeletePartialMatch(Labels{"l2": "v3"}), 0; got != want {
+		t.Errorf("got %v, want %v", got, want)
+	}
```
stone-z, line 188: I have a difficult time wrapping my head around the currying behavior, so I added some more tests and would like some clarification here. When we curry, we set the value of a label at some index to some value. Is that now _always_ the definitive value for that index, even if the vector was created before currying? 

In these tests, either line 182 or line 186 can succeed, but not both. In this test flow, we create a metric like
`{"l1": "v3", "l2": "v3"} 168`
Then, we create a new vector c2 where we curry with `{"l2": "l2CurriedValue"}`:
`{"l1": "11", "l2": "l2CurriedValue"} 1`

When it comes time to delete based on the `l2` label, the underlying c2 metrics _still_ contain 
`{"l1": "v3", "l2": "v3"}` and
`{"l1": "11", "l2": "l2CurriedValue"}`

And `curry` contains the `l2CurriedValue` at index 1.

I _think_ I should pretend as if the `v3` value does not exist anymore. Is that correct?

I guess for other deletions this is handled (or at least hidden) by the hash function, but how should it be handled in this case?

prometheus/vec.go
```diff
@@ -99,6 +99,16 @@ func (m *MetricVec) Delete(labels Labels) bool {
 	return m.metricMap.deleteByHashWithLabels(h, labels, m.curry)
 }
 
+// DeletePartialMatch deletes all metrics where the variable labels contain all of those
+// passed in as labels. The order of the labels does not matter.
+// It returns the number of metrics deleted.
+
+// This method deletes a metric if the given label: value pair is found in that metric.
+// To delete a metric if a particular value is found associated with any label, use DeletePartialMatchLabelValues().
```
beorn7, line 107: This line is obsolete now.

prometheus/vec_test.go
beorn7: My memory about the currying is quite distant (I implemented it long ago…), but as far as I do remember, the currying is like a "view" on the underlying original vector where the curried labels behave more or less like constant labels.

From that perspective, I think both line 182 and 186 should be considered invalid as they refer to a label that has been curried away. Note that the `Delete` method errors out if you include the curried label because the cardinality doesn't match anymore. I guess we want to avoid giving `DeletePartialMatch` an error return. In that case, I suggest to treat the curried labels the same as non-existing labels when `DeletePartialMatch` is called on a curried vector.

stone-z: I was thinking it makes sense to include curried labels in `DeletePartialMatch`, otherwise those metrics would still need to be entirely reconstructed in order to delete. Then it also doesn't change the behavior of `DeletePartialMatch` based on the workflow before it is called.

For example, a user could construct two vectors which are identical from the scrape perspective, but one which is curried and another which is not. It would be confusing if they then call `DeletePartialMatch` and only the non-curried vector disappears despite the curried one reporting the same label. With `DeletePartialMatch` the cardinality effectively doesn't matter anymore since users are expected to specify only a subset of the labels to begin with.

But if I'm understanding your response correctly, in any case the `v3` value should be concealed by currying, and calls to delete based on that value should not have any effect.

> From that perspective, I think both line 182 and 186 should be considered invalid as they refer to a label that has been curried away.

Based on the above, I think line 182 would then be the expected behavior (one metric would be deleted because the current representation would be `{"l1": "11", "l2": "l2CurriedValue"}` and the user has asked for deletion of all metrics with `{"l2": "l2CurriedValue"}`)

beorn7: > For example, a user could construct two vectors which are identical from the scrape perspective, but one which is curried and another which is not. It would be confusing if they then call DeletePartialMatch and only the non-curried vector disappears despite the curried one reporting the same label.

I think the terminology is a bit off here. A vector is the whole thing. What you delete are elements in the vector, AKA metrics.

If you curry a vector, the curried vector still refers to the same vector as the uncurried vector. If you delete a metric from the uncurried vector, it will also appear from the curried vector, because they are still the same vector. The curried vector is just a "view" on the uncurried vector.

The doc comment says: "CurryWith returns a vector curried with the provided labels, i.e. the returned vector has those labels pre-set for all labeled operations performed on it." I wouldn't like to change this semantics for the new `DeletePartialMatch`. Therefore, `curriedVec.DeletePartialMatch(Labels{"l1": "11"})` should have the exact same effect as `baseVec.DeletePartialMatch(Labels{"l1": "11", "l2": "l2CurriedValue"})`. This pattern should also be the one you usually use. It could be argued what should happen if the curried label is included. `curriedVec.DeletePartialMatch(Labels{"l1": "11", "l2": "l2CurriedValue"}})` could be seen as redundant and therefore should still have the same effect, while `curriedVec.DeletePartialMatch(Labels{"l1": "11", "l2": "somethingelse"}})` could then be seen as a contradiction, which should be a no-op or even an error. _However_ I think this view is at odds with the behavior of the plain `Delete` where `curriedVec.Delete(Labels{"l1": "11", "l2": "l2CurriedValue"}})` is _not_ seen as redundant but results in an error. Therefore, I prefer to treat a specification of the curried label the same as specifying a non-existing label, resulting in a no-op. (It would be even more consistent to return an error, but I do understand that we wouldn't like an error return just for this one purpose.)


> With DeletePartialMatch the cardinality effectively doesn't matter anymore since users are expected to specify only a subset of the labels to begin with.

Yes, but my point isn't so much that the plain `Delete` returns a cardinality error. It's that the curried label is added implicitly to all labeled operation before applying them to the underlying vector. If we allow a redundant specification of the curried label in `DeletePartialMatch`, we should allow the same for `Delete`, which would also resolve the cardinality error. But that would be a change in semantics (and arguably a confusing one).

prometheus/vec_test.go
```diff
@@ -125,6 +125,87 @@ func testDeleteLabelValues(t *testing.T, vec *GaugeVec) {
...
+	c1 := vec.MustCurryWith(Labels{"l1": "v1"})
+	c1.WithLabelValues("2").Inc()
+
+	// Try to delete nonexistent label lx with existent value v1.
+	if got, want := c1.DeletePartialMatch(Labels{"lx": "v1"}), 0; got != want {
+		t.Errorf("got %v, want %v", got, want)
+	}
+
+	// Delete with valid pair l1: v1.
+	if got, want := c1.DeletePartialMatch(Labels{"l1": "v1"}), 1; got != want {
```
beorn7, line 154: Following my previous line of argument, this should _not_ delete anything. (But `vec.DeletePartialMatch(Labels{"l1": "v1"})` should, which could be tested right after.)

prometheus/vec_test.go
```diff
@@ -125,6 +125,87 @@ func testDeleteLabelValues(t *testing.T, vec *GaugeVec) {
...
+	c3 := vec.MustCurryWith(Labels{"l2": "l2C3CurriedValue"}) // Used below
+	vec.With(Labels{"l1": "v3", "l2": "v3"}).(Gauge).Set(168)
+	if got, want := vec.DeletePartialMatch(Labels{"l1": "v1"}), 2; got != want {
+		t.Errorf("got %v, want %v", got, want)
+	}
+
+	// Try to delete a value which shouldn't be in our base vector (only the curried one c3).
+	if got, want := vec.DeletePartialMatch(Labels{"l2": "l2C3CurriedValue"}), 0; got != want {
+		t.Errorf("got %v, want %v", got, want)
+	}
```
beorn7, line 176: Note that this doesn't delete anything because there is no matching element neither in the base vector nor in the curried vector (which is only a "view" on the base vector). Currying alone doesn't create an element in the vector. You needed to call something like `c3.With(Labels{"l1": "something"})` first. After that, the test as written here (`vec.DeletePartialMatch(Labels{"l2": "l2C3CurriedValue"})`) _would_ actually delete an element.

prometheus/vec_test.go
```diff
@@ -125,6 +125,87 @@ func testDeleteLabelValues(t *testing.T, vec *GaugeVec) {
...
+
+	c2 := vec.MustCurryWith(Labels{"l2": "l2CurriedValue"})
+	c2.With(Labels{"l1": "11"}).Inc()
+
+	// Delete with valid curried pair l2: l2CurriedValue.
+	if got, want := c2.DeletePartialMatch(Labels{"l2": "l2CurriedValue"}), 1; got != want {
```
beorn7, line 182: Again, this shouldn't delete anything, IMHO.

prometheus/vec_test.go
```diff
@@ -125,6 +125,87 @@ func testDeleteLabelValues(t *testing.T, vec *GaugeVec) {
...
+	}
+
+	c3.With(Labels{"l1": "11"}).Inc()
+
+	// Try to delete with invalid curried pair l1: v1.
+	if got, want := c3.DeletePartialMatch(Labels{"l1": "v1"}), 0; got != want {
+		t.Errorf("got %v, want %v", got, want)
+	}
+	// Delete valid curried pair l2: l2C3CurriedValue.
+	if got, want := c3.DeletePartialMatch(Labels{"l2": "l2C3CurriedValue"}), 1; got != want {
```
beorn7, line 193: Same, shouldn't delete anything.

prometheus/vec_test.go
```diff
@@ -125,6 +125,87 @@ func testDeleteLabelValues(t *testing.T, vec *GaugeVec) {
...
+		t.Errorf("got %v, want %v", got, want)
+	}
+
+	// Try to delete with a label value from before currying.
+	if got, want := c2.DeletePartialMatch(Labels{"l2": "v3"}), 0; got != want {
```
beorn7, line 198: It's correct that this doesn't delete anything. But the reason is not that the label is "from before currying". The vector element `{"l1": "v3", "l2": "v3"}` still exists. It's just not accessible through the curried view (because that one hardcodes `l2` to `l2CurriedValue`, and therefore any attempt to specify the same or even a different value for `l2` should be considered invalid and not do anything).

beorn7 (changes requested):

taiyang-li:
> Nice! It looks quite good to me however, I personally did not have a code that needs functionality like this. This is why it would be amazing to have opinions from potential users: @taiyang-li @pstibrany
> 
> I added a small comment though, I think the `DeletePartialMatchLabelValues(values..)` semantics is too open. I wonder if `DeletePartialMatchLabelValues(name, values ...)` make more sense (:

This is the exact function I needed .

taiyang-li:
> > Nice! It looks quite good to me however, I personally did not have a code that needs functionality like this. This is why it would be amazing to have opinions from potential users: @taiyang-li @pstibrany
> > I added a small comment though, I think the `DeletePartialMatchLabelValues(values..)` semantics is too open. I wonder if `DeletePartialMatchLabelValues(name, values ...)` make more sense (:
> 
> This is the exact function I needed before.

prometheus/vec_test.go
stone-z: I've had some other priorities come up but I'm still planning to come back here.

I'll take more time to review your comment here and try to reflect it with test cases. I think I understand where you're coming from now, I guess what's confusing to me is the difference between the user's intent and the implementation when calling `DeletePartialMatch`. If my vector is curried with some value, and I call `DeletePartialMatch` with that label value, I'd expect that I would no longer see metrics containing that value. It might be redundant because by definition all metrics curried with that value will match, but I'm considering cases where a client calls `DeletePartialMatch` on a list of vectors, or where they can't reconstruct the other (uncurried) values for their vector. The alternative would be to call `DeletePartialMatch` on the base vector instead, but that may not be available to a caller.

However, I'm open to implementing it the way you've described. In that case, if only uncurried labels are supported for partial deletion, it may be worth adding an error to remain consistent with `Delete` and differentiate between situations where a user passed a valid label which matched zero metrics and those where a user passed an invalid (perhaps curried) value. What would your preference be on the error?

prometheus/vec_test.go
```diff
@@ -125,6 +125,87 @@ func testDeleteLabelValues(t *testing.T, vec *GaugeVec) {
...
 
+func TestDeletePartialMatch(t *testing.T) {
+	vec := NewGaugeVec(
+		GaugeOpts{
+			Name: "test",
+			Help: "helpless",
+		},
+		[]string{"l1", "l2"},
+	)
+	testDeletePartialMatch(t, vec)
```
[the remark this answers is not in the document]
stone-z: It was copied from another test. Removed now

stone-z:
Ok, I've updated this PR with what I think @beorn7 had in mind. I am open to returning an error if a user passes a curried label to `DeletePartialMatch`, but I checked and this would actually be different from `Delete` and `DeleteLabelValues`, which just return booleans. So I'm happy with this if the maintainers are as well. Thanks for your patience :pray:

beorn7:
I'm aware, and this is on my review queue, but the queue is very long and progressing slowly. I'll get to this eventually, but feel free to make the call without me.

bwplotka:
I think it's safe to merge as the API got smaller and very sensible, but waiting for your eyes @beorn7 if you want to check it still. I happy with this (:

beorn7 (approved):
Semantics looks OK to me now.

kakkoyun (approved):
LGTM

prometheus/vec.go
```diff
@@ -99,6 +99,16 @@ func (m *MetricVec) Delete(labels Labels) bool {
...
 
+// DeletePartialMatch deletes all metrics where the variable labels contain all of those
+// passed in as labels. The order of the labels does not matter.
+// It returns the number of metrics deleted.
+//
+// Note that curried labels will never be matched if deleting from the curried vector.
+// To match curried labels with DeletePartialMatch, it must be called on the base vector.
+func (m *MetricVec) DeletePartialMatch(labels Labels) int {
+	return m.metricMap.deleteByLabels(labels, m.curry)
+}
```
maelvls, line 110: What is the computational complexity of this function? I imagine it is `O(metrics count)`, right?

stone-z: Correct. I hoped to find a constant time implementation but didn't find one (at least at the time)

PR #1014, opened 2022-03-28
Update query API to support timeouts

### LINKED ISSUE #999: Allow passing timeout parameter in Prometheus queries
Prometheus HTTP API allows to pass `timeout` parameter for queries: https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries but API doesn't allow to pass it.
One easy way would be to add a new parameter to Query and QueryRange but that would be incompatible.
Another way could to to check if passed context has a deadline and set a timeout smaller than that deadline, like:

```
func (h *httpAPI) Query(ctx context.Context, query string, ts time.Time) (model.Value, Warnings, error) {
	u := h.client.URL(epQuery, nil)
	q := u.Query()

	q.Set("query", query)
	if !ts.IsZero() {
		q.Set("time", formatTime(ts))
	}
	if deadline, ok := ctx.Deadline(); ok {
		q.Set("timeout", fmt.Sprintf("%ds", int(time.Until(deadline).Truncate(time.Millisecond*500).Seconds())))
	}
[...]
```

but smaller by what duration? So one way or another extra information is needed I think.

JosephWoodward:
Fixes #999

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -240,7 +240,7 @@ type API interface {
 	// Query performs a query for the given time.
 	Query(ctx context.Context, query string, ts time.Time, opts ...Option) (model.Value, Warnings, error)
 	// QueryRange performs a query for the given range.
-	QueryRange(ctx context.Context, query string, r Range) (model.Value, Warnings, error)
+	QueryRange(ctx context.Context, query string, r Range, opts ...Option) (model.Value, Warnings, error)
 	// QueryExemplars performs a query for exemplars by the given query and time range.
 	QueryExemplars(ctx context.Context, query string, startTime time.Time, endTime time.Time) ([]ExemplarQueryResult, error)
 	// Buildinfo returns various build information properties about the Prometheus server
@@ -859,7 +859,7 @@ func (h *httpAPI) Query(ctx context.Context, query string, ts time.Time, opts ..
 	return model.Value(qres.v), warnings, json.Unmarshal(body, &qres)
 }
 
-func (h *httpAPI) QueryRange(ctx context.Context, query string, r Range) (model.Value, Warnings, error) {
+func (h *httpAPI) QueryRange(ctx context.Context, query string, r Range, opts ...Option) (model.Value, Warnings, error) {
 	u := h.client.URL(epQueryRange, nil)
 	q := u.Query()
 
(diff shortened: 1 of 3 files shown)


kakkoyun:
Hey @JosephWoodward, we plan to release soonish? Would you be able to have a look at this so that we can include it in the release?

JosephWoodward:
@kakkoyun Yes, no problem. IIRC the change is almost there, I'll have a look tonight.

api/prometheus/v1/api.go
```diff
@@ -818,10 +818,33 @@ func (h *httpAPI) LabelValues(ctx context.Context, label string, matches []strin
...
+
 	u := h.client.URL(epQuery, nil)
 	q := u.Query()
 
+	opt := &apiOptions{}
+	for _, o := range opts {
+		o(opt)
+	}
+
+	t := opt.timeout.Milliseconds()
```
kakkoyun, line 843:
```suggestion
	d := opt.timeout.Milliseconds()
```
nit: `d` would be more appropriate since it's a duration.

api/prometheus/v1/api.go
```diff
@@ -818,10 +818,33 @@ func (h *httpAPI) LabelValues(ctx context.Context, label string, matches []strin
...
+	if t > 0 {
+		q.Set("timeout", fmt.Sprint(t))
```
kakkoyun, line 845:
```suggestion
		q.Set("timeout", t.String())
```
I think we can just pass the duration, and let the server handle it https://github.com/prometheus/prometheus/blob/0b41fd6e716f3e8b79016f73b6dd77fdf5fcdced/web/api/v1/api.go#L1598-L1610

https://github.com/prometheus/common/blob/0c7319ad920e3239143af6c9f87149ff05faf2ab/model/time.go#L188-L234

kakkoyun (commented):
Looking good. Thanks a lot.

Would you consider adding the timeout to `QueryRange` as well? It's the same API.
https://prometheus.io/docs/prometheus/latest/querying/api/#range-queries

JosephWoodward:
> Looking good. Thanks a lot.
> 
> Would you consider adding the timeout to `QueryRange` as well? It's the same API. https://prometheus.io/docs/prometheus/latest/querying/api/#range-queries

Ah yes, thanks. I missed that one.

JosephWoodward:
@kakkoyun This is ready for another review. I've added support for timeouts to the `QueryRange` API too. Also, I noticed that the tests currently don't validate the request parameters the HTTP client sends, is there an issue to fix this or is there more context I'm unaware of here?

kakkoyun:
> Also, I noticed that the tests currently don't validate the request parameters the HTTP client sends, is there an issue to fix this or is there more context I'm unaware of here?

I believe I don't have the context either. Feel free to open an issue so that we can discuss.

kakkoyun (approved):
LGTM. 
Thanks for the contribution.

bboreham:
At first glance I thought this would set the context timeout. I think there should at least be a documentation comment saying what the intended behaviour is.

PR #1021, opened 2022-04-07
Refactor apiClientImpl.DoGetFallback

dohnto:
Make apiClientImpl.DoGetFallback more idiomatic and efficient:
 * Save result of args.Encode() operation as it might be used 2 times in
   the function and due to looping and sorting it might be heavy.
 * Follow line-of-sight practise and therefore simplify the code.

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -1133,7 +1133,8 @@ func (h *apiClientImpl) Do(ctx context.Context, req *http.Request) (*http.Respon
 // DoGetFallback will attempt to do the request as-is, and on a 405 or 501 it
 // will fallback to a GET request.
 func (h *apiClientImpl) DoGetFallback(ctx context.Context, u *url.URL, args url.Values) (*http.Response, []byte, Warnings, error) {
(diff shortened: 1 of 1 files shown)


api/prometheus/v1/api.go
```diff
@@ -1133,27 +1133,23 @@ func (h *apiClientImpl) Do(ctx context.Context, req *http.Request) (*http.Respon
...
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
 	resp, body, warnings, err := h.Do(ctx, req)
 	if resp != nil && (resp.StatusCode == http.StatusMethodNotAllowed || resp.StatusCode == http.StatusNotImplemented) {
-		u.RawQuery = args.Encode()
+		u.RawQuery = encodedArgs
 		req, err = http.NewRequest(http.MethodGet, u.String(), nil)
 		if err != nil {
 			return nil, nil, warnings, err
```
dohnto, line 1148: Also I am not sure if it is correct here to return `warnings`
```suggestion
			return nil, nil, nil, err
```

kakkoyun (approved):
LGTM :+1: Thanks for the contribution.

I'll go ahead and merge it since we have tests for this method and there doesn't seem to any regressions present.

Issue #1020, opened 2022-04-06
TCP Reset breaks connection without a retry


dohnto:
# My setup
We run a Grafana which queries Prometheus. Inbetween Grafana and Prometheus, there is a Google Cloud Load Balancer L7 which terminates SSL:

>   Grafana --> GCLB --> Prometheus

# Issue
Grafana user opens a dashboard and Grafana opens a long running TCP connection. Rarely we observe a TCP Reset sent by the Load Balancer, which Grafana displays to user as 5xx error. The request **is not retried**.

# Background

Grafana utilizes `client_golang` library and in particular its [`httpapi.QueryRange`](https://github.com/prometheus/client_golang/blob/v1.11.1/api/prometheus/v1/api.go#L821) method which relies on [`apiClientImpl.DoGetFallback`](https://github.com/prometheus/client_golang/blob/989baa30fe956631907493ccee1f8e7708660d96/api/prometheus/v1/api.go#L1100). This method tries to perform the query using POST request and in case of specific issues (405) it fallbacks to GET.

Prometheus' client [prefers](https://github.com/prometheus/client_golang/issues/801#issuecomment-689601312) to use POST over GET, but this causes inconsistencies as POST is not consider idempotent.

Quote from [`net/http`](https://pkg.go.dev/net/http)

> Transport only retries a request upon encountering a network error if the request is idempotent and either has no body or has its Request.GetBody defined. HTTP requests are considered idempotent if they have HTTP methods GET, HEAD, OPTIONS, or TRACE; or if their Header map contains an "Idempotency-Key" or "X-Idempotency-Key" entry. If the idempotency key value is a zero-length slice, the request is treated as idempotent but the header is not sent on the wire.

More information about this retry feature can be found in https://go-review.googlesource.com/c/go/+/3210/9//COMMIT_MSG . 
I found it really confusing, that when the implementation of  `apiClientImpl.DoGetFallback` fallbacks to GET request it might recover from some connectivity issues while with default POST request, it will just error.

# Expectations

For me ideally I would prefer making the POST request idempotent (as I believe they are; using one of the `Idempotency-*` header?). I am not sure if this would be acceptable as this changes the behavior of the client a bit. 

# Reproducer

None yet.

dohnto:
This is the related code which decides, whether a request is idempotent https://github.com/golang/go/blob/eca0d44cec58951fb716e540dcc21c0f527686d5/src/net/http/request.go#L1420-L1433

[...]

PR #1022, opened 2022-04-07
Make Query requests idempotent

dohnto:
Address #1020.

Signed-off-by: Tomáš Dohnálek <dohnto@gmail.com>

This was challenging to test. I used [toxiproxy](https://github.com/Shopify/toxiproxy) to inject connectivity issues. I build a [custom Toxic](https://github.com/Shopify/toxiproxy/blob/master/CREATING_TOXICS.md) which counts HTTP requests a closes every even request.

Without the header set, the conversation was following and my range query ended with
> Error querying Prometheus: Post "http://localhost:8080/api/v1/query_range": EOF

Screenshot from 2022-04-07 13-51-44

When I set the idempotency header, the conversation was equal to the case when GET is used - the connection is reopen and the request is retried.

Screenshot from 2022-04-07 13-53-40

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -1138,8 +1138,13 @@ func (h *apiClientImpl) DoGetFallback(ctx context.Context, u *url.URL, args url.
 		return nil, nil, nil, err
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
-	// Underlying `net.http` library automatically retries` idempotent requests when connectivity issues are hit.
-	// POST requests are not considered idempotent by default, so we need to explicitly mark them as such.
+	// Following comment originates from https://pkg.go.dev/net/http#Transport
+	// Transport only retries a request upon encountering a network error if the request is
+	// idempotent and either has no body or has its Request.GetBody defined. HTTP requests
+	// are considered idempotent if they have HTTP methods GET, HEAD, OPTIONS, or TRACE; or
+	// if their Header map contains an "Idempotency-Key" or "X-Idempotency-Key" entry. If the
+	// idempotency key value is a zero-length slice, the request is treated as idempotent but
+	// the header is not sent on the wire.
 	req.Header["Idempotency-Key"] = nil
 
 	resp, body, warnings, err := h.Do(ctx, req)
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -1140,7 +1140,7 @@ func (h *apiClientImpl) DoGetFallback(ctx context.Context, u *url.URL, args url.
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 	// Underlying `net.http` library automatically retries` idempotent requests when connectivity issues are hit.
 	// POST requests are not considered idempotent by default, so we need to explicitly mark them as such.
-	req.Header.Set("Idempotency-Key", "TODO")
+	req.Header["Idempotency-Key"] = nil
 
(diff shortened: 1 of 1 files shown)


api/prometheus/v1/api.go
```diff
@@ -1138,6 +1138,9 @@ func (h *apiClientImpl) DoGetFallback(ctx context.Context, u *url.URL, args url.
 		return nil, nil, nil, err
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	// Underlying `net.http` library automatically retries` idempotent requests when connectivity issues are hit.
+	// POST requests are not considered idempotent by default, so we need to explicitly mark them as such.
+	req.Header.Set("Idempotency-Key", "TODO")
```
dohnto, line 1143: Not sure which value we should set. It doesn't really matter as the go implementation only checks for the [header presence](https://github.com/golang/go/blob/eca0d44cec58951fb716e540dcc21c0f527686d5/src/net/http/request.go#L1429).

On the other hand this [IEFT draft](https://tools.ietf.org/id/draft-idempotency-header-01.html#name-uniqueness-of-idempotency-k) document claims that it must be unique and it is recommended to use UUID or similar function.

I made some research and this project doesn't use any UUIC library. [`mysql_exporter`](https://github.com/prometheus/mysqld_exporter/blob/main/go.mod#L7) utilizes `github.com/google/uuid` so that might be a good candidate. Another option would be to utilize [xxhash](https://github.com/prometheus/client_golang/blob/main/go.mod#L5) module, which is already utilized in `client_golang`.

bboreham: I would suggest `req.Header["Idempotency-Key"] = nil`, given the [docs ](https://pkg.go.dev/net/http#Transport) talking about "value is a zero-length slice"

dohnto: I missed that part. Thanks, updated my PR.

api/prometheus/v1/api.go
```diff
@@ -1138,6 +1138,9 @@ func (h *apiClientImpl) DoGetFallback(ctx context.Context, u *url.URL, args url.
 		return nil, nil, nil, err
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	// Underlying `net.http` library automatically retries` idempotent requests when connectivity issues are hit.
+	// POST requests are not considered idempotent by default, so we need to explicitly mark them as such.
+	req.Header["Idempotency-Key"] = nil
```
kakkoyun, line 1143: Could you please at the docs directly here?
```
Transport only retries a request upon encountering a network error if the request is idempotent and either has no body or has its Request.GetBody defined. HTTP requests are considered idempotent if they have HTTP methods GET, HEAD, OPTIONS, or TRACE; or if their Header map contains an "Idempotency-Key" or "X-Idempotency-Key" entry. If the idempotency key value is a zero-length slice, the request is treated as idempotent but the header is not sent on the wire.
```

And a link to the docs?

kakkoyun (commented):
This looks reasonable. Let's document more, and I'm happy to accept it.

api/prometheus/v1/api.go
dohnto: PTAL.

I think we should squash before merging. Not sure if I am suppose to do that, or the Github's UI is used for it?

kakkoyun (approved):

kakkoyun:
@dohnto Amazing work. Thanks for taking care of it.

bboreham:
Looking forward to having this change in a release.

PR #1025, opened 2022-04-11
client: Allow configuration of Client

yolossn:
Signed-off-by: yolossn <nssvlr@gmail.com>


api/client.go
```diff
@@ -52,6 +54,16 @@ func (cfg *Config) roundTripper() http.RoundTripper {
 	return cfg.RoundTripper
 }
 
+func (cfg *Config) client() *http.Client {
```
kakkoyun, line 57: I would rather prefer to have the same pattern we used for the `RounTripper`. It should be passed as part of configuration, if it doesn't present we can use the default. Also, we need to decide and document what happens when both `RoundTripper` and `Client` used at the same time. I'm in favor of returning an error.

kakkoyun:
@yolossn Could you also handle DCO failures?

kakkoyun:
@yeya24 @dsonck92 Does this fulfill your requirements?

kakkoyun (approved):
LGTM. Thanks for the contribution.

api/client.go
```diff
@@ -52,6 +62,13 @@ func (cfg *Config) roundTripper() http.RoundTripper {
 	return cfg.RoundTripper
 }
 
+func (cfg *Config) client() http.Client {
```
kakkoyun, line 65: This method needs to take account of the round tripper. We should consume the given roundtripper through config, no? Consider calling `cfg.roundTripper()` accordingly.

kakkoyun (changes requested):

api/client.go
yolossn: Hey, sorry there seems to be a confusion. The first solution was to use the RoundTripper provided in the configuration and you suggested to follow same pattern as `roundTripper()`
https://github.com/prometheus/client_golang/pull/1025#discussion_r847285988

kakkoyun: Sorry, for creating the confusion. I guess I've missed the fact when I wrote it :see_no_evil: We should use when the roundTripper specified.

dsonck92:
> @yeya24 @dsonck92 Does this fulfill your requirements?

Considering it's exactly as the potential implementation, yes :smile:

kakkoyun:
@yolossn PTAL https://github.com/yolossn/client_golang/pull/1

kakkoyun (approved):

kakkoyun (approved):

kakkoyun:
@bwplotka Any ideas why CircleCI is not properly functioning? Any way to re-run the checks?

yolossn:
@bwplotka  looks like the maintainer has to click Approve and Run since this is a first time contribution. 

image

kakkoyun:
Somehow, CircleCI ignores this PR. I'll merge it and if builds fail I'll fix it.

kakkoyun:
I can't even do that @bwplotka could you please try to merge this?

bwplotka:
Thanks!

PR #1028, opened 2022-04-12
push: Add PushContext and AddContext to Pusher

tatsuhiro-t:
Add PushContext and AddContext to Pusher, which are context-aware
version of Push and Add respectively.  They give a caller the ability
to cancel an HTTP request.

### DIFF
--- a/prometheus/push/push.go
+++ b/prometheus/push/push.go
@@ -36,6 +36,7 @@ package push
 
 import (
 	"bytes"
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
LGTM. It seems like a needed addition.

Thanks for taking care of it.

bwplotka (approved):
Thanks!

Issue #983, opened 2022-02-05
Add option to Go process metrics to fallback to pre v1.12.0 behaviour.


bwplotka:
AC:
* As per [comment](https://github.com/prometheus/client_golang/issues/967#issuecomment-1021650447), there is a variadic option on `func NewGoCollector() prometheus.Collector`, so  `func NewGoCollector(opts...GoOption) prometheus.Collector` where `GoOption` is some `func(o *goOptions)` and then we can have `WithMemStatsEnabled(bool)` `WithRuntimeMetricsEnabled(bool)` and both true by default, which allows to customize metric behaviour.

cc @kakkoyun @mknyszek

Issue #1026, opened 2022-04-11
New GoCollector metrics in 1.12 do not pass `promtool check metrics` lint


chancez:
Using git tag  v1.12.1, if we run `promtool check metrics` on our metrics we get the following failure:

```
go_gc_heap_allocs_by_size_bytes_total non-counter metrics should not have "_total" suffix
go_gc_heap_frees_by_size_bytes_total non-counter metrics should not have "_total" suffix
go_gc_pauses_seconds_total non-counter metrics should not have "_total" suffix
```

It seems that new metrics were potentially introduced in https://github.com/prometheus/client_golang/pull/955

I also filed https://github.com/prometheus/prometheus/issues/10574 before realizing these are new metrics as part of a library upgrade, though it still seems useful in general.

bwplotka:
Hmm, those are indeed histogram metrics, which are derived from `runtime/metrics` - thanks for pointing.

I would really consider renaming it now, given it's a fresh change (but this will break some compatibility though). Let me look on it.

Issue #995, opened 2022-03-04
Invalid le label value on go_gc_pauses_seconds_total_bucket


prymitive:
Reporting https://github.com/prometheus/client_golang/issues/967#issuecomment-1028850776 as a separate issue so it won't get lost

`go_gc_pauses_seconds_total_bucket` ends up with a negative `le` bucket and a `NaN` value for `go_gc_pauses_seconds_total_sum`
```
# HELP go_gc_pauses_seconds_total Distribution individual GC-related stop-the-world pause latencies.
# TYPE go_gc_pauses_seconds_total histogram
go_gc_pauses_seconds_total_bucket{le="-5e-324"} 0
go_gc_pauses_seconds_total_bucket{le="9.999999999999999e-10"} 0
go_gc_pauses_seconds_total_bucket{le="9.999999999999999e-09"} 0
go_gc_pauses_seconds_total_bucket{le="1.2799999999999998e-07"} 0
go_gc_pauses_seconds_total_bucket{le="1.2799999999999998e-06"} 0
go_gc_pauses_seconds_total_bucket{le="1.6383999999999998e-05"} 0
go_gc_pauses_seconds_total_bucket{le="0.00016383999999999998"} 3307
go_gc_pauses_seconds_total_bucket{le="0.0020971519999999997"} 6712
go_gc_pauses_seconds_total_bucket{le="0.020971519999999997"} 6809
go_gc_pauses_seconds_total_bucket{le="0.26843545599999996"} 6812
go_gc_pauses_seconds_total_bucket{le="+Inf"} 6812
go_gc_pauses_seconds_total_sum NaN
go_gc_pauses_seconds_total_count 6812
```

kakkoyun:
I guess there's a legitimate answer to that in the same thread by @mknyszek  https://github.com/prometheus/client_golang/issues/967#issuecomment-1029260834

That being said, we could consider an absorbing the negative values here.

fly3366:
Calculate quantile need bucket include Inf, but main branch has exclude Inf. So quantile return NaN. if someone need a patch, I will pr.

kakkoyun:
@fly3366 Feel free to contribute. The contributions are always welcome :hugs:  Please note we have reverted new enabling new collector by default and this is a bug in that. See #1033. There are several other issues we need to take care of for the new collector, so again help is very much appriciated.

kakkoyun:
Fixed with https://github.com/prometheus/client_golang/pull/1049

Issue #994, opened 2022-03-04
go metrics are tightly coupled with used go version


prymitive:
I've upgraded from Go 1.17.7 to 1.17.8 and I've noticed that exported metrics labels changed as a result.
le label on some histogram uses now different values than it used to:

```diff
--- a/metrics.txt
+++ b/metrics.txt
@@ -105,13 +105,13 @@ go_gc_heap_tiny_allocs_objects_total
 go_gc_pauses_seconds_total_bucket{le="-5e-324"}
 go_gc_pauses_seconds_total_bucket{le="9.999999999999999e-10"}
 go_gc_pauses_seconds_total_bucket{le="9.999999999999999e-09"}
-go_gc_pauses_seconds_total_bucket{le="1.2799999999999998e-07"}
-go_gc_pauses_seconds_total_bucket{le="1.2799999999999998e-06"}
-go_gc_pauses_seconds_total_bucket{le="1.6383999999999998e-05"}
-go_gc_pauses_seconds_total_bucket{le="0.00016383999999999998"}
-go_gc_pauses_seconds_total_bucket{le="0.0020971519999999997"}
-go_gc_pauses_seconds_total_bucket{le="0.020971519999999997"}
-go_gc_pauses_seconds_total_bucket{le="0.26843545599999996"}
+go_gc_pauses_seconds_total_bucket{le="9.999999999999998e-08"}
+go_gc_pauses_seconds_total_bucket{le="1.0239999999999999e-06"}
+go_gc_pauses_seconds_total_bucket{le="1.0239999999999999e-05"}
+go_gc_pauses_seconds_total_bucket{le="0.00010239999999999998"}
+go_gc_pauses_seconds_total_bucket{le="0.0010485759999999998"}
+go_gc_pauses_seconds_total_bucket{le="0.010485759999999998"}
+go_gc_pauses_seconds_total_bucket{le="0.10485759999999998"}
 go_gc_pauses_seconds_total_bucket{le="+Inf"}
 go_gc_pauses_seconds_total_sum NaN
 go_gc_pauses_seconds_total_count
@@ -240,13 +240,13 @@ go_sched_goroutines_goroutines
 go_sched_latencies_seconds_bucket{le="-5e-324"}
 go_sched_latencies_seconds_bucket{le="9.999999999999999e-10"}
 go_sched_latencies_seconds_bucket{le="9.999999999999999e-09"}
-go_sched_latencies_seconds_bucket{le="1.2799999999999998e-07"}
[...]

PR #1031, opened 2022-04-12
gocollector: Added options to Go Collector for changing the

bwplotka:
Fixes https://github.com/prometheus/client_golang/issues/983

Also:

* Improved docs
* fixed TestMemStatsEquivalence, it was noop before (:
* Removed gc_cpu_fraction metric completely, since it's not working completely for Go1.17+


bwplotka:
cc @mknyszek - we would love your eyes on this  (:

prometheus/go_collector_latest_test.go
```diff
@@ -227,9 +251,15 @@ func TestMemStatsEquivalence(t *testing.T) {
 	for i := 0; i < msRealValue.NumField(); i++ {
 		fr := msRealValue.Field(i)
 		ff := msFakeValue.Field(i)
-		switch typ.Kind() {
```
bwplotka, line 230: Fun fact. This test was noop, because `typ.Kind` was always `struct` - we want field kind instead 🙈

bwplotka:
After this is merged I will attempt to fix other issues:

https://github.com/prometheus/client_golang/issues/994
https://github.com/prometheus/client_golang/issues/995
https://github.com/prometheus/client_golang/issues/1026

I would also (in separate PR) propose to default to just `goRuntimeMemStats` for now instead of `goRuntimeMemStats | goRuntimeMetrics` set. I think migration to new metrics is important, but it has to be opt in for now IMO. We impacted a lot ppl by adding two digit more of metrics  and I am kind of missing the reason (users don't know how to leverage them yet) 🤔 Not sure, yet, let's discuss. (:

kakkoyun (commented):
Overall looks good.

`prometheus/collectors/go_collector.go` file missing build flags to exclude versions older than <1.17, I guess, considering CI failures.

bwplotka:
Fixed @kakkoyun 

@mknyszek I don't know much about `runtime/metrics.Read` and nothing about "stop-the-world" problem was written into its documentation (go docs). I assume we can remove section about the consequences of stop-the-world logic from Go Collector, right? Could you confirm? (:  I mean the following one:

```
// NewGoCollector returns a collector that exports metrics about the current Go
// process. This includes memory stats. To collect those, runtime.ReadMemStats
// is called. This requires to “stop the world”, which usually only happens for
// garbage collection (GC). Take the following implications into account when
// deciding whether to use the Go collector:
//
// 1. The performance impact of stopping the world is the more relevant the more
// frequently metrics are collected. However, with Go1.9 or later the
// stop-the-world time per metrics collection is very short (~25µs) so that the
// performance impact will only matter in rare cases. However, with older Go
// versions, the stop-the-world duration depends on the heap size and can be
// quite significant (~1.7 ms/GiB as per
// https://go-review.googlesource.com/c/go/+/34937).
//
// 2. During an ongoing GC, nothing else can stop the world. Therefore, if the
// metrics collection happens to coincide with GC, it will only complete after
// GC has finished. Usually, GC is fast enough to not cause problems. However,
// with a very large heap, GC might take multiple seconds, which is enough to
// cause scrape timeouts in common setups. To avoid this problem, the Go
// collector will use the memstats from a previous collection if
// runtime.ReadMemStats takes more than 1s. However, if there are no previously
// collected memstats, or their collection is more than 5m ago, the collection
// will block until runtime.ReadMemStats succeeds.
```

kakkoyun (approved):

prometheus/go_collector_latest_test.go
mknyszek: oops! good catch.

mknyszek:
Hey sorry for the delay. If it matters anymore, LGTM! Thanks, and sorry I never got around to this.

PR #925, opened 2021-10-19
Update /api/v1/status/tsdb to include headStats

prymitive:
Current structs for v1 HTTP API assume that there is chunkCount and timeSeriesCount on /api/v1/status/runtimeinfo, which seems to be gone for at least a few releases. /api/v1/status/tsdb now holds an extra headStats block with chunkCount, numSeries and a few other fields.
Update API models and tests to reflect this change.

Signed-off-by: Łukasz Mierzwa <l.mierzwa@gmail.com>

PR #1035, opened 2022-04-22
Add WithQueryStats option

prymitive:
https://github.com/prometheus/prometheus/pull/10369 was part of v2.35.0 release and it exposes more stats.
There's currently no way to access those stats since Query() and RangeQuery() returns only a subset of Prometheus API response.
To avoid bloating Query() return values I'm adding this as a new Option where you can pass a pointer to a QueryStats struct that will hold all recorded stats,
instead of adding another return value.

Signed-off-by: Lukasz Mierzwa <l.mierzwa@gmail.com>

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -421,10 +421,36 @@ type Metadata struct {
 	Unit string     `json:"unit"`
 }
 
+type stepStat struct {
+	T int64
+	V float64
+}
+
+type queryTimings struct {
+	EvalTotalTime        float64 `json:"evalTotalTime"`
+	ResultSortTime       float64 `json:"resultSortTime"`
+	QueryPreparationTime float64 `json:"queryPreparationTime"`
+	InnerEvalTime        float64 `json:"innerEvalTime"`
+	ExecQueueTime        float64 `json:"execQueueTime"`
+	ExecTotalTime        float64 `json:"execTotalTime"`
+}
+
+type querySamples struct {
+	TotalQueryableSamplesPerStep []stepStat `json:"totalQueryableSamplesPerStep,omitempty"`
+	TotalQueryableSamples        int        `json:"totalQueryableSamples"`
+	PeakSamples                  int        `json:"peakSamples"`
+}
+
+type QueryStats struct {
+	Timings queryTimings `json:"timings"`
+	Samples querySamples `json:"samples"`
+}
+
 // queryResult contains result data for a query.
 type queryResult struct {
 	Type   model.ValueType `json:"resultType"`
 	Result interface{}     `json:"result"`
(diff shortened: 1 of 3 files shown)


api/prometheus/v1/api.go
```diff
@@ -845,6 +883,10 @@ func (h *httpAPI) Query(ctx context.Context, query string, ts time.Time, opts ..
 		q.Set("timeout", d.String())
 	}
 
+	if opt.perStepStats {
```
bwplotka, line 886: I think it would be nicer to stick to some enum like `all` and `minimal` instead of `opt.perStepStats` bool - we will be resilient to adding any other option.

api/prometheus/v1/api.go
```diff
@@ -830,6 +861,13 @@ func WithTimeout(timeout time.Duration) Option {
 	}
 }
 
+func WithQueryStats(s *QueryStats, perStep bool) Option {
```
bwplotka, line 864: This is bit unexpected as a start what will return value via variadic option (: So let's add a comment at least.

But I would be curious if we can actually change `model.Value` with some interface that is like model.Value but also returns stats 🤔 Will be much nicer for users.

And this package is in experimental stage so we can change API btw.

bwplotka (commented):
Nice! 

2 comments to discuss, otherwise good 👍🏽

api/prometheus/v1/api.go
prymitive: It's mostly boolean because I based this on the PR for Prometheus that adds it, and it had a bool - https://github.com/prometheus/prometheus/blob/ce6a643ee88fba7c02fbd0459c4d0ac498f512dd/promql/engine.go#L130

The only supported value here is "all", at least at the moment.
https://github.com/prometheus/prometheus/blob/a64b9fe323cf913c212c1829fe93d1a9dbc63348/web/api/v1/api.go#L433
Not sure if it's worth to create a client abstraction for it but happy to do so.

IMHO ideally it should be an enum in Prometheus source, since that's where it's consumed and used, but I imagine we want to avoid prometheus & client_golang circular dependencies, so it seems like we'll need to duplicate supported values in client_golang.

Side note: I do find it awkward at time to use client_golang to interact with Prometheus, it's easy for client_golang to fall behind Prometheus (structs for JSON might not get updated, #925) so I wonder if in the future it would make sense to move API client libs to Prometheus repo itself, especially now that it's versioned better and so it's easier to import and use.

api/prometheus/v1/api.go
prymitive: > And this package is in experimental stage so we can change API btw.

Brilliant, I'll try to come up with an API change that makes it easier to consume this.

Issue #685, opened 2019-11-28
OpenMetrics creation time support (in v1)


beorn7:
OpenMetrics has the option of marking creation times of metrics. As this doesn't happen as part of some 1st class metadata but as an additional time series with a magic suffix `_created`, we cannot just transparently support it, but have to make it configurable: Either don't use them at all or (opt in) use them but error out if a name collision happens.

Issue #144, opened 2015-07-15
Either bring Makefile in line with other projects or remove it altogether.


beorn7:
The `Makefile` is only used to run the examples and tests. It's usefulness has been disputed. Even in our own public instructions to prepare for Prometheus workshops, we tell the users to run `go get` and `go build` instead of `make`.

So we could either get rid of the `Makefile` (with the only downside that the CI will fail all of a sudden if a breaking change of a dependency happens - which could also be considered an advantage), or we should at least bring the Makefile in line with other repos, i.e. use the common Makefile from the `utils` repo.

beorn7:
Fixed.

PR #1044, opened 2022-05-03
add _created support: add createdTime() method to counter

andysim3d:
first step of issue #685 

Signed-off-by: Pengfei Zhang <zhangpengfei@google.com>

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -41,9 +41,6 @@ type Counter interface {
 	// Add adds the given value to the counter. It panics if the value is <
 	// 0.
 	Add(float64)
-
-	// Get _created suffixed metric. Records when it got first exposed.
-	Created() float64
 }
 
 // ExemplarAdder is implemented by Counters that offer the option of adding a
@@ -164,10 +161,6 @@ func (c *counter) updateExemplar(v float64, l Labels) {
 		panic(err)
 	}
 	c.exemplar.Store(e)
-}
-
-func (c *counter) Created() float64 {
-	return float64(c.createdTime)
 }
 
 func (c *counter) initCreated() {
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -41,6 +41,9 @@ type Counter interface {
 	// Add adds the given value to the counter. It panics if the value is <
 	// 0.
 	Add(float64)
+
+	// Get _created suffixed metric. Records when it got first exposed.
+	Created() float64
 }
 
 // ExemplarAdder is implemented by Counters that offer the option of adding a
@@ -79,6 +82,7 @@ func NewCounter(opts CounterOpts) Counter {
 		opts.ConstLabels,
 	)
 	result := &counter{desc: desc, labelPairs: desc.constLabelPairs, now: time.Now}
+	result.initCreated()
 	result.init(result) // Init self-collection.
 	return result
 }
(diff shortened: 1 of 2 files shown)


andysim3d:
@beorn7 Please take a look since you are the original owner of that issue.

beorn7:
While I opened this issue, this has to be figured out between the OpenMetrics experts and the current maintainers of prometheus/client_golang.

From a very general perspective, I don't think `Created()` should be a method (similar to not making the counter value itself accessible via some kind of `Get()` method).

prometheus/counter.go
```diff
@@ -41,6 +41,9 @@ type Counter interface {
 	// Add adds the given value to the counter. It panics if the value is <
 	// 0.
 	Add(float64)
+
+	// Get _created suffixed metric. Records when it got first exposed.
+	Created() float64
```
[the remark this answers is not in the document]
beorn7: Indeed. I guess "when it got initialized" might be a more accurate description.

kakkoyun (commented):
We haven't discussed OpenMetrics support among maintainers yet. cc @bwplotka 
I'll add this to the milestone. And ask for your further patience for the decision.

pintohutch:
Hello,

Is this still being discussed?

logicalhan:
Can we please do this? `_created` timeseries is a necessary prerequisite for upstream Kubernetes to even consider supporting prometheus' open-metrics format.

prometheus/counter.go
```diff
@@ -88,8 +89,9 @@ type counter struct {
 	// valInt stores values that are exact integers. Both have to go first
 	// in the struct to guarantee alignment for atomic operations.
 	// http://golang.org/pkg/sync/atomic/#pkg-note-BUG
-	valBits uint64
-	valInt  uint64
+	valBits     uint64
+	valInt      uint64
+	createdTime uint64
```
bwplotka, line 94: I don't think this field is ever used, no?

bwplotka (commented):
Hi, 

Thanks for this. We decided on Prometheus DevSummit to pursue proto version of OpenMetrics format as the main one. So I would implement that first.

For OM v1.0 text format, there are many weird conflicts and incompatibilities (e.g spam of extra metric on all Prometheus versions), so we can't really support it in client_golang and definitely not by default. I would suggest waiting on text improvement for OM e.g in OM 2.0 and rely on proto only. 

WDYT?

prometheus/counter.go
pintohutch: It's initialized on L166 in `initCreated()` call.

Presumably, a forthcoming change would use it for serving in a Prometheus counter metric.

bwplotka: Yup, but we would need this to be either in this PR ideally so we see how end solution would look like 🤗 

Definitely it's good direction!

bwplotka:
BTW, @ArthurSens, our mentee works on adding created timestamp with tests https://github.com/prometheus/client_golang/pull/1044

bwplotka:
Done in https://github.com/prometheus/client_golang/pull/1044, so closing this PR. Thanks anyway!

logicalhan:
> Done in #1044, so closing this PR. Thanks anyway!

this is PR #0144...

ArthurSens:
I believe he meant #1313

PR #1048, opened 2022-05-06
prometheus: Fix convention violating names for generated collector metrics

### LINKED ISSUE #1026: New GoCollector metrics in 1.12 do not pass `promtool check metrics` lint

kakkoyun:

Fixes #1026 

cc @chancez

### Test

```
> go run examples/gocollector/main.go
> curl http://localhost:8080/metrics | promtool check metrics
```


bwplotka:
1.16 is not happy ): 
 
image

PR #1056, opened 2022-05-21
Enable same linters as the Prometheus repo itself

sazary:
This will enable same linters in `golangci-lint` as the base [Prometheus repo](https://github.com/prometheus/prometheus), and fixes current issues that linters complain about.  
I've copied the config in that repo, specifically [this version](https://github.com/prometheus/prometheus/blob/d3c9c4f57488d924cf9a6ba3b1a7e98f6cf35464/.golangci.yml) and removed some parts:

 - I've removed `depguard` because I don't know what are the policies in this repo about the dependencies, and the requirements of the source repo are not met here
 - In the source config the generated files are all excluded from linting, but we don't have a good naming convention for filenames of generated files here.

I've split the changes applied for each linter into separate commits so that it's easier for review. I think the most critical changes that need review are in the 3dfd8d0 where I've fixed `errcheck` issues.

Signed-off-by: sazary <soroosh@azary.ir>


sazary:
@bwplotka @kakkoyun could you please take the time to review this? thanks

.golangci.yml
```diff
@@ -1,5 +1,31 @@
...
+    - gofumpt
+    - goimports
+    - revive
+    - misspell
+
+issues:
+  max-same-issues: 0
+  exclude-rules:
+    - path: _test.go
+      linters:
```
kakkoyun, line 20: These linters don't seem to be enabled, no?

prometheus/collectors/go_collector_latest.go
```diff
@@ -72,7 +72,8 @@ const (
 //
 // The current default is GoRuntimeMemStatsCollection, so the compatibility mode with
 // client_golang pre v1.12 (move to runtime/metrics).
-func WithGoCollections(flags GoCollectionOption) goOption {
+//nolint:staticcheck // Ignore SA1019 until v2.
```
kakkoyun, line 75: Is staticcheck enabled? From the config above, it doesn't seem like it.

kakkoyun (commented):
This looks great! I'm happy to merge it. And sorry for the late response :)

I have a couple of questions left. I would be glad if you can clarify those.

.golangci.yml
sazary: `golangci-lint` has a list of default linters that are enabled by default, unless being disabled explicitly or implicitly. in the previous config they were disabled implicitly by using `disable-all: true` in [this line](https://github.com/prometheus/client_golang/pull/1056/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9L5)

the [default linters](https://golangci-lint.run/usage/linters/#enabled-by-default) list is:

  - deadcode
  - errcheck
  - gosimple
  - govet
  - ineffassign
  - staticcheck
  - structcheck
  - typecheck
  - unused
  - varcheck

since in the main prometheus repo they left them enabled implicitly i did the same too but i understand that it could reduce readability. 

do you want me to enable them in the config explicitly or leave them as they are?

kakkoyun: Having them explicit would be amazing. Also, this would prevent any unwanted issues, if we upgrade golangci-lint and decide to change their default preset.

kakkoyun (approved):
LGTM.

Let's merge this and see how it goes!

PR #1057, opened 2022-05-24
Add BasicAuth in promhttp handler

alan7yg:
I'm having trouble while try to protect metrics from my application. 
Since I'm working on Kubernetes with Prometheus-Operator, BasiceAuth is my first plan. I implemented the BasicAuth authorization in promhttp package, as a new HandlerOpts parameter member.

Signed-off-by: alan7yg <alanlyg@gmail.com>

### DIFF
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -119,6 +119,18 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 	}
 
 	h := http.HandlerFunc(func(rsp http.ResponseWriter, req *http.Request) {
+		if opts.HandlerBasicAuthOpts != nil {
+			username, password, ok := req.BasicAuth()
+			if !ok {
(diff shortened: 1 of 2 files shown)


SuperQ:
If you use the [`exporter-toolkit`](https://github.com/prometheus/exporter-toolkit) there is a web package that does this.

alan7yg:
@SuperQ Good to know. But I was supposed to enhance the usage of promhttp.Handler from controller-runtime project, code goes from here: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/internal.go#L326
So, I'm not sure if importing [exporter-toolkit](https://github.com/prometheus/exporter-toolkit) into controller-runtime package is appropriate, since it is still experimental.

kakkoyun:
@alan7yg Thanks for the PR and contribution. As suggested in the comments, it's better to have an authentication layer on top of the SDK.

Issue #491, opened 2018-11-05
Feature: Expose a request path label in the `http_request_*` metric by default


zephinzer:
Currently, metrics are exposed as:

```
http_request_duration_microseconds{handler="users",quantile="0.5"} 401.237
...
http_request_duration_microseconds{handler="users",quantile="0.5"} 652.793
...
http_request_duration_microseconds{handler="users",quantile="0.5"} 652.793
```

I would like to add on to this by also allowing for a `path` label to be added so that each route does not need to be manually instrumented with a handler name. I'm proposing that the default collection should result in:

```
http_request_duration_microseconds{handler="users",path="/users/logout",quantile="0.5"} 401.237
...
http_request_duration_microseconds{handler="users",path="/users/login",quantile="0.5"} 652.793
...
http_request_duration_microseconds{handler="users",path="/users/status",quantile="0.5"} 652.793
...
# other examples I think would be useful
http_requests_total{code="200",handler="users",path="/users/status",method="get"} 3
promhttp_metric_handler_requests_total{path="/some/other/endpoint",code="200"} 2
promhttp_metric_handler_requests_total{path="/another/endpoint",code="500"} 0
promhttp_metric_handler_requests_total{path="/yet/another/endpoint",code="503"} 0
```

This could be done by adding `r.URL.Path` to the labels creation as a `path` label.

I'm new to Golang but would be happy to attempt this with some guidance!

Any opinions?

stuartnelson3:
That was our concern. Some users could be putting user IDs or some other non-bounded set of data in the path, so we decided to avoid potentially destroying our users' prometheus servers.

beorn7:
Prometheus itself would suffer. The TSDB of Prometheus has a very low price per sample (usually between one and two bytes) but a much much higher price per serios (kilobytes). Thus, your server would explode.

I'm not suggesting to use a 3rd party package. All the tools you need are included in client_golang. You should just not use the convenience-middlewares but the usual instrumentation primitives as described in the [linked video](https://youtu.be/HkEZ1LJ7kzQ?t=617).

zephinzer:
cool, thanks for the advice on the potential effects on tsdb @beorn7! I gave it a watch and have managed to get up my own middleware to do something like what I was requesting for, am closing this-

but if anyone else is looking for such a convenience middleware despite the above warnings, I've published it at [https://github.com/zephinzer/ezpromhttp](https://github.com/zephinzer/ezpromhttp)!

[...]

PR #1066, opened 2022-06-16
Add possibility to dynamically get label values for http instrumentation

Okhoshi:
Signed-off-by: Quentin Devos <4972091+Okhoshi@users.noreply.github.com>

I need to partition the metrics generated by promhttp by the value coming in a specific header, and while I appreciate the design of promhttp is made to be simple and cover only the base use cases, it would be great if we could have a bit more of flexibility.

It would also give a way to solve #491 elegantly while letting the user control the cardinality of the label they add.

@bwplotka  @kakkoyun

### DIFF
--- a/prometheus/promhttp/instrument_client.go
+++ b/prometheus/promhttp/instrument_client.go
@@ -68,16 +68,17 @@ func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.Rou
 		o.apply(rtOpts)
 	}
 
-	code, method := checkLabels(counter)
+	// Curry the counter with dynamic labels before checking the remaining labels.
+	code, method := checkLabels(counter.MustCurryWith(rtOpts.emptyDynamicLabels()))
 
 	return func(r *http.Request) (*http.Response, error) {
 		resp, err := next.RoundTrip(r)
 		if err == nil {
-			addWithExemplar(
-				counter.With(labels(code, method, r.Method, resp.StatusCode, rtOpts.extraMethods...)),
-				1,
-				rtOpts.getExemplarFn(r.Context()),
-			)
+			l := labels(code, method, r.Method, resp.StatusCode, rtOpts.extraMethods...)
+			for label, resolve := range rtOpts.extraLabelsFromCtx {
+				l[label] = resolve(resp.Request.Context())
+			}
+			addWithExemplar(counter.With(l), 1, rtOpts.getExemplarFn(r.Context()))
 		}
 		return resp, err
 	}
@@ -110,17 +111,18 @@ func InstrumentRoundTripperDuration(obs prometheus.ObserverVec, next http.RoundT
 		o.apply(rtOpts)
 	}
 
-	code, method := checkLabels(obs)
+	// Curry the observer with dynamic labels before checking the remaining labels.
+	code, method := checkLabels(obs.MustCurryWith(rtOpts.emptyDynamicLabels()))
 
 	return func(r *http.Request) (*http.Response, error) {
 		start := time.Now()
 		resp, err := next.RoundTrip(r)
 		if err == nil {
-			observeWithExemplar(
-				obs.With(labels(code, method, r.Method, resp.StatusCode, rtOpts.extraMethods...)),
-				time.Since(start).Seconds(),
-				rtOpts.getExemplarFn(r.Context()),
-			)
+			l := labels(code, method, r.Method, resp.StatusCode, rtOpts.extraMethods...)
+			for label, resolve := range rtOpts.extraLabelsFromCtx {
+				l[label] = resolve(resp.Request.Context())
+			}
+			observeWithExemplar(obs.With(l), time.Since(start).Seconds(), rtOpts.getExemplarFn(r.Context()))
 		}
 		return resp, err
 	}
--- a/prometheus/promhttp/option.go
+++ b/prometheus/promhttp/option.go
@@ -24,14 +24,32 @@ type Option interface {
 	apply(*options)
 }
 
+// LabelValueFromCtx are used to compute the label value from request context.
+// Context can be filled with values from request through middleware.
+type LabelValueFromCtx func(ctx context.Context) string
+
 // options store options for both a handler or round tripper.
 type options struct {
-	extraMethods  []string
-	getExemplarFn func(requestCtx context.Context) prometheus.Labels
+	extraMethods       []string
+	getExemplarFn      func(requestCtx context.Context) prometheus.Labels
+	extraLabelsFromCtx map[string]LabelValueFromCtx
 }
 
 func defaultOptions() *options {
-	return &options{getExemplarFn: func(ctx context.Context) prometheus.Labels { return nil }}
+	return &options{
+		getExemplarFn:      func(ctx context.Context) prometheus.Labels { return nil },
+		extraLabelsFromCtx: map[string]LabelValueFromCtx{},
+	}
+}
+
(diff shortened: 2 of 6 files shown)


Okhoshi:
Gently repinging @bwplotka and @kakkoyun for a review of this PR ?

Let me know if it doesn't have any chance to be merged, so that I don't wait for it 👍 

Also, I saw it's quite similar to #1104, probably share the same concern (that user can now add as many label values as it wants) but I also have the same thought than @chradcliffe on the fact that a user is knowingly added a high cardinality label by using the option, and therefore it's not promhttp responsibility.

Note that I could work with #1104 even though using context would force me to add another middleware to push request/response value into the context.

Okhoshi:
I added a way to constraint the values any dynamic label can take, such that the cardinality has to be known in advance. Would that address your concerns @bwplotka ?

prometheus/promhttp/instrument_client.go
```diff
@@ -110,17 +111,22 @@ func InstrumentRoundTripperDuration(obs prometheus.ObserverVec, next http.RoundT
 		o.apply(rtOpts)
 	}
 
-	code, method := checkLabels(obs)
+	dynamicLabels := prometheus.Labels{}
+	for label := range rtOpts.labelResolverFns {
+		dynamicLabels[label] = "dummy"
```
bwplotka, line 116: Do we want this in production code? 🤔

prometheus/promhttp/instrument_client.go
```diff
@@ -68,16 +68,17 @@ func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.Rou
 		o.apply(rtOpts)
 	}
 
-	code, method := checkLabels(counter)
+	// Curry the counter with dynamic labels before checking the remaining labels
```
bwplotka, line 71:
```suggestion
	// Curry the counter with dynamic labels before checking the remaining labels.
```
And below.

prometheus/promhttp/option.go
```diff
@@ -56,3 +76,24 @@ func WithExemplarFromContext(getExemplarFn func(requestCtx context.Context) prom
 		o.getExemplarFn = getExemplarFn
 	})
 }
+
+// WithLabelResolver register a label for dynamic resolution with access to request information.
+// Values returned by the resolver that would not be present in allowedValues would be replaced
+// with "unknown" value instead. Values in allowedValues are used as is.
+// See the example for ExampleInstrumentHandlerWithLabelResolver for example usage
+func WithLabelResolver(label string, allowedValues []string, resolver LabelResolver) Option {
```
bwplotka, line 84: I think I like it, but I think perhaps we could embed that rather into each metric implementation. Some options:

1.  Add `SafeWith` and `SafeWithLabelValues` in e.g. `Observer` (not sure if `Safe` is a good word)

```
type ObserverVec interface {
	GetMetricWith(Labels) (Observer, error)
	GetMetricWithLabelValues(lvs ...string) (Observer, error)
	With(Labels) Observer
	WithLabelValues(...string) Observer
        SafeWith(Labels) Observer // <---
	SafeWithLabelValues(...string) Observer // <---
	CurryWith(Labels) (ObserverVec, error)
	MustCurryWith(Labels) ObserverVec

	Collector
}
```

2. Let metric constructor option for adding label values that are "enforced" or "safe"

If we like this approach, I think I am ok in doing this in another PR and adding some dynamic labels options in middleware and tripperware here.

WDYT @kakkoyun?

prometheus/promhttp/option.go
```diff
@@ -24,14 +26,32 @@ type Option interface {
 	apply(*options)
 }
 
+// LabelResolver are used to compute the label value from request or response.
+// Response will be provided only when used on a round tripper
+type LabelResolver func(*http.Request, *http.Response) string
```
bwplotka, line 31: Hm, to be consistent. Ideally, we want to have RespWriter here as well - which might mean separate abstractions. Also, this is not ideal as somebody might want to edit those as well (I guess we don't have any option here - we can't mark them for read-only - unless we want to somehow pack them in interfaces, which might be a fuss).

That's why in the way... https://github.com/prometheus/client_golang/pull/1104 change reduces the surface of error - one has to put some value in to context in extra middleware if they want it to get some label in and makes this dynamic function read only. It also means more code to user.

bwplotka (commented):
Thanks for this!

I think with more and more requests to add dynamic labels to `promhttp` - we should do it.

I like your idea for enforcing label values, but it might be better to have something built-in metric so the same enforcing or categorization logic can be reused across anyone who uses `prometheus.Observer` or `prometheus.Adder`

With this direction, I would be happy with solutions like yours or https://github.com/prometheus/client_golang/pull/1104 to add dynamic labels without verification.

As per ctx vs pointers structs, the ctx will be safer for this library in some sense but tedious for users... I think I would favour your idea (but perhaps separate for client and server?) Read only thing is nice, but there is more risk for users to add new middleware, makes sure you don't forget to add it and put correct context value that does not conflict and test etc than risk of someone modifying request or response in our  `dynLabelValues(req, resp) []string`

prometheus/promhttp/option.go
```diff
@@ -24,14 +24,32 @@ type Option interface {
...
+// Context can be filled with values from request through middleware.
+type LabelResolver func(ctx context.Context) string
+
 // options store options for both a handler or round tripper.
 type options struct {
-	extraMethods  []string
-	getExemplarFn func(requestCtx context.Context) prometheus.Labels
+	extraMethods     []string
+	getExemplarFn    func(requestCtx context.Context) prometheus.Labels
+	labelResolverFns map[string]LabelResolver
```
bwplotka, line 35: What about 👇🏽 

```suggestion
	extraLabelsFromCtx map[string]LabelResolver
```

prometheus/promhttp/option.go
```diff
@@ -24,14 +24,32 @@ type Option interface {
 	apply(*options)
 }
 
+// LabelResolver are used to compute the label value from request context.
+// Context can be filled with values from request through middleware.
+type LabelResolver func(ctx context.Context) string
```
bwplotka, line 29:
```suggestion
type LabelValueFromCtx func(ctx context.Context) string
```

prometheus/promhttp/option.go
```diff
@@ -56,3 +74,11 @@ func WithExemplarFromContext(getExemplarFn func(requestCtx context.Context) prom
 		o.getExemplarFn = getExemplarFn
 	})
 }
+
+// WithLabelResolver register a label for dynamic resolution with access to request information.
+// See the example for ExampleInstrumentHandlerWithLabelResolver for example usage
+func WithLabelResolver(label string, resolver LabelResolver) Option {
```
bwplotka, line 80:
```suggestion
func WithLabelFromCtx(name string, valueFn LabelValueFromCtx) Option {
```

bwplotka (approved):
Looks good, just some ideas for more explicit naming, WDYT?

Also tests fail 🤗

Okhoshi:
> Looks good, just some ideas for more explicit naming, WDYT?

That naming is way better than mine 💯 Changed.

> Also tests fail 🤗

Yes, something weird about not being able to find some git references 🤷

bwplotka (approved):
LGTM, solid work. Thanks for patience 💪🏽 (:

jcass8695:
Hey gang, I was wondering if there was any estimate on when there could be a release cut with this PR included? My team are keen to use this great feature. TIA 😌

PR #1067, opened 2022-06-17
Add MustCollector API for pusher

oiooj:
@kakkoyun PTAL

### DIFF
--- a/prometheus/push/push.go
+++ b/prometheus/push/push.go
@@ -168,6 +168,16 @@ func (p *Pusher) Collector(c prometheus.Collector) *Pusher {
 	return p
 }
 
+// MustCollector works like Collector but registers any number of
+// Collectors and panics upon the first registration that causes an
+// error.
+//
+// For convenience, this method returns a pointer to the Pusher itself.
+func (p *Pusher) MustCollector(c ...prometheus.Collector) *Pusher {
+	p.registerer.MustRegister(c...)
+	return p
+}
+
 // Grouping adds a label pair to the grouping key of the Pusher, replacing any
 // previously added label pair with the same label name. Note that setting any
 // labels in the grouping key that are already contained in the metrics to push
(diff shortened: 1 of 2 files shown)


kakkoyun:
@oiooj Could you please explain the justification behind this change? It's hard to understand without any context.

prometheus/push/push.go
```diff
@@ -168,6 +168,16 @@ func (p *Pusher) Collector(c prometheus.Collector) *Pusher {
 	return p
 }
 
+// MustCollector works like Collector but registers any number of
+// Collectors and panics upon the first registration that causes an
+// error.
+//
+// For convenience, this method returns a pointer to the Pusher itself.
+func (p *Pusher) MustCollector(c ...prometheus.Collector) *Pusher {
+	p.registerer.MustRegister(c...)
```
kakkoyun, line 177: I don't think adding an API that panics to the pusher is the right solution here.
If I'm not mistaken, I understand from you comment that you want to have an API to add multiple collectors. Would a simple for loop not satisfy your needs?

```
p := push.New("http://example.org/metrics", "my_job").
for _, c := range []prometheus.Collector{myCollector1, myCollector2} {
    p.Collector(myCollector1)
} 
p.Grouping("zone", "xy").Client(&myHTTPClient).BasicAuth("top", "secret").Add()
```

If somehow you don't want to use this, I might consider making `Collector` method variadic.
```
func (p *Pusher) Collector(collectors ...prometheus.Collector) *Pusher {
      if p.error != nil {
            return p
      }
       for _, c := range collectors {
            p.error = p.registerer.Register(c)		
            if p.error != nil {
                  return p
            }
	}
	return p
}
```

oiooj: Is myCollector1 registered successfully? We want to know the result when the collectors is being registered. Maybe that is why `Registerer` has the `MustRegister` API.

kakkoyun: I see what you mean. We can consider adding a method the obtain the error from outside. I just don't want to add a method that panics in the pusher. None of the other APIs in the package do it. I'd prefer to keep the consistent behaviour.

oiooj:
New  API #1075 merged, this PR closed.

PR #1079, opened 2022-07-06
Merge release-1.12 branch back into main

beorn7:
So that all the commits in the release branch are also in main.

(That's what we do in prometheus/prometheus. I got very confused when I noticed that the commit released as 1.12.2 is not in the main branch. It is unfortunate that so many commits get duplicated without a change in the code, but that's what you get when you cherrypick/rebase.)


codesome (approved):

kakkoyun (approved):

beorn7:
@kakkoyun you didn't properly merge this, you did "squash and merge". Which created new commits all over again.

I'll do it again, creating a new PR, but this time you really have to merge.

kakkoyun:
> @kakkoyun you didn't properly merge this, you did "squash and merge". Which created new commits all over again.
> 
> I'll do it again, creating a new PR, but this time you really have to merge.

@beorn7  Sorry for that. @bwplotka and I decided to only have the squash merge. And it's the only option now.
And I don't have the necessary access to it. Could you please open that first?

We can also revert the previous merge to keep things clean.

Again sorry for the inconvenience.

beorn7:
I think this deserves a wider discussion within the community (I think we should try to be consistent in this regard between repos, but even if the community as a whole doesn't agree, at least the decision that we don't need consistency should be a collective one ;). Thus, this is a perfect topic for the dev summit. I put a suggestion into the doc.

bwplotka:
Sorry that this comes with a surprise and that we break your existing workflows. 

We just had some debate in DM with @beorn7, which touched on existential reasons and misunderstanding of Git, conventional semantics of commit and so on. While I still think we should squash everything for simplicity and buildable commits, squash-only logic needs redesign in our release process, which we did NOT do. For this reason, I would propose reenabling that option back it has proven to have a significant negative effect on [Go modules](https://github.com/prometheus/client_golang/issues/1083).

PR #1081, opened 2022-07-06
Merging release branch back to main again

beorn7:
Again, zero diff. Actually _almost_ zero, because by now we have the fixed VERSION file in it.

Please do a real merge this time. Otherwise, we will just have a new commit in main and not the commits that actually got released.


kakkoyun:
Please also see https://github.com/prometheus/client_golang/pull/1079#issuecomment-1176332301

beorn7:
As discussed over in #1079, closing this.

Issue #1083, opened 2022-07-07
Tag and commit for v1.12.2 does not belong to any branch


beorn7:
Sorry for touching the topic again, but I think this is actually a problem:

The v1.12.2 tag points to commit e203144f43306c1f344fbc548fd02c4b79962e30 , but this commit is actually not in any branch in this repo. If you go to the commit in the GH web UI, you get:
![image](https://user-images.githubusercontent.com/5609886/177802843-4cc6247a-1c76-4352-b846-ac349fe9fa89.png)

The problem is that you will usually not fetch this commit if you clone the repo. When Go Modules fetches the version, it apparently fetches the commit directly without paying attention to branches. I tried to merge v1.12.2 into the `sparsehistogram` branch so that Go Modules would see it as an upgrade to v1.12.2, not a downgrade (which causes all kind of Go Modules pain when trying to run anything with the new sparse histograms). It took me a while to understand that I still didn't have the right commit in my branch despite merging in the `release-1.12` branch.

In the end, I fetched that commit directly into a pseudo remote branch:
```
$ git fetch origin e203144f43306c1f344fbc548fd02c4b79962e30:refs/remotes/origin/foobar
```

I could then merge that branch into the `sparsehistogram` branch, so that Go Modules works again with the `sparsehistogram` branch. Ironically, a byproduct of this is that now commit e203144f43306c1f344fbc548fd02c4b79962e30 actually belongs to a branch again, but not the release branch (which might be confusing to users in yet another way).

My guess is that v1.12.2 was tagged in a review branch before it got merged into the release branch via squashing (thereby rewriting the commit with a different sha). Then, the review branch got deleted, and the tagged commit was lost without branch.

Not sure how to fix this in the best way. Maybe merge (this time for real) e203144f43306c1f344fbc548fd02c4b79962e30 into branch `release-1.12` ?

bwplotka:
As per https://github.com/prometheus/client_golang/pull/1079#issuecomment-1177872946 agree - sorry for the confusion.

beorn7:
The tagged commit is at least in the release-1.12 branch now.

@kakkoyun @bwplotka I leave it to you if you want to merge that commit into main, too. Some users will expect that (including me), but an argument could be made to not do it.

PR #1084, opened 2022-07-07
Merge commit tagged as v1.12.2 into the release branch

beorn7:
This also removes leftovers from merge conflicts that somehow made it into the release branch (but were never released, because the released commit wasn't in the release branch in the first place ;).


beorn7:
Fixes #1083

kakkoyun:
> Not sure where the lint errors are coming from. Besides the updated `VERSION`, there is no difference between this and the released code.

I think it's because we have enabled additional linters after these commits https://github.com/prometheus/client_golang/pull/1056

kakkoyun (approved):

PR #1088, opened 2022-07-14
Adding +inf bucket exemplar

arun-shopify:
- explicitly add +inf bucket, if there is an exemplar for it.
- If there are multiple exemplars outside of the max bucket bound, choose one for the +inf bucket and end the loop.

### DIFF
--- a/prometheus/metric.go
+++ b/prometheus/metric.go
@@ -186,12 +186,14 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
 			if i < len(pb.Histogram.Bucket) {
 				pb.Histogram.Bucket[i].Exemplar = e
 			} else {
-				b := &dto.Bucket{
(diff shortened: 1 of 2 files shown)


kakkoyun:
Hey @arun-shopify, it would be nice to open an issue to describe your problem and need before sending the PR. Maybe we could have discussed the details. Could you please explain your motivation in the PR description?

arun-shopify:
@kakkoyun hey sorry, this was supposed to be a PR to our fork (before making a PR here), I opened it in the main repo by mistake. But here is the actual problem I am trying to address: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9945 TLDR; Prometheus exporter in Open telemetry collector panics when it finds an exemplar that is outside of the bucket bounds. Closing this PR for now.

Issue #977, opened 2022-01-24
Client API: don't read entire response into a buffer before parsing it


bboreham:
The API client functions all follow this pattern, where they make an http call to get back a `[]byte` then decode the JSON:
https://github.com/prometheus/client_golang/blob/f63e219e6b9074f8a55c8475e7b11720bdfc3737/api/prometheus/v1/api.go#L830-L836

For larger responses, this buffer gets quite expensive (see #976).

I propose that instead we parse JSON from the response body as it comes in.
I can see that this would make handling timeouts more complicated.

kakkoyun:
Thanks for bringing it up. I'm happy to accept any changes for the API part. It is still considered experimental and we have wiggle room to experiment.
@bboreham Do you need a hand on this or do you plan to take care of it?

bboreham:
Right now I don't expect to start on this.  It's quite a big job and not critical to what I'm doing.

There was some discussion in #763 of code-generating the API, which seems an attractive way to make such structural changes.

importhuman:
Hey @kakkoyun, can I take this up?

This is basically changing the methods to return parsed data instead of `[]byte`, or something else?

bboreham:
The current code calls `h.client.DoGetFallback()` which returns a `[]byte`, via `bytes.Buffer.ReadFrom()`.
(This call is made in multiple places, for each different API method).

My idea was to parse JSON from `resp.Body`, and not create the buffer.

importhuman:
The `[]byte` is being returned by a `Do` method [in client.go](https://github.com/prometheus/client_golang/blob/f63e219e6b9074f8a55c8475e7b11720bdfc3737/api/client.go#L96). Did you mean changing this method to just return the response and not return `[]byte` (and consequently wherever changes are required), or did I go too deep?

Full flow through which I reached here: [DoGetFallback](https://github.com/prometheus/client_golang/blob/f63e219e6b9074f8a55c8475e7b11720bdfc3737/api/prometheus/v1/api.go#L1142) -> [apiClientImpl.Do](https://github.com/prometheus/client_golang/blob/f63e219e6b9074f8a55c8475e7b11720bdfc3737/api/prometheus/v1/api.go#L1087) -> [httpClient.Do](https://github.com/prometheus/client_golang/blob/f63e219e6b9074f8a55c8475e7b11720bdfc3737/api/client.go#L96)

kakkoyun:
[...]

PR #1090, opened 2022-07-18
Parse HTTP response directly

importhuman:
Was something like this required for #977? (I'm aware tests are failing, will look into it more)

Got the following benchmark results (used the existing benchmark test in `client_test.go`, not sure why delta is nil)

```
name              old time/op    new time/op    delta
Client/4KB-12       62.7µs ± 0%   137.4µs ± 0%   ~     (p=1.000 n=1+1)
Client/50KB-12       119µs ± 0%     132µs ± 0%   ~     (p=1.000 n=1+1)
Client/1000KB-12    1.44ms ± 0%    0.13ms ± 0%   ~     (p=1.000 n=1+1)
Client/2000KB-12    2.68ms ± 0%    0.14ms ± 0%   ~     (p=1.000 n=1+1)

name              old alloc/op   new alloc/op   delta
Client/4KB-12       20.5kB ± 0%    18.2kB ± 0%   ~     (p=1.000 n=1+1)
Client/50KB-12       136kB ± 0%      18kB ± 0%   ~     (p=1.000 n=1+1)
Client/1000KB-12    2.11MB ± 0%    0.02MB ± 0%   ~     (p=1.000 n=1+1)
Client/2000KB-12    4.21MB ± 0%    0.02MB ± 0%   ~     (p=1.000 n=1+1)

name              old allocs/op  new allocs/op  delta
Client/4KB-12         74.0 ± 0%     128.0 ± 0%   ~     (p=1.000 n=1+1)
Client/50KB-12         101 ± 0%       137 ± 0%   ~     (p=1.000 n=1+1)
Client/1000KB-12       595 ± 0%       135 ± 0%   ~     (p=1.000 n=1+1)
Client/2000KB-12     1.10k ± 0%     0.14k ± 0%   ~     (p=1.000 n=1+1)
```

Thank you!


importhuman:
cc @bboreham @kakkoyun can you please review if this is along correct lines?

bboreham:
No, I don't expect the code to get much shorter when implementing this.
In particular code like this which you removed: https://github.com/prometheus/client_golang/pull/1090/files#diff-aa9bfd1a638fbb706f8e8920297902937011160319d9679add5dca56e5ab8277L145-L153
is important to handle timeouts, and will need some functional replacement.

Possibly calling `resp.Body.Close()` will still work, but I think you will need a goroutine watching for context cancellation in order to do this.

importhuman:
@bboreham will this work? I didn't use a `select` statement since this flow doesn't use a `done` channel. Apologies for the late update.

api/client.go
```diff
@@ -130,27 +129,16 @@ func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response,
...
-	case <-ctx.Done():
-		<-done
+		<-ctx.Done()
 		err = resp.Body.Close()
 		if err == nil {
 			err = ctx.Err()
 		}
-	case <-done:
-	}
+	}()
```
bboreham, line 141: Nothing is now waiting for the goroutine to finish, so assigning the error is pointless.
(This may not be important, but stylistically the code should not do pointless things)

bboreham (commented):
> will this work?

I don't know. It is closer to what I thought might work than last time.

According to CI it doesn't compile, but let's assume you are working on that.

Eventually, you need to convince reviewers (not just me) that it works. E.g. perform experiments and/or add unit tests to demonstrate that it works.

importhuman:
Benchmark results:

`main` branch:
```
goos: linux
goarch: amd64
pkg: github.com/prometheus/client_golang/api
cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
BenchmarkClient/4KB-12         	   19268	     55355 ns/op	   20532 B/op	      74 allocs/op
BenchmarkClient/50KB-12        	   10000	    100465 ns/op	  135893 B/op	     101 allocs/op
BenchmarkClient/1000KB-12      	    1003	   1201450 ns/op	 2112968 B/op	     595 allocs/op
BenchmarkClient/2000KB-12      	     542	   2535752 ns/op	 4214170 B/op	    1097 allocs/op
PASS
ok  	github.com/prometheus/client_golang/api	5.644s
```

This branch:
```
goos: linux
goarch: amd64
pkg: github.com/prometheus/client_golang/api
cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
BenchmarkClient/4KB-12         	    8221	    137427 ns/op	   18519 B/op	     131 allocs/op
BenchmarkClient/50KB-12        	    8016	    136149 ns/op	   18442 B/op	     139 allocs/op
BenchmarkClient/1000KB-12      	    7760	    136296 ns/op	   18307 B/op	     138 allocs/op
BenchmarkClient/2000KB-12      	    8731	    137935 ns/op	   18318 B/op	     138 allocs/op
PASS
ok  	github.com/prometheus/client_golang/api	5.793s
```

Benchstat:
```
name              old time/op    new time/op    delta
Client/4KB-12       55.4µs ± 0%   137.4µs ± 0%   ~     (p=1.000 n=1+1)
Client/50KB-12       100µs ± 0%     136µs ± 0%   ~     (p=1.000 n=1+1)
Client/1000KB-12    1.20ms ± 0%    0.14ms ± 0%   ~     (p=1.000 n=1+1)
Client/2000KB-12    2.54ms ± 0%    0.14ms ± 0%   ~     (p=1.000 n=1+1)

name              old alloc/op   new alloc/op   delta
Client/4KB-12       20.5kB ± 0%    18.5kB ± 0%   ~     (p=1.000 n=1+1)
Client/50KB-12       136kB ± 0%      18kB ± 0%   ~     (p=1.000 n=1+1)
Client/1000KB-12    2.11MB ± 0%    0.02MB ± 0%   ~     (p=1.000 n=1+1)
Client/2000KB-12    4.21MB ± 0%    0.02MB ± 0%   ~     (p=1.000 n=1+1)

name              old allocs/op  new allocs/op  delta
Client/4KB-12         74.0 ± 0%     131.0 ± 0%   ~     (p=1.000 n=1+1)
Client/50KB-12         101 ± 0%       139 ± 0%   ~     (p=1.000 n=1+1)
Client/1000KB-12       595 ± 0%       138 ± 0%   ~     (p=1.000 n=1+1)
Client/2000KB-12     1.10k ± 0%     0.14k ± 0%   ~     (p=1.000 n=1+1)
```

Slightly worse for 4 and 50 KB, much better for 1000 and 2000 KB.

---
Failing tests:

Currently, they are failing due to a different error message. I have slightly changed the flow of the code to be in line what is currently in `main` branch, reading the response body after checking the status code. 
Are we allowed to change the expected error message? My concern is doing it would impact users who have structured their code with the current expected error messages.

api/client.go
importhuman: 1. Do we need to do `resp.Body.Close()` now, as we are no longer creating a buffer?
2. The error is being returned in the end, should we still not assign it?

importhuman:
cc @bboreham @kakkoyun

bboreham (commented):
> Currently, they are failing due to a different error message.

It's much more than that.  Some are failing because they "expected error, but got none".
As this PR stands, each test case is initializing a `response` member which is never used.
The point of `TestAPIClientDo` was to stub out the lower layer http call and check what the higher layer did with it. This may need re-thinking now that the http call works differently.

> Are we allowed to change the expected error message? 

If it preserves the basic semantics, sure.  For instance, it must still differentiate client errors from server errors (400 from 500 series).

Separately, let me again empahasise that the point of the buffer and goroutine in `httpClient.Do()` was to allow requests to be cancelled. You must show that this works in your new implementation.
For example, write a unit test which cancels a request.

api/client_test.go
```diff
@@ -152,7 +152,7 @@ func BenchmarkClient(b *testing.B) {
 			}
 			b.ResetTimer()
 			for i := 0; i < b.N; i++ {
-				_, _, err := client.Do(ctx, req)
+				_, err := client.Do(ctx, req)
```
bboreham, line 155: This is now starting the call but never reading the response. It's no longer benchmarking anything interesting.

`BenchmarkClient` was added to illustrate the difference between ways of filling the buffer; To illustrate the difference between buffer and no buffer you have to go to to a higher level and benchmark reading an entire Prometheus API response, including JSON parsing.

bboreham:
I've just noticed that the code calls the JSON parser twice to read each request.  Once in `apiClientImpl.Do:
https://github.com/prometheus/client_golang/blob/618194de6ad3db637313666104533639011b470d/api/prometheus/v1/api.go#L1140

and once in the individual API methods like `Query`:
https://github.com/prometheus/client_golang/blob/618194de6ad3db637313666104533639011b470d/api/prometheus/v1/api.go#L859

Then, different data types make nested calls:
https://github.com/prometheus/client_golang/blob/618194de6ad3db637313666104533639011b470d/api/prometheus/v1/api.go#L584

I suspect the first two could be merged without major change, but the last kind needs the code rewritten in "streaming" style to avoid building an intermediate buffer.

importhuman:
Thanks for the detailed feedback! I don't have the answers currently, will update as I look into these issues more.

PR #1091, opened 2022-07-19
Raise exemplar labels limit from 64 to 128

bboreham:
In line with the OpenMetrics spec:
> The combined length of the label names and values of an Exemplar's LabelSet MUST NOT exceed 128 UTF-8 character code points

https://github.com/OpenObservability/OpenMetrics/blob/98ae26c87b/specification/OpenMetrics.md#exemplars

I suspect this number changed since the discussion in #706, but I can't see that change in the git history of the spec.

### DIFF
--- a/prometheus/counter_test.go
+++ b/prometheus/counter_test.go
@@ -16,6 +16,7 @@ package prometheus
 import (
 	"fmt"
 	"math"
+	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


kakkoyun:
@beorn7 Do you have any comments on this?

beorn7:
This looks fine.

PR #1094, opened 2022-07-21
explicitly add +inf bucket in withExemplarsMetric

arun-shopify:
### Problem:
Currently if there are exemplar values that are outside the maximum bucket bounds for a histogram metric, client_go lang will panic https://github.com/prometheus/client_golang/blob/main/prometheus/metric.go#L186-L189

When using this client in [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector), this panic leads to a crash and therefore we are unable to use client_golang in [prometheus exporter](https://github.com/Shopify/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter) in Open Telemetry Collector. There is a discussion about this issue in our [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/9945) in that repository.

### Proposed solution:

In this PR we are proposing a solution to this issue by explicitly adding `+inf` bucket when there are exemplar values outside the max bucket bound and picking one of the exemplars if there are more than one in that range. 

@bwplotka @kakkoyun

### DIFF
--- a/prometheus/metric.go
+++ b/prometheus/metric.go
@@ -184,7 +184,8 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
 			})
 			if i < len(pb.Histogram.Bucket) {
 				pb.Histogram.Bucket[i].Exemplar = e
-			} else { // +inf bucket should be explicitly added if there is an exemplar for it.
+			} else { 
+			    // The +Inf bucket should be explicitly added if there is an exemplar for it, similar to non-const histogram logic in https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L357-L365.
 				b := &dto.Bucket{
 					CumulativeCount: proto.Uint64(pb.Histogram.Bucket[len(pb.Histogram.GetBucket())-1].GetCumulativeCount()),
 					UpperBound:      proto.Float64(math.Inf(1)),
--- a/prometheus/metric.go
+++ b/prometheus/metric.go
@@ -15,6 +15,7 @@ package prometheus
 
 import (
 	"errors"
+	"math"
 	"sort"
 	"strings"
 	"time"
@@ -183,9 +184,16 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
 			})
 			if i < len(pb.Histogram.Bucket) {
 				pb.Histogram.Bucket[i].Exemplar = e
-			} else {
-				// This is not possible as last bucket is Inf.
-				panic("no bucket was found for given exemplar value")
+			} else { // +inf bucket should be explicitly added if there is an exemplar for it.
+				b := &dto.Bucket{
+					CumulativeCount: proto.Uint64(pb.Histogram.Bucket[len(pb.Histogram.GetBucket())-1].GetCumulativeCount()),
+					UpperBound:      proto.Float64(math.Inf(1)),
+					Exemplar:        e,
+				}
+				pb.Histogram.Bucket = append(pb.Histogram.Bucket, b)
+				break
+				// end looping after creating +inf bucket and adding one exemplar.
+				// there could be other exemplars that are in the "inf" range but those will be ignored.
 			}
 		}
 	default:
--- a/prometheus/metric.go
+++ b/prometheus/metric.go
@@ -184,8 +184,8 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
 			})
 			if i < len(pb.Histogram.Bucket) {
 				pb.Histogram.Bucket[i].Exemplar = e
-			} else { 
-			    // The +Inf bucket should be explicitly added if there is an exemplar for it, similar to non-const histogram logic in https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L357-L365.
+			} else {
+				// The +Inf bucket should be explicitly added if there is an exemplar for it, similar to non-const histogram logic in https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L357-L365.
 				b := &dto.Bucket{
 					CumulativeCount: proto.Uint64(pb.Histogram.Bucket[len(pb.Histogram.GetBucket())-1].GetCumulativeCount()),
 					UpperBound:      proto.Float64(math.Inf(1)),
@@ -193,8 +193,7 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
 				}
 				pb.Histogram.Bucket = append(pb.Histogram.Bucket, b)
 				break
-				// end looping after creating +inf bucket and adding one exemplar.
-				// there could be other exemplars that are in the "inf" range but those will be ignored.
+				// Terminating the loop after creating the +Inf bucket and adding one exemplar, if there are other exemplars in the +Inf bucket range they will be ignored.
 			}
 		}
 	default:
--- a/prometheus/metric_test.go
+++ b/prometheus/metric_test.go
@@ -68,7 +68,7 @@ func TestWithExemplarsMetric(t *testing.T) {
 			t.Errorf("want %v, got %v", want, got)
 		}
 
-		// when there are more exemplars than there are buckets, a +inf bucket will be created and the last exemplar value will be added to the +inf bucket.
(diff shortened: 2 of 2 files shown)


arun-shopify:
@bwplotka @kakkoyun bumping this PR for a review :)

prometheus/metric_test.go
```diff
@@ -56,16 +57,19 @@ func TestWithExemplarsMetric(t *testing.T) {
...
 		if err := m.Write(&metric); err != nil {
 			t.Fatal(err)
 		}
-		if want, got := 4, len(metric.GetHistogram().Bucket); want != got {
+		if want, got := 5, len(metric.GetHistogram().Bucket); want != got {
 			t.Errorf("want %v, got %v", want, got)
 		}
 
-		expectedExemplarVals := []float64{24.0, 42.0, 100.0, 157.0}
+		// when there are more exemplars than there are buckets, a +inf bucket will be created and the last exemplar value will be added to the +inf bucket.
```
bwplotka, line 71: Do you mind making comments a full sentence for readability?

```suggestion
		// When there are more exemplars than there are buckets, a +Inf bucket will be created and the last exemplar value will be added.
```

prometheus/metric_test.go
```diff
@@ -74,5 +78,17 @@ func TestWithExemplarsMetric(t *testing.T) {
...
 			}
 		}
+
+		infBucket := metric.GetHistogram().Bucket[len(metric.GetHistogram().Bucket)-1].GetUpperBound()
+
+		if infBucket != math.Inf(1) {
+			t.Errorf("want %v, got %v", math.Inf(1), infBucket)
+		}
+
+		infBucketValue := metric.GetHistogram().Bucket[len(metric.GetHistogram().Bucket)-1].GetExemplar().GetValue()
```
bwplotka, line 88: This was already checked above, so we can remove.

prometheus/metric_test.go
```diff
@@ -74,5 +78,17 @@ func TestWithExemplarsMetric(t *testing.T) {
 				t.Errorf("%v: want %v, got %v", i, want, got)
 			}
 		}
+
+		infBucket := metric.GetHistogram().Bucket[len(metric.GetHistogram().Bucket)-1].GetUpperBound()
```
bwplotka, line 82: 👍🏽

prometheus/metric.go
```diff
@@ -183,9 +184,16 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
 			})
 			if i < len(pb.Histogram.Bucket) {
 				pb.Histogram.Bucket[i].Exemplar = e
-			} else {
-				// This is not possible as last bucket is Inf.
-				panic("no bucket was found for given exemplar value")
+			} else { // +inf bucket should be explicitly added if there is an exemplar for it.
```
bwplotka, line 187: Can we make it a full sentence?
```suggestion
			} else { 
			    // The +Inf bucket should be explicitly added if there is an exemplar for it, similar to non-const histogram logic in https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L357-L365.
```

prometheus/metric.go
```diff
@@ -183,9 +184,16 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
...
+				b := &dto.Bucket{
+					CumulativeCount: proto.Uint64(pb.Histogram.Bucket[len(pb.Histogram.GetBucket())-1].GetCumulativeCount()),
+					UpperBound:      proto.Float64(math.Inf(1)),
+					Exemplar:        e,
+				}
+				pb.Histogram.Bucket = append(pb.Histogram.Bucket, b)
+				break
+				// end looping after creating +inf bucket and adding one exemplar.
+				// there could be other exemplars that are in the "inf" range but those will be ignored.
```
bwplotka, line 196: If we want those comments, let's make them full sentence, but also I am not sure this is safe. Nowhere in the interface/signature we mention that exemplars will be sorted by anything. I think there is no harm to continue the loop, unless we want to optimize this some day. WDYT?

```suggestion
```

bwplotka (approved):
Thanks for this! Some comments, but generally I love this idea - thanks for troubleshooting this 💪🏽

prometheus/metric_test.go
arun-shopify: removed

prometheus/metric.go
arun-shopify: Yes you are right that there is no mention (or expectation) of sorting exemplars. If we leave the loop running we pick the last exemplar and if we leave it as is and terminate it here, we pick the first exemplar (in that range) - sort order of the exemplar value being arbitrary.  I would consider terminating the loop as it would just avoid running through the remaining exemplars that we are not going to use anyway. But please let me know if you think otherwise or see any safety concerns, I can make the change. 

Also, I made the comment into a sentence.

bwplotka: This was not addressed completely. What do you about lack of sorted order invariant?

> Nowhere in the interface/signature we mention that exemplars will be sorted by anything. I think there is no harm to continue the loop, unless we want to optimize this some day. WDYT?

prometheus/metric.go
```diff
@@ -184,8 +185,15 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
...
-				// This is not possible as last bucket is Inf.
-				panic("no bucket was found for given exemplar value")
+				// The +Inf bucket should be explicitly added if there is an exemplar for it, similar to non-const histogram logic in https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L357-L365.
+				b := &dto.Bucket{
+					CumulativeCount: proto.Uint64(pb.Histogram.Bucket[len(pb.Histogram.GetBucket())-1].GetCumulativeCount()),
+					UpperBound:      proto.Float64(math.Inf(1)),
+					Exemplar:        e,
+				}
+				pb.Histogram.Bucket = append(pb.Histogram.Bucket, b)
+				break
```
bwplotka, line 195: Thanks for addressing this, but I think one thing is still not addressed: https://github.com/prometheus/client_golang/pull/1094#discussion_r931227718

prometheus/metric.go
arun-shopify: For all the buckets we are comparing the exemplar values to the bucket bound and matching the right exemplar to right bucket (this is the exiting logic): 

specifically the check here to get the index of the right bucket for that exemplar: 

https://github.com/prometheus/client_golang/blob/a528affed905f64ed33ed2f52b7880ed7247f99e/prometheus/metric.go#L182

And assigning the exemplar to the right bucket here:

https://github.com/prometheus/client_golang/blob/a528affed905f64ed33ed2f52b7880ed7247f99e/prometheus/metric.go#L185
 

At a high level the only change in this PR is that instead of a panic in the [else](https://github.com/prometheus/client_golang/blob/a528affed905f64ed33ed2f52b7880ed7247f99e/prometheus/metric.go#L186) condition, we add the `+Inf` bucket and add one exemplar that is outside of all previous bucket range and break the loop.

If there are multiple exemplars for the `+Inf` bucket, we could pick the exemplar that is more representative of the group such as a median - a future improvement, would require further discussion. Currently we are just picking the first in the array in the +inf bucket range.

I hope that clarifies your concern, if I understood your question correctly.

arun-shopify: @bwplotka waiting for your response :) Would be nice if we could include this in the upcoming release.

prometheus/metric.go
bwplotka: will fix that for you if you don't mind - I am preparing next release

bwplotka: merging and will fix in separate PR.

arun-shopify: sure, also response in https://github.com/prometheus/client_golang/pull/1094#discussion_r931227718

prometheus/metric.go
bwplotka: Missed this discussion, sorry. 

So I like the idea about median or tuning the exemplar we take from those belonging to +Inf. My only problem is that those inputs can be not sorted, that's it. Hope this PR https://github.com/prometheus/client_golang/pull/1100 makes sense to you.

PR #1095, opened 2022-07-27
Update documentation for exemplar label limit

fredr:
The limit where raised from 64 to 128 runes in #1091, this reflect the change in the documentation comments.


bwplotka (approved):
Thanks!

PR #1096, opened 2022-07-29
Remove ioutil

inosato:
`ioutil` is deprecated from go1.16 (ref: https://go.dev/doc/go1.16#ioutil).
So removed them.

Signed-off-by: inosato <si17_21@yahoo.co.jp>

### DIFF
--- a/api/prometheus/v1/api_test.go
+++ b/api/prometheus/v1/api_test.go
@@ -17,7 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


bwplotka:
Lol. I did not know that, well spotted! (still was using `ioutil` everywhere).

Perhaps we could set up faillint to avoid that problem in future https://github.com/fatih/faillint

bwplotka (approved):
LGTM, thanks

Issue #945, opened 2021-12-09
Builder-like pattern to simplify creation of metrics containers 


rfratto:
## Background

It is [common](https://github.com/prometheus/prometheus/blob/c965a7555b7ffcee1a127d782abd5bb478a16750/tsdb/db.go#L198-L308) for Prometheus client code to create a container which holds metrics pertaining to a specific subsystem. Custom containers are useful for avoiding pollution of the global registry and allow for tracking independent sets of metrics when multiple instances of a subsystem exist. 

Today, I find that creating metrics containers is largely repetitive. A metrics container must always both create the inner metrics and use some mechanism to [expose them all](https://github.com/prometheus/client_golang/blob/6d5cf25fcc344ed30a0623e36c73639bfac112b4/prometheus/collectors/dbstats_collector.go#L94-L119).  

Ideally, metrics containers expose their inner metrics through the prometheus.Collector interface. However, I find that this is not how most people implement these containers. Instead, developers are more likely to avoid implementing the interface by registering all the metrics at [creation time](https://github.com/prometheus/prometheus/blob/c965a7555b7ffcee1a127d782abd5bb478a16750/tsdb/db.go#L290-L306). I view this as an anti-pattern, as the caller potentially loses the ability to unregister those metrics. This has propagated to projects like Grafana Agent, which needed to hackily implement [some way](https://github.com/grafana/agent/blob/a9bc23c74f3727399e4f720abbbe9a3b5fc8b973/pkg/util/unregisterer.go#L53-L63) of unregistering metrics from packages that do this.  

## Proposal 

I propose a new type, `prometheus.Container`, to reduce the amount of boilerplate required to create these metric containers. By making it easier to have a metrics container that implements prometheus.Collector, I hope it will also dissuade people from registering metrics from the container's constructor.  

The type will track a set of inner Collectors, and expose a similar API to the promauto package:

```go
type Container struct {
  cs []Collector 
}

func (c *Container) NewCounter(opts CounterOpts) Counter {
  m := NewCounter(opts) 
  c.cs = append(c.cs, m)
  return m 
}

// Omitted for brevity: NewCounterVec, NewGauge, etc..

func (c *Container) Describe(ch <-chan *Desc) {
  for _, m := range c.cs {
    m.Describe(ch)
  }
}

func (c *Container) Collect(ch <-chan Metric) {
  for _, m := range c.cs {
    m.Collect(ch)
  }
}
```

## Example Usage 

```go
[...]

PR #1103, opened 2022-08-03
Extend prometheus.Registry to implement Collector

rfratto:
This change allows Registries to be used as Collectors.

This enables new instances of Registry to be passed to ephemeral subroutines for collecting metrics from subroutines which are still running:

```go
package main

import (
  "fmt"

)

func main() {
  globalReg := prometheus.NewRegistry()

  for i := 0; i < 100; i++ {
    workerReg := prometheus.WrapRegistererWith(prometheus.Labels{
      // Add an ID label so registered metrics from workers don't
      // collide.
      "worker_id": fmt.Sprintf("%d", i),
    }, prometheus.NewRegistry())

    globalReg.MustRegister(workerReg)

    go func(i int) {
      runWorker(workerReg)

      // Unregister any metrics the worker may have created.
      globalReg.Unregister(workerReg)
    }(i)
  }
}

// runWorker runs a worker, registering worker-specific metrics.
func runWorker(reg *prometheus.Registry) {
  // ... register metrics ...
  // ... do work ...
}
```

This change makes it easier to avoid leaking metrics from subroutines which do not consistently properly unregister metrics.

This would primarily make things easier for Grafana Agent, where it can spawn different subroutines throughout the process lifetime which have different sets of metrics (i.e., remote_write). Ensuring that metrics from these subroutines consistently get unregistered can be a source of bugs, and something like this would make it easier for us to avoid leaking no-longer-needed metrics.

I'm not completely sure this change works as-is (it seems reasonable), but I wanted to open it as a PR to spawn discussion about the concept.

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -255,6 +255,9 @@ func (errs MultiError) MaybeUnwrap() error {
 // them into MetricFamilies for exposition. It implements Registerer, Gatherer,
 // and Collector. The zero value is not usable. Create instances with
 // NewRegistry or NewPedanticRegistry.
+//
+// Registry implements Collector to allow it to be used for creating groups of
+// metrics. See the Grouping example for how this can be done.
 type Registry struct {
 	mtx                   sync.RWMutex
 	collectorsByID        map[uint64]Collector // ID is a hash of the descIDs.
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -252,8 +252,8 @@ func (errs MultiError) MaybeUnwrap() error {
 }
 
 // Registry registers Prometheus collectors, collects their metrics, and gathers
-// them into MetricFamilies for exposition. It implements both Registerer,
-// Gatherer, and Collector. The zero value is not usable. Create instances with
+// them into MetricFamilies for exposition. It implements Registerer, Gatherer,
+// and Collector. The zero value is not usable. Create instances with
 // NewRegistry or NewPedanticRegistry.
 type Registry struct {
 	mtx                   sync.RWMutex
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -252,9 +252,9 @@ func (errs MultiError) MaybeUnwrap() error {
 }
 
 // Registry registers Prometheus collectors, collects their metrics, and gathers
-// them into MetricFamilies for exposition. It implements both Registerer and
-// Gatherer. The zero value is not usable. Create instances with NewRegistry or
-// NewPedanticRegistry.
+// them into MetricFamilies for exposition. It implements both Registerer,
+// Gatherer, and Collector. The zero value is not usable. Create instances with
+// NewRegistry or NewPedanticRegistry.
 type Registry struct {
 	mtx                   sync.RWMutex
 	collectorsByID        map[uint64]Collector // ID is a hash of the descIDs.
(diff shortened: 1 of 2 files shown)


rfratto:
I'm not sure this change works as-is (I'm double checking), but I wanted to open it as a PR to spawn discussion about the concept. 

This would make things easier for us in Grafana Agent, where we have many goroutines (i.e., spawning a Prometheus remote_Write) which don't always last throughout the process lifetime, and we have a need to make it easier to not leak any metrics that get created.

rfratto:
This would also make #945 no longer needed.

kakkoyun (commented):
I really like the pattern. I can see the places where I'd use this pattern.
I want to think about it for a second and check if there's any downside (possible abuse?) to complying with `Collector` interface :thinking:

bwplotka:
Hm, so essentially you want to be able to group collectors better and collect, register and unregister them in one go. I think it's pretty nice characteristics, although at first it the `Registry` might feel like a separate functionality. The only downside I can see is that we put too many functionalities into one struct which can cause a bit of confusion and many ways of doing same thing.

The alternative is to create dedicated structure `MultiRegistry` (or multiCollector) attempt I did in https://github.com/prometheus/client_golang/pull/929/files#diff-6f0796727ecbd7f1d226589f6cf66cbb90a2c30d1823ce35d7a480e2561358d6R966 or something like that... But given this code is so little and clean I would be happy with your proposal @rfratto 💪🏽

prometheus/registry.go
```diff
@@ -252,9 +252,9 @@ func (errs MultiError) MaybeUnwrap() error {
 }
 
 // Registry registers Prometheus collectors, collects their metrics, and gathers
-// them into MetricFamilies for exposition. It implements both Registerer and
-// Gatherer. The zero value is not usable. Create instances with NewRegistry or
-// NewPedanticRegistry.
+// them into MetricFamilies for exposition. It implements Registerer, Gatherer,
+// and Collector. The zero value is not usable. Create instances with
+// NewRegistry or NewPedanticRegistry.
```
bwplotka, line 257: Do you mind adding a paraph about implementing Collector, it's behavior and what it enables?

Ideally we could add short `ExampleRegistryGrouping`  example in test file too! 🤗

bwplotka (approved):
LGMT, I like this logic 👍🏽 

A bit more docs would be amazing if you have time (:

rfratto:
It's a good thing @bwplotka suggested writing an Example function, because I definitely missed something on my first pass at this :) 

The snippet I gave in the PR description doesn't work; `prometheus.WrapRegistererWith` returns an instance of the `Registerer` interface, which doesn't implement Collector. 

I'm not sure what we should do about this in that case; it's a lot more limiting when most of the API erases the Collector implementation from the Registry. 

There could be a new interface:

```go
type RegistererCollector interface {
  Registerer 
  Collector 
}
```

which methods like `WrapRegistererWith` can be updated to return, but I'm not sure if this would bloat the proposal too much.

bwplotka:
Hmmm what about casting to `Collector` interface in your code for now?

rfratto:
Added more docs with an example, PTAL :) 

Also, `prometheus.WrapRegistererWith` _would_ work, it's just not very pretty:

```go
package main

import (
  "fmt"

)

func main() {
  globalReg := prometheus.NewRegistry()

  for i := 0; i < 100; i++ {
    workerMetrics := prometheus.NewRegistry() 
    globalReg.MustRegister(workerMetrics)

    go func(i int) {
      // Wrap worker metrics with a unique label so workers don't 
      // collide. 
      runWorker(prometheus.WrapRegistererWith(prometheus.Labels{
        "worker_id": fmt.Sprintf("%d", i),
      }, workerMetrics)

      // Unregister any metrics the worker may have created.
      globalReg.Unregister(workerMetrics)
    }(i)
  }
}

// runWorker runs a worker, registering worker-specific metrics.
func runWorker(reg prometheus.Registerer) {
  // ... register metrics ...
  // ... do work ...
}
```

bwplotka:
I like it 👍🏽 Again, casting to different interface would also work just fine. Thanks!

PR #1118, opened 2022-08-18
Fix double-counting bug in promhttp.InstrumentRoundTripperCounter

### LINKED ISSUE #1117: bug: Metric is double-incremented in promhttp.InstrumentRoundTripperCounter
It looks like v0.13.0 introduced a bug in `promhttp.InstrumentRoundTripperCounter` (from #1055, it appears).

Instead of replacing the `Inc()` call with a wrapped call to `exemplarAdd(...)`, the `Inc()` call was left in:

https://github.com/prometheus/client_golang/pull/1055/files#diff-7871a83741d29b5f29e3c47d1c284f948c258a0c6fc18895f0c2d4204cc46054R76-R81

The impact is that counter metrics are now double-counted when instrumented with `promhttp.InstrumentRoundTripperCounter`.

The bug can be observed with this test:

```go
func TestBug(t *testing.T) {
	counter := prometheus.NewCounterVec(
		prometheus.CounterOpts{Name: "requests_total", Help: "A request counter"},
		[]string{"method", "code"},
	)

	reg.MustRegister(counter)

	rt := promhttp.InstrumentRoundTripperCounter(counter, http.DefaultTransport)
	c := &http.Client{Transport: rt}

	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(http.StatusOK)
	}))

	req, _ := http.NewRequestWithContext(context.Background(), http.MethodGet, srv.URL, nil)
	resp, _ := c.Do(req)
	defer resp.Body.Close()

	expected := `
		# HELP requests_total A request counter
		# TYPE requests_total counter
		requests_total{code="200",method="get"} 1
	`

	if err := testutil.GatherAndCompare(reg, strings.NewReader(expected),
		"requests_total",
	); err != nil {
		t.Fatal(err)
	}
}
```

The output is:

```
[...]
        Diff:
        --- metric output does not match expectation; want
        +++ got:
        @@ -2,3 +2,3 @@
         # TYPE requests_total counter
        -requests_total{code="200",method="get"} 1
        +requests_total{code="200",method="get"} 2
```

I'll try to issue a fix for this today!

hairyhenderson:
Fixes #1117

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/prometheus/promhttp/instrument_client.go
+++ b/prometheus/promhttp/instrument_client.go
@@ -78,7 +78,6 @@ func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.Rou
 				1,
 				rtOpts.getExemplarFn(r.Context()),
 			)
-			counter.With(labels(code, method, r.Method, resp.StatusCode, rtOpts.extraMethods...)).Inc()
 		}
 		return resp, err
 	}
--- a/prometheus/promhttp/instrument_client_test.go
+++ b/prometheus/promhttp/instrument_client_test.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
 
 	dto "github.com/prometheus/client_model/go"
 	"google.golang.org/protobuf/proto"
@@ -249,6 +250,19 @@ func TestClientMiddlewareAPI_WithRequestContext(t *testing.T) {
 		if len(mf.Metric) == 0 {
 			t.Errorf("metric family %s must not be empty", mf.GetName())
 		}
+	}
+
+	// make sure counters aren't double-incremented (see #1117)
+	expected := `
+		# HELP client_api_requests_total A counter for requests from the wrapped client.
+		# TYPE client_api_requests_total counter
+		client_api_requests_total{code="200",method="get"} 1
+	`
+
+	if err := testutil.GatherAndCompare(reg, strings.NewReader(expected),
+		"client_api_requests_total",
+	); err != nil {
+		t.Fatal(err)
 	}
 }
 


hairyhenderson:
cc @bwplotka @kakkoyun - LMK if you want anything more in this, or if there's a better fix!

prometheus/promhttp/instrument_client.go
```diff
@@ -78,7 +78,6 @@ func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.Rou
 				1,
 				rtOpts.getExemplarFn(r.Context()),
 			)
-			counter.With(labels(code, method, r.Method, resp.StatusCode, rtOpts.extraMethods...)).Inc()
```
pracucci, line 81: Thinking loudly: I think the function name `exemplarAdd()` doesn't make it obvious the metric is incremented. Maybe renaming it to `addWithExemplar()` may clarify it?

pracucci (approved):

prometheus/promhttp/instrument_client.go
hairyhenderson: Agreed! My goal here was to fix the bug with as few changes as possible so I didn't consider changing the function name... IMO that'd best be deferred to a separate PR.

kakkoyun (approved):
LGTM. Thanks for fixing it.

I agree with the naming. Let's fix it as well. I'll create an issue for it.

prometheus/promhttp/instrument_client.go
kakkoyun: https://github.com/prometheus/client_golang/issues/1120

dlahyani:
Thanks for the fix.

bboreham:
Hi, please can we get a release with this fix in.

prometheus/promhttp/instrument_client.go
bwplotka: 👍🏽 Thanks for noticing this

PR #1123, opened 2022-08-22
Assure we exhaust bodies of all HTTP responses.

bwplotka:
Otherwise HTTP library might not be able to reuse TCP connections.

Signed-off-by: bwplotka <bwplotka@gmail.com>

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -24,6 +24,8 @@ import (
 	"path"
 	"strings"
 	"time"
+
+	"github.com/prometheus/client_golang/internal/errcapture"
 )
 
 // DefaultRoundTripper is used if no RoundTripper is set in Config.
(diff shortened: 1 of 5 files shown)


api/client.go
```diff
@@ -118,39 +120,17 @@ func (c *httpClient) URL(ep string, args map[string]string) *url.URL {
...
-		err = resp.Body.Close()
-		if err == nil {
-			err = ctx.Err()
-		}
-	case <-done:
-	}
-
-	return resp, body, err
+	var buf bytes.Buffer
+	_, err = buf.ReadFrom(resp.Body)
```
wking, line 134: not having a way for `ctx` to cancel a slow read seems dangerous.

internal/errcapture/errcapture.go
```diff
@@ -0,0 +1,61 @@
...
+	"os"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type doFunc func() error
+
+// Do runs function and on error return error by argument including the given error (usually
+// from caller function).
+func Do(err *error, doer doFunc, format string, a ...interface{}) {
```
kakkoyun, line 30:
```suggestion
func do(err *error, doer doFunc, format string, a ...interface{}) {
```
No need to export.

api/client.go
kakkoyun: I think we can have a context-aware cancellable reader as an internal thing and utilize it here. The functionality was previously there. 

https://pace.dev/blog/2020/02/03/context-aware-ioreader-for-golang-by-mat-ryer.html
https://benjamincongdon.me/blog/2020/04/23/Cancelable-Reads-in-Go
https://github.com/dolmen-go/contextio

kakkoyun (commented):
Overall LGTM. Happy to merge.

We have only a couple of nits to handle.

api/client.go
bboreham: BTW (as far as I know) the only reason a `bytes.Buffer` was ever used in this function is to allow cancellation.

This is the subject of #977.

PR #1133, opened 2022-09-13
Fix float64 comparison test failure on archs using FMA

sbunce:
Background
=========
Architectures using FMA optimization yield slightly different results so we cannot assume floating point values will be precisely the same across different architectures.
https://github.com/prometheus/client_golang/pull/899#issuecomment-1244506390

Change
======
The solution in this change is to check "abs(a-b) < tolerance" instead of comparing the exact values. This will give us confidence that the histogram buckets are near identical.

Testing
=====
I have not tested this change on one of the previously failing architectures. Help from someone with access to one of these architectures would be appreciated.

### DIFF
--- a/prometheus/histogram_test.go
+++ b/prometheus/histogram_test.go
@@ -355,13 +355,12 @@ func TestBuckets(t *testing.T) {
 
 	got = ExponentialBucketsRange(1, 100, 10)
 	want = []float64{
-		1.0, 1.6681005372000588, 2.782559402207125,
-		4.641588833612779, 7.742636826811273, 12.915496650148842,
-		21.544346900318846, 35.93813663804629, 59.94842503189414,
-		100.00000000000007,
+		1.0, 1.6681, 2.7825, 4.6415, 7.7426, 12.9154, 21.5443,
+		35.9381, 59.9484, 100.0000,
 	}
-	if !reflect.DeepEqual(got, want) {
-		t.Errorf("exponential buckets range: got %v, want %v", got, want)
+	const tolerance = 0.0001
+	if !equalFloat64s(got, want, 0.0001) {
+		t.Errorf("exponential buckets range: got %v, want %v (tolerance %f)", got, want, tolerance)
 	}
 }
 
@@ -466,4 +465,24 @@ func TestHistogramExemplar(t *testing.T) {
 			t.Errorf("expected exemplar %s, got %s.", expected, got)
 		}
 	}
+}
+
+// equalFloat64 returns true if a and b are within the tolerance. We have this
+// because float comparison varies on different architectures. For example,
+// architectures which do FMA yield slightly different results.
+// https://github.com/prometheus/client_golang/pull/899#issuecomment-1244506390
+func equalFloat64(a, b, tolerance float64) bool {
+	return math.Abs(a-b) < tolerance
+}
+
+func equalFloat64s(a, b []float64, tolerance float64) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
+		if !equalFloat64(a[i], b[i], tolerance) {
+			return false
+		}
+	}
+	return true
 }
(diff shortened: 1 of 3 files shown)


prometheus/histogram_test.go
```diff
@@ -355,13 +355,12 @@ func TestBuckets(t *testing.T) {
...
-		4.641588833612779, 7.742636826811273, 12.915496650148842,
-		21.544346900318846, 35.93813663804629, 59.94842503189414,
-		100.00000000000007,
+		1.0, 1.6681, 2.7825, 4.6415, 7.7426, 12.9154, 21.5443,
+		35.9381, 59.9484, 100.0000,
 	}
-	if !reflect.DeepEqual(got, want) {
-		t.Errorf("exponential buckets range: got %v, want %v", got, want)
+	const tolerance = 0.0001
+	if !equalFloat64s(got, want, 0.0001) {
```
dswarbrick, line 362: Wouldn't it make sense to avoid the repetition of the tolerance constant value here?
```suggestion
	if !equalFloat64s(got, want, tolerance) {
```

sbunce: Oops, yes. That should be used.

beorn7:
There are better ways of a tolerant comparison. We already have https://github.com/prometheus/prometheus/blob/main/promql/test.go#L637 in the Prometheus code base. Perhaps that's easy enough to just copy over here.

beorn7:
Or depend on https://github.com/beorn7/floats :o)

dswarbrick:
> Or depend on https://github.com/beorn7/floats :o)

I would prefer not to have to package yet another tiny dependency to make this build on Debian (and derivatives), so I would advocate just copying that function.

Or use something like https://pkg.go.dev/github.com/stretchr/testify/assert?utm_source=godoc#InDeltaSlice

sbunce:
Sounds like there's consensus on avoiding the extra dependency. Reminds me of that go proverb, "a little copying is better than a little dependency".

I copied the contents of https://github.com/beorn7/floats to the prometheus/internal package. I made a few small changes.
- I renamed "ε" to "epsilon". I feel like this improves readability for non-math people slightly. I can switch back to the original variable name if people want.
- I modified the comment to reference the original repo.
- I added variants which operate on slices of floats.

Couple points I anticipate there may be disagreement about. Calling these out.
- I haven't copied over the tests. Seems like vendoring doesn't usually grab the tests, but this is a slightly different situation.
- I copied over the float32 variant of things, even though we don't use it. Let me know if I should remove those.

prometheus/internal/difflib.go
```diff
@@ -46,6 +47,97 @@ func calculateRatio(matches, length int) float64 {
 	return 1.0
 }
 
+var (
```
beorn7, line 50: I wouldn't put this in difflib.go because that file is really a (possibly modified) copy of https://github.com/pmezard/go-difflib/blob/master/difflib/difflib.go . Better to use a separate file for the AlmostEqual code.

beorn7 (commented):
Looks good to me in general. Modulo location in file, see comment. Maintainers need to make the final call. @bwplotka @kakkoyun

beorn7:
> I renamed "ε" to "epsilon".

I love the ε, but I am well aware that I'm a small minority here. (Insert lament about decline of classical education…)

beorn7:
> I copied over the float32 variant of things.

I think it is overwhelmingly likely that we'll never use the float32 here. I would prefer to remove this variant. It's easy to add later should we really need it, thanks to the reference back to the source repo.

All other points you made seem fine to me.

Thanks for your contribution and your patience.

sbunce:
You're welcome. I appreciate your feedback.

- I removed the float32 variants. I bet you're right that this will never be used.
- I moved the function to a new "almost_equal.go" file. I didn't realize the difflib file was a fork of a vendored dependency. Makes sense to not mix those.

beorn7 (approved):
Thanks again.

@bwplotka @kakkoyun please have a final look.

kakkoyun (approved):
LGTM :) Thanks for review @beorn7

PR #1139, opened 2022-09-23
expose a request path label (prototype)

empire:
This pull request demonstrates that the request path label can be supported without any security concerns.


empire:
I've opened a conversation at the [google group](https://groups.google.com/g/prometheus-developers/c/gRgSwiu3qmI).

PR #1141, opened 2022-09-28
Fix ±Inf handling in sparse buckets

### LINKED ISSUE #1131: histograms: Verify the handling of ±Inf observations
The idea so far was to not put observations of `+Inf`, `-Inf`, and `NaN` into any bucket, but the [code](https://github.com/prometheus/client_golang/blob/95cf173f1965388665dcb2a28971f35af280e3a5/prometheus/histogram.go#L589-L590) doesn't seem to do that. (Also, that's just an idea so far, heavily underspecified…)

Using `math.MaxInt32` as the bucket index in the code above will create an overflow when encoding the spans and a bucket with a negative index is followed by this "Infinity" bucket.

If we want to have an index for an "infinity" bucket, it had to be half of `math.MaxInt32` at most. This deserves a bit of thought… (also because of the resolution limit we bake into the spec via the index (and the delta between indices!) being an sint32 protobuf type). OTel came up with use cases of extremely high resolution…

beorn7:
__This is only for the sparsehistogram branch.__

Fixes #1131.

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -577,7 +577,9 @@ func (hc *histogramCounts) observe(v float64, bucket int, doSparse bool) {
 		atomic.AddUint64(&hc.buckets[bucket], 1)
 	}
 	atomicAddFloat(&hc.sumBits, v)
-	if doSparse {
+
+	// Only handle sparse buckets if requested, and only if v is a regular number.
+	if doSparse && !math.IsInf(v, 0) && !math.IsNaN(v) {
 		var (
 			sparseKey           int
 			sparseSchema        = atomic.LoadInt32(&hc.sparseSchema)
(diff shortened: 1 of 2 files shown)


fstab (approved):
The intention of this PR is to ignore `+Inf`, `-Inf`, and `NaN` in sparse histogram buckets. I don't see any bug in the implementation, so from my point of view it can be merged.

However, there are some implications of this approach:

* If we plan a backwards-compatible way to expose sparse histograms in the current text format for older Prometheus servers, conversion will be inconsistent, because current histograms have a `+Inf` bucket.
* Prometheus' `histogram_quantile()` will be off if lots of `+Inf` and `-Inf` values are observed.

These are of course corner cases that won't matter much in practice. However, these corner cases would not exist if we had explicit `+Inf` and `-Inf` buckets in sparse histograms.

Issue #1137, opened 2022-09-20
Support for multiple samples within same metric 'NewMetricWithTimestamp'


machadovilaca:
This might not even be considered a bug, but I think this is a very annoying behavior.

When using `NewMetricWithTimestamp` in a Collector as in:

```golang
// {
//     Name: "metric_test_1", Description: "metric_test_1 description",
//     Values: [ { value: 1,  timestamp: 1663713926158 }, { value: 2,  timestamp: 1663713937491 } ] 
// } 

desc := prometheus.NewDesc(m.Name, m.Description, nil, nil)

for _, value := range m.Values {
  ch <- prometheus.NewMetricWithTimestamp(
    value.timestamp,
    prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, value.value),
  )
}
```

The client is running [checkMetricConsistency](https://github.com/prometheus/client_golang/blob/main/prometheus/registry.go#L869) for each of the pushed metrics which is throwing the following error

```
...
* collected metric "metric_test_1" { gauge:<value:2 > timestamp_ms:1663713937491 } was collected before with the same name and label values
...
```

I think the validation should be aware of the metrics timestamps and not throw an error in this case because the metrics belong to different point in times, and there are use cases where it makes sense to allow multiple ones to be pushed in the same scrape call.

bwplotka:
Hey, I think his is not desired functionality, I am afraid. ):

Prometheus clients are optimized to present the current state (or last known state) of data to a potential collector/scraper/Prometheus. Even if we somewhat hack it to put multiple samples of the same series, Prometheus and other scrapers will most likely fail. In other words, you would need to propose a such a change to the https://github.com/OpenObservability/OpenMetrics standard. Only if it's there, will we implement it such extension (which I doubt will be accepted).

Perhaps you might need to look in Otel Go SDK and push the metric paradigm to have functionality like this, but you need to understand the huge impact of buffering multiple samples in your clients.

Hope that makes sense. I will close this issue to avoid confusion but feel free to discuss - maybe I missed something. We can always reopen. Thanks!

machadovilaca:
@bwplotka I was just checking the OpenMetrics standard specification. In the 
[Gauge MetricPoint section](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#gauge-1), isn't what I proposed on of the use cases they show?

> An example with a Metric with no labels and two MetricPoints with timestamps:
> ~~~~
[...]

PR #1142, opened 2022-09-29
Check metric uniqueness against timestamp

Elecon-rou:
It is possible to pass multiple value-timestamp pairs of the same metric name and same labels to Prometheus. If its not something horrible to do, then maybe we could add some kind of check for that.
Possible fix for (#1137)

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"unicode/utf8"
@@ -931,6 +932,10 @@ func checkMetricConsistency(
 		h.WriteString(lp.GetName())
 		h.Write(separatorByteSlice)
 		h.WriteString(lp.GetValue())
+		h.Write(separatorByteSlice)
+	}
+	if dtoMetric.TimestampMs != nil {
+		h.WriteString(strconv.FormatInt(*(dtoMetric.TimestampMs), 10))
 		h.Write(separatorByteSlice)
 	}
 	hSum := h.Sum64()


beorn7:
> It is possible to pass multiple value-timestamp pairs of the same metric name and same labels to Prometheus.

While that's true, it's not a behavior covered by the specification of the exposition format. (Arguably, Prometheus is following the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) here. Prometheus could stop this behavior at any time and would still be compliant. Instrumentation libraries must not depend on this behavior to exist. (Or to continue following the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) narrative: The instrumentation libraries should stick to the spec as closely as possible, even though (or actually: precisely because) Prometheus is tolerant in what it accepts.)

bwplotka:
Hey, I looked, and this is not desired functionality, I am afraid. ):

Prometheus clients are optimized to present the current state (or last known state) of data to a potential collector/scraper/Prometheus. Even if we somewhat hack it to put multiple samples of the same series, Prometheus and other scrapers will most likely fail. In other words, you would need to propose such change to the https://github.com/OpenObservability/OpenMetrics standard. Only if it's there, will we implement it such extension (which I doubt will be accepted).

Perhaps you might need to look in Otel Go SDK and pushing metric paradigm to have functionality like this, but you need to understand huge impact of buffering multiple samples in your clients.

Hope that makes sense. I will close this PR to avoid confusion but feel free discuss - maybe I missed something. We can always reopen. Thanks!

bwplotka:
Discussion moved forwared in the issue, I think we could change it as in PR. Can we at least add unit test that proves multiple samples of same metric are possible?

Issue #765, opened 2020-06-04
Flaky test `TestSummaryDecay`


beorn7:
Recently, `TestSummaryDecay` started to become flaky on CircleCI. It's a timing dependent test. It used to work well, but I guess something has changed in CircleCI. Ideally, the test should be made less timing sensitive.

beorn7:
Hasn't happened in a while. Will close this (although a proper solution would be better of course).

bwplotka:
Happened again 2x times on new Go 1.20 https://app.circleci.com/pipelines/github/prometheus/client_golang/902/workflows/1a9e0222-09a4-4f9d-8c76-89441995bb24/jobs/3496

![image](https://user-images.githubusercontent.com/6950331/226727044-f2023c8d-27cb-4bd5-b58b-ada294df3608.png)

imorph:
fixed in #1672

PR #1143, opened 2022-10-03
Filter expected metrics as well in CollectAndCompare

DariaKunoichi:
**Problem**:
Implementation for GatherAndCompare does not match documentation. Documentation says that if you provide any metricNames it filters both provided and expected metrics. But only provided metrics are filtered in `compareMetricFamilies` function.

Here is my **usecase**: https://github.com/fluxninja/aperture/blob/main/pkg/otelcollector/metricsprocessor/processor_test.go#L154

I manually add `<split>` in the expectedMetrics string so I don't have to provide multiple expectedMetrics variables.

### DIFF
--- a/prometheus/testutil/testutil_test.go
+++ b/prometheus/testutil/testutil_test.go
@@ -331,6 +331,31 @@ func TestScrapeAndCompare(t *testing.T) {
 	}
 }
 
+func TestScrapeAndCompareWithMultipleExpected(t *testing.T) {
+	const expected = `
+		# HELP some_total A value that represents a counter.
+		# TYPE some_total counter
+
+		some_total{ label1 = "value1" } 1
+
+		# HELP some_total2 A value that represents a counter.
+		# TYPE some_total2 counter
+
+		some_total2{ label1 = "value1" } 1
+	`
+
+	expectedReader := strings.NewReader(expected)
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, expected)
+	}))
+	defer ts.Close()
+
+	if err := ScrapeAndCompare(ts.URL, expectedReader, "some_total2"); err != nil {
+		t.Errorf("unexpected scraping result:\n%s", err)
+	}
+}
+
 func TestScrapeAndCompareFetchingFail(t *testing.T) {
 	err := ScrapeAndCompare("some_url", strings.NewReader("some expectation"), "some_total")
 	if err == nil {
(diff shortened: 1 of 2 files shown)


DariaKunoichi:
@bwplotka Please take a look at this quick fix.

kakkoyun:
@DariaKunoichi It would be nice to have some context in the PR description. What's the expected behavior and what is not working?

bwplotka (commented):
Awesome, thanks for contributing! ❤️ 

A description of the PR would be nice, as now we have to guess what you are trying to fix here. 🙈 

I guess that you would like the provided `expected` slice to be affected by filtering. Thanks to that, functions like `ScrapeAndCompare` could accept any bigger, arbitrary `expected` stream of bytes. However, given expected stream is provided usually as synthetic sting (hand crafted), you can filter it on your own - that's why it was filtered before.

So to push this change, ideally:

* Please tell us what use case you have, it will help motivate this change.
* Can we modify the `TestScrapeAndCompare` unit test to test this change?

DariaKunoichi:
Thanks for the pointers, guys. I need to remember what was the issue and I'll add the description and modify the PR if it's needed.

bwplotka:
Thx and sorry for lag, will check today (:

kakkoyun (approved):
LGTM 🚀

prometheus/testutil/testutil_test.go
```diff
@@ -331,6 +331,31 @@ func TestScrapeAndCompare(t *testing.T) {
...
+		# HELP some_total2 A value that represents a counter.
+		# TYPE some_total2 counter
+
+		some_total2{ label1 = "value1" } 1
+	`
+
+	expectedReader := strings.NewReader(expected)
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, expected)
```
bwplotka, line 350: Can we expose different some_total{} so we can actually test filtering? Otherwise without your change this test would pass as well if I am understanding this correctly.

bwplotka (commented):
Thanks! One important improved to test is needed if I am not wrong 🤗

prometheus/testutil/testutil_test.go
DariaKunoichi: Ok, sure, will change.

bwplotka (approved):
Thanks!

Great job!

bwplotka:
We have one flaky test https://github.com/prometheus/client_golang/issues/765, merging

Issue #1145, opened 2022-10-07
Race condition with Exemplar in Counter


fstab:
I found an inconsistent concurrency behavior in `client_java`, and I think that `client_golang` has the same issue. Here's the current implementation of the counter's `Write()` function:

```golang
func (c *counter) Write(out *dto.Metric) error {
	val := c.get()

	var exemplar *dto.Exemplar
	if e := c.exemplar.Load(); e != nil {
		exemplar = e.(*dto.Exemplar)
	}

	return populateMetric(CounterValue, val, c.labelPairs, exemplar, out)
}
```

Counter implements `selfCollector`. If I understand correctly, this means that the counter continues to be updated while `Write()` is being called.

So if an update happens after reading `val` but before reading `exemplar`, you might see an examplar for an observation that's not represented in `val`.

I guess to fix this it would be sufficient to read `exemplar` first and then `val`, because then the worst case you get an exemplar for a previous observation, but you won't get an exemplar for a not-yet-existing observation.

fstab:
PR for the trivial fix. I hope I'm not missing anything #1146.

PR #1146, opened 2022-10-07
Fix race condition with Exemplar in Counter

fstab:
Potential fix for #1145.

Signed-off-by: Fabian Stäber <fabian@fstab.de>

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -140,7 +140,8 @@ func (c *counter) get() float64 {
 }
 
 func (c *counter) Write(out *dto.Metric) error {
-
+	// Read the Exemplar first and the value second. This is to avoid a race condition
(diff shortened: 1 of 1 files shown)


fstab:
Build fails with
```
Error: File is not `gofumpt`-ed with `-extra` (gofumpt)
```
Was that me, or is that an issue with the build pipeline?

bboreham (commented):
Suggest a code comment stating that the ordering is essential, otherwise someone may swap it back.
Or even better, a test.

Also note the linter complained.

fstab:
Thanks Bryan. I added a comment. Not sure why the linter had failed, but now it's green again.

bboreham (approved):
lgtm

SD145 (approved):
Last comment

Issue #1147, opened 2022-10-12
Exemplar for `+Inf` bucket adds bucket with wrong value


balintzs:
When writing histogram metrics, and there is an exemplar for the `+Inf` bucket, then the `+Inf` bucket is added with the value of the previous bucket:
https://github.com/prometheus/client_golang/blob/main/prometheus/metric.go#L188

This is incorrect, as the cumulative count of the `+Inf` bucket should instead be added with the total count of the histogram datapoint.

The above behaviour results in invalid values being reported for the `+Inf` bucket.

balintzs:
I'll open a PR shortly

bwplotka:
Thanks! Approved fix with some small nit.

balintzs:
Thanks for merging it, @bwplotka. When can we expect it to get released?

bwplotka:
Can you just use commit sha for now? We will release soon, but don't know when (KubeCom + PromCon is coming)

bboreham:
Note your link to the problem now shows the fixed code, because it references `main`.
You can get a fixed link by pressing 'Y' in the GitHub UI, e.g.:
https://github.com/prometheus/client_golang/blob/3faf3bae7076031e56adc1f76d0c2d817410597e/prometheus/metric.go#L188-L190

PR #1148, opened 2022-10-12
Fix `CumulativeCount` value of `+Inf` bucket created from exemplar

balintzs:
When writing histogram metrics, and there is an exemplar for the +Inf bucket, then the +Inf bucket is added with the value of the previous bucket.

This is incorrect, as the cumulative count of the +Inf bucket should instead be added with the total count of the histogram datapoint.

The above behaviour results in invalid values being reported for the +Inf bucket.

See issue: https://github.com/prometheus/client_golang/issues/1147

@bwplotka @kakkoyun

### DIFF
--- a/prometheus/metric_test.go
+++ b/prometheus/metric_test.go
@@ -81,11 +81,11 @@ func TestWithExemplarsMetric(t *testing.T) {
 
 		infBucket := metric.GetHistogram().Bucket[len(metric.GetHistogram().Bucket)-1]
 
-		if want, got := infBucket.GetUpperBound(), math.Inf(1); want != got {
+		if want, got := math.Inf(1), infBucket.GetUpperBound(); want != got {
 			t.Errorf("want %v, got %v", want, got)
 		}
 
-		if want, got := infBucket.GetCumulativeCount(), uint64(4711); want != got {
+		if want, got := uint64(4711), infBucket.GetCumulativeCount(); want != got {
 			t.Errorf("want %v, got %v", want, got)
 		}
 	})
--- a/prometheus/metric_test.go
+++ b/prometheus/metric_test.go
@@ -79,10 +79,14 @@ func TestWithExemplarsMetric(t *testing.T) {
 			}
 		}
 
-		infBucket := metric.GetHistogram().Bucket[len(metric.GetHistogram().Bucket)-1].GetUpperBound()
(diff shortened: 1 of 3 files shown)


prometheus/metric_test.go
```diff
@@ -79,10 +79,14 @@ func TestWithExemplarsMetric(t *testing.T) {
...
+		infBucket := metric.GetHistogram().Bucket[len(metric.GetHistogram().Bucket)-1]
 
-		if infBucket != math.Inf(1) {
-			t.Errorf("want %v, got %v", math.Inf(1), infBucket)
+		if upperBound := infBucket.GetUpperBound(); upperBound != math.Inf(1) {
+			t.Errorf("want %v, got %v", math.Inf(1), upperBound)
+		}
+
+		if cumulativeCount := infBucket.GetCumulativeCount(); cumulativeCount != 4711 {
+			t.Errorf("want %v, got %v", 4711, cumulativeCount)
```
bwplotka, line 89: Can we stay consistent with code above?

```suggestion
		if want, got := infBucket.GetUpperBound(), math.Inf(1); want != got {
			t.Errorf("want %v, got %v", want, got)
		}

		if want, got := infBucket.GetCumulativeCount(), uint64(4711); want != got {
			t.Errorf("want %v, got %v", want, got)
```

bwplotka (approved):
I think you are right, thanks!

The confusion was from the fact that const histogram `count` includes all in last bucket PLUS what is in Inf  but it's not in others. This means `Inf` should be equal to `count`. 

Can you help reduce confusion by changing the comment in `histogram.go:615` from 

```
// buckets is a map of upper bounds to cumulative counts, excluding the +Inf
// bucket.
```

To something like:

```
// buckets is a map of upper bounds to cumulative counts, excluding the +Inf
// bucket. The +Inf bucket is implicit, and its value is equal to the provided count.
//
```

bwplotka:
Fixes: https://github.com/prometheus/client_golang/issues/1147

prometheus/metric_test.go
balintzs: I pushed another commit to fix the assignment, as I didn't notice when I committed this that `want` and `got` were mixed up.

bwplotka (approved):
Thanks 💪🏽

PR #1151, opened 2022-10-21
Add constrained labels and Constrained variant for all MetricVecs

Okhoshi:
Constrained variants of MetricVec provides a way to restrict the possible values a label can take, to prevent cardinality explosion when the label value comes from a non-trusted source (as a user input or HTTP header).

Extracted from discussions on #1066 

@bwplotka @kakkoyun

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -59,6 +59,18 @@ type ExemplarAdder interface {
 // CounterOpts is an alias for Opts. See there for doc comments.
 type CounterOpts Opts
 
+// CounterVecOpts bundles the options to create a CounterVec metric.
+// It is mandatory to set CounterOpts, see there for mandatory fields. VariableLabels
+// is optional and can safely be left to its default value.
+type CounterVecOpts struct {
+	CounterOpts
+
+	// VariableLabels are used to partition the metric vector by the given set
+	// of labels. Each label value will be constrained with the optional Contraint
+	// function, if provided.
+	VariableLabels ConstrainableLabels
+}
+
 // NewCounter creates a new Counter based on the provided CounterOpts.
 //
 // The returned implementation also implements ExemplarAdder. It is safe to
@@ -174,16 +186,24 @@ type CounterVec struct {
 // NewCounterVec creates a new CounterVec based on the provided CounterOpts and
 // partitioned by the given label names.
 func NewCounterVec(opts CounterOpts, labelNames []string) *CounterVec {
-	desc := NewDesc(
+	return V2.NewCounterVec(CounterVecOpts{
+		CounterOpts:    opts,
+		VariableLabels: UnconstrainedLabels(labelNames),
+	})
+}
+
+// NewCounterVec creates a new CounterVec based on the provided CounterVecOpts.
+func (v2) NewCounterVec(opts CounterVecOpts) *CounterVec {
+	desc := V2.NewDesc(
 		BuildFQName(opts.Namespace, opts.Subsystem, opts.Name),
 		opts.Help,
-		labelNames,
+		opts.VariableLabels,
 		opts.ConstLabels,
 	)
 	return &CounterVec{
 		MetricVec: NewMetricVec(desc, func(lvs ...string) Metric {
 			if len(lvs) != len(desc.variableLabels) {
-				panic(makeInconsistentCardinalityError(desc.fqName, desc.variableLabels, lvs))
+				panic(makeInconsistentCardinalityError(desc.fqName, desc.variableLabels.labelNames(), lvs))
 			}
 			result := &counter{desc: desc, labelPairs: MakeLabelPairs(desc, lvs), now: time.Now}
 			result.init(result) // Init self-collection.
--- a/prometheus/desc.go
+++ b/prometheus/desc.go
@@ -53,9 +53,9 @@ type Desc struct {
 	// constLabelPairs contains precalculated DTO label pairs based on
 	// the constant labels.
 	constLabelPairs []*dto.LabelPair
-	// variableLabels contains names of labels for which the metric
-	// maintains variable values.
-	variableLabels []string
+	// variableLabels contains names of labels and normalization function for
+	// which the metric maintains variable values.
+	variableLabels ConstrainedLabels
 	// id is a hash of the values of the ConstLabels and fqName. This
 	// must be unique among all registered descriptors and can therefore be
 	// used as an identifier of the descriptor.
@@ -79,10 +79,24 @@ type Desc struct {
 // For constLabels, the label values are constant. Therefore, they are fully
 // specified in the Desc. See the Collector example for a usage pattern.
 func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *Desc {
+	return V2.NewDesc(fqName, help, UnconstrainedLabels(variableLabels), constLabels)
+}
+
+// NewDesc allocates and initializes a new Desc. Errors are recorded in the Desc
+// and will be reported on registration time. variableLabels and constLabels can
+// be nil if no such labels should be set. fqName must not be empty.
+//
+// variableLabels only contain the label names and normalization functions. Their
+// label values are variable and therefore not part of the Desc. (They are managed
+// within the Metric.)
+//
+// For constLabels, the label values are constant. Therefore, they are fully
+// specified in the Desc. See the Collector example for a usage pattern.
+func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, constLabels Labels) *Desc {
 	d := &Desc{
 		fqName:         fqName,
 		help:           help,
-		variableLabels: variableLabels,
+		variableLabels: variableLabels.constrainedLabels(),
 	}
 	if !model.IsValidMetricName(model.LabelValue(fqName)) {
 		d.err = fmt.Errorf("%q is not a valid metric name", fqName)
@@ -92,7 +106,7 @@ func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *
 	// their sorted label names) plus the fqName (at position 0).
 	labelValues := make([]string, 1, len(constLabels)+1)
 	labelValues[0] = fqName
-	labelNames := make([]string, 0, len(constLabels)+len(variableLabels))
+	labelNames := make([]string, 0, len(constLabels)+len(d.variableLabels))
 	labelNameSet := map[string]struct{}{}
 	// First add only the const label names and sort them...
 	for labelName := range constLabels {
@@ -117,13 +131,13 @@ func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *
 	// Now add the variable label names, but prefix them with something that
 	// cannot be in a regular label name. That prevents matching the label
 	// dimension with a different mix between preset and variable labels.
-	for _, labelName := range variableLabels {
-		if !checkLabelName(labelName) {
-			d.err = fmt.Errorf("%q is not a valid label name for metric %q", labelName, fqName)
+	for _, label := range d.variableLabels {
+		if !checkLabelName(label.Name) {
+			d.err = fmt.Errorf("%q is not a valid label name for metric %q", label.Name, fqName)
 			return d
 		}
-		labelNames = append(labelNames, "$"+labelName)
-		labelNameSet[labelName] = struct{}{}
+		labelNames = append(labelNames, "$"+label.Name)
+		labelNameSet[label.Name] = struct{}{}
 	}
 	if len(labelNames) != len(labelNameSet) {
 		d.err = fmt.Errorf("duplicate label names in constant and variable labels for metric %q", fqName)
--- a/prometheus/labels.go
+++ b/prometheus/labels.go
@@ -32,6 +32,78 @@ import (
 // create a Desc.
 type Labels map[string]string
 
+// ConstrainedLabels represents a label name and its constrain function
+// to normalize label values. This type is commonly used when constructing
+// metric vector Collectors.
+type ConstrainedLabel struct {
+	Name       string
+	Constraint func(string) string
+}
+
+func (cl ConstrainedLabel) Constrain(v string) string {
+	if cl.Constraint == nil {
+		return v
+	}
+	return cl.Constraint(v)
+}
+
+// ConstrainableLabels is an interface that allows creating of labels that can
+// be optionally constrained.
+//
+//	prometheus.V2().NewCounterVec(CounterVecOpts{
+//	  CounterOpts: {...}, // Usual CounterOpts fields
+//	  VariableLabels: []ConstrainedLabels{
+//	    {Name: "A"},
+//	    {Name: "B", Constraint: func(v string) string { ... }},
+//	  },
+//	})
+type ConstrainableLabels interface {
+	constrainedLabels() ConstrainedLabels
+	labelNames() []string
+}
+
+// ConstrainedLabels represents a collection of label name -> constrain function
+// to normalize label values. This type is commonly used when constructing
+// metric vector Collectors.
+type ConstrainedLabels []ConstrainedLabel
+
+func (cls ConstrainedLabels) constrainedLabels() ConstrainedLabels {
+	return cls
+}
+
+func (cls ConstrainedLabels) labelNames() []string {
+	names := make([]string, len(cls))
+	for i, label := range cls {
+		names[i] = label.Name
+	}
+	return names
+}
+
+// UnconstrainedLabels represents collection of label without any constraint on
+// their value. Thus, it is simply a collection of label names.
+//
+//	UnconstrainedLabels([]string{ "A", "B" })
+//
+// is equivalent to
+//
+//	ConstrainedLabels {
+//	  { Name: "A" },
+//	  { Name: "B" },
+//	}
+type UnconstrainedLabels []string
+
+func (uls UnconstrainedLabels) constrainedLabels() ConstrainedLabels {
+	constrainedLabels := make([]ConstrainedLabel, len(uls))
+	for i, l := range uls {
+		constrainedLabels[i] = ConstrainedLabel{Name: l}
+	}
+	return constrainedLabels
+}
+
+func (uls UnconstrainedLabels) labelNames() []string {
+	return uls
+}
+
 // reservedLabelPrefix is a prefix which is not legal in user-supplied
 // label names.
 const reservedLabelPrefix = "__"
--- a/prometheus/vnext.go
+++ b/prometheus/vnext.go
@@ -1,0 +1,23 @@
+// Copyright 2022 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package prometheus
+
+type v2 struct{}
+
+// V2 is a struct that can be referenced to access experimental API that might
+// be present in v2 of client golang someday. It offers extended functionality
+// of v1 with slightly changed API. It is acceptable to use some pieces from v1
+// and e.g `prometheus.NewGauge` and some from v2 e.g. `prometheus.V2.NewDesc`
+// in the same codebase.
+var V2 = v2{}
--- a/prometheus/counter_test.go
+++ b/prometheus/counter_test.go
@@ -102,7 +102,7 @@ func TestCounterVecGetMetricWithInvalidLabelValues(t *testing.T) {
 			Name: "test",
 		}, []string{"a"})
 
-		labelValues := make([]string, len(test.labels))
+		labelValues := make([]string, 0, len(test.labels))
 		for _, val := range test.labels {
 			labelValues = append(labelValues, val)
 		}
--- a/prometheus/examples_test.go
+++ b/prometheus/examples_test.go
@@ -294,9 +294,9 @@ func ExampleRegister() {
 
 	// Output:
 	// taskCounter registered.
-	// taskCounterVec not registered: a previously registered descriptor with the same fully-qualified name as Desc{fqName: "worker_pool_completed_tasks_total", help: "Total number of tasks completed.", constLabels: {}, variableLabels: [worker_id]} has different label names or a different help string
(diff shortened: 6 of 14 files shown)


prometheus/desc.go
```diff
@@ -161,6 +163,62 @@ func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *
...
+			if labelSet[label] {
+				variableLabels = append(variableLabels, label)
+			}
+		}
+	}
+
+	d := NewDesc(fqName, help, variableLabels, constLabels)
+	if len(constrainedLabels) > 0 {
+		d.labelConstraints = make([]func(string) string, len(d.variableLabels))
+		for i, label := range d.variableLabels {
```
bwplotka, line 207: Not sure why we do anything with  variable labels in terms of constraints

prometheus/desc.go
```diff
@@ -161,6 +163,62 @@ func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *
...
+				d.labelConstraints[i] = fn
+			} else {
+				d.labelConstraints[i] = identityConstraint
+			}
+		}
+	}
+	return d
+}
+
+func identityConstraint(lv string) string {
```
bwplotka, line 218: What does it mean?

bwplotka: Do we need this? Why not just leaving non constrained labels alone - not do anything special, just use old flow. Then only constrained will execute function

bwplotka (commented):
Generally great! Just some suggestions ideas for simpler API. Thanks!

prometheus/desc.go
```diff
@@ -161,6 +163,62 @@ func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *
...
+func NewConstrainedDesc(
+	fqName, help string,
+	variableLabels []string,
+	constLabels Labels,
+	constrainedLabels ConstrainedLabels,
+) *Desc {
+	if len(constrainedLabels) > 0 {
+		// Merge constrainedLabels into variableLabels as a source for variable
+		// labels in the Desc
+		labelSet := map[string]bool{}
```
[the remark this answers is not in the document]
Okhoshi: The bool value let us know which value of the label set is already present in the variable label list. This allows to preserve the order of the variable labels set by the user.

prometheus/desc.go
Okhoshi: The easiest way to integrate with the current inner working was to keep the contrains in an array that would have the exact same elements as the variable labels one. Because some Desc functions only receive the label values, it's not possible to do lookups by label name easily.

prometheus/desc.go
Okhoshi: This is an identity function that always returns the string it receives. This reason we need this is in the [comment above](https://github.com/prometheus/client_golang/pull/1151#discussion_r1003624180)

prometheus/desc.go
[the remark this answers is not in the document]
bwplotka: Right... I think we should never have overlap there. Labels are sorted at the end anyway, so specifying same label name as const AND dynamic AND constrained should return err

prometheus/desc.go
bwplotka: Yes, again, see comment above - what if we have contained label names in separate set.

prometheus/desc.go
bwplotka: proposed idea that perhaps removes this.

bwplotka:
Some alternatives we might have:

```go
func TestConstrained(t *testing.T) {
	constrained := ConstrainedLabels{
		"c": func(s string) string {
			switch s {
			case "value1":
				return s
			default:
				return "unspecified"
			}
		},
	}

	// Option 1: Wrap registry.
	// Cons: We can't validate WithLabelValues(...)
	{
		r := NewRegistry()
		r = WrapRegistererWithConstrained(constrained, r)
		c := NewCounterVec(CounterOpts{Name: "yolo"}, []string{"a"})
		c2 := NewCounter(CounterOpts{Name: "yolo2"})
		r.MustRegister(c, c2)
	}

	// Option 2a: New function
	// Cons: Weird when we want to have metric with only constrained dynamic labels
	{
		r := NewRegistry()
		c := NewConstrainedCounterVec(CounterOpts{Name: "yolo"}, []string{"a"}, constrained)
		c2 := NewConstrainedCounterVec(CounterOpts{Name: "yolo2"}, nil, constrained)
		r.MustRegister(c, c2)
	}
	// Option 2b: New function with mapping (initial idea)
	// Cons: Complex init code, extra TWO functions to each metric type, not easy to configure mapping between dynamic label name and its constraints, weird UX - might be not clear we are mapping values not adding new ones with the set.
	{
		r := NewRegistry()
		c := NewConstrainedCounterVec(CounterOpts{Name: "yolo"}, []string{"a", "c"}, constrained)
		c2 := NewConstrainedCounterVec(CounterOpts{Name: "yolo2"}, []string{"a"}, constrained)
		r.MustRegister(c, c2)
	}

	// Option 3a: WrapCounter (func WrapCounterWithConstrainedLabel(c interface{}, labels ConstrainedLabels) *CounterVec {)
	// Pros: One wrapper per metric type.
	// Cons: Yet another way to define new label
	{
		r := NewRegistry()
		c := WrapCounterWithConstrainedLabels(NewCounterVec(CounterOpts{Name: "yolo"}, []string{"a"}), constrained)
		c2 := WrapCounterWithConstrainedLabel(NewCounter(CounterOpts{Name: "yolo2"}), constrained)
		r.MustRegister(c, c2)
	}
	// Option 3b: WrapCounter with mapping
	// Pros: One wrapper per metric type.
	// Cons: Yet another way to define new label, same as 2b
	{
		r := NewRegistry()
		c := WrapCounterWithConstrainedLabels(NewCounterVec(CounterOpts{Name: "yolo"}, []string{"a", "b"}), constrained)
		c2 := WrapCounterWithConstrainedLabel(NewCounter(CounterOpts{Name: "yolo2"}), []string{"a"}, constrained)
		r.MustRegister(c, c2)
	}

	// Option 4: New function with redefined API - variadic (perhaps in another package).
	{
		r := NewRegistry()
		c := NewCounterVec2(CounterOpts{Name: "yolo"}, WithCounterVecLabels("a", "b"), WithCounterVecConstrainedLabels(constrained))
		c2 := NewCounterVec2(CounterOpts{Name: "yolo2"}, WithCounterVecLabels("a"), WithCounterVecConstrainedLabels(constrained))
		r.MustRegister(c, c2)
	}

	// Option 5: New function with redefined API - struct (perhaps in another package).
	{
		r := NewRegistry()
		c := NewCounterVec3(CounterVecOpts{Name: "yolo", VariableLabels: []string{"a", "b"}, ConstrainedLabels: constrained})
		c2 := NewCounterVec3(CounterVecOpts{Name: "yolo2", VariableLabels: []string{"a"}, ConstrainedLabels: constrained})
		r.MustRegister(c, c2)
	}
}

type ConstrainedLabel func(string) string
type ConstrainedLabels map[string]ConstrainedLabel

type CounterVecOpts struct {
	Namespace string
	Subsystem string
	Name      string

	Help string

	ConstLabels       Labels
	VariableLabels    []string
	ConstrainedLabels ConstrainedLabels
}

```

None of the ideas are perfect (: However Option 5 seems to be the cleanest and following clent golang patterns 🤔

prometheus/counter.go
```diff
@@ -174,11 +174,23 @@ type CounterVec struct {
 // NewCounterVec creates a new CounterVec based on the provided CounterOpts and
 // partitioned by the given label names.
 func NewCounterVec(opts CounterOpts, labelNames []string) *CounterVec {
-	desc := NewDesc(
+	return NewConstrainedCounterVec(opts, labelNames, nil)
```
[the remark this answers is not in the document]
bwplotka: Continuing in main thread: https://github.com/prometheus/client_golang/pull/1151#issuecomment-1297304754

bwplotka:
Apart from those options I wonder if `func(string) string` is enough - I wonder if we want another logic to panic/error on certain cardinality. I guess It's fine, error might be handled by implementation of `func(string) string` through some side effect.

Okhoshi:
> Apart from those options I wonder if `func(string) string` is enough - I wonder if we want another logic to panic/error on certain cardinality. I guess It's fine, error might be handled by implementation of `func(string) string` through some side effect.

For me, the logic to not return error was that we always reduce to some known values. If some values are completely unacceptable, there's still the option to panic inside the `func(string) string` or to validate the values before passing them to the `WithLabels`.

Okhoshi:
> Some alternatives we might have:
> 
> [code stripped for readability]
> 
> None of the ideas are perfect (: However Option 5 seems to be the cleanest 🤔 arguments for:
> 
> Pros:
> 
> * Follow client golang patterns (configuring things via opts)
> * Extensible
> 
> We can also do 5b - so where `ConstainedLabels` could be change to `Constrains` of existing Labels in `VariableLabels`, but to me this is more confusing and harder to implement.
> 
> WDYT?

I like Option 4 and 5, for their extensibility and that they are known patterns. The choice between the two is essentially a question of personal taste I guess. We could have them in a separate package until v2 where we could make them default, WDYT ?

I also agree on the better readability if we don't merge `variableLabels` and `constrainedLabels`, having them completely separated and panicking if they overlap seems the most predictable and understandable way to manage them.
In that case, we only need to define the behaviour of `WithLabelValues` since one doesn't the label names, only the values. The most logical way would be to keep the ordering of `variableLabels` and then the ordering of `constrainedLabels`.

Just to make sure I'm on the same page as you, `constrainedLabels` are not applicable to the single Metrics types (Counter, Gauge, ...), are they ?

bwplotka:
> I like Option 4 and 5, for their extensibility and that they are known patterns. The choice between the two is essentially a question of personal taste I guess. We could have them in a separate package until v2 where we could make them default, WDYT ?

Yea, I think Opt 5 is equally explicit and less verbose perhaps, so probably better?

> In that case, we only need to define the behaviour of WithLabelValues since one doesn't the label names, only the values. The most logical way would be to keep the ordering of variableLabels and then the ordering of constrainedLabels.

You are totally right, this might cause confusion... I thought as you proposed, but it's not very explicit (someone might add constainedLabel fields above variable label fields in their definition). This is definitely Pros of your approach of setting constraints to existing variable labels. Another trade-off (: 

> Just to make sure I'm on the same page as you, constrainedLabels are not applicable to the single Metrics types (Counter, Gauge, ...), are they ?

If we assume non-mapping approach (separate label names to variableLabels) - they are not applicable to non Vectors.

Okhoshi:
> Yea, I think Opt 5 is equally explicit and less verbose perhaps, so probably better?

I agree, let's start with Opt5 then 👍 

> You are totally right, this might cause confusion... I thought as you proposed, but it's not very explicit (someone might add constainedLabel fields above variable label fields in their definition). This is definitely Pros of your approach of setting constraints to existing variable labels. Another trade-off (:

Definitely not an easy task to define this new API 😅 
Instead of splitting them completely, we could also merge them completely. And since we opt for Opt5, it gives us some freedom to redefine how we achieve this.

```go
type ConstrainedLabel struct {
  Name         string
  Constraint func(string)string
}

type CounterVecOpts struct {
  CounterOpts
  VariableLabels []ConstrainedLabel
}

// ---

prometheus.V2().NewCounterVec(CounterVecOpts{
  ..., // Usual CounterOpts stuff
  VariableLabels: []ConstrainedLabels{
    {"A"},
    {"B", func(v string) string { ... }},
  },
})
```

WDYT ?

prometheus/desc.go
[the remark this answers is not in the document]
Okhoshi: Resolved as we completely changed the principle

beorn7:
I was too busy to look at this all week. Maybe I'll find time next week. But feel free to move forward without me.

beorn7:
Finally I had a bit of time to look at it.

I haven't checked correctness (i.e. "is it doing what we expect it to do?"). I assume you are on top of that anyway.

Interesting approach with the `V2` variable. Do you know of any other applications of this pattern? It looks intriguing, but it would be even better if this has already been tried elsewhere.

In general, I'm OK with this. I'll elaborate on one more idea in the next comment, but feel free to go ahead as is.

beorn7:
This whole thing reminds me of ancient thoughts I had about revamping labels. One was about somehow automating the instantiation of all the possible vector elements, which is something that is recommended to do manually right now. Here is an example from the client_golang code itself:

```go
	cnt := prometheus.NewCounterVec(
		prometheus.CounterOpts{
			Name: "promhttp_metric_handler_requests_total",
			Help: "Total number of scrapes by HTTP status code.",
		},
		[]string{"code"},
	)
	// Initialize the most likely HTTP status codes.
	cnt.WithLabelValues("200")
	cnt.WithLabelValues("500")
	cnt.WithLabelValues("503")
```

Now imagine the `code` label above being a `ConstrainedLabel` that normalizes all label values to either `200`, `500`, `503`, or `other`.

Imagine further that the `ConstrainedLabel` type looks like this:

```go
type ConstrainedLabel struct {
	Name       string
	Constraint func(string) string
	Values     func() []string
}
```

If `Values` returns a non-empty slice, the slice contains all possible label values.

And now we have all the parts in place: If a vector has only constant and constrained labels, of which all have a `Values` function that returns a non-empty slice, the code can automatically create the cartesian product of all labels and initialize all vector elements.

WDYT?

Okhoshi:
> And now we have all the parts in place: If a vector has only constant and constrained labels, of which all have a `Values` function that returns a non-empty slice, the code can automatically create the cartesian product of all labels and initialize all vector elements.
> 
> WDYT?

I really loves the idea ! Instantiating all the possibles combinations always add some boilerplate code we could indeed automate in this case.

However, I'm not really sure to see how we could enforce that `Constraint` won't ever return something else than an element of `Values` ?
And is there any reason to make `Values` a function that returns `[]string` instead of the slice itself ?

Also, there's situation where the Cartesian product of all labels is larger than the actual set of label combinations, because some combinations may be impossible, we couldn't do anything about that though.

beorn7:
> However, I'm not really sure to see how we could enforce that Constraint won't ever return something else than an element of Values ?

That would be the contract of the interface. Go's type system is not powerful enough to let the compiler enforce it, but that shouldn't keep us from using such a contract, should it?

> And is there any reason to make Values a function that returns []string instead of the slice itself ?

I guess the reason is that I thought of `ConstrainedLabel` as an interface, but it is, in fact, a `struct`. (Maybe it should be an interface? Not sure…) So yes, if `ConstrainedLabel` remains a `struct`, with the name as a field (rather than a method), then `Values` should just be a field, too.

> On top of that, automation feels like the second step. The first should be to define the API for constraints.

Sure. But we should design the API now in a way that it allows this kind of automation without a breaking change.
(Which means we might need a boolean to switch the behavior on explicitly if we introduce it later.)

> Do we really need to force everyone to expose hundreds status code metrics, because constraints allow max 100, when during execution only 200 and 500 codes were seen for weeks, so only 2 metrics could suffice?

First of all, we could make the auto-initialization an opt-in, see above.

But to think a bit further: In the cleanest Prometheus semantics, you want to initialize all metrics in a vector in advance. Having "rare" status codes only show up when they actually occur is indeed a smell (the "springing into existence with a value of one" problem is not the only problem with that). In practice, we cannot always predict all future label values in advance, so we have to live with the situation. However, with constrained labels in the game, we _do_ know all possible label values in advance, so we _can_ initialize the full metrics vector. I would also argue that the situation where you require the label values to be constrained is likely also a situation where you want clean semantics.

On the other hand, if you want to use a constrained label merely to check if an HTTP status code is "syntactically correct" (three digit number), then you wouldn't populate the `Values` slice anyway, and the auto-init wouldn't kick in.

If you want to use a constrained label to make sure it only has a very constrained set of label values (e.g. `200`, `500`, `other`), then you would populate the `Values` slice and get the auto-init out of the box.

So maybe we don't need an opt-in flag, if we say you "opt in" by populating the `Values` slice.

Okhoshi:
> I guess the reason is that I thought of ConstrainedLabel as an interface, but it is, in fact, a struct. (Maybe it should be an interface? Not sure…) So yes, if ConstrainedLabel remains a struct, with the name as a field (rather than a method), then Values should just be a field, too.

I don't really see any reason to make ConstrainedLabel an interface, but would be happy to be proven wrong 👍 

> But we should design the API now in a way that it allows this kind of automation without a breaking change.

It seems to me that the current design is flexible enough to be extended in the future to support such automation.

Could we go further with this MR as is then ?

beorn7:
> Could we go further with this MR as is then ?

Seems fine to me.

prometheus/vnext.go
```diff
@@ -0,0 +1,18 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package prometheus
+
+type v2 struct{}
+
+var V2 = v2{}
```
bwplotka, line 18: Nice. Let's add some commentary on this exported variable. For example, the fact we are using this as a potential early experimentation of potential v2 API? Should it have an experimental mark on, so we can change it later or never in v1? 

I think I like the alias *I think* it makes it impossible for callers to change this variable globally which is great. I wonder why this and not `func V2() v2`?

prometheus/labels.go
```diff
@@ -31,6 +31,64 @@ import (
...
+}
+
+func (cl ConstrainedLabel) Constrain(v string) string {
+	if cl.Constraint != nil {
+		return cl.Constraint(v)
+	}
+	return v
+}
+
+// ConstrainableLabels is the interface to create label constraints
```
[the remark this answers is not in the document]
bwplotka: Let's also add your example of how to use it in the comment... Otherwise, it might feel tedious. https://github.com/prometheus/client_golang/pull/1151/#issuecomment-1298340233

prometheus/labels.go
```diff
@@ -31,6 +31,64 @@ import (
...
+// to normalize label values. This type is commonly used when constructing
+// metric vector Collectors.
+type ConstrainedLabels []ConstrainedLabel
+
+// constrainedLabels implements ConstrainableLabels
+func (cls ConstrainedLabels) constrainedLabels() ConstrainedLabels {
+	return cls
+}
+
+// labelNames implements ConstrainableLabels
```
bwplotka, line 65: No need for this comment, it's not exported method and nothing useful here: 🙈

prometheus/labels.go
```diff
@@ -31,6 +31,64 @@ import (
...
+	constrainedLabels() ConstrainedLabels
+	labelNames() []string
+}
+
+// ConstrainedLabels represents a collection of label name -> constrain function
+// to normalize label values. This type is commonly used when constructing
+// metric vector Collectors.
+type ConstrainedLabels []ConstrainedLabel
+
+// constrainedLabels implements ConstrainableLabels
```
bwplotka, line 60: No need for this comment, it's not exported method and nothing useful here: 🙈

prometheus/labels.go
```diff
@@ -31,6 +31,64 @@ import (
...
+		names[i] = label.Name
+	}
+	return names
+}
+
+// UnconstrainedLabels represents collection of label without any constraint on
+// their value. Thus, it is simply a collection of label names.
+type UnconstrainedLabels []string
+
+// constrainedLabels implements ConstrainableLabels
```
bwplotka, line 78: ditto

prometheus/labels.go
```diff
@@ -31,6 +31,64 @@ import (
...
+func (uls UnconstrainedLabels) constrainedLabels() ConstrainedLabels {
+	constrainedLabels := make([]ConstrainedLabel, len(uls))
+	for i, l := range uls {
+		constrainedLabels[i] = ConstrainedLabel{Name: l}
+	}
+	return constrainedLabels
+}
+
+// labelNames implements ConstrainableLabels
```
bwplotka, line 87: ditto

prometheus/labels.go
```diff
@@ -31,6 +31,64 @@ import (
...
+	names := make([]string, len(cls))
+	for i, label := range cls {
+		names[i] = label.Name
+	}
+	return names
+}
+
+// UnconstrainedLabels represents collection of label without any constraint on
+// their value. Thus, it is simply a collection of label names.
+type UnconstrainedLabels []string
```
bwplotka, line 76: Worth to comment that its an equivalent of `[]ConstrainedLabels{{Name:"a"}, ... {Name:"z"}}` ?

bwplotka (approved):
Looks clean! Some comments otherwise LGTM. I asked on Gopher slack about out `V2` pattern -  maybe we can get some quick feedback.

Thanks 💪🏽 

cc @kakkoyun

prometheus/vnext.go
Okhoshi: Good question... I guess mainly because I prefer the way `V2.bla()` looks compared to `V2().bla()`, but I don't really have any other argument. It also avoids the function call 🤷 

FWIW, you can't replace the value of `V2`, as it is a private type, and it doesn't accept `nil`.

prometheus/labels.go
Okhoshi: Done 👍

prometheus/labels.go
Okhoshi: Removed 🗑️

prometheus/labels.go
Okhoshi: Removed 🗑️

prometheus/labels.go
Okhoshi: Removed 🗑️

prometheus/labels.go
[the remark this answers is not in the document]
Okhoshi: Does that sound good to you ?

prometheus/vnext.go
bwplotka: 👍🏽 All good then

bwplotka, line 18: Can we add comment about experimental feature?

// V2 is a struct that can be referenced to access experimental API that might be present in v2 of client golang someday. It offers extended functionality of v1 with slightly changed API. It is acceptable to use some pieces from v1 and e.g `prometheus.NewGauge` and some from  v2 e.g. `prometheus.V2.NewDesc` in the same codebase.

bwplotka, line 18: Can we add comment about experimental feature?

```suggestion
// V2 is a struct that can be referenced to access experimental API that might be present in v2 of client golang someday. It offers extended functionality of v1 with slightly changed API. It is acceptable to use some pieces from v1 and e.g `prometheus.NewGauge` and some from  v2 e.g. `prometheus.V2.NewDesc` in the same codebase.
var V2 = v2{}
```

bwplotka (approved):
Amazing work, thanks!

Important comment about V2 struct is missing, otherwise it's good to merge, thanks!

bwplotka (approved):
💪🏽

PR #1168, opened 2022-11-11
Bump Go to 1.19

lucacome:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/prometheus/client_golang
 
-go 1.17
+go 1.19
```
kakkoyun, line 3: This value is for the minimum required version https://github.com/golang/go/wiki/Modules#version-selection

It should only be bumped if the module depends on an API that has introduced this version and exists for later version. 

We don't have any such dependencies, and we already bumped our CI, and it's tested against go 1.19.

kakkoyun (commented):
Thanks for the PR. However, it's not needed at this stage. Please see my comment.

PR #1174, opened 2022-11-29
promhttp: Gracefully instrument nil RoundTrippers

treuherz:
When a caller of the InstrumentRoundTripper methods has other
middlewares in their stack, the way the provided examples wrap
`http.DefaultTransport` don't quite work, and you have to do something
like this whenever you call it

```go
if httpClient.Transport == nil {
	httpClient.Transport = http.DefaultTransport
}
httpClient.Transport = promhttp.InstrumentRoundTripperDuration(httpClient.Transport)
```

This change checks for and handles a nil `next` round-tripper, meaning
the nil check in this example isn't needed. This means callers can use
a passed-in HTTP client without having to worry about whether its
transport has been set already.

Signed-off-by: Eli Treuherz <et@arenko.group>

@bwplotka  @kakkoyun

### DIFF
--- a/prometheus/promhttp/instrument_client.go
+++ b/prometheus/promhttp/instrument_client.go
@@ -63,6 +63,10 @@ func InstrumentRoundTripperInFlight(gauge prometheus.Gauge, next http.RoundTripp
 //
 // See the example for ExampleInstrumentRoundTripperDuration for example usage.
 func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.RoundTripper, opts ...Option) RoundTripperFunc {
+	if next == nil {
+		next = http.DefaultTransport
+	}
+
 	rtOpts := defaultOptions()
 	for _, o := range opts {
 		o.apply(rtOpts)
@@ -105,6 +109,10 @@ func InstrumentRoundTripperCounter(counter *prometheus.CounterVec, next http.Rou
 // Note that this method is only guaranteed to never observe negative durations
 // if used with Go1.9+.
 func InstrumentRoundTripperDuration(obs prometheus.ObserverVec, next http.RoundTripper, opts ...Option) RoundTripperFunc {
+	if next == nil {
+		next = http.DefaultTransport
+	}
+
 	rtOpts := defaultOptions()
 	for _, o := range opts {
 		o.apply(rtOpts)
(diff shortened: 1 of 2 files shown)


bwplotka:
Thanks a lot for this proposal! 

I am not sure if this is needed. It feels like the inconvenience of doing 

```
if httpClient.Transport == nil {
	httpClient.Transport = http.DefaultTransport
}
```
Only happens in very simple and trivial setups. This is because usually there are multiple round-trippers or/and custom HTTP transports (default one is not recommended for all productions setup - you might need to adjust timeouts for your use case).

On top of that user of `InstrumentRoundTripperDuration` might be surprised as if they pass `nil` by accident, instead of panic (fail fast!) somehow our function is now `DefaultHTTPTransportWithInstrumentedRoundTripperDuration`. This is why unless we have a good reason there might be some little harm in this change.

How often you have this use case, how bad you need this? 🤗

treuherz:
That's very fair, and I understand if you decide it's not worth including. You're right that the added boilerplate isn't very much. For me it's more of a matter of making the user less likely to overwrite a custom roundtripper that was already there.

The reason I wrote this PR in the first place was because I was looking a colleague's code, which they had taken written by looking at `ExampleInstrumentRoundTripperDuration()`, and they had done exactly what it does there, which is

```
rt := InstrumentRoundTripperDuration(http.DefaultTransport)
```

This overwrote the opentelemetry roundtripper that had already been placed on the client further up the stack. What I hoped to do was avoid that happening by changing the behaviour so that a user can _always_ do `InstrumentRoundTripperDuration(client.Transport)` without having to check first. This also follows the behaviour of [OpenTelemetry's](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/instrumentation/net/http/otelhttp/v0.37.0/instrumentation/net/http/otelhttp/transport.go#L50) wrapper so it seemed tidy to me.

That said, you're totally right that it could mask other mistakes. Those seem less likely to me than clobbering it, but I understand if you feel the implicit behaviour is more dangerous.

If you choose not to take on this MR, would you mind me submitting a follow-up to add the check to the examples so that it's clearer that it's something users should check?

```
if httpClient.Transport == nil {
	httpClient.Transport = http.DefaultTransport
}
httpClient.Transport = promhttp.InstrumentRoundTripperDuration(httpClient.Transport)
```

PR #1177, opened 2022-12-05
Add context to "duplicate label names" to enable debugging

SpencerMalone:
This error is really hard to debug as it provides no dynamic context, and "duplicate label names" on it's own can provide confusion as to what the error is. I tried to include a little more context on "what the source of duplication was" as well as "what metric caused this" to help in debugging.

### DIFF
--- a/prometheus/desc.go
+++ b/prometheus/desc.go
@@ -14,7 +14,6 @@
 package prometheus
 
 import (
-	"errors"
 	"fmt"
 	"sort"
 	"strings"
--- a/prometheus/desc.go
+++ b/prometheus/desc.go
@@ -127,7 +127,7 @@ func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *
 		labelNameSet[labelName] = struct{}{}
 	}
 	if len(labelNames) != len(labelNameSet) {
-		d.err = errors.New("duplicate label names")
+		d.err = fmt.Errorf("duplicate label names in constant and variable labels for metric %q", fqName)
 		return d
 	}
 


SpencerMalone:
I'm gonna @bwplotka per https://github.com/prometheus/client_golang/blob/main/CONTRIBUTING.md based on the existing committers to this file, if you want me to @ someone else, just LMK!

prometheus/desc.go
```diff
@@ -127,7 +126,7 @@ func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *
 		labelNameSet[labelName] = struct{}{}
 	}
 	if len(labelNames) != len(labelNameSet) {
-		d.err = errors.New("duplicate label names")
+		d.err = fmt.Errorf("duplicate label names in constant and variable labels for metric %q", fqName)
```
bwplotka, line 129: Should we also print `labelNames`? (:

bwplotka (approved):
Nice, looks good, just one suggestion.

Thanks!

prometheus/desc.go
SpencerMalone: I wasn't sure, I was trying to walk the line between specific enough to give you a place to debug, without giving a flood of unstructured logging. If you think printing the full array is the right move, I'm happy to include that.

bwplotka: IMO it's useful, usually metric name might be enough, but there might be duplicates of labels too - hard to debug without extra entry.

We can add in next PR though, thanks!

PR #1181, opened 2022-12-09
Support for multiple samples within same metric

### LINKED ISSUE #1137: Support for multiple samples within same metric 'NewMetricWithTimestamp'

machadovilaca:
Fixes #1137

- adds the metrics timestamp to the metric repetition validation hash
- adds a unit test for metric consistency check

Signed-off-by: João Vilaça <jvilaca@redhat.com>

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"unicode/utf8"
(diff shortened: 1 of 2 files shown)


bwplotka (approved):
LGTM, thanks!

PR #1182, opened 2022-12-13
client-golang minor fix

Shivkumar13:
Came across this while going through the README, fix for a better reading experience.

@bwplotka


bboreham:
Thanks, but it was a duplicate of #1178 already posted.

Issue #1175, opened 2022-12-02
Stop importing deprecated github.com/golang/protobuf


bwplotka:
AC:
* Move to https://pkg.go.dev/google.golang.org/protobuf API

See https://github.com/prometheus/common/issues/317

bwplotka:
Fixed by https://github.com/prometheus/client_golang/pull/1183

zhsj:
@bwplotka since github.com/golang/protobuf is still in go.mod, thus I didn't close it by commit message. Probably keep it open until the tests have migrated as well?

SuperQ:
I agree, we still need to clean up the tests so the Go modules dependencies are fixed.

drobazko:
Hey folks.
Any updates on this issue?

rufreakde:
@SuperQ any updates?

ash2k:
I found https://github.com/prometheus/common/pull/509 that should help remove the deprecated proto package from the transitive dependency list.

ash2k:
Another chain of dependencies is via the `github.com/prometheus/common` too:

```
go mod why -m github.com/golang/protobuf
# github.com/golang/protobuf
github.com/prometheus/common/config
golang.org/x/oauth2
golang.org/x/oauth2/internal
google.golang.org/appengine/urlfetch
github.com/golang/protobuf/proto
```

Work to address that is tracked in https://github.com/golang/appengine/issues/228. After it's done, someone will need to update golang.org/x/oauth2 to use the new version (there is a `google.golang.org/appengine/v2` already, so a major version bump will be needed).

PR #1183, opened 2022-12-22
Replace deprecated github.com/golang/protobuf package

zhsj:
This replaces usage of proto.{Float64,Int32,Int64,String,Uint32,Uint64}, which doesn't break the interface.

And remove usage of proto.MarshalTextString in wrap_test.go

Updates: #1175

Signed-off-by: Shengjing Zhu <zhsj@debian.org>

### DIFF
--- a/prometheus/counter_test.go
+++ b/prometheus/counter_test.go
@@ -20,11 +20,9 @@ import (
 	"testing"
 	"time"
 
-	//nolint:staticcheck // Ignore SA1019. Need to keep deprecated package for compatibility.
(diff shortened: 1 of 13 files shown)


zhsj:
This only lefts usage of `github.com/golang/protobuf/proto` in example tests. However it's just test, and shouldn't block other packages to get rid of `github.com/golang/protobuf`.

bwplotka (approved):
Thanks! 💪🏽

ahrycej:
Which version of prometheus will have this fix?

Issue #1017, opened 2022-03-28
`Gather` does not include CounterVec unless it has had `WithLabelValues` called


samsondav:
If I register as such:

```go
var (
	simpleCounter = promauto.NewCounter(prometheus.CounterOpts{
		Name: "simple_counter",
		Help: "Some help message",
	})

	simpleCounterVec = promauto.NewCounterVec(prometheus.CounterOpts{
		Name: "simple_counter_vec",
		Help: "Some help message",
	}, []string{"label1", "label2"})
)

func init() {
	// prometheus.DefaultRegisterer.MustRegister(simpleCounterVec)

	metricsFamilies, err := prometheus.DefaultGatherer.Gather()
	if err != nil {
		panic(err)
	}
	for _, f := range metricsFamilies {
		fmt.Println(f)
	}
}
```

Then only `simple_counter` is included in the metrics. `simpleCounterVec` is ignored. I know it has been registered because calling `MustRegister` panics.

I would expect all prometheus metrics to be included in the result of Gather, both regular counters and vector counters.

beorn7:
An empty vector on its own isn't producing any metric yet. Without any known label values, how should such a metric look like?

samsondav:
We want a way to export all metrics so that it could be easily seen by a consumer which metrics are available.

Perhaps we would need to implement a custom `Register` that wraps the `DefaultRegisterer` ?

PR #1215, opened 2023-01-13
promauto: initialise vector labels at time of creation

kerolasa:
Vector metrics do not appear in gather results unless these metrics are either initialised to zero or incremented.  Upside of automatically initialising to zero is least surprise with metrics that just exist regardless how common or uncommon increments are.

Fixes: https://github.com/prometheus/client_golang/issues/1017
Signed-off-by: Sami Kerola <kerolasa@cloudflare.com>

### DIFF
--- a/prometheus/promauto/auto.go
+++ b/prometheus/promauto/auto.go
@@ -173,7 +173,9 @@ func NewCounter(opts prometheus.CounterOpts) prometheus.Counter {
 // prometheus.DefaultRegisterer. If the registration fails, NewCounterVec
 // panics.
 func NewCounterVec(opts prometheus.CounterOpts, labelNames []string) *prometheus.CounterVec {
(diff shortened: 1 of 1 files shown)


prometheus/promauto/auto.go
```diff
@@ -173,7 +173,9 @@ func NewCounter(opts prometheus.CounterOpts) prometheus.Counter {
 // prometheus.DefaultRegisterer. If the registration fails, NewCounterVec
 // panics.
 func NewCounterVec(opts prometheus.CounterOpts, labelNames []string) *prometheus.CounterVec {
-	return With(prometheus.DefaultRegisterer).NewCounterVec(opts, labelNames)
+	v := With(prometheus.DefaultRegisterer).NewCounterVec(opts, labelNames)
+	v.WithLabelValues(labelNames...)
```
beorn7, line 177: You are using the label _names_ as label _values_ here. That can't be right, can it?

PR #1230, opened 2023-03-11
api: enable passing bearer token

vbatts:


bwplotka:
Thanks for this. However, my understanding is that you can add whatever header you want via custom Client (implementing roundTripper). Otherwise we would need to implement all different auth and TLS permutations.  At some point we could add some example of doing this if not written anywhere yet.

vbatts:
@bwplotka I started down that path, but it seemed like a lot of complication just to authenticate

PR #1238, opened 2023-03-21
Only set start/end if time is not Zero

### LINKED ISSUE #621: Add option to client_golang to *not* send the time on a Series() call
The [series call](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers) doesn't *require* a time, it would be nice if we had some mechanism in the go API to *not* set a time. I had opened https://github.com/prometheus/client_golang/pull/615 towards this, but I'm not sure this is a great approach, maybe we instead make the `time.Time` pointers?

jacksontj:
This is an updated PR of #615 -- based on discussion in #621

Fixes #621


bwplotka:
You most likely have to rebase, that's it.

jacksontj:
Ah, it changed from `master` to `main` (was rebased on `master` but not `main`). that should do it.

bwplotka:
Do you mind making other endpoints consistent too, while we are this?

api/prometheus/v1/api.go
```diff
@@ -1139,8 +1139,12 @@ func (h *httpAPI) Series(ctx context.Context, matches []string, startTime, endTi
 		q.Add("match[]", m)
 	}
 
-	q.Set("start", formatTime(startTime))
-	q.Set("end", formatTime(endTime))
+	if !startTime.IsZero() {
```
bwplotka, line 1142: Can we add Series method comment to mention this behavior?

jacksontj:
> Do you mind making other endpoints consistent too, while we are this?

Definitely; PR updated with those updates (didn't see them before with the wrong base).

api/prometheus/v1/api.go
jacksontj: I can, although the `Query` method (which already has this behavior) doesn't document it either. Do we still want to document this up (since this would be the case for all methods with a time.Time argument (exception being `Range` as it is required).

PR #1252, opened 2023-04-16
Switch to POST for LabelNames, Series, and QueryExemplars to DoGetFallback

jacksontj:
The upstream prometheus HTTP API supports POSTS for these methods (the same as Query and QueryRange). Similar to the original issue (https://github.com/prometheus/client_golang/issues/428) we can hit 414 errors with these other APIs. This change simply duplicates the logic to these other endpoints

Related to: https://github.com/jacksontj/promxy/issues/588


bwplotka (approved):
Thanks!

PR #1257, opened 2023-04-21
Fix tutorial on WSL-based systems

marevers:
@bwplotka I noticed while doing the tutorial on a WSL2 based system the application URL generated is invalid. This fix aims to detect whether its being ran on a WSL-based system and if so, gets the IPv4 address of `eth0` and uses that as the 'whatsup' address in the Prometheus configuration.


bwplotka:
Thanks, epic work!

PR #1263, opened 2023-04-30
graphite: refactor error handling

ifireice:
Removed Logger interface and make error handling more flexible.

For example, we could use our own logger implementation and have more control over the execution flow

```
package main

import (
	"context"
	"time"

	"github.com/prometheus/client_golang/prometheus/graphite"
	"go.uber.org/zap"
)

func main() {
	logger, _ := zap.NewProduction()
	defer logger.Sync()

	ctx := context.Background()
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()

	c := &Config{
			URL:              "graphite.example.org:3099",
			Gatherer:         prometheus.DefaultGatherer,
			Prefix:           "prefix",
			Interval:          5 * time.Second,
			Timeout:           2 * time.Second,
			ErrorHandling:     testCase.errorHandling,
			ErrorCallbackFunc: func(err error) {  if err != nil { logger.Error("run", zap.Error(err)); cancel() } },
		}


	b, err := graphite.NewBridge(c)
	if err != nil {
			t.Fatal(err)
		}

	b.Run(ctx)
}
```

### DIFF
--- a/examples/exemplars/main.go
+++ b/examples/exemplars/main.go
@@ -17,6 +17,54 @@
 package main
 
 import (
+	"context"
+	"fmt"
+	"github.com/prometheus/client_golang/prometheus/collectors"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"log"
+	"math/rand"
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/graphite"
+	"go.uber.org/zap"
+)
+
+func main() {
+	logger, _ := zap.NewProduction()
+	defer logger.Sync()
+
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	c := &Config{
+		URL:              "graphite.example.org:3099",
+		Gatherer:         prometheus.DefaultGatherer,
+		Prefix:           "prefix",
+		Interval:          5 * time.Second,
+		Timeout:           2 * time.Second,
+		ErrorHandling:     testCase.errorHandling,
+		ErrorCallbackFunc: func(err error) {  if err != nil { logger.Error("run", zap.Error(err)); cancel() } },
+	}
+
+
+	b, err := graphite.NewBridge(c)
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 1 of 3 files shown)


bwplotka:
I think this is fair improvement, but can we break compatibility there? I don't think so ): cc @beorn7 ?

ifireice:
Made error interception native to existing implementation

bwplotka (commented):
Amazing! It looks now much better, suggested one thing, let me know what you think 🤗

examples/exemplars/main.go
```diff
@@ -14,6 +14,54 @@
 // A simple example of how to record a latency metric with exemplars, using a fictional id
 // as a prometheus label.
 
+package main
```
bwplotka, line 17: Hm, we can't have another package main here 🤔

prometheus/graphite/bridge.go
```diff
@@ -79,6 +79,9 @@ type Config struct {
 	// logged regardless of the configured ErrorHandling provided Logger
 	// is not nil.
 	ErrorHandling HandlerErrorHandling
+
+	// ErrorCallbackFunc is a callback function that can be executed when error is occurred
+	ErrorCallbackFunc ErrorCallbackFunc
```
bwplotka, line 84: 👍🏽

prometheus/graphite/bridge.go
```diff
@@ -164,19 +175,29 @@ func (b *Bridge) Run(ctx context.Context) {
...
+	err := b.push()
+	if b.errorCallbackFunc != nil {
+		b.errorCallbackFunc(err)
+	}
+	switch b.errorHandling {
+	case AbortOnError:
+		return err
+	case ContinueOnError:
+		if b.logger != nil {
+			b.logger.Println("continue on error:", err)
```
bwplotka, line 187: I think it would be safe to leave that panic in, especially if we would add validation to NewBridge for it, WDYT?

bwplotka (approved):
Thanks! Sorry for lag and sounds like some unwanted changes on exemplars got in? Do you mind removing them, checking my comments and rebasing? I think it's good to go otherwise!

bwplotka:
👋🏽 Moved to draft, since there are some comments to address, but otherwise it would be nice to have this! Thanks!

PR #1264, opened 2023-05-01
Cleanup proto use in tests

SuperQ:
Switch tests over to new proto libary, eliminating the direct dependency on the old proto library. Note the indirect requirement is still needed due to upstream use by the client_model package.

Related: https://github.com/prometheus/client_golang/issues/1175

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,6 @@ require (
 	github.com/beorn7/perks v1.0.1
 	github.com/cespare/xxhash/v2 v2.2.0
 	github.com/davecgh/go-spew v1.1.1
-	github.com/golang/protobuf v1.5.3
 	github.com/json-iterator/go v1.1.12
 	github.com/prometheus/client_model v0.3.0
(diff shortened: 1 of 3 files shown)


SuperQ:
Note, it would be useful to add https://github.com/prometheus/client_model/pull/64, so we can use `prototext.Format(metric)` rather than `.String()`.

go.mod
```diff
@@ -16,6 +15,7 @@ require (
 )
 
 require (
+	github.com/golang/protobuf v1.5.3 // indirect
```
bwplotka, line 18: I mean it's still there 🙃

bwplotka (approved):
Thanks!

SuperQ:
Oops, I should have put this in draft, I actually got the client_model fixed up.

go.mod
SuperQ: Yea, there's some circular dependencies to cleanup.

PR #1272, opened 2023-05-13
Reduce constrainLabels allocations

khasanovbi:
Hi, in my application prometheus allocates a lot of memory in constrainLabels. 
I suggest replacing labels inplace instead of creating a new map.
image


prometheus/vec.go
```diff
@@ -646,16 +646,12 @@ func inlineLabelValues(lvs []string, curry []curriedLabelValue) []string {
...
-			constrainedValues[l] = desc.variableLabels[i].Constrain(v)
-			continue
+			labels[l] = desc.variableLabels[i].Constrain(v)
 		}
-		constrainedValues[l] = v
 	}
-	return constrainedValues
 }
 
 func constrainLabelValues(desc *Desc, lvs []string, curry []curriedLabelValue) []string {
```
bwplotka, line 657: Should do similar here for consistency?

bwplotka (commented):
Thanks for this!

Hm, this seems great initially, but we need to double check if it's not breaking change or problematic for people or our code. Generally reusing input memory is considered as a side effect of the function, which can cause hard to find bugs for people (modifying in place).

cc @beorn7 do you know from the top of your head if this semantics trigger some bad memories ? (((:

khasanovbi:
> Generally reusing input memory is considered as a side effect of the function, which can cause hard to find bugs for people (modifying in place).
Ok, looks like it better to use sync.Pool to prevent side effects.

khasanovbi:
in main branch:
```
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkMetricVecWithBasic
BenchmarkMetricVecWithBasic-12    	 1982983	       628.3 ns/op	     400 B/op	       4 allocs/op
```
in PR:
```
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
BenchmarkMetricVecWithBasic
BenchmarkMetricVecWithBasic-12    	 3176462	       389.7 ns/op	      64 B/op	       2 allocs/op
```

beorn7:
We dave certainly done a bunch of quite daring optimizations in client_golang in the past. It is, after all, often in a very hot code path of a lot of programs.

I have to leave the detailed code review to you, but at first glance, I don't see any red flags.

bwplotka (approved):
This is great iteration and safe. We can experiment with more aggressive label reuse, but this is already a big improvement. 

Also, I just noticed: Constrained labels is very rarely used feature (new one). I believe there is room to never allocate anything and do it only if metric HAS some constrained labels defined. WDYT? It's essentially a little bit of regression and thanks for noticing this!

Thanks 💪🏽

bwplotka (approved):
This is great iteration and safe. We can experiment with more aggressive label reuse, but this is already a big improvement. 

Also, I just noticed: Constrained labels is very rarely used feature (new one). I believe there is room to never allocate anything and do it only if metric HAS some constrained labels defined. WDYT? It's essentially a little bit of regression and thanks for noticing this!

Thanks 💪🏽

bboreham:
Historical note: this code was changed again in #1296

bwplotka:
Thanks for bringing my reviewing mistake again =D (we missed we allocate a lot of memory even if constrained feature is not used)

bboreham:
No shade intended; just took me a little time to trace the history so I added a breadcrumb.

PR #1279, opened 2023-05-25
Fix bug in bucket key calculation

beorn7:
The current code doesn't work fork negative schemas if the observed
value should go into a bucket with a non-positive key.

In the 1st commit, tests are added to expose the bug.

In the 2st commit, the bug is fixed. The expression smells a bit like it should be possible to simplify. But I couldn't come up with a simpler form. Let me know if you have any ideas.

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -639,8 +639,8 @@ func (hc *histogramCounts) observe(v float64, bucket int, doSparse bool) {
 			if frac == 0.5 {
 				key--
 			}
-			div := 1 << -schema
-			key = (key + div - 1) / div
+			offset := (1 << -schema) - 1
+			key = (key + offset) >> -schema
 		}
 		if isInf {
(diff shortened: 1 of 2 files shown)


beorn7:
So you could write the two-liner as `key = (key + (1 << -schema) - 1) >> -schema`.

If you inverted the sign of `schema` first, it would look more benign:

```Go
key = (key + (1 << schema) - 1) >> schema
```

I thought this is a perfect thing to throw at ChatGPT:

image

That's obviously garbage. Just in case anyone was afraid to lose their job…

beorn7:
The fun thing is that you can argue with ChatGPT:

image

So the improvement is that the "simplified" expression is actually less simple but also still wrong… :rofl:

fstab (approved):
Thanks for the quick fix. I ran a brute force test with random numbers and it's all green now.

PR #1298, opened 2023-06-17
Fix top-level examples in push.go.

eduarrrd:
As mentioned in the func doc for New(): "However, do not include the “/metrics/jobs/…” part."

### DIFF
--- a/prometheus/push/push.go
+++ b/prometheus/push/push.go
@@ -16,10 +16,10 @@
 // by using its methods, finally calling Add or Push, like this:
 //
 //	// Easy case:
-//	push.New("http://example.org/metrics", "my_job").Gatherer(myRegistry).Push()
+//	push.New("http://example.org/", "my_job").Gatherer(myRegistry).Push()
 //
 //	// Complex case:
-//	push.New("http://example.org/metrics", "my_job").
+//	push.New("http://example.org/", "my_job").
 //	    Collector(myCollector1).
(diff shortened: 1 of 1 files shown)


prometheus/push/push.go
```diff
@@ -16,10 +16,10 @@
 // by using its methods, finally calling Add or Push, like this:
 //
 //	// Easy case:
-//	push.New("http://example.org/metrics", "my_job").Gatherer(myRegistry).Push()
+//	push.New("http://example.org/", "my_job").Gatherer(myRegistry).Push()
```
bwplotka, line 19:
```suggestion
//	push.New("http://pushgateway.example.org", "my_job").Gatherer(myRegistry).Push()
```

prometheus/push/push.go
```diff
@@ -16,10 +16,10 @@
...
 //
 //	// Complex case:
-//	push.New("http://example.org/metrics", "my_job").
+//	push.New("http://example.org/", "my_job").
```
bwplotka, line 22:
```suggestion
//	push.New("http://pushgateway.example.org", "my_job").
```

bwplotka (approved):
Proposed more specific example URL, but you are right, thanks!

bwplotka:
👋🏽 Moved to draft, since there are some comments to address, but otherwise it would be nice to have this! Thanks!

PR #1309, opened 2023-07-11
Get variable label names from metric descriptors

sagmor:
This change introduces the `*Desc.VariableLabelNames()` function to fetch a slice with the names of all Variable Labels.

### DIFF
--- a/prometheus/desc.go
+++ b/prometheus/desc.go
@@ -205,3 +205,12 @@ func (d *Desc) String() string {
 		strings.Join(vlStrings, ","),
 	)
 }
+
+// VariableLabelNames returns the names of all variable labels of the descriptor.
+func (d *Desc) VariableLabelNames() []string {
+	if d.err != nil {
+		return []string{}
+	}
+
+	return append([]string{}, d.variableLabels.labelNames()...)
+}
--- a/prometheus/desc_test.go
+++ b/prometheus/desc_test.go
@@ -28,3 +28,26 @@ func TestNewDescInvalidLabelValues(t *testing.T) {
 		t.Errorf("NewDesc: expected error because: %s", desc.err)
 	}
 }
+
+func TestDescGetVariableLabels(t *testing.T) {
+	desc := NewDesc(
+		"sample_label",
+		"sample label",
+		[]string{"a", "b"},
+		nil,
+	)
+
+	labels := desc.VariableLabelNames()
+
+	if len(labels) != 2 {
+		t.Errorf("Desc.VariableLabelNames: expected 2 variable label, got %d", len(labels))
+	}
+	if labels[0] != "a" {
+		t.Errorf("Desc.VariableLabelNames: expected variable label 'a', got %s", labels[0])
+	}
+
+	labels[0] = "c"
+	if desc.variableLabels.labelNames()[0] != "a" {
+		t.Errorf("Desc.VariableLabelNames: returned slice can mutate Desc properties")
+	}
+}


sagmor:
CC @bwplotka as specified in CONTRIBUTING.md

I think it's a minor change but let me know if you think I should follow a different procedure 👍

sagmor:
I have a system that updates different metrics on different conditions, each of them has different labels.

With this i can have a "Bag of labels" (`map[string]string`) with all the possible labels from which I can just fetch the combination I need for that specific metric.


For Example:
```golang
func Counter(vector *prometheus.CounterVec, labelBag map[string]string) prometheus.Counter {
	descChan := make(chan *prometheus.Desc, 1)
	vector.Describe(descChan)
	desc := <-descChan

	keys := desc.VariableLabelNames()
	labels := make([]string, 0, len(keys))

	for _, key := range keys {
		labels = append(labels, labelBag[key])
	}

	return vector.WithLabelValues(labels...)
}
```

Allows me to turn something like:
```golang
metric1.WithLabelValues(labels["labelA"], labels["labelB"], labels["labelC"], labels["labelD"]).Inc()
metric2.WithLabelValues(labels["labelA"], labels["labelC"], labels["labelD"], labels["labelE"]).Inc()
metric3.WithLabelValues(labels["labelB"], labels["labelC"]).Inc()
metric4.WithLabelValues(labels["labelA"], labels["labelB"], labels["labelC"], labels["labelD"]).Inc()
metric5.WithLabelValues(labels["labelB"], labels["labelD"]).Inc()
```

into:
```golang
Counter(metric1, labels).Inc()
Counter(metric2, labels).Inc()
Counter(metric3, labels).Inc()
Counter(metric4, labels).Inc()
Counter(metric5, labels).Inc()
```

or even:
```golang
for _, metric := range metrics {
  Counter(metric, labels).Inc()
}
```

bwplotka:
Thanks! I think I would really try to understand why you need that map of labels. It sounds like an unnecessary abstraction. Why not, whenever you update that bag of labels with another value you simply increment/set new value to the metric? And if there is no change in labels, you can simply get yourself the pointer to the metrics with certain label values e.g.

```go
metric3ForAB := metric3.WithLabelValues("a", "b")

// ...
metric3ForAB.Inc()
```

Or if you have stable some labels, then dynamic others using [MustCurryWith](https://pkg.go.dev/github.com/prometheus/client_golang@v1.16.0/prometheus#CounterVec.MustCurryWith):

```go
metric3ForA := metric3.MustCurryWith(map[string]string{"labelA": "a"})

// ...
metric3ForA.WithLabelValues("b").Inc()
```

This way you improve efficiency, reliability, but even more important - correctness due to some type safety instead of dynamic map. Let us know if that helps  (:

bwplotka:
Essentially we optimized this library, so you can keep that label state in the optimized metric type, not twice -- in metric and in your map.

sagmor:
I guess my problem is that labels are already on the map, but since `metric.With` requires a map with the exact same keys I can't just use it (each metric needs a different set) so i need to either craft a new map with the filtered keys OR use `metric.WithLabelValues` fetching each key from the map.

One advantage of this approach is that it allows me to arbitrarily change each metric variable labels in one place (the metric) without touching each place the metric is modified.

I guess I could store the list of labels in a constant to reduce some duplication of label keys, but that's when I questioned myself on why not just ask the vector "What are your labels?" 🤷

sagmor:
If you don't think this is a good idea though I can refactor my code, don't have a strong requirement on "dynamic" labels and if I do, i guess I could just wrap the whole metric with something that remembers the labels and allows for bigger `With` calls.

My use case might be a bit specific anyways 👍

bwplotka:
Wonder if @beorn7 has any thoughts on this. I guess the main question remains: Why those labels are in one big map then? Does it really help you in anything? Or it's requirement coming from above?

If you are absolutely blocked, you can create thin wrapper on NewCounterVec too that saves those label keys additionally, but again--I feel there is some odd use case here - you should be able to replace that bag of labels with prometheus metric pointers, but we don't have big picture to tell.

If you could represent your full solution with some bigger examples on why big labels map pattern is useful, we are happy to reconsider something that supports it. But our hypothesis right now is that it's an anti-pattern (: Fully understand if you don't have time for this though.

beorn7:
The longer history here is that I was never really happy with `Desc`. It felt like an implementation detail that the user should never see. Therefore, I've tried to keep it as much out of the way of the users as possible, and thought a lot about how to change the use cases where you still have to touch it in the future v2 version of this library.

Adding exported methods to `Desc` would go into the exact opposite direction.

Another fundamental concern is that this library already has a lot of knobs for the users to turn. Adding more knobs requires strong justification, like simplifying a fairly common use case.

In this case, it feels we would help a relatively small number of users to simplify a relatively niche use case that could also be solved with a relatively simple work-around. All of that at the cost of increasing the cognitive load for the many many users of this library that do not need it.

bwplotka:
Thanks @beorn7 . It seems like we would prefer to not change the Desc for this use case.

Thanks @sagmor anyway!

bwplotka:
Closing for now, we can always reopen.

PR #1311, opened 2023-07-13
Add support for custom validations in promlint

machadovilaca:


prometheus/testutil/lint.go
```diff
@@ -18,12 +18,13 @@ import (
...
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/testutil/promlint"
+	"github.com/prometheus/client_golang/prometheus/testutil/promlint/validations"
 )
 
 // CollectAndLint registers the provided Collector with a newly created pedantic
 // Registry. It then calls GatherAndLint with that Registry and with the
 // provided metricNames.
-func CollectAndLint(c prometheus.Collector, metricNames ...string) ([]promlint.Problem, error) {
+func CollectAndLint(c prometheus.Collector, metricNames ...string) ([]validations.Problem, error) {
```
bwplotka, line 27: Can we avoid this? 

client_golang is 1.0 and generally we can't break compatibility.

testutil might be experimental (?) -  not sure if we ever mentioned it's compatibility level, but if we can-- let's not break the compatibility here.  Let's move Problem back to promlint and work from there (perhaps that might require validations to be  promlint package).

WDYT?

bwplotka (changes requested):
Thanks! 

I like this idea, but wonder if we have to break compatibility here. If we can let's avoid it - once this is sorted will review closer (:

prometheus/testutil/lint.go
machadovilaca: I understand your concerns, I will update it and see if I can get a clean solution

machadovilaca: @bwplotka updated

prometheus/testutil/promlint/promlint.go
```diff
@@ -110,276 +99,25 @@ func (l *Linter) Lint() ([]Problem, error) {
...
-
-// lintMetricTypeInName detects when metric types are included in the metric name.
-func lintMetricTypeInName(mf *dto.MetricFamily) []Problem {
-	var problems []Problem
-	n := strings.ToLower(mf.GetName())
-
-	for i, t := range dto.MetricType_name {
-		if i == int32(dto.MetricType_UNTYPED) {
-			continue
+	for _, fn := range defaultValidations {
```
bwplotka, line 105: Why, why not merging those slices (default and custom) and not repeating the code?

Not a blocker, that's the only nit I would improve, merging anyway.

bwplotka (approved):
Thanks! Epic 💪🏽

bwplotka:
Thanks! I would also recommend adding useful validators to our repo (:

PR #1313, opened 2023-07-19
Extend Counters, Summaries and Histograms with creation timestamp

ArthurSens:
Following our [design doc](https://docs.google.com/document/d/1kakDVn8aP1JerimLeazuMfy2jaB14W2kZDHMRokPcv4/edit), this PR aims to add created timestamp to counters, summaries, and histograms when using Prometheus protobuf.

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -115,7 +115,8 @@ type counter struct {
 	labelPairs []*dto.LabelPair
 	exemplar   atomic.Value // Containing nil or a *dto.Exemplar.
 
-	now func() time.Time // For testing, all constructors put time.Now() here.
+	// now is for testing purposes, by default it's time.Now.
+	now func() time.Time
 }
 
 func (c *counter) Desc() *Desc {
@@ -165,9 +166,7 @@ func (c *counter) Write(out *dto.Metric) error {
 		exemplar = e.(*dto.Exemplar)
 	}
 	val := c.get()
-
-	ct := c.createdTs.AsTime()
-	return populateMetric(CounterValue, val, c.labelPairs, exemplar, out, &ct)
+	return populateMetric(CounterValue, val, c.labelPairs, exemplar, out, c.createdTs)
 }
 
 func (c *counter) updateExemplar(v float64, l Labels) {
@@ -215,7 +214,7 @@ func (v2) NewCounterVec(opts CounterVecOpts) *CounterVec {
 			if len(lvs) != len(desc.variableLabels.names) {
 				panic(makeInconsistentCardinalityError(desc.fqName, desc.variableLabels.names, lvs))
 			}
-			result := &counter{desc: desc, labelPairs: MakeLabelPairs(desc, lvs), now: time.Now}
+			result := &counter{desc: desc, labelPairs: MakeLabelPairs(desc, lvs), now: opts.now}
 			result.init(result) // Init self-collection.
 			result.createdTs = timestamppb.New(opts.now())
 			return result
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	dto "github.com/prometheus/client_model/go"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 // Counter is a Metric that represents a single numerical value that only ever
@@ -90,8 +91,12 @@ func NewCounter(opts CounterOpts) Counter {
 		nil,
 		opts.ConstLabels,
 	)
-	result := &counter{desc: desc, labelPairs: desc.constLabelPairs, now: time.Now}
+	if opts.now == nil {
+		opts.now = time.Now
+	}
+	result := &counter{desc: desc, labelPairs: desc.constLabelPairs, now: opts.now}
 	result.init(result) // Init self-collection.
+	result.createdTs = timestamppb.New(opts.now())
 	return result
 }
 
@@ -106,10 +111,11 @@ type counter struct {
 	selfCollector
 	desc *Desc
 
+	createdTs  *timestamppb.Timestamp
 	labelPairs []*dto.LabelPair
 	exemplar   atomic.Value // Containing nil or a *dto.Exemplar.
 
-	now func() time.Time // To mock out time.Now() for testing.
+	now func() time.Time // For testing, all constructors put time.Now() here.
 }
 
 func (c *counter) Desc() *Desc {
@@ -160,7 +166,8 @@ func (c *counter) Write(out *dto.Metric) error {
 	}
 	val := c.get()
 
-	return populateMetric(CounterValue, val, c.labelPairs, exemplar, out)
+	ct := c.createdTs.AsTime()
+	return populateMetric(CounterValue, val, c.labelPairs, exemplar, out, &ct)
 }
 
 func (c *counter) updateExemplar(v float64, l Labels) {
@@ -200,6 +207,9 @@ func (v2) NewCounterVec(opts CounterVecOpts) *CounterVec {
 		opts.VariableLabels,
 		opts.ConstLabels,
 	)
+	if opts.now == nil {
+		opts.now = time.Now
+	}
 	return &CounterVec{
 		MetricVec: NewMetricVec(desc, func(lvs ...string) Metric {
 			if len(lvs) != len(desc.variableLabels.names) {
@@ -207,6 +217,7 @@ func (v2) NewCounterVec(opts CounterVecOpts) *CounterVec {
 			}
 			result := &counter{desc: desc, labelPairs: MakeLabelPairs(desc, lvs), now: time.Now}
 			result.init(result) // Init self-collection.
+			result.createdTs = timestamppb.New(opts.now())
 			return result
 		}),
 	}
--- a/prometheus/counter_test.go
+++ b/prometheus/counter_test.go
@@ -27,12 +27,12 @@ import (
 
 func TestCounterAdd(t *testing.T) {
 	now := time.Now()
-	nowFn := func() time.Time { return now }
+
 	counter := NewCounter(CounterOpts{
 		Name:        "test",
 		Help:        "test help",
 		ConstLabels: Labels{"a": "1", "b": "2"},
-		now:         nowFn,
+		now:         func() time.Time { return now },
 	}).(*counter)
 	counter.Inc()
 	if expected, got := 0.0, math.Float64frombits(counter.valBits); expected != got {
@@ -71,7 +71,7 @@ func TestCounterAdd(t *testing.T) {
 		},
 		Counter: &dto.Counter{
 			Value:            proto.Float64(67.42),
-			CreatedTimestamp: timestamppb.New(nowFn()),
+			CreatedTimestamp: timestamppb.New(now),
 		},
 	}
 	if !proto.Equal(expected, m) {
@@ -146,11 +146,11 @@ func expectPanic(t *testing.T, op func(), errorMsg string) {
 
 func TestCounterAddInf(t *testing.T) {
 	now := time.Now()
-	nowFn := func() time.Time { return now }
+
 	counter := NewCounter(CounterOpts{
 		Name: "test",
 		Help: "test help",
-		now:  nowFn,
+		now:  func() time.Time { return now },
 	}).(*counter)
 
 	counter.Inc()
@@ -183,7 +183,7 @@ func TestCounterAddInf(t *testing.T) {
 	expected := &dto.Metric{
 		Counter: &dto.Counter{
 			Value:            proto.Float64(math.Inf(1)),
-			CreatedTimestamp: timestamppb.New(nowFn()),
+			CreatedTimestamp: timestamppb.New(now),
 		},
 	}
 
@@ -194,11 +194,11 @@ func TestCounterAddInf(t *testing.T) {
 
 func TestCounterAddLarge(t *testing.T) {
 	now := time.Now()
-	nowFn := func() time.Time { return now }
+
 	counter := NewCounter(CounterOpts{
 		Name: "test",
 		Help: "test help",
-		now:  nowFn,
+		now:  func() time.Time { return now },
 	}).(*counter)
 
 	// large overflows the underlying type and should therefore be stored in valBits.
@@ -217,7 +217,7 @@ func TestCounterAddLarge(t *testing.T) {
 	expected := &dto.Metric{
 		Counter: &dto.Counter{
 			Value:            proto.Float64(large),
-			CreatedTimestamp: timestamppb.New(nowFn()),
+			CreatedTimestamp: timestamppb.New(now),
 		},
 	}
 
@@ -228,12 +228,13 @@ func TestCounterAddLarge(t *testing.T) {
 
 func TestCounterAddSmall(t *testing.T) {
 	now := time.Now()
-	nowFn := func() time.Time { return now }
+
 	counter := NewCounter(CounterOpts{
 		Name: "test",
 		Help: "test help",
-		now:  nowFn,
+		now:  func() time.Time { return now },
 	}).(*counter)
+
 	small := 0.000000000001
 	counter.Add(small)
 	if expected, got := small, math.Float64frombits(counter.valBits); expected != got {
@@ -249,7 +250,7 @@ func TestCounterAddSmall(t *testing.T) {
 	expected := &dto.Metric{
 		Counter: &dto.Counter{
 			Value:            proto.Float64(small),
-			CreatedTimestamp: timestamppb.New(nowFn()),
+			CreatedTimestamp: timestamppb.New(now),
 		},
 	}
 
@@ -264,8 +265,8 @@ func TestCounterExemplar(t *testing.T) {
 	counter := NewCounter(CounterOpts{
 		Name: "test",
 		Help: "test help",
+		now:  func() time.Time { return now },
 	}).(*counter)
-	counter.now = func() time.Time { return now }
 
 	ts := timestamppb.New(now)
 	if err := ts.CheckValid(); err != nil {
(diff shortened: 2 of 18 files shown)


examples/exemplars/main.go
```diff
@@ -34,6 +34,14 @@ func main() {
 		Help:    "A histogram of the HTTP request durations in seconds.",
 		Buckets: prometheus.ExponentialBuckets(0.1, 1.5, 5),
 	})
+	requestsTotal := prometheus.NewCounter(prometheus.CounterOpts{
```
TheSpiritXIII, line 37: I'm not sure I understand how the changes in this file relate to the rest of the PR.

prometheus/counter.go
```diff
@@ -121,6 +123,10 @@ func (c *counter) Add(v float64) {
 		panic(errors.New("counter cannot decrease in value"))
 	}
 
+	if c.createdTs == nil {
+		c.createdTs = timestamppb.Now()
```
TheSpiritXIII, line 127: Do we maybe want a `c.timestampPbNow` in a similar fasion to `c.now` for testing?

TheSpiritXIII (changes requested):
Good start! I'm going to have to research `MetricVec` myself...

prometheus/counter.go
ArthurSens: Oh, now I get how this could be tested! I was so confused how I could create a deterministic test with something that depends on machine clock 😭

examples/exemplars/main.go
bwplotka: I would recommend writing a unit tests(s) that will verify those instead. Manual checks won't get us far, ideally we want next commits to be also validated against potential issues with created timestamps.

bwplotka (commented):
Thanks! So far so good, @TheSpiritXIII comments are solid 👍🏽 

Let's make sure that when OpenMetrics text format is used we DON'T create separate series unless instrumentation explicitly enable that in the promhttp config.

examples/exemplars/main.go
ArthurSens: tests added :)

prometheus/value.go
```diff
@@ -110,7 +110,7 @@ func NewConstMetric(desc *Desc, valueType ValueType, value float64, labelValues
 	}
 
 	metric := &dto.Metric{}
-	if err := populateMetric(valueType, value, MakeLabelPairs(desc, labelValues), nil, metric); err != nil {
+	if err := populateMetric(valueType, value, MakeLabelPairs(desc, labelValues), nil, metric, timestamppb.Now()); err != nil {
```
ArthurSens, line 113: I'm very confused about whether a ConstMetric should have a created timestamp or not because I'm not sure how this is supposed to be used 🤔 

Is it supposed to be used as some kind of stateless collector? I.e., would we have a new created timestamp every time we trigger the metric collection?

ArthurSens: Again I'm struggling to implement a test to verify 😵

bwplotka: Ack, let's discuss it on our mtg today (:

bwplotka: ConstMetric is interesting as indeed it's stateless - if we can't track long term creation time, perhaps we should ask called to provide it? Changing it every sample sounds wrong. ConstMetrics often represent external counters (if counter anyway).

prometheus/promhttp/http_test.go
```diff
@@ -331,3 +333,55 @@ func TestHandlerTimeout(t *testing.T) {
 
 	close(c.Block) // To not leak a goroutine.
 }
+
+func TestHandlerWithCreatedTimestamps(t *testing.T) {
+	reg := prometheus.NewRegistry()
+	handlerWithCT := HandlerFor(reg, HandlerOpts{EnableCreatedTimestamps: true})
```
bwplotka, line 339: This test screams for [table test](https://thanos.io/tip/contributing/coding-style-guide.md/#table-tests) using t.Run, to clearly separate test cases (with and without CT) (:  Otherwise it's a little bit chaotic

prometheus/promhttp/http_test.go
```diff
@@ -331,3 +333,55 @@ func TestHandlerTimeout(t *testing.T) {
...
+		Name: "test",
+		Help: "help test",
+	}, []string{"label"})
+
+	reg.MustRegister(counter)
+	counter.WithLabelValues("test1").Inc()
+	counter.WithLabelValues("test2").Add(2)
+
+	request, _ := http.NewRequest("GET", "/", nil)
+	request.Header.Add("Accept", "application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited;q=0.8")
```
bwplotka, line 353: :+1:

prometheus/promhttp/http_test.go
```diff
@@ -331,3 +333,55 @@ func TestHandlerTimeout(t *testing.T) {
...
+	}
+
+	var mfWithCT, mfWithoutCT dto.MetricFamily
+	err = proto.Unmarshal(bodyWithCT, &mfWithCT)
+	if err != nil {
+		t.Errorf("Error unmarshaling response: %s", err.Error())
+	}
+
+	err = proto.Unmarshal(bodyWithoutCT, &mfWithoutCT)
+	if err != nil {
```
bwplotka, line 376:
```suggestion
	if err = proto.Unmarshal(bodyWithoutCT, &mfWithoutCT); err != nil {
```

prometheus/summary.go
```diff
@@ -304,10 +309,19 @@ func (s *summary) Observe(v float64) {
 	if len(s.hotBuf) == cap(s.hotBuf) {
 		s.asyncFlush(now)
 	}
+	if s.createdTs == nil {
+		s.setCreatedTimestamp()
+	}
+}
+
+func (s *summary) setCreatedTimestamp() {
```
bwplotka, line 317: Do we really need this method that is used only once? 

It's a little bit too shallow to me (perhaps adds more complexity not saves complexity).

See: https://thanos.io/tip/contributing/coding-style-guide.md/#keep-the-interface-narrow-avoid-shallow-functions

prometheus/counter.go
```diff
@@ -121,6 +123,10 @@ func (c *counter) Add(v float64) {
 		panic(errors.New("counter cannot decrease in value"))
 	}
 
+	if c.createdTs == nil {
```
bwplotka, line 126: I think this is incorrect. The metric is created on `WithLabelValues` and family. This is leveraged when people are dealing with init problem now. They do:

1. Create metric counterVec (e.g. with label "code")
2. Register it
3. do `metric.WithLabelValues("200")` (no inc or add)

This behaviour is confirmed by [docs](https://github.com/prometheus/client_golang/blob/main/prometheus/counter.go#L215): 

```
// GetMetricWithLabelValues returns the Counter for the given slice of label
// values (same order as the variable labels in Desc). If that combination of
// label values is accessed for the first time, a new Counter is created.
```

(WithLabelValues and family invokes GetMetricWithLabelValues)

It feels we should add timestamp on GetMetricWithLabelValues of MetricVec or deeper/shallower, not when the metric gets incremented? 

Also test for that would be epic (another case)

bwplotka (commented):
Looks good so far, some comments to make it even more correct and readable. Let me know if those make sense! 

Thanks!

prometheus/summary.go
ArthurSens: Yeah good point

prometheus/counter.go
ArthurSens: Oh I didn't know that, thanks for the info :) 
I'll fix and add a test case.

But what about counters that are not part of a CounterVec? The only way we know they are initialized is once they are incremented for the first time, isn't it? Should we keep that conditional in `Add()` and `Inc()` and just add another to `WithLabelValues`?

ArthurSens: Ok, just saw https://github.com/prometheus/client_golang/pull/1044. Looks like the correct implementation :P

prometheus/counter_test.go
```diff
@@ -273,3 +273,52 @@ func TestCounterExemplar(t *testing.T) {
...
 }
+
+func TestCounterCreatedTimestamp(t *testing.T) {
+	now := time.Now()
+
+	counter := NewCounter(CounterOpts{
+		Name: "test",
+		Help: "test help",
+	}).(*counter)
+	counter.now = func() time.Time { return now }
```
bwplotka, line 284: I don't get this test. We should test NewCounter giving us correct CT. Instead we manually override it, which won't test real scenario. Does it make sense? 🤔

ArthurSens: Exactly, that's why I pinged you in slack 😅


`NewCounter` doesn't have access to counter.now, so I can't crest a counter with a mocked now function. I'm struggling to find a good way to test NewCounter an NewCounterVec for that reason

prometheus/histogram.go
```diff
@@ -747,9 +749,10 @@ func (h *histogram) Write(out *dto.Metric) error {
 	waitForCooldown(count, coldCounts)
 
 	his := &dto.Histogram{
-		Bucket:      make([]*dto.Bucket, len(h.upperBounds)),
-		SampleCount: proto.Uint64(count),
-		SampleSum:   proto.Float64(math.Float64frombits(atomic.LoadUint64(&coldCounts.sumBits))),
+		Bucket:           make([]*dto.Bucket, len(h.upperBounds)),
+		SampleCount:      proto.Uint64(count),
+		SampleSum:        proto.Float64(math.Float64frombits(atomic.LoadUint64(&coldCounts.sumBits))),
+		CreatedTimestamp: h.createdTs,
```
dashpole, line 755: @bwplotka how should this handle native histogram resets?  The correct start time for a fixed-bucket histogram is the instrument creation time, but the correct (I think?) start time for native histograms is the last reset time.

ArthurSens: I still haven't had the chance to work with Native Histograms... May I ask why do they need to be reset in the first place?

prometheus/counter_test.go
bwplotka: Yea, we have to do bit of improvisation here. One idea is to have package-level, private variable we will replace in tests and it's always time.Now for production.

prometheus/histogram.go
bwplotka: Yup, didn't think about it yet. I would say scoping this to exclude native histograms (explicitly) could be  one solution. 

Asking @beorn7 is another solution =D

I am diving into native histograms more next week, so will know more.

prometheus/counter_test.go
bwplotka: var timeNow func() time.Time = time.Now

bwplotka:
```
counter := NewCounter(CounterOpts{
		Name: "test",
		Help: "test help",

                 timeNow: func() time.Time { return now }
	})
```

prometheus/counter_test.go
```diff
@@ -273,3 +273,52 @@ func TestCounterExemplar(t *testing.T) {
...
+	ts := timestamppb.New(now)
+	if err := ts.CheckValid(); err != nil {
+		t.Fatal(err)
+	}
+
+	expectedCounter := &dto.Counter{
+		CreatedTimestamp: ts,
+	}
+
+	if counter.createdTs.AsTime() != expectedCounter.CreatedTimestamp.AsTime() {
```
bwplotka, line 321: Can we instead use counter.Write (or some form of it), and check if resulted DTO has CT set.

prometheus/promhttp/http.go
```diff
@@ -379,6 +379,9 @@ type HandlerOpts struct {
 	// NOTE: This feature is experimental and not covered by OpenMetrics or Prometheus
 	// exposition format.
 	ProcessStartTime time.Time
+	// If true, Counters, Summaries and Histograms will be exposed with additional
```
bwplotka, line 382: If we focus on proto, I think we can skip that and always send CT, no? (: (discussed in the mtg sync)

bwplotka (commented):
Looks good so far, module some comments.

prometheus/histogram.go
ArthurSens: > reseting CT on native histogram reset sounds trivial

I know I said that during our sync today, but now that I tried to implement it doesn't look so simple 😅. It could be just me not understanding the concepts of native histograms though.

I can reset functions being called in `histogram.Write()`, does that mean that histograms are reset on every scrape? I feel like I'm missing something here 🤔

ArthurSens:
Thanks for all the reviews so far!

@bwplotka, @TheSpiritXIII, @macxamin, it should be ready for another round of reviews :)

prometheus/counter.go
```diff
@@ -207,6 +213,7 @@ func (v2) NewCounterVec(opts CounterVecOpts) *CounterVec {
 			}
 			result := &counter{desc: desc, labelPairs: MakeLabelPairs(desc, lvs), now: time.Now}
```
bwplotka, line 214: Something is odd here - I expect `now`: being passed from CounterVecOpts, so you can test this code.

prometheus/counter.go
```diff
@@ -90,8 +91,12 @@ func NewCounter(opts CounterOpts) Counter {
 		nil,
 		opts.ConstLabels,
 	)
-	result := &counter{desc: desc, labelPairs: desc.constLabelPairs, now: time.Now}
+	if opts.now == nil {
+		opts.now = time.Now
+	}
+	result := &counter{desc: desc, labelPairs: desc.constLabelPairs, now: opts.now}
```
bwplotka, line 97: Does "counter" type require "now"? We don't need to save it really, do we?

(Same with other metrics)

prometheus/histogram.go
beorn7: Native histograms are reset if they contain too many buckets _and_ a sufficient time has passed since the last reset. All of this is configurable, see https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#HistogramOpts .

I'm fairly certain that we should reset the CT together with the histogram. After all, one of the cited benefits of the CT is to detect resets reliably.

prometheus/counter.go
ArthurSens: Counters still need `now` for exemplars.

prometheus/counter.go
ArthurSens: the `now` function from `counter` is not used to initialize the created timestamp. It is used only for exemplars.

The `now` function from `CounterVecOpts` in the other hand is used further below

ArthurSens:
@kakkoyun, the PR should be basically ready! There is only one test that I can't figure out what it does and how to fix though 😢 

When you have the time, could you take a look at TestHandler and provide any suggestions on how to fix it?

prometheus/counter.go
bwplotka: This is still quite wrong. Do we need multiple now functions?

prometheus/examples_test.go
```diff
@@ -319,6 +319,8 @@ func ExampleSummary() {
 	// internally).
 	metric := &dto.Metric{}
 	temps.Write(metric)
+	// We remove CreatedTimestamp just to make sure the assert below works.
+	metric.Summary.CreatedTimestamp = nil
```
bwplotka, line 323: Alternatively we could mock `now: func() time.Time { return time.Time{} }` and use `0001-01-01 00:00:00 +0000 UTC` string (or similar in proto)

prometheus/metric.go
```diff
@@ -92,6 +92,8 @@ type Opts struct {
 	// machine_role metric). See also
 	// https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels
 	ConstLabels Labels
+
+	now func() time.Time // To mock out time.Now() for testing.
```
bwplotka, line 96:
```suggestion
	now func() time.Time // For testing, all constructors put time.Now() here.
```

prometheus/summary.go
```diff
@@ -145,6 +146,8 @@ type SummaryOpts struct {
 	// is the internal buffer size of the underlying package
 	// "github.com/bmizerany/perks/quantile").
 	BufCap uint32
+
+	now func() time.Time // To mock out time.Now() for testing.
```
bwplotka, line 150: ditto and everywhere else.

prometheus/value.go
```diff
@@ -153,11 +191,17 @@ func populateMetric(
 	labelPairs []*dto.LabelPair,
 	e *dto.Exemplar,
 	m *dto.Metric,
+	createdTimestamp *time.Time,
 ) error {
+	var ct *timestamppb.Timestamp
```
bwplotka, line 196: A little bit more readable code is to add this in m.Counter.CreatedTimestamp ... once m.Counter is created.

bwplotka (commented):
Nice! Almost there. Is test handler still a problem? I can see tests passed?

prometheus/counter.go
bwplotka: Hw, can't find any usages of counter.now anywhere, what am I missing? 🤔

prometheus/counter.go
ArthurSens: I believe we do, since `counter` doesn't have access to the `counterOpts` object that created it.

[Counter still needs to call `now()` when updating exemplars](https://github.com/prometheus/client_golang/blob/51d24f86807c24ce866355d0c82258420f571c17/prometheus/counter.go#L170)

prometheus/counter.go
ArthurSens: Here it is: https://github.com/prometheus/client_golang/blob/51d24f86807c24ce866355d0c82258420f571c17/prometheus/counter.go#L170

😬

prometheus/counter_test.go
```diff
@@ -26,10 +26,12 @@ import (
 )
 
 func TestCounterAdd(t *testing.T) {
+	now := func() time.Time { ct := timestamppb.Timestamp{Seconds: 1, Nanos: 1}; return ct.AsTime() }
```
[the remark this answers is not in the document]
ArthurSens: Good point, not sure what I was thinking here 😅.

prometheus/examples_test.go
ArthurSens: but `now` is an unexported attribute of `SummaryOpts` since the only goal of the now function is to enable good tests (we don't want people setting weird functions in their instrumentation).

examples_test lives in a different package, therefore we can't set a mocked now function.

ArthurSens:
Thanks for all the suggestions again bartek! I agree it looks ugly to have duplicated `now` functions between Counter/CounterOpts and similars, but so far I couldn't find a way to remove it from one of them since it is used by both in different situations. `.*Opts` use it for CreatedTimestamp initiation while `counter|histograms` use them for exemplars


I've also realized that I'm still not updating CreatedTimestamps when Histograms are reset. I'm hoping this could be done in a separate PR though 😅

prometheus/histogram.go
bwplotka: Fixed in PR to this PR https://github.com/ArthurSens/client_golang/pull/1 (:

bwplotka (approved):
It needs rebase, otherwise LGTM! 💪🏽

ArthurSens:
The failing test seems to be unrelated, but I'm not sure 🤨

bwplotka:
It's fixed here https://github.com/prometheus/client_golang/pull/1348, so your PR should be good to go!

PR #1314, opened 2023-07-20
histogram: Enable detection of a native histogram without observations

### LINKED ISSUE #1127: histogram: Find a way of marking a Native Histograms without observations as such
It is a good thing if the format can expose the conventional and the native version of a histogram side by side. But if a native histogram has a zero threshold of zero and zero observations yet, it might easily look like it’s not a native histogram at all, so Prometheus might ingest it assuming it’s a conventional histogram, which gets weird once it has actually received observations and Prometheus starts to ingest it as a native histogram. See [code](https://github.com/prometheus/prometheus/blob/5937b4f5d4f658758f4aa6b9ab803f022e6e5497/model/textparse/protobufparse.go#L468-L482) for details.

This is P2 as explained in the [milestone's description](https://github.com/prometheus/client_golang/milestone/6).

beorn7:
Fixes #1127.

If a native histogram has no observations and a zero threshold of
zero, then it is indistinguishable from a classic histogram. To give
scrapers a hint that it is indeed a native histogram, we add a no-op
span.

This needs follow-up PRs in prometheus/prometheus and
prometheus/client_model.

(The 1st commit is just a cleanup of doc comments and long lines.)

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -428,12 +428,12 @@ type HistogramOpts struct {
 	// a major version bump.
 	NativeHistogramBucketFactor float64
 	// All observations with an absolute value of less or equal
-	// NativeHistogramZeroThreshold are accumulated into a “zero”
-	// bucket. For best results, this should be close to a bucket
-	// boundary. This is usually the case if picking a power of two. If
+	// NativeHistogramZeroThreshold are accumulated into a “zero” bucket.
+	// For best results, this should be close to a bucket boundary. This is
+	// usually the case if picking a power of two. If
 	// NativeHistogramZeroThreshold is left at zero,
-	// DefNativeHistogramZeroThreshold is used as the threshold. To configure
-	// a zero bucket with an actual threshold of zero (i.e. only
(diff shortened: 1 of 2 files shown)


beorn7:
Will create companion PR in prometheus/prometheus once https://github.com/prometheus/prometheus/pull/12583 has been merged.

beorn7:
@bwplotka I'll assign you to the companion PRs, too. They are all quite easy.

beorn7:
@bwplotka any chance you could have a quick look at this and the companion PRs? I suppose this is a light-weight and uncontroversial change.

bwplotka (approved):
Wow, quite deep edge cases. Looks like a decent solution - thanks!

Sorry for lag.

fstab:
@beorn7 does that mean client libraries need to create the no-op span for native-only histograms?

beorn7:
They SHOULD (cf. RFC 2119) create the no-op span if the histogram is meant to be a native histogram (not necessarily native-only) AND it has a zero threshold of zero AND it has received no observations yet.

beorn7:
I will write down all those guidelines soon™.

PR #1323, opened 2023-08-07
Bump client_model

ArthurSens:
Opening this PR to make the review of https://github.com/prometheus/client_golang/pull/1313 easier.

This one is only about bumping the version of client_model in go.mod. This version upgrade is particularly difficult because, in client_model, it updates the version of protoc which changes how proto messages are cast into strings:

1. '<' and '>' characters were replaced with '{' and '}' respectively.
2. The text format is non-deterministic. More information in https://github.com/golang/protobuf/issues/1121
3. proto3 adds a new field to proto.Message(sizeCache) that breaks `reflect.DeepEqual`. Forcing us to use `proto.Equal` now.

### DIFF
--- a/prometheus/utils_test.go
+++ b/prometheus/utils_test.go
@@ -1,0 +1,43 @@
+// Copyright 2018 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package prometheus_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+// printlnNormalized is a helper function to compare proto messages in json format.
+// Without removing brittle, we can't assert that two proto messages in json/text format are equal.
+// Read more in https://github.com/golang/protobuf/issues/1121
+func printlnNormalized(m proto.Message) {
+	fmt.Println(protoToNormalizedJSON(m))
+}
+
+// protoToNormalizedJSON works as printlnNormalized, but returns the string instead of printing.
+func protoToNormalizedJSON(m proto.Message) string {
+	mAsJSON, err := protojson.Marshal(m)
+	if err != nil {
+		panic(err)
+	}
+
+	buffer := new(bytes.Buffer)
+	if err := json.Compact(buffer, mAsJSON); err != nil {
+		panic(err)
+	}
+	return buffer.String()
+}
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.2.0
 	github.com/davecgh/go-spew v1.1.1
 	github.com/json-iterator/go v1.1.12
-	github.com/prometheus/client_model v0.3.0
+	github.com/prometheus/client_model v0.4.0
 	github.com/prometheus/common v0.42.0
 	github.com/prometheus/procfs v0.11.1
 	golang.org/x/sys v0.10.0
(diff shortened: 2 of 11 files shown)


ArthurSens:
The only test I still couldn't figure out is `TestHandler` in prometheus/register_test.go, which compares multiline proto messages. ANy help appreciated 😅

bwplotka:
Thanks!

Can we double check that:

* The newest protoc version has the `{` and that's the decision for future? Perhaps someone used old protoc and we will have to switch back ;p
* Are we sure this does not introduce random char discussed in https://github.com/prometheus/client_golang/pull/1270#issuecomment-1572024397? Since tests are mostly passing, I assume this does not.

What's the problem with multiline? If you need help could give us more details on why it's tricky?

beorn7:
I'm pretty sure this will run into the test flakiness problem because the new protobuf code intentionally randomizes the text output.

ArthurSens:
Ok, as discussed in slack, I started to refactor all tests that were using proto-text for comparison. Now they are using `proto.Equal`.

There are a few situations that I'm not sure how to solve yet though.

A few tests are Examples. Those examples compare what is being printed to what magic comment `// Output:`. For example:

```go
func ExampleMetricVec() {
	infoVec := NewInfoVec(
		"library_version_info",
		"Versions of the libraries used in this binary.",
		[]string{"library", "version"},
	)

	infoVec.WithLabelValues("prometheus/client_golang", "1.7.1")
	infoVec.WithLabelValues("k8s.io/client-go", "0.18.8")

	// Just for demonstration, let's check the state of the InfoVec by
	// registering it with a custom registry and then let it collect the
	// metrics.
	reg.MustRegister(infoVec)

	metricFamilies, err := reg.Gather()
	if err != nil || len(metricFamilies) != 1 {
		panic("unexpected behavior of custom test registry")
	}
	fmt.Println(metricFamilies[0].String())

	// Output:
	// name:"library_version_info" help:"Versions of the libraries used in this binary." type:GAUGE metric:<label:<name:"library" value:"k8s.io/client-go" > label:<name:"version" value:"0.18.8" > gauge:<value:1 > > metric:<label:<name:"library" value:"prometheus/client_golang" > label:<name:"version" value:"1.7.1" > gauge:<value:1 > >
}
```

I'm not familiar with tests functions called `Example.*`... What can we do here?

ArthurSens:
Also, `testHandler` in `prometheus/registry_test.go` is such an extensive test. I'm having a hard time understanding how to fix that one 😓

bwplotka:
This is a very common example on why random character decision is damaging.

I would propose creating "derandomizeProtoString" and remove all instances of double whitespace into one (: and use it in examples (only ideally)

beorn7:
Maybe we could also convert protobuf to the Prometheus text format? In the spirit of github.com/prometheus/client_golang/prometheus/testutil .

SuperQ:
For the error content in the Prometheus text output format I would suggest switching to [protojson](https://pkg.go.dev/google.golang.org/protobuf@v1.31.0/encoding/protojson)  formatter.

ArthurSens:
> For the error content in the Prometheus text output format I would suggest switching to [protojson](https://pkg.go.dev/google.golang.org/protobuf@v1.31.0/encoding/protojson) formatter.

Unfortunately protojson adds the same brittle 😭

prometheus/utils_test.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+)
+
+// removeBrittleJSON is a helper function to compare proto messages in json format.
+// Without removing brittle, we can't assert that two proto messages in json/text format are equal.
+// Read more in https://github.com/golang/protobuf/issues/1121
+func removeBrittleJSON(j []byte) string {
```
bwplotka, line 24: Not sure where you got this "brittle" word from. (: Let's be specifc and call it "normalization" rather?

Also printing errors is odd if nothing guards callers from using this function for other things than printing.

Finally, we have opportunity to reduce boilerplate of marshalling too and make this function generally type safe. We know we want to use it only for printing in example, let's block other use cases, so further devs will be forced to produce golden tests instead (: 

I would propose:

```suggestion
func printlnNormalized(t testing.TB, msg proto.Message) {
  t.Helper()
  
   // marshall, do t.Fatal on error
   
   // compact json
   
   // print to stdout (e.g. fmt.Println)
}
```

WDYT?

bwplotka (commented):
Nice, epic work! 

Thanks for caring about bigger problem here.

I have suggestion for that special function type safety, readability etc. Once that is addressed - LGTM! Thanks!

prometheus/utils_test.go
ArthurSens: > Not sure where you got this "brittle" word from. (: Let's be specifc and call it "normalization" rather?

I guess I'm not 100% fluent in English and sometimes some words just get stuck in my head 😅. Agreed that normalization sounds better
 
> Also printing errors is odd if nothing guards callers from using this function for other things than printing.
>
> Finally, we have opportunity to reduce boilerplate of marshalling too and make this function generally type safe. We know we want to use it only for printing in example, let's block other use cases, so further devs will be forced to produce golden tests instead (:
> 
> I would propose:
> 
> WDYT?

Sounds good! Thanks for the suggestions

ArthurSens: Actually, examples don't have parameter testing.T, so calling 'printlnNormalized(**t testing.TB**...' doesn't seem possible?

I'm panic'ing instead of calling t.Fatal, but let me know if I missed something here

bwplotka: 👍🏽 Good point, panic is fine

bwplotka (approved):
Thanks! Huge amount of work 💪🏽 

LGTM

PR #1324, opened 2023-08-10
histogram docs: Fixed minor nit.

bwplotka:
I was a bit confused by this "inaccuracy" (?), WDYT?


beorn7 (approved):
It wasn't really wrong before, but with your change, _n_ is the same as the internally used `schema`.

bwplotka:
Yea, rather "inconsistency". Thanks!

PR #1327, opened 2023-08-12
Add HandlerWithAllowlist which can control exposed metrics

CatherineF-dev:
After adding HandlerWithAllowlist, cilium or other components can exclude high cardinality metrics instead of disabling metrics collection in a running cluster.  

context: https://github.com/cilium/cilium/issues/27471

I notice this pr introduces a new dependency (dto "github.com/prometheus/client_model/go"), maybe need to move HandlerWithAllowlist to another file.

### DIFF
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -37,6 +37,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -45,6 +46,8 @@ import (
 	"github.com/prometheus/common/expfmt"
 
 	"github.com/prometheus/client_golang/prometheus"
+
+	dto "github.com/prometheus/client_model/go"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


bwplotka (commented):
Thanks!

Normally this could be done on registration, so you only register what you want.

I guess you want to register all and filter afterwards in one place. This has consequences, one would check in the code that certain metric is instrumented but it's actually ignored near the handler.

Still, I think it's fine, however there is no need to add complexity to handler AND add new gatherer. We could simply add new gatherer similar to what you have here. Anyone can then wrap their registry with it.

prometheus/promhttp/http.go
```diff
@@ -381,6 +393,35 @@ type HandlerOpts struct {
 	ProcessStartTime time.Time
 }
 
+type FilteredDefaultGatherer struct {
```
bwplotka, line 396: I would just add `FilteredGatherer` that wraps registry and pass through only matching metrics. Matchers can be done with something like https://github.com/prometheus/client_golang/blob/main/prometheus/collectors/go_collector_latest.go#L93 - regexp. 

This allows us to allow list those metrics you wanted to hardcode here, outside of `FilteredGatherer` - on caller side (:

prometheus/promhttp/http.go
```diff
@@ -381,6 +393,35 @@ type HandlerOpts struct {
...
+	allowedMetrics map[string]bool
+}
+
+func (f *FilteredDefaultGatherer) Gather() ([]*dto.MetricFamily, error) {
+	results := []*dto.MetricFamily{}
+	metrics, err := prometheus.DefaultGatherer.Gather()
```
bwplotka, line 402: Let's wrap explicit gatherer - caller can then pass `prometheus.DefaultGatherer`

CatherineF-dev: Done.

prometheus/promhttp/http.go
```diff
@@ -90,6 +93,13 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 	return HandlerForTransactional(prometheus.ToTransactionalGatherer(reg), opts)
 }
 
+// HandlerWithAllowlist remove metrics whose name can't match with metricNameMatcher
+func HandlerWithAllowlist(metricNameMatcher *regexp.Regexp) http.Handler {
```
bwplotka, line 97: This is nice, but we might want to discuss exact API. e.g. do we want only metric names or also labels? What would be flexible enough?

We think it might be useful to tackle that with OpenMetrics changes and have same API for per scrape or per handler options 🤔

bwplotka (commented):
Is it still needed? I noticed the [CFP/feature request was closed](https://github.com/cilium/cilium/issues/27471).

Nevertheless it feels useful, just we might want to discuss an exact API. Thanks! 🤗

PR #1359, opened 2023-10-04
Configure CI to automerge dependabot PRs

ArthurSens:
Aiming to reduce toil on PR reviews, this workflow will automatically approve and merge dependabot PRs for minor and patch upgrades.

*Warning*: It requires that auto-merge is enabled in the repository settings (I don't have access to enable it)

### DIFF
--- a/.github/workflows/automerge-dependabot.yml
+++ b/.github/workflows/automerge-dependabot.yml
@@ -1,0 +1,23 @@
+name: Dependabot auto-merge
+on: pull_request
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


bwplotka (approved):
LGTM, does it ensure green CI? I think so, right?

ArthurSens:
It respects branch protections, but I've just noticed that there aren't any workflows marked as required...so it would merge without waiting for tests 😓

bwplotka:
We can, but be careful as Prometheus really wanted to avoid that red button when tests are not passing for some reasons. We are fine to change it here I think if we want.

bwplotka:
Should be fixed.

bwplotka:
hmm auto-merge - is it really needed if you use gh client here?

ArthurSens:
yeah, because we still want to wait for the tests to pass and not just merge immediately :P

PR #1367, opened 2023-10-14
histograms: Add timer to reset ASAP after bucket limiting has happened

### LINKED ISSUE #1248: histograms: Reset ASAP after resolution reduction
While analysing real-world production data, I had an insight:

If we hit the configured `NativeHistogramMaxBucketNumber` _before_ the configured `NativeHistogramMinResetDuration` has passed since the last reset, we'll reduce the resolution instead. So far so good.

_However_, once we have reduced the resolution, it is much less likely to ever hit `NativeHistogramMaxBucketNumber` again. Therefore, we might very well stay at the reduced resolution forever (i.e. for the lifetime of the binary). It would be better to reset the histogram immediately after `NativeHistogramMinResetDuration` has passed, because this will reinstate the originally configured resolution.

More context: In the real-world production scenario I have analysed, it is indeed a very rare even that the resolution has to be reduced, mostly triggered by short peaks of more intense traffic. However, because of the reason explained above, it essentially means that a single occurrence will cause the affected binary to expose a reduced-resolution histogram for the rest of its lifetime. Since histograms are often aggregated together, this needlessly reduced the resolution of the aggregated results with quite some blast radius.

beorn7:
Fixes #1248. See issue description for all the details.

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -475,6 +475,9 @@ type HistogramOpts struct {
 
 	// now is for testing purposes, by default it's time.Now.
 	now func() time.Time
+
+	// afterFunc is for testing purposes, by default it's time.AfterFunc.
+	afterFunc func(time.Duration, func()) *time.Timer
 }
 
 // HistogramVecOpts bundles the options to create a HistogramVec metric.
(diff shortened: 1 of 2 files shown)


beorn7:
@bwplotka @ArthurSens could you have a look at this? It fixes a rather annoying bug.

ArthurSens:
Yes, will do! I returned from a 1 week PTO today, I'll schedule some time today or tomorrow to take a look at this PR

PR #1375, opened 2023-11-01
Add example of NewConstMetricWithCreatedTimestamp

### LINKED ISSUE #1364: Ability to override Created Timestamp on new metric or vector instance.
Currently `NewCounter` (and similar for other metrics with created timestamp) and `CounterVec.With..` family use time.Now() to obtain created timestamp. 

This will work in majority of cases, but it will be a problem for exporters which might want to use those "normal" metrics (as opposed to collectors with e.g. const metrics) to report values they don't control (e.g cadvisor showing Linux counters). The counters they might expose can be started back in time, so created timestamp can't be "now" in the moment of exporter creating that metric.

We have to figure out API that would allow us to override time.Now() logic. This problem is a little bit related to https://github.com/prometheus/client_golang/issues/1354 but for production use case of modifying created timestamp. Some yolo ideas (didn't think about this through):

* Don't change API, but update documentation that for those cases one should use `NewConstMetricWithCreatedTimestamp`. Alternatively ask in those cases to reset metric to 0 - both kind of bad UX.
* Add to CounterOpts some `CreatedTimestampFunc` (prone to so many issues) or simply `CreatedTimestamp time.Time`. The latter is nice, but won't work with Vec easily 🤔 
* Add `WithLabelValues...CT` (and 3 others) that will allow specifying labels/label values and time.Time Created timestamp.
* Invest in v2 API which will allow this if compatibility guarantee blocks us.

Acceptance Criteria:
* Users have clear guidance (and possibility) to provide custom created timestamp to their counter based metrics. 
* Override has to be atomic (e.g. we can't just add "SetCreatedTimestamp" method to `Counter` as scrape can happen anytime between registration and `SetCreatedTimestamp).

cc @beorn7 @ArthurSens

ArthurSens:
Fixes #1364

Adds an example for NewConstMetricWithCreatedTimestamp.

### DIFF
--- a/prometheus/examples_test.go
+++ b/prometheus/examples_test.go
@@ -655,3 +655,27 @@ func ExampleNewMetricWithTimestamp() {
 	// Output:
 	// {"gauge":{"value":298.15},"timestampMs":"1257894000012"}
 }
+
+func ExampleNewConstMetricWithCreatedTimestamp() {
+	// Here we have a metric that is reported by an external system.
+	// Besides providing the value, the external system also provides the
+	// timestamp when the metric was created.
+	desc := prometheus.NewDesc(
+		"time_since_epoch_seconds",
+		"Current epoch time in seconds.",
+		nil, nil,
+	)
+
+	timeSinceEpochReportedByExternalSystem := time.Date(2009, time.November, 10, 23, 0, 0, 12345678, time.UTC)
+	epoch := time.Unix(0, 0).UTC()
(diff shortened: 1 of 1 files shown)


ArthurSens:
I think that an example would be more helpful than just an entry in README 🤔, but curious about what others think...

I'm also open to a better example than the one proposed :P

bwplotka (approved):
That's epic, thanks!

PR #1378, opened 2023-11-10
Normalize empty help values in CollectAndCompare

### LINKED ISSUE #1308: testutil.CollectAndCompare fails for metrics with empty help string
**Version:** 1.16.0

While trying to write a test for a custom collector, I discovered that the `testutil.CollectAndCompare` function does not work with metrics that have an empty help string. The issue appears to be a discrepancy between the encoding and decoding of the text format:

1. When encoding metrics from a registry (i.e. the "got" side), the registry [always sets `Help` to a non-nil value](https://github.com/prometheus/client_golang/blob/v1.16.0/prometheus/registry.go#L686).
2. If the `Help` field is non-nil, the [text encoding writes a `HELP` line](https://github.com/prometheus/common/blob/v0.42.0/expfmt/text_create.go#L95)
3. When parsing metrics from text (i.e. the "want" side), the text encoding does not populate the `Help` field if [the line only contains whitespace after the keyword](https://github.com/prometheus/common/blob/v0.42.0/expfmt/text_parse.go#L207-L220).

This example program shows how the comparison fails regardless of whether the `HELP` comment appears in the expected text:

<details>
  <summary>Example</summary>

```go
package main

import (
	"fmt"
	"strings"

	"github.com/prometheus/client_golang/prometheus/testutil"
)

type Collector struct{}

func (c Collector) Describe(ch chan<- *prometheus.Desc) {}

func (c Collector) Collect(ch chan<- prometheus.Metric) {
	ch <- prometheus.MustNewConstMetric(
		prometheus.NewDesc("sample_metric", "", nil, nil),
		prometheus.GaugeValue,
		0,
	)
}

func main() {
	withHelp := `
# HELP sample_metric
# TYPE sample_metric gauge
sample_metric 0
`

	withoutHelp := `
# TYPE sample_metric gauge
sample_metric 0
`
	fmt.Println("--- With HELP ---")
	fmt.Print(testutil.CollectAndCompare(Collector{}, strings.NewReader(withHelp)))

	fmt.Println()
	fmt.Println("--- Without HELP ---")
	fmt.Print(testutil.CollectAndCompare(Collector{}, strings.NewReader(withoutHelp)))
}
```
</details>

My guess is that the test functions should normalize empty help fields before comparing metrics, rather than changing how encoding or decoding works. If that sounds reasonable, I can look at submitting a PR.

bluekeyes:
Due to an inconsistency in the text protocol between encoding and decoding, it was not possible to use the `testutil.CollectAndCompare` function to test metrics with empty help values. To fix this, normalize empty help values from the expected/want side of the test so that they compare correctly with empty values on the actual/got side of the test.

Fixes #1308.

### DIFF
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -47,6 +47,7 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/internal"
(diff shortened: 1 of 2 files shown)


bwplotka (approved):
👍🏽 Thanks!

PR #1383, opened 2023-11-20
This way the import is more readable

manas-rust:
@ArthurSens


bwplotka (approved):

bwplotka:
I thought we had lint for that.. sounds like it does not work anymore?

PR #1387, opened 2023-11-22
change api http.client to interface

tsipo:
Hi there,
The current implementation uses http.Client
The only function which is called for it is `Do(req *http.Request) (*http.Response, error)`
It is desirable to allow other Client implementations, e.g. `github.com/hashicorp/go-retryablehttp/Client` which allows retries on various HTTP status codes like 429 (it has its own `Request` struct but it's easy to wrap it in another struct that will implement the above-mentioned interface). Many commercial solutions implementing Prometheus API (e.g. Grafana Cloud) have rate limiting. Thank you.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -36,14 +36,18 @@ var DefaultRoundTripper http.RoundTripper = &http.Transport{
 	TLSHandshakeTimeout: 10 * time.Second,
 }
 
+type HttpClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
 // Config defines configuration parameters for a new client.
 type Config struct {
 	// The address of the Prometheus to connect to.
 	Address string
 
 	// Client is used by the Client to drive HTTP requests. If not provided,
-	// a new one based on the provided RoundTripper (or DefaultRoundTripper) will be used.
-	Client *http.Client
+	// a new http.Client based on the provided RoundTripper (or DefaultRoundTripper) will be used.
+	Client HttpClient
 
 	// RoundTripper is used by the Client to drive HTTP requests. If not
 	// provided, DefaultRoundTripper will be used.
@@ -57,13 +61,13 @@ func (cfg *Config) roundTripper() http.RoundTripper {
 	return cfg.RoundTripper
 }
 
-func (cfg *Config) client() http.Client {
+func (cfg *Config) client() HttpClient {
 	if cfg.Client == nil {
-		return http.Client{
+		return &http.Client{
 			Transport: cfg.roundTripper(),
 		}
 	}
-	return *cfg.Client
+	return cfg.Client
 }
 
 func (cfg *Config) validate() error {
@@ -101,7 +105,7 @@ func NewClient(cfg Config) (Client, error) {
 
 type httpClient struct {
 	endpoint *url.URL
-	client   http.Client
+	client   HttpClient
 }
 
 func (c *httpClient) URL(ep string, args map[string]string) *url.URL {
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -105,7 +105,7 @@ func TestClientURL(t *testing.T) {
 
 		hclient := &httpClient{
 			endpoint: ep,
-			client:   http.Client{Transport: DefaultRoundTripper},
+			client:   &http.Client{Transport: DefaultRoundTripper},
 		}
 
 		u := hclient.URL(test.endpoint, test.args)


ArthurSens:
Sounds like a valid use case :)

Could you sign your commit to make CI happy?

tsipo:
@ArthurSens Apologies, I just saw the CI complaint and did that :-)

api/client.go
```diff
@@ -57,13 +61,13 @@ func (cfg *Config) roundTripper() http.RoundTripper {
 	return cfg.RoundTripper
 }
 
-func (cfg *Config) client() http.Client {
+func (cfg *Config) client() HttpClient {
```
kakkoyun, line 64:
```suggestion
func (cfg *Config) client() *http.Client {
```
Let's return to the concrete type. So we don't restrict the caller and ourselves from introducing any breaking changes to the interfaces.

https://github.com/golang/go/wiki/CodeReviewComments#interfaces

kakkoyun (changes requested):

api/client.go
tsipo: Hi @kakkoyun,

I am sorry, but I think there is a misunderstanding. The change request:

1. Does not compile - `api/client.go l. 70: Cannot use 'cfg.Client' (type HttpClient) as the type *http.Client`
2. Even if it would compile, it would practically nullify this PR

The PR I issued follows the [Interfaces guidelines](https://github.com/golang/go/wiki/CodeReviewComments#interfaces) to the letter. 
The **producer** package is `net/http` . The **consumer** package is `github.com/prometheus/client_golang/api`.

We need an _HTTP client implementation_ to perform exactly one function:
`Do(req *http.Request) (*http.Response, error)`
and we call it exactly once, inside `func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte, error)`
(note that both `httpClient` and `httpClient.client` are unexported internal structs).

This is exactly why I have defined the interface in the **consumer** package, and made sure that the native `http.Client` struct complies with this interface (of course). The idea is to enable replacing the native client with implementations which have retries (on rate limiting, network errors etc.), something that the native client does not support (PS this cannot be done in a round tripper, you need a client implementation for that - e.g. `github.com/hashicorp/go-retryablehttp/Client`).

The unexported function `(cfg *Config) client()` you requested to change is called exactly once, from the exported function
`func NewClient(cfg Config) (Client, error)`
If we change the signature back to return the concrete type (even if it would compile), we render this PR meaningless as it will mean you CANNOT replace the native client with another implementation, which was the motivation for this entire PR.

Thank you,
Roy

tsipo: See this example code to use with native http.Client and custom HTTP Client implementations:

```
package main

import (
	"fmt"
	rhttp "github.com/hashicorp/go-retryablehttp"
	"net/http"
)

type retryClient struct {
	c *rhttp.Client
}

func (rc *retryClient) Do(req *http.Request) (*http.Response, error) {
	if r, err := rhttp.FromRequest(req); err == nil {
		return rc.c.Do(r)
	} else {
		return nil, err
	}
}

func newRetryClient(rt http.RoundTripper) *retryClient {
	rc := rhttp.NewClient()
	rc.HTTPClient = &http.Client{Transport: rt}
	return &retryClient{c: rc}
}

const (
	address = "http://localhost:9090"
)

func main() {
	cfg1 := api.Config{Address: address, RoundTripper: http.DefaultTransport}
	cfg2 := api.Config{Address: address, Client: http.DefaultClient}
	cfg3 := api.Config{Address: address, Client: newRetryClient(http.DefaultTransport)}
	var cfgs = []api.Config{cfg1, cfg2, cfg3}
	for _, cfg := range cfgs {
		c, err := api.NewClient(cfg)
		fmt.Printf("client: %v, err: %v\n", c, err)
		// ...
	}
}
```

tsipo:
Hello all, can anyone please update on the status of this one? Thanks.

ArthurSens (approved):
LGTM

I believe the only way we can break this is if we end up adding new methods to the HttpClient interface. I don't think we'll need anything more than `Do()` in the near future though 🤔 

I'll keep this open for another week just in case @kakkoyun or @bwplotka want to express a different opinion

tsipo:
> LGTM
> 
> I believe the only way we can break this is if we end up adding new methods to the HttpClient interface. I don't think we'll need anything more than `Do()` in the near future though 🤔
> 
> I'll keep this open for another week just in case @kakkoyun or @bwplotka want to express a different opinion

Thank you @ArthurSens.

FWIW I have tested it recently quite a bit using the commit to `densify-dev/retryable-prometheus-client` (referred to above) and the sample code I have included in my comment on @kakkoyun 's requested change. It works just fine.

In case you need to add new functions to the HttpClient interface you are most welcome to ask for my assistance with it, shouldn't be too difficult :smile: .

api/client.go
```diff
@@ -36,14 +36,18 @@ var DefaultRoundTripper http.RoundTripper = &http.Transport{
 	TLSHandshakeTimeout: 10 * time.Second,
 }
 
+type HttpClient interface {
```
bwplotka, line 39: Why not using simply `http.RoundTripper`?

bwplotka: Also if anything this is not following style guide (should be `HTTPClient`) and does not have commentary as public interface

bwplotka: It's bit different method name, but still.. `RoundTrip(*Request) (*Response, error)`

In theory the api/http packages are experimental so we can change them with breaking changes.

bwplotka: But even with all of this the use case mentioned was possible without this change I believe: 

There is even documentation how to do it here: [github.com/hashicorp/go-retryablehttp/Client](https://github.com/hashicorp/go-retryablehttp?tab=readme-ov-file#getting-a-stdlib-httpclient-with-retries)

bwplotka (commented):
Hm, I think we could revert this. Additional interface is not needed, http.Client has RoundTripper which is replaceable. Plus the mentioned hashicorp project has a converter to http.Client: https://github.com/hashicorp/go-retryablehttp?tab=readme-ov-file#getting-a-stdlib-httpclient-with-retries

bwplotka:
It's easy to miss but it's very common to create a new http (std Go) Client with Roundtripper which has the same signature as Do (just different name):

```
&http.Client{Transport: <round tripper with `RoundTrip(*Request) (*Response, error)` method>}
```

api/client.go
tsipo: Hi @bwplotka , thanks for your comments.

To start with, you cannot get HTTP retries just by replacing the `RoundTripper` (I've tried that...), you need an **alternative client**. The retryable client (like hashicorp's) does more stuff than roundtrips, like draining the response body after each retry etc.

Your suggested solution (sorry, somehow I missed that in their docs) to use their `StandardClient()` [converter](https://github.com/hashicorp/go-retryablehttp?tab=readme-ov-file#getting-a-stdlib-httpclient-with-retries) to the library's `http.Client` works - I have just tried that out and saw retries. So the use-case is possible without this change.

In view of that, you can decide. I still don't think that adding the interface is a bad idea as this is what the Prometheus client requires (only the `Do()` function). It's true that hashicorp's client gives you a converter to the standard client, but you are now relying on that. I'll leave it up to you.

ArthurSens:
Thanks Bartek for taking a deeper look at this, and apologies for also missing that the use case was already possible 😅 

I agree that having multiple ways for doing the same thing confusing, let's revert it 👍

tsipo:
Thank you @bwplotka and @ArthurSens for your comments.
I think the main reason is confusion / ambiguity in the design decisions of Go HTTP client and the way these are used, in general and in this repository too.

It's true - as @bwplotka pointed out - that you can easily wrap a fully-fledged `Client` struct with a `RoundTripper` which delegates its `RoundTrip()` function to the Client's `Do()` function, and then use that `RoundTripper` as a new `Client`'s `Transport`. That's exactly what hashicorp's `StandardClient()` [function](https://github.com/hashicorp/go-retryablehttp?tab=readme-ov-file#getting-a-stdlib-httpclient-with-retries) does BTW.

The question is - and that is maybe a bit philosophical - is that the right thing to do?

The documentation of [RoundTripper.RoundTrip()](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/net/http/client.go) says (l. 118-119):
`	// RoundTrip executes a single HTTP transaction, returning`
`	// a Response for the provided Request.`
Moreover, its [documentation](https://pkg.go.dev/net/http#hdr-Clients_and_Transports) states 
`For control over HTTP client headers, redirect policy, and other settings, create a Client`
and
`For control over proxies, TLS configuration, keep-alives, compression, and other settings, create a Transport`

In practice we see many cases, including in this repo, where a `RoundTripper` is used to set HTTP headers (e.g. various authz headers).
The retry `RoundTripper` takes it one step further. Should it still be considered **a single HTTP transaction**? From the PoV of the user probably yes, as they don't care the `Client` may need to retry till they get the _final_ response. But from the PoV of the `Client` itself? It may be multiple HTTP requests, after each one the response is drained before the retry takes place. Question of transactions and meta-transactions, I guess.

Merry Christmas!

bwplotka:
Oh yea, I agree there might be something to improve in Go, even if it's just documentation. I read the [documentation](https://pkg.go.dev/net/http#hdr-Clients_and_Transports) you referred differently (maybe you are right)--I read it as:

> For control over HTTP client headers, redirect policy, and other settings, create a Client

 Those are the things to supply via request with configurable headers and some specific implementation of `Client` settings. This is what you can configure on client and this is the "source" of settings.

> For control over proxies, TLS configuration, keep-alives, compression, and other settings, create a Transport

This is what you can configure with `specific implementation of transport.`

However... we don't talk about any std mechanisms here -- we are extending the client and transport and here. I think this part of docs does not mention extensibility topic at all, only what's in std types.

The extensibility is done via [RoundTripper](https://pkg.go.dev/net/http#RoundTripper). Historically it was used for much more than in the docs, perhaps overused, but it allowed huge innovation e.g. [client side load balancing](https://github.com/improbable-eng/kedge/blob/master/pkg/kedge/http/lbtransport/transport.go#L104) and other crazy things (e.g. hashicorp retries).  

Plus if you look on http.Client implementation it only checks http.Request request, sets std headers, sets deadlines [1](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/net/http/client.go;l=595 )[2](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/net/http/client.go;l=212), calls RoundTripper and handles response, redirects.

Thus, generally it makes no point to extend client by interfacing `Do(...)`--in fact it will also limit extensibility as RoundTrippers are commonly chained, and http.Client creates an amazing starting point of that chain via Do, so RoundTrippers can be chained. (: 

Merry Christmas!

PR #1389, opened 2023-11-23
gocollector: Add regex option to allow collection of debug runtime metrics

ArthurSens:
At the same time, it also solves the linting problem we're getting on main

### DIFF
--- a/prometheus/collectors/go_collector_go120_test.go
+++ b/prometheus/collectors/go_collector_go120_test.go
@@ -112,3 +112,7 @@ func withSchedulerMetrics() []string {
 		"go_sched_latencies_seconds",
 	})
 }
+
+func withDebugMetrics() []string {
+	return withBaseMetrics([]string{})
+}
--- a/prometheus/collectors/gen_go_collector_set.go
+++ b/prometheus/collectors/gen_go_collector_set.go
@@ -128,7 +128,7 @@ func rm2prom(d metrics.Description) string {
 func groupMetrics(metricsList []string) []metricGroup {
 	var groupedMetrics []metricGroup
 	for _, group := range metricGroups {
-		var matchedMetrics []string
+		matchedMetrics := make([]string, 0)
 		for _, metric := range metricsList {
 			if group.Regex == nil || group.Regex.MatchString(metric) {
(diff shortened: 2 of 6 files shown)


kakkoyun:
> This is a quick easy fix, but maybe we could extend the Go collector to allowList debug metrics instead(Consequentially we'd start using the `withDebugMetrics` method)?

👍

prometheus/collectors/go_collector_go121_test.go
```diff
@@ -149,26 +149,3 @@ func withSchedulerMetrics() []string {
 		"go_threads",
 	}
 }
-
-func withDebugMetrics() []string {
```
kakkoyun, line 153: Let's utilize them, in a unit test rather than removing them.

kakkoyun:
@ArthurSens, shall we revive this?

ArthurSens:
Yeah, it's on my list for quite a while now and I don't even remember what blocked me 😓 

I'll try to get back to this at some point this week

ArthurSens:
Alright, finally found time to come back to this :) 

I had to modify the approach a bit after #1476. I'm now generating the `withDebugMetrics()` function with the script that @SachinSahu431 created. 

There aren't any debug metrics for older debug metrics, so I've updated the script to also generate the function even if empty.

prometheus/collectors/go_collector_go120_test.go
```diff
@@ -33,6 +30,9 @@ func withAllMetrics() []string {
 		"go_cpu_classes_scavenge_total_cpu_seconds_total",
 		"go_cpu_classes_total_cpu_seconds_total",
 		"go_cpu_classes_user_cpu_seconds_total",
+		"go_gc_cycles_automatic_gc_cycles_total",
+		"go_gc_cycles_forced_gc_cycles_total",
+		"go_gc_cycles_total_gc_cycles_total",
```
ArthurSens, line 35: After running the script, the order changed 😬

SachinSahu431: The metrics were not sorted in the previous files

https://github.com/prometheus/client_golang/pull/1483#discussion_r1545534373

kakkoyun (approved):

PR #1392, opened 2023-11-26
Optionally add OM unit 

### LINKED ISSUE #684: Unit support (in v1)
OpenMetrics offers units as part of the metadata. We can easily add them in the various `...Opts` next to `Help`. They would only be used if OpenMetrics is negotiated as the output format (but they will also implicitly show up in protobuf, cf. https://github.com/prometheus/common/issues/214 ).

vesari:
This PR adds support for unit for Open Metrics. It is ready for review, but not ready to merge (see my long comment below). Fixes #684.


ArthurSens:
Hey @vesari , just passing by to ask if there is anything we could do to unblock you 👋

vesari:
> Hey @vesari , just passing by to ask if there is anything we could do to unblock you 👋

@ArthurSens hello! I think I'm just using a Go version which is too new, I plan on fixing that later today. Thanks a lot for checking on me, much appreciated! :)

vesari:
@ArthurSens hello again! Here are a few considerations to give you some insight about what I’ve done here and why.
I had to upgrade Go in the GitHub workflows so that the version is at least 1.21, as this is needed by `common`  0.50.0 which, as you know, contains the necessary implementation for unit. I also changed the `Makefile` and the `go.mod` and `main.go` files in the `tutorial` folder accordingly. I have not touched the `.bingo` file but I guess the version should be upgraded there too.
However, `Tests(1.20)` is still expected to run among the checks, and, as you can see, just hangs, but I suppose that’s because of the repo settings, so I’ll need your help there. 

The lint was failing on a supposedly unused test function unrelated to my changes (`withDebugMetrics`), so I removed it. Let me know if I should instead just have the lint ignore it.

I made the changes needed where an OM format can be passed to `expfmt.NewEncoder`, and also all the others stemming from the field `Unit` being added to the type `prometheus.Desc`. In some cases, I chose to populate this field with the unit which was already included in the metric family name, but I could have also just put an empty string, and  I’m not sure what the preferred behaviour would be or if it doesn’t make any difference. I’m talking especially about the various collectors files, like `prometheus/collector/dbstats_collector.go` , `prometheus/go_collector.go` or `prometheus/go_collector_latest.go` etc.  I’d be grateful if you could advise me on this.

Also, so far I haven’t added any new tests to `promhttp/http_test.go`, considering not even the `EnableOpenMetrics` option has been tested there. Let me know if I should.

Lastly, I happened to push a lot of commits, let me know if you want me to rebase and squash them.

As per usual, many thanks for your time and help! :)

ArthurSens:
Oh sorry for the silence here, somehow I missed your last comment.

> I had to upgrade Go in the GitHub workflows so that the version is at least 1.21, as this is needed by common 0.50.0 which, as you know, contains the necessary implementation for unit.

Yeah, that's a problem actually. We promise support for the 3 latest go versions, so we can't bump our go version to 1.21 until mid-August when go 1.23 is released. I've started [conversations](https://github.com/prometheus/common/issues/603) about how we can improve this situation with common.

> The lint was failing on a supposedly unused test function unrelated to my changes (withDebugMetrics), so I removed it. Let me know if I should instead just have the lint ignore it.

[This is another problem that I've been delaying for way too long](https://github.com/prometheus/client_golang/pull/1389) 😬  From my understanding it doesn't block PRs though. But anyway, I'll try to get back to this soon, no need to remove it now :)

---

Regarding the changes that are related to OpenMetrics units, I haven't taken a deep look yet 😬. I'm mostly concerned with enabling the PR first and will review the changes once we solve all the blockers

vesari:
Thanks a lot for the detailed update! Much appreciated

ArthurSens:
Hi @vesari, all blockers have been resolved finally :) 

The [unused function is now used](https://github.com/prometheus/client_golang/pull/1389), and we can [update prometheus/common without giving up support of go 1.20](https://github.com/prometheus/common/pull/617). If you rebase your PR it should be fine :)

prometheus/promhttp/http.go
```diff
@@ -379,6 +383,9 @@ type HandlerOpts struct {
 	// NOTE: This feature is experimental and not covered by OpenMetrics or Prometheus
 	// exposition format.
 	ProcessStartTime time.Time
+	// If true, WithUnit adds the unit to the encoder options, ultimately allowing the
+	// unit into the final output, provided that EnableOpenMetrics is also true.
+	WithUnit bool
```
ArthurSens, line 388: In https://github.com/prometheus/client_golang/pull/1408, we're changing a bit the structure of the HandlerOpts regarding OpenMetrics functionality. We'll probably want something similar here

vesari:
Awesome! Thank you very much! I'm on it.

prometheus/promhttp/http.go
vesari: @ArthurSens I made some changes to my code accordingly, but I think it would be better for your PR to be merged before mine, as yours has already been approved and also because in yours you explain the rationale behind the new HandlerOpts struct and what options to prefer. I can then take it from there and finish mine. What do you think? :)

vesari:
@ArthurSens , I've marked this as "ready for review", but I'm still assuming [your PR](https://github.com/prometheus/client_golang/pull/1408) should be merged first?

prometheus/promhttp/http.go
```diff
@@ -381,6 +387,14 @@ type HandlerOpts struct {
 	ProcessStartTime time.Time
 }
 
+type OpenMetricsOptions struct {
+	Enable bool
```
bwplotka, line 391: Perhaps we could remove this field and use `OpenMetricsOptions *OpenMetricsOptions` to ensure it's optional? We cannot remove `EnableOpenMetrics` anyway so we could still use that, OR when OpenMetricsOptions is allocated assume OM is enabled? 🤔

prometheus/promhttp/http.go
```diff
@@ -381,6 +387,14 @@ type HandlerOpts struct {
...
+	// If true, EnableUnit adds the unit to the encoder options, ultimately allowing the
```
bwplotka, line 392:
```suggestion
	// EnableUnit adds the unit to the encoder options, ultimately allowing the
```

(this is to comply with "godoc" https://tip.golang.org/doc/comment)

prometheus/desc.go
```diff
@@ -75,8 +77,8 @@ type Desc struct {
 //
 // For constLabels, the label values are constant. Therefore, they are fully
 // specified in the Desc. See the Collector example for a usage pattern.
-func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *Desc {
-	return V2.NewDesc(fqName, help, UnconstrainedLabels(variableLabels), constLabels)
+func NewDesc(fqName, help, unit string, variableLabels []string, constLabels Labels) *Desc {
```
bwplotka, line 80: Unfortunately this is a breaking change. We can't change this function.

Not sure how to solve it. Probably NewDescWithUnit, or changing v2 only would make sense.

bwplotka (changes requested):
Thanks for this!

I don't think this PR is blocked on https://github.com/prometheus/client_golang/pull/1408, but I think I would challenge adding units in this way (or adding units at all).

First of all, we can't change `NewDesc`, it's breaking change.

Then the explicit unit in v2.NewDesc and all metric type constructors' options. Essentially, are we sure asking users to add unit to all their definitions (given majority of the ecosystem metrics have unit in the metric name already) is a good use of their time for the benefit (and what is the benefit actually?)

BTW @vesari or anyone, feel free to challenge me! I might miss some use cases etc! 🤗 

So there are two questions to answer:

1. What's the benefit of having UNIT entry in metadata for OpenMetrics? Essentially **what users would gain by using OpenMetrics with Units metadata entry**? Right now, I see near zero value 🤔 
   a. They would gain "formal compatibility with OpenMetrics 1.0, which means little (no scraper blocks client_golang OpenMetrics at the moment, so what's the point?) and we break compatibility for reasons (we think it would be damaging for Go Prometheus ecosystem to add e.g. _created metrics by default).
   b. Some third party systems would use it. Can we wait for actual use cases come to us? or maybe you have use case?

Wonder, what's your motivation to add units?

2. What's the additional friction for defining metrics. How we can reduce it or automate it. Howe can reduce the things they need to touch, not increase in their code?

Just looking on changing client golang metrics, we have dozens of changes like this:

image

Also:

image

Now, do we think ALL of [118,491](https://github.com/prometheus/client_golang/network/dependents) repositories will check and be happy with extra line to every metric definition that literally duplicate a string. That's like billion code lines in total if everyone would just add one line to every metric with some unit across all our users, JUST in open source. We have to be careful here.

One idea would be to work on some autodetection mechanism that would try to detect units in metric names IF somebody enabled UNITs AND OpenMetrics format 🤔 Ideally we could have some bigger design/proposal for it. However, first question - remains. Do we need UNIT metadata entry?

To sum up, sorry for those challenging questions. Your code is top quality, thanks for massive work on this. It's just a controversial change a bit, and we have large user base, so we have to be more careful here 🤗 DMed you on Slack too, feel bad a bit I didn't respond earlier.

bwplotka:
I realized we already kind of agreed to supporting this... https://github.com/prometheus/client_golang/issues/684 so probably too late to challenge it cc @beorn7 @ArthurSens - do you remember what's the benefit other than "compatibility"? 

The friction comment and breaking change stays. I feel we could do better than adding `Unit` field and asking everybody to duplicate this information 🤔

vesari:
@bwplotka thanks, I'll await @beorn7 and @ArthurSens 's feedback on your questions, as they know the original motivation :)

ArthurSens:
> I realized we already kind of agreed to supporting this... https://github.com/prometheus/client_golang/issues/684 so probably too late to challenge it cc @beorn7 @ArthurSens - do you remember what's the benefit other than "compatibility"?

At this point, I believe the main benefit is only compatibility with OpenMetrics 1.0, yes. I totally understand that OpenMetrics 1.0 has many details that hinge adoption in the Prometheus ecosystem, but I think we can't ignore that OM 1.0 exists and is used out there.

Prometheus already [parses OM units](https://github.com/prometheus/prometheus/blob/3b8b57700c469c7cde84e1d8f9d383cb8fe11ab0/model/textparse/openmetricsparse.go#L144-L149), and [stores it as metadata](https://github.com/prometheus/prometheus/blob/3b8b57700c469c7cde84e1d8f9d383cb8fe11ab0/scrape/scrape.go#L1547-L1549). So my rationale to accept this was to finally close the gap between go applications that wanted this information in Prometheus.

With that said, I admit I'm not aware of how units are used in Prometheus besides easy identification when a human reads PromQL. I was not part of the team when OpenMetrics became a thing, but I had the impression that `units` were added to the data model for a reason, and that work was being done (or planned to) to make use of units when making operations between metrics that were exposed with different units, e.g. `http_duration_seconds + http_duration_miliseconds`. I recognize that I could have done a better job here and questioned my own assumptions, but I can also see how units are important for Prometheus now that we're investing in OTLP compatibility. [The metric data model for OTLP uses Unit as a metric identifier](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#opentelemetry-protocol-data-model), which shows the importance for us to support units in client_golang

.github/workflows/golangci-lint.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: install Go
         uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
-          go-version: 1.20.x
+          go-version: 1.21.x
```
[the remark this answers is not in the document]
ArthurSens: I wonder why this change was even needed 😬, could you explain why you decided to bump the go version here?

prometheus/desc.go
ArthurSens: Agreed, we can't make this change.

I wonder what would be a good option though, maybe something like below?

```go
func (d *Desc) WithUnit(u string) *Desc {
  d.unit = u
  return d
}
```

A new function `NewDescWithUnit` also sounds doable, but honestly, a v2 would be the cleaner option. I'm not sure how viable v2 is though

prometheus/example_metricvec_test.go
```diff
@@ -52,8 +52,8 @@ type InfoVec struct {
 	*prometheus.MetricVec
 }
 
-func NewInfoVec(name, help string, labelNames []string) *InfoVec {
-	desc := prometheus.NewDesc(name, help, labelNames, nil)
+func NewInfoVec(name, help, unit string, labelNames []string) *InfoVec {
+	desc := prometheus.NewDesc(name, help, unit, labelNames, nil)
```
ArthurSens, line 56: Similar problem here, we can't change Public function's signatures

tutorial/whatsup/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/prometheus/client_golang/tutorial
 
-go 1.20
+go 1.21
```
ArthurSens, line 3: I believe we want to keep 1.20 until the next go release

ArthurSens (commented):
If we decide to support units, let's think of alternative that isn't breaking changes to downstream projects :)

tutorial/whatsup/go.mod
vesari: Oops. Yes, absolutely. Just a leftover from when I reverted the changes I had made before `common` was fixed in this sense.

prometheus/promhttp/http.go
vesari: [Here](https://github.com/prometheus/client_golang/pull/1392/files#diff-8dc7897d4d9d5bf1c91a5fe1c931db556eabc1a6458dfce5ddd087b0bc5eec78R163-L167) I was kinda tying the possibility of enabling the optional unit to OM been already negotiated as the output format, but maybe there are better ways to ensure that.

prometheus/desc.go
vesari: Sure, I'll revert the breaking changes and explore the alternatives you both suggested.

.github/workflows/golangci-lint.yml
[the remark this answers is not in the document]
vesari: Same as with the `go.mod` below. Just some change I should have reverted, sorry.

vesari:
> If we decide to support units, let's think of alternative that isn't breaking changes to downstream projects :)

Of course! I'll await your decision whether to support them and, if so, I'll be happy to continue working on it the way you suggested :)

beorn7:
> I realized we already kind of agreed to supporting this... https://github.com/prometheus/client_golang/issues/684 so probably too late to challenge it cc @beorn7 @ArthurSens - do you remember what's the benefit other than "compatibility"?

It's hard for me to give a fair judgement here (because of my difficult history with OM). Having said that, the OM approach to units strikes me as weird. 

I understand the historical Prometheus approach of "including the unit in the metric name" as a pragmatic solution to the lack of proper metadata support in Prometheus. It should be noted that it always has been just a convention.

OM added the unit metadata field, presumably because others asked for it (the OTel data model didn't exist back then, but I assume that the presence of a unit field in OTel comes from the same school of thinking). Now you might think, if OM has it as an explicit field, you don't have to include it in the metric name anymore. But OM did the opposite: If there is a unit field, the same string MUST be in the metric name, otherwise it is not valid OM. Personally, I think it should be _possible_ to include the unit in the metric name (for those that like that for all the situation where you do not want to lookup the metadata even if you could), but given that others want to live in a world where metadata is always available at your fingertips, _enforcing_ it in the metric name feels redundant.

Sadly, OTel took its guidance for Prometheus compatibility from OM, so the way the OTel data model is converted to Prometheus is already following the OM requirements, although there was never a Prometheus consensus about it.

Looping back to the question: How should we deal with the situation in client_golang? Given how far unit support has already seeped into various parts of the wider ecosystem, it's tough to say "this is all f*** up, let's just ignore it".
I think we should make it possible to expose the unit field for those that want it, but I would, for now, not enforce that the unit is also in the metric name (therefore accepting that we might generate invalid OM 1.0, but anticipating that this requirement is lifted in OM 2.0). It could be a note in the doc comment, like: "If you want to create valid OM 1.0, include the unit in the metric name, but this is not enforced." (I guess the OM vision was that you specify the name without the unit in instrumentation, then specify the unit separately (but only once), and then the actually exposed metric name includes the metric?)

How "1st class" adding the unit should be, is up to debate. I definitely wouldn't require all users to now add a unit. It should be an option for those that want it.

WRT to `NewDesc`: I once had an ambitious plan to get rid of (user-visible) `Desc`, see also #222 for issues with `Desc`. Not sure how to add unit without making everything even worse. Maybe `NewDescWithUnit` is OK for now.

vesari:
Thank you all very much for your considerations and suggestions! In order not to enforce that the unit is also in the metric name when the unit field is populated, the logic in `common` has to be modified accordingly, by reverting some of the most recent changes that enforce exactly that. Of course, I'll be happy to do so. After that, I can work on the `NewDescWithUnit` approach for `NewDesc`. Let me know if I can proceed @ArthurSens  @bwplotka  @beorn7 :)

beorn7:
To be clear: @ArthurSens @bwplotka @kakkoyun should make the call here. I'm just providing my personal context.

ArthurSens:
Hi @vesari!

The team met recently and one of the things we discussed was whether we accept OpenMetrics units or not. 
A TL;DR; is: **_We won't accept this until Prometheus has a clear use case for unit metadata._**

---

Reading the OpenMetrics specification, we can read[[1](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily)][[2](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unit)]:
```
A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, HELP, TYPE, and UNIT metadata. Every Metric within a MetricFamily MUST have a unique LabelSet.
.
.
.
Unit specifies MetricFamily units. If non-empty, it MUST be a suffix of the MetricFamily name separated by an underscore. Be aware that further generation rules might make it an infix in the text format.
```

We clearly understand that units are **required** by OpenMetrics 1.0, and we clearly understand that units must be used as metric name suffixes. However, at this point, if we accept this the only benefit would be to say "We're OpenMetrics 1.0 compatible". Being OpenMetrics 1.0 compatible is not a bad thing per se, but Prometheus hasn't used the Unit metadata for anything yet, so it brings Client_golang users little benefit while making our API a tiny bit more complex.

If in the future Prometheus introduces the usage of Unit metadata, e.g. to properly sum metrics with different but compatible units (one in seconds and the other in milliseconds), we'd be happy to resurrect this.

In the name of the team, we apologize for taking this long to make a decision. Your work was quite involved, spreading contributions across multiple repositories and communicating with several stakeholders. We'd be happy to continue working with you if that's also of your interest :)

ArthurSens:
Maybe we could start brainstorming how to use the unit metadata in Prometheus? Although this would be a high-complexity project that I would also need a lot of time to understand properly😬

vesari:
> Hi @vesari!
> 
> The team met recently and one of the things we discussed was whether we accept OpenMetrics units or not. A TL;DR; is: **_We won't accept this until Prometheus has a clear use case for unit metadata._**
> 
> Reading the OpenMetrics specification, we can read[[1](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metricfamily)][[2](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unit)]:
> 
> ```
> A MetricFamily MAY have zero or more Metrics. A MetricFamily MUST have a name, HELP, TYPE, and UNIT metadata. Every Metric within a MetricFamily MUST have a unique LabelSet.
> .
> .
> .
> Unit specifies MetricFamily units. If non-empty, it MUST be a suffix of the MetricFamily name separated by an underscore. Be aware that further generation rules might make it an infix in the text format.
> ```
> 
> We clearly understand that units are **required** by OpenMetrics 1.0, and we clearly understand that units must be used as metric name suffixes. However, at this point, if we accept this the only benefit would be to say "We're OpenMetrics 1.0 compatible". Being OpenMetrics 1.0 compatible is not a bad thing per se, but Prometheus hasn't used the Unit metadata for anything yet, so it brings Client_golang users little benefit while making our API a bit more complex.
> 
> If in the future Prometheus introduces the usage of Unit metadata, e.g. to properly sum metrics with different but compatible units (one in seconds and the other in milliseconds), we'd be happy to resurrect this.
> 
> In the name of the team, we apologize for taking this long to make a decision. Your work was quite involved, spreading contributions across multiple repositories and communicating with several stakeholders. We'd be happy to continue working with you if that's also of your interest :)

@arthursens, no worries at all, I totally understand. Of course I’d be delighted to continue working with the team, thanks for suggesting that!

vesari:
> Maybe we could start brainstorming how to use the unit metadata in Prometheus? Although this would be a high-complexity project that I would also need a lot of time to understand properly😬

Either this or anything the team deem useful. Whatever you think needs attention, I’ll be happy to work on :) Feel free to DM me on Slack if necessary. Thank you!

vesari:
As agreed with @bwplotka, we want to add units eventually, but this needs more thought on how to avoid adding significant boilerplate to all users for that; help wanted, closing for now.

ArthurSens:
We might want to ressurect this PR now that https://github.com/prometheus/proposals/pull/39 adds a purpose to units in Prometheus :)

vesari:
I'll be happy to do so

vesari:
@kakkoyun ignore the request for review for now, as it was sent automatically once I reopened the PR. There's still a lot to catch up on here and work to do.

prometheus/promhttp/http.go
```diff
@@ -215,11 +215,14 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
...
 
 		var enc expfmt.Encoder
+		var encOpts []expfmt.EncoderOption
 		if opts.EnableOpenMetricsTextCreatedSamples {
-			enc = expfmt.NewEncoder(w, contentType, expfmt.WithCreatedLines())
-		} else {
-			enc = expfmt.NewEncoder(w, contentType)
+			encOpts = append(encOpts, expfmt.WithCreatedLines())
+		}
+		if opts.EnableOpenMetricsUnit {
```
bwplotka: Do we need this manual flag?

Why can't we simply add UNIT line if someone enabled OM? Shouldn't this be safe?

bwplotka: We added flag for `EnableOpenMetricsTextCreatedSamples` because we believe CT line is potentially breaking to the ecosystem (lot's of problems). But with UNIT metadata, I'd expect OM compatible parsers to either ignore it or parse. Even TEXT formats should work, because it's just a `# .. ` comment.

The only potential risk is the slight increase of payload size, but this is actually limited by the user who need to manually add `Unit: <....>` fields to their metric definitions, no?

prometheus/promhttp/http.go
```diff
@@ -215,11 +215,14 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
...
+			encOpts = append(encOpts, expfmt.WithUnit())
```
bwplotka: I wonder, does this print UNIT line if UNIT is unknown (empty string?)

Can we ensure this does NOT add UNIT if it's unknown? Just to safe space and limit the risks mentioned in https://github.com/prometheus/client_golang/pull/1392/files#r2580622660

prometheus/promhttp/http.go
```diff
@@ -215,11 +215,14 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		}
 
 		var enc expfmt.Encoder
+		var encOpts []expfmt.EncoderOption
```
bwplotka:
```suggestion
var (
enc expfmt.Encoder
encOpts []expfmt.EncoderOption
)
```

prometheus/promhttp/http.go
```diff
@@ -435,6 +438,9 @@ type HandlerOpts struct {
 	// Prometheus introduced the feature flag 'created-timestamp-zero-ingestion'
 	// in version 2.50.0 to handle this situation.
 	EnableOpenMetricsTextCreatedSamples bool
+	// EnableOpenMetricsUnit enables unit metadata in the OpenMetrics output format.
+	// This is only applicable when OpenMetrics format is negotiated.
+	EnableOpenMetricsUnit bool
```
bwplotka: again, I wonder if we could skip this (reduce complexity and effort for everyone, users already need to manually add `Unit` field).

prometheus/promhttp/http_test.go
```diff
@@ -159,11 +159,11 @@ func TestHandlerErrorHandling(t *testing.T) {
 		t.Fatalf("unexpected number of done invokes, want 0, got %d", got)
 	}
 
-	wantMsg := `error gathering metrics: error collecting metric Desc{fqName: "invalid_metric", help: "not helpful", constLabels: {}, variableLabels: {}}: collect error
+	wantMsg := `error gathering metrics: error collecting metric Desc{fqName: "invalid_metric", help: "not helpful", unit: "", constLabels: {}, variableLabels: {}}: collect error
```
bwplotka, line 162: Can we ensure everywhere we don't print empty string? (and just skip it?). It does not give a lot of info in an empty string form

prometheus/promhttp/http_test.go
```diff
@@ -549,6 +549,40 @@ func TestNegotiateEncodingWriter(t *testing.T) {
 	}
 }
 
+func TestHandlerWithUnit(t *testing.T) {
+	reg := prometheus.NewRegistry()
+
+	counter := prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "http_request_duration_seconds_total",
+		Help: "Total time spent handling HTTP requests.",
+		Unit: "seconds",
```
bwplotka, line 558: Great! Can we add test for empty unit as well? Also another edge case would be a very long string and crazy UTF-8 emoji string (just to show what to expect).

prometheus/counter.go
```diff
@@ -203,6 +204,7 @@ func (v2) NewCounterVec(opts CounterVecOpts) *CounterVec {
 	desc := V2.NewDesc(
 		BuildFQName(opts.Namespace, opts.Subsystem, opts.Name),
 		opts.Help,
+		// opts.Unit, // QUESTION: shall I put this for last?
```
bwplotka: I think here is ok, or even after name.

prometheus/desc.go
```diff
@@ -47,6 +47,8 @@ type Desc struct {
 	fqName string
 	// help provides some helpful information about this metric.
 	help string
+	// unit provides the unit of this metric.
```
bwplotka, line 50: ditto on https://github.com/prometheus/client_golang/pull/1392/files#r2580684116

prometheus/desc.go
```diff
@@ -75,8 +84,8 @@ type Desc struct {
 //
 // For constLabels, the label values are constant. Therefore, they are fully
 // specified in the Desc. See the Collector example for a usage pattern.
-func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *Desc {
-	return V2.NewDesc(fqName, help, UnconstrainedLabels(variableLabels), constLabels)
+func NewDesc(fqName, help string, variableLabels []string, constLabels Labels, unit ...string) *Desc {
```
bwplotka: Nice! I was about to comment we can't break compatibility, but you found a way.

I am not a fan as it's a very one-time hack (: 

Some alternatives:

A. Do `descOpts` variadic parameter so we are flexible.
B. Don't change `NewDesc` but instead use `V2.NewDesc` everywhere internally. This should suffice for all usages, no?

prometheus/desc.go
```diff
@@ -89,11 +98,12 @@ func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *
 //
 // For constLabels, the label values are constant. Therefore, they are fully
 // specified in the Desc. See the Collector example for a usage pattern.
-func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, constLabels Labels) *Desc {
+func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, constLabels Labels, unit ...string) *Desc {
```
bwplotka: Here, I wonder if we don't want to break compatibility. 🤔  It's V2 and I wished we  marked it as experimental somewhere visibility (it was meant to experiment on a new API). However, I bet downstream ecosystem already started using it.

I'd suggest, (see A, B options in https://github.com/prometheus/client_golang/pull/1392/files#r2580668173)
* For `NewDesc` we do (B)
* For V2 we do (A) ?

prometheus/desc.go
```diff
@@ -150,10 +160,11 @@ func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, const
 	d.id = xxh.Sum64()
 	// Sort labelNames so that order doesn't matter for the hash.
 	sort.Strings(labelNames)
-	// Now hash together (in this order) the help string and the sorted
+	// Now hash together (in this order) the help string, the unit string and the sorted
 	// label names.
 	xxh.Reset()
 	xxh.WriteString(help)
+	xxh.WriteString(optionalUnitValue(unit...))
```
bwplotka: let's eval unit as soon as we have it from `descV2Opts`  or something

prometheus/desc.go
```diff
@@ -202,9 +213,10 @@ func (d *Desc) String() string {
 		}
 	}
 	return fmt.Sprintf(
-		"Desc{fqName: %q, help: %q, constLabels: {%s}, variableLabels: {%s}}",
+		"Desc{fqName: %q, help: %q, unit: %q, constLabels: {%s}, variableLabels: {%s}}",
```
bwplotka, line 232: ditto, optionally print it?

prometheus/go_collector_latest.go
```diff
@@ -210,16 +210,19 @@ func NewGoCollector(opts ...func(o *internal.GoCollectorOptions)) Collector {
 		sampleBuf = append(sampleBuf, metrics.Sample{Name: d.Name})
 		sampleMap[d.Name] = &sampleBuf[len(sampleBuf)-1]
 
+		// Extract unit from the runtime/metrics name (e.g., "/gc/heap/allocs:bytes" -> "bytes")
```
bwplotka, line 213: That's nice!

prometheus/metric.go
```diff
@@ -81,6 +81,9 @@ type Opts struct {
 	// string.
 	Help string
 
+	// Unit provides the unit of this metric.
```
bwplotka: Shall we mention https://prometheus.io/docs/specs/om (this link always brings latest OM).
```suggestion
	// Unit provides the unit of this metric as per https://prometheus.io/docs/specs/om
```

vesari:
> This is amazing. Solid work, thanks for finding way to not break compatibility.
> 
> Added a few suggestions and ideas, hope those help! Overall this goes in a good direction 👍🏽

Thank you very much, I'm working on addressing your requests for changes and your observations. I'll get back to you once I'm done :)

prometheus/promhttp/http.go
vesari: I see your point. However, as things stand, the flag is necessary because `prometheus/common`'s `expfmt` encoder requires explicit opt-in via `WithUnit()` to output units. Without it, [units are silently ignored even if set in the MetricFamily](https://github.com/prometheus/common/blob/d80d8544703e59a080a204b6f7429ac6561fb24f/expfmt/openmetrics_create.go#L54). I think this was to avoid potential changes to metric names to happen automatically. In fact, [because of this logic](https://github.com/prometheus/common/blob/d80d8544703e59a080a204b6f7429ac6561fb24f/expfmt/openmetrics_create.go#L154) 

  if we:
  - Opt in with `WithUnit()`
  - Have a metric `foo_total` with Unit: `"seconds"`
  - The metric name doesn't already end with `_seconds`

The metric name becomes `foo_seconds_total`.  I believe the consensus back then was this should be the result of a very explicit decision by the user. 
Of course, if  - for example - we decide we’re allowed not to enforce the unit suffix anymore or that we deem acceptable to silently add unit suffixes, then the flag becomes superfluous, and can be removed as soon as the related code in common gets modified accordingly. I’ll be happy to do that myself, if consensus is ever reached in that sense.

vesari:
This is now ready for review :)

prometheus/promhttp/http.go
bwplotka: I think we agree it's a bit toxic to change name depending on the unit, even though it's what OM 1.0 kind of requires (but there tons of other way to handle this e.g. return error on registration e.g. using pedantic registry).


To proceed I think we need to simplify this for users and make sure there's a mode in expfmt that do not change metric names, perhaps even to _total. Then we also need to reduce risk this mode will be not alter.

The problem with adding a feature flag "for now", it's because it's for forever (v1) and it's also adding a flag with something that we don't recommend and can have consequences. So let's try to alter common then?

vesari: Sure thing. Opened the corresponding [PR on common](https://github.com/prometheus/common/pull/877) today.

prometheus/desc.go
```diff
@@ -150,10 +166,11 @@ func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, const
 	d.id = xxh.Sum64()
 	// Sort labelNames so that order doesn't matter for the hash.
 	sort.Strings(labelNames)
-	// Now hash together (in this order) the help string and the sorted
+	// Now hash together (in this order) the help string, the unit string and the sorted
 	// label names.
 	xxh.Reset()
 	xxh.WriteString(help)
+	xxh.WriteString(unit)
```
bwplotka:
```suggestion
	xxh.WriteString(help)
	xxh.Write(separatorByteSlice)
	xxh.WriteString(d.unit)
```

prometheus/desc.go
```diff
@@ -66,6 +68,15 @@ type Desc struct {
 	err error
 }
 
+type DescOpt func(*Desc)
```
bwplotka, line 72: Probably needs commentary (public structure)

prometheus/desc.go
```diff
@@ -89,12 +100,17 @@ func NewDesc(fqName, help string, variableLabels []string, constLabels Labels) *
...
 	d := &Desc{
 		fqName:         fqName,
 		help:           help,
 		variableLabels: variableLabels.compile(),
 	}
+
+	for _, opt := range opts {
+		opt(d)
+	}
+	unit := d.unit
```
bwplotka: I'd inline this

prometheus/desc.go
```diff
@@ -210,10 +227,15 @@ func (d *Desc) String() string {
 			}
 		}
 	}
+	unitStr := ""
+	if d.unit != "" {
+		unitStr = fmt.Sprintf(", unit: %q", d.unit)
```
bwplotka: I'd add this inline and show unit is empty? We expanded the data structure so we can expand the error too

prometheus/example_metricvec_test.go
```diff
@@ -52,8 +52,12 @@ type InfoVec struct {
 	*prometheus.MetricVec
 }
 
-func NewInfoVec(name, help string, labelNames []string) *InfoVec {
-	desc := prometheus.NewDesc(name, help, labelNames, nil)
+func NewInfoVec(name, help string, labelNames []string, unit ...string) *InfoVec {
```
bwplotka: Info likely does not have unit?

prometheus/go_collector_latest.go
```diff
@@ -210,16 +210,19 @@ func NewGoCollector(opts ...func(o *internal.GoCollectorOptions)) Collector {
...
+		unit := d.Name[strings.IndexRune(d.Name, ':')+1:]
```
bwplotka: As found in https://gist.github.com/bwplotka/a38c33a74fad88a46c4623e924fa58c8#file-prometheusclient_golang1392-md, I think indeed we should handle the case of no unit here gracefully (-1 index response):


```
> Unsafe Substring Slice during Runtime Metric Unit Extraction

File with exact line number: prometheus/go_collector_latest.go:L213
Link to a file line on GitHub PR: https://github.com/prometheus/client_golang/pull/1392/files#diff-8360f0c0ae2f0d98fb9ecdf37119d6d333dc17ec23f0547055c1ddfdf4ed0759R213
Problem: The unit is extracted using d.Name[strings.IndexRune(d.Name, ':')+1:]. If a runtime metric is introduced without a :, strings.IndexRune evaluates to -1 resulting in -1+1 = 0. The unit silently turns into the entire metric name, bypassing obvious fault detection. (Note: this extraction assumption existed before the PR for histograms, but was generalized here).
Suggestion: Consider checking if strings.ContainsRune(d.Name, ':') before slicing to future-proof the collector against misconfigured runtime metrics.
````

**Comment by vesari** [comment]
Ready for review.

File: prometheus/promhttp/http.go [hunk 28]
```diff
@@ -226,12 +226,14 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
...
+			enc     expfmt.Encoder
+			encOpts []expfmt.EncoderOption
+		)
 		if opts.EnableOpenMetricsTextCreatedSamples {
-			enc = expfmt.NewEncoder(w, contentType, expfmt.WithCreatedLines())
-		} else {
-			enc = expfmt.NewEncoder(w, contentType)
+			encOpts = append(encOpts, expfmt.WithCreatedLines())
 		}
+		enc = expfmt.NewEncoder(w, contentType, encOpts...)
```
bwplotka, line 236: Unrelated change, but if we want this we could simplify further: 

```suggestion
		var 	encOpts []expfmt.EncoderOption
		if opts.EnableOpenMetricsTextCreatedSamples {
			encOpts = append(encOpts, expfmt.WithCreatedLines())
		}
		enc := expfmt.NewEncoder(w, contentType, encOpts...)
```

prometheus/promhttp/http_test.go
bwplotka: Do you plan to address this?

prometheus/promhttp/http_test.go
bwplotka: ping

prometheus/promhttp/http_test.go
vesari: I had originally addressed your Dec 2 comment asking to skip printing the unit when it's an empty string. But then in your March 9 review on desc.go:232, you asked to show the unit inline even when empty ("I'd add this inline and show unit is empty? We expanded the data structure so we can expand the error too"). The current `unit: ""` in these test expectations follows that guidance. Maybe I misunderstood something?

prometheus/promhttp/http_test.go
vesari: I think I already addressed this? TestHandlerWithEmptyUnit (line 586), TestHandlerWithLongUnit (line 615),and TestHandlerWithEmojiUnit (line 648) are all there. Are they not what you expected?

vesari:
@bwplotka  About your comment: "do you plan to address this?"
I had originally addressed your Dec 2 comment asking to skip printing the unit when it's an empty string. But then in your March 9 review on desc.go:232, you asked to show the unit inline even when empty ("I'd add this inline and show unit is empty? We expanded the data structure so we can expand the error too"). The current unit: "" in these test expectations follows that guidance. Maybe I misunderstood something?

vesari:
@bwplotka about your "ping" comment: 
I think I already addressed this? TestHandlerWithEmptyUnit (line 586), TestHandlerWithLongUnit (line 615),and TestHandlerWithEmojiUnit (line 648) are all there. Are they not what you expected?

prometheus/promhttp/http_test.go
bwplotka: No, you are right. I am (and was) referring to desc.String. Error is ok to have it, super useful, but it's literally same thing.

Sorry!

prometheus/promhttp/http_test.go
bwplotka: Ack, I just saw it's unresolved, so wanted to double check!

TestHandlerWithEmptyUnit looks great!

prometheus/promhttp/http.go
bwplotka: Let's do in separate PR

PR #1408, opened 2023-12-12
Optionally print OM created lines

ArthurSens:
This PR adds an extra option to print _created lines when OpenMetrics is the chosen exposition format. It's implemented following the conditional option added in https://github.com/prometheus/common/pull/504.

### DIFF
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -207,7 +207,13 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		if encodingHeader != string(Identity) {
 			rsp.Header().Set(contentEncodingHeader, encodingHeader)
 		}
-		enc := expfmt.NewEncoder(w, contentType)
+
+		var enc expfmt.Encoder
+		if opts.EnableOpenMetricsTextCreatedSamples {
+			enc = expfmt.NewEncoder(w, contentType, expfmt.WithCreatedLines())
+		} else {
+			enc = expfmt.NewEncoder(w, contentType)
+		}
 
 		// handleError handles the error according to opts.ErrorHandling
 		// and returns true if we have to abort after the handling.
@@ -408,6 +414,21 @@ type HandlerOpts struct {
 	// (which changes the identity of the resulting series on the Prometheus
 	// server).
 	EnableOpenMetrics bool
+	// EnableOpenMetricsTextCreatedSamples specifies if this handler should add, extra, synthetic
+	// Created Timestamps for counters, histograms and summaries, which for the current
+	// version of OpenMetrics are defined as extra series with the same name and "_created"
+	// suffix. See also the OpenMetrics specification for more details
+	// https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#counter-1
+	//
+	// Created timestamps are used to improve the accuracy of reset detection,
+	// but the way it's designed in OpenMetrics 1.0 it also dramatically increases cardinality
+	// if the scraper does not handle those metrics correctly (converting to created timestamp
+	// instead of leaving those series as-is). New OpenMetrics versions might improve
+	// this situation.
+	//
+	// Prometheus introduced the feature flag 'created-timestamp-zero-ingestion'
+	// in version 2.50.0 to handle this situation.
+	EnableOpenMetricsTextCreatedSamples bool
 	// ProcessStartTime allows setting process start timevalue that will be exposed
 	// with "Process-Start-Time-Unix" response header along with the metrics
 	// payload. This allow callers to have efficient transformations to cumulative
--- a/examples/createdtimestamps/main.go
+++ b/examples/createdtimestamps/main.go
@@ -1,0 +1,60 @@
+// Copyright 2022 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// A simple example of how to exposed created timestamps in OpenMetrics format.
+
+package main
+
+import (
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+func main() {
+	requestDurations := prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name:    "http_request_duration_seconds",
+		Help:    "A histogram of the HTTP request durations in seconds.",
+		Buckets: prometheus.ExponentialBuckets(0.1, 1.5, 5),
+	})
+
+	// Create non-global registry.
+	registry := prometheus.NewRegistry()
+	registry.MustRegister(
+		requestDurations,
+	)
+
+	go func() {
+		for {
+			// Record fictional latency.
+			now := time.Now()
+			requestDurations.Observe(time.Since(now).Seconds())
+			time.Sleep(600 * time.Millisecond)
+		}
+	}()
+
+	// Expose /metrics HTTP endpoint using the created custom registry.
+	http.Handle(
+		"/metrics", promhttp.HandlerFor(
+			registry,
+			promhttp.HandlerOpts{
+				EnableOpenMetrics:                   true,
+				EnableOpenMetricsTextCreatedSamples: true,
+			}),
+	)
+	// To test: curl -H 'Accept: application/openmetrics-text' localhost:8080/metrics
+	log.Fatalln(http.ListenAndServe(":8080", nil))
+}


prometheus/promhttp/http.go
```diff
@@ -371,6 +376,8 @@ type HandlerOpts struct {
 	// (which changes the identity of the resulting series on the Prometheus
 	// server).
 	EnableOpenMetrics bool
+	// TODO: document this properly
+	EnableOpenMetricsCreatedLines bool
```
bwplotka, line 380:
```suggestion
	EnableOpenMetricsCreatedMetrics bool
```

Nice, let's do extensive docs here on how to use it and what are the consequences.

bwplotka: Also I still think this is dangerous option due to name conflicts and size of scrape 🙈

ArthurSens: > Also I still think this is dangerous option due to name conflicts and size of scrape 🙈

Not sure what can we do about that besides documenting it properly 😬

prometheus/promhttp/http.go
```diff
@@ -371,6 +376,21 @@ type HandlerOpts struct {
...
+	// but be aware that it also increases the size of the payload.
+	//
+	// Prometheus introduced the feature flag 'created-timestamp-zero-ingestion'
+	// in version 2.50.0, but with support limited to the Prometheus protobuf
+	// format. Starting in Prometheus XXXX, the feature flag will be extended
+	// to the OpenMetrics text format. If using Prometheus XXXX or later, it
+	// is recommended to enable the feature flag in Prometheus, otherwise enabling
+	// _created lines will result in increased cardinality and no improvements
+	// in reset detection.
+	EnableOpenMetricsCreatedMetrics bool
```
bwplotka, line 393: Nice. I wonder if we want to keep this or perhaps deprecate "EnableOpenMetrics" and have "OpenMetrics: *OpenMetricOpts" to add more stuff. We might need this to control:

* Created metrics in text
* Version
* ?

just an idea to not have trillion bool flags.

bwplotka (approved):
Thanks! One minor nit, otherwise well done!

prometheus/promhttp/http.go
ArthurSens: Good idea 👍

prometheus/promhttp/http.go
```diff
@@ -381,6 +391,34 @@ type HandlerOpts struct {
...
+	// If true, the experimental OpenMetrics encoding is added to the
+	// possible options during content negotiation. Note that Prometheus
+	// 2.5.0+ will negotiate OpenMetrics as first priority. OpenMetrics is
+	// the only way to transmit exemplars. However, the move to OpenMetrics
+	// is not completely transparent. Most notably, the values of "quantile"
+	// labels of Summaries and "le" labels of Histograms are formatted with
+	// a trailing ".0" if they would otherwise look like integer numbers
+	// (which changes the identity of the resulting series on the Prometheus
+	// server).
+	EnableOpenMetrics bool
```
bwplotka, line 404:
```suggestion
	Enable bool
```

prometheus/promhttp/http.go
```diff
@@ -381,6 +391,34 @@ type HandlerOpts struct {
 	ProcessStartTime time.Time
 }
 
+type OpenMetricsOptions struct {
+	// If true, the experimental OpenMetrics encoding is added to the
```
bwplotka, line 395:
```suggestion
	// Enable specifies if the experimental OpenMetrics encoding is added to the
```

prometheus/promhttp/http.go
```diff
@@ -370,7 +375,12 @@ type HandlerOpts struct {
 	// a trailing ".0" if they would otherwise look like integer numbers
 	// (which changes the identity of the resulting series on the Prometheus
 	// server).
```
bwplotka, line 377: Let's remove all the commentary above and leave only Deprecated note. Users can navigate to that field for the description.

prometheus/promhttp/http.go
```diff
@@ -370,7 +375,12 @@ type HandlerOpts struct {
...
+	//
+	// Deprecated: Use OpenMetricsOptions.EnableOpenMetrics instead.
```
bwplotka, line 379:
```suggestion
	// Deprecated: Use OpenMetricsOptions.Enable instead.
```

prometheus/promhttp/http.go
```diff
@@ -381,6 +391,34 @@ type HandlerOpts struct {
...
+type OpenMetricsOptions struct {
+	// If true, the experimental OpenMetrics encoding is added to the
+	// possible options during content negotiation. Note that Prometheus
+	// 2.5.0+ will negotiate OpenMetrics as first priority. OpenMetrics is
+	// the only way to transmit exemplars. However, the move to OpenMetrics
+	// is not completely transparent. Most notably, the values of "quantile"
+	// labels of Summaries and "le" labels of Histograms are formatted with
+	// a trailing ".0" if they would otherwise look like integer numbers
+	// (which changes the identity of the resulting series on the Prometheus
+	// server).
```
bwplotka, line 403:
```suggestion
	// possible options during content negotiation. 
	//
	// Note that Prometheus 2.5.0+ might negotiate OpenMetrics Text format 
	// as first priority unless user uses custom scrape protocol prioritization or
	// histograms feature is enabled (then Prometheus proto format is prioritized,
	// which client_golang supports).
	//
	// The main reason for enabling OpenMetrics Text was to enable exemplars,
	// but Prometheus proto has that as feature as well.
	// 
	// However, the move to OpenMetrics is not completely transparent. Most notably,
	// the values of "quantile" labels of Summaries and "le" labels of Histograms are
	// formatted with a trailing ".0" if they would otherwise look like integer numbers
	// (which changes the identity of the resulting series on the Prometheus
	// server).
	//
	// See other options in OpenMetricsOptions to learn how to enable some special
	// features e.g. potentially dangerous created timestamp series.
```

prometheus/promhttp/http.go
```diff
@@ -381,6 +391,34 @@ type HandlerOpts struct {
...
+	// but be aware that it also increases the size of the payload.
+	//
+	// Prometheus introduced the feature flag 'created-timestamp-zero-ingestion'
+	// in version 2.50.0, but with support limited to the Prometheus protobuf
+	// format. Starting in Prometheus XXXX, the feature flag will be extended
+	// to the OpenMetrics text format. If using Prometheus XXXX or later, it
+	// is recommended to enable the feature flag in Prometheus, otherwise enabling
+	// _created lines will result in increased cardinality and no improvements
+	// in reset detection.
+	EnableOpenMetricsCreatedMetrics bool
```
bwplotka, line 419:
```suggestion
	EnableCreatedSeries bool
```

prometheus/promhttp/http.go
```diff
@@ -381,6 +391,34 @@ type HandlerOpts struct {
...
+	// Created timestamps are used to improve the accuracy of reset detection,
+	// but be aware that it also increases the size of the payload.
+	//
+	// Prometheus introduced the feature flag 'created-timestamp-zero-ingestion'
+	// in version 2.50.0, but with support limited to the Prometheus protobuf
+	// format. Starting in Prometheus XXXX, the feature flag will be extended
+	// to the OpenMetrics text format. If using Prometheus XXXX or later, it
+	// is recommended to enable the feature flag in Prometheus, otherwise enabling
+	// _created lines will result in increased cardinality and no improvements
+	// in reset detection.
```
bwplotka, line 418:
```suggestion
	// EnableCreatedSeries specifies if this handler should add, extra, synthetic
	// '_created' series for counters, histograms and summaries, as defined in
	// the OpenMetrics specification (see
	// https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#counter-1).
	// 
	// Created timestamps are used to improve the accuracy of reset detection,
	// but the way it's designed in OpenMetrics 1.0 it also dramatically increases cardinality
	// if the scraper does not handle those metrics correctly (converting to created timestamp
	// instead of leaving those series as-is). New OpenMetrics versions might improve
	// this situation.
	//
	// Prometheus introduced the feature flag 'created-timestamp-zero-ingestion'
	// in version 2.50.0, but only for the Prometheus protobuf format. Starting in 
	// future Prometheus version, the feature flag will be extended to the OpenMetrics
	// text format, thus safe to be enabled to improve accuracy of counters in Prometheus.
```

bwplotka (approved):
Epic, proposed some improvements, thanks!

prometheus/promhttp/http.go
ArthurSens:
```
	// The main reason for enabling OpenMetrics Text was to enable exemplars,
	// but Prometheus proto has that as feature as well.
```

this part seems a bit out of place, not adding much to the comment 🤔, is it ok to not add this part?

prometheus/promhttp/http.go
ArthurSens: Hmmmmm, future versions of OM might not expose them as a series. What do you think about `EnableCreatedTimestamps`?

ArthurSens:
We need prometheus/common v0.49.0 for this feature, but this version of prometheus/common requires go 1.21.

Therefore, the PR is blocked until we drop support for go 1.20

kakkoyun (approved):
LGTM

ArthurSens:
After quite some work to unblock this, this is finally ready :)

Do you want to review again or should I just merge?

ArthurSens:
Now that https://github.com/prometheus/prometheus/pull/14356 was merged, should we revive this? :)

ArthurSens:
PR rebased on main. Ready for another review just in case @kakkoyun @bwplotka @vesari

ArthurSens:
Since we have approvals from the past, I'll take the liberty to merge this PR next week :)

prometheus/promhttp/http.go
```diff
@@ -418,6 +419,43 @@ type HandlerOpts struct {
 	ProcessStartTime time.Time
 }
 
+type OpenMetricsOptions struct {
```
bwplotka, line 422: Hm.. want to play with different ideas for this options first 🤔 Even keeping `EnableOpenMetrics` and then `OpenMetricsOptions.AddV1CreatedSeries` might a better choice? 🤔

bwplotka (changes requested):
Let's timebox this, but wanted to try last more time to consider an improvement to the config structure, otherwise LGTM!

prometheus/promhttp/http.go
ArthurSens: > Hm.. want to play with different ideas for this options first 🤔 Even keeping `EnableOpenMetrics` and then `OpenMetricsOptions.AddV1CreatedSeries` might a better choice? 🤔

Interesting! I assume that this suggestion comes from not liking the idea of mentioning that "Created" is a "Series", this is something we'd like to change in the next version to something closer to metadata.

What do you think of `EnableCreatedTimestamps` instead? 

I think mentioning versions on fields is not the best, in the future I can see a `version` field that controls how each feature is exposed instead of `AddV1Created..`, `AddV2Created...` and then handling mutually exclusiveness

Maniktherana (commented):
should we add summaries and counters in this example too?

prometheus/promhttp/http.go
ArthurSens: Oh lol, I just realized I suggested what's already in the code 🤦

prometheus/promhttp/http.go
bwplotka: If they are not expanded as series they are not a problem and will be fine by default (sorry for missing this commment)

prometheus/promhttp/http.go
bwplotka: I understand the confusion, here is why: https://github.com/prometheus/client_golang/pull/1408#discussion_r1774821420

We will likely not have "enable X Y Z" unless protocol will make a challenging decision like with CT being a series. So ideally we only have this "unsafe 1.0 Ct as series" flag for now, but do CT by default if they are safe to use (they don't increase 50-30% of Prometheus memory for no reason). (:

bwplotka: We might have some more flags in future for units too though, so some plan for those would be nice. For CT as series that an OpenMetrics option, but for units (e.g. AutogenerateUnits)... it's kind of general option? maybe not even in promhttp but in registry or something?

Since we now little and we already make it complex I would vote for keeping `EnableOpenMetrics` and adding `EmitOpenMetricsTextCreatedSamples bool` would be perfect for now. Samples not series as this is how it's referenced in OpenMetrics  https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#counter-1

Simple, and effective, then we probably should fail or ignore this if EnableOpenMetrics is not true.

WDYT

vesari: On a loosely relate note, I would agree on a general option for units. Happy to resume the work on that one whenever (if ever) needed.

ArthurSens:
> should we add summaries and counters in the example too?

So far all our examples just add one or two metrics, not sure if that's needed 🤷

prometheus/promhttp/http.go
ArthurSens: I'm still confused here haha

`EmitOpenMetricsTextCreatedSamples bool`, this means there's no distinction between OM v1 and v2, do I understand that correctly? If I'm following you correctly, if we introduce this flag now with `false` as the default value, we won't be able to change it once OM v2 is out, right?

bwplotka: That's exactly why I call this option "CreatedSamples", so the not ideal _created lines. Once in V2 we have that in metadata or as annotation, even with this option false we will emit CT semantics by default IMO. 

This the same as with exemplars - enable open metrics, enables exemplars, no questions asked, because they were designed to NOT spam TSDB with odd series (:

Perhaps let's chat offline if this is still not clear/I am missing something? (:

ArthurSens:
@bwplotka should be ready for review again!

prometheus/promhttp/http.go
```diff
@@ -408,6 +414,21 @@ type HandlerOpts struct {
...
+	//
+	// Created timestamps are used to improve the accuracy of reset detection,
+	// but the way it's designed in OpenMetrics 1.0 it also dramatically increases cardinality
+	// if the scraper does not handle those metrics correctly (converting to created timestamp
+	// instead of leaving those series as-is). New OpenMetrics versions might improve
+	// this situation.
+	//
+	// Prometheus introduced the feature flag 'created-timestamp-zero-ingestion'
+	// in version 2.50.0 to handle this situation.
+	EmitOpenMetricsTextCreatedSamples bool
```
krajorama, line 431: I feel that including `Emit` is a bit strong word because switching to OM v2 if  OM v2 has the metadata support for created timestamp AND someone set `EmitOpenMetricsTextCreatedSamples`  to true, they will no longer see the `_created` samples with 2.0 , only 1.0 - because we want to move CT into metadata. Which would mean we contradict the call to action `Emit!`.  Maybe calling it `EnableOpenMetricsTextCreatedSamples` would be slightly better to not contradict ourselves in 2.0.

bwplotka: +1 cc @ArthurSens WDYT?

ArthurSens: Yep, sounds better!

krajorama (approved):
SGTM

bwplotka (commented):
LGTM, thanks!

bwplotka (approved):

PR #1414, opened 2023-12-27
Merging 1.18 back to main

ArthurSens:


api/client.go
```diff
@@ -36,18 +36,14 @@ var DefaultRoundTripper http.RoundTripper = &http.Transport{
...
 // Config defines configuration parameters for a new client.
 type Config struct {
 	// The address of the Prometheus to connect to.
 	Address string
 
 	// Client is used by the Client to drive HTTP requests. If not provided,
-	// a new http.Client based on the provided RoundTripper (or DefaultRoundTripper) will be used.
-	Client HttpClient
+	// a new one based on the provided RoundTripper (or DefaultRoundTripper) will be used.
+	Client *http.Client
```
bwplotka, line 46: This is weird, why we see this change again given https://github.com/prometheus/client_golang/commit/96f1aec289bfec34242831d74d450c3d0089050f

Like I see on main we have *http.Client as it should be already https://github.com/prometheus/client_golang/blob/main/api/client.go#L46 

🤔 

Something is off

bwplotka: Let's try anyway.

Issue #1294, opened 2023-06-15
CI: defining minimal workflow permissions


diogoteles08:
Hi!

I'm here to suggest that you set minimal permissions to your workflow [go.yml](https://github.com/prometheus/client_golang/blob/main/.github/workflows/go.yml), because currently it doesn't specify the permissions for its jobs and their privileges are being determined by GitHub's defaults.

As the change would be really simple and [this very similar PR](https://github.com/prometheus/client_golang/pull/1180) was already merged, I'll take the liberty and already open a PR with the suggested changes 😊

#### Context
I'm Diogo and I work on Google's Open Source Security Team([GOSST](https://opensource.googleblog.com/2023/04/googles-open-source-security-upstream-team-one-year-later.html)) in cooperation with the Open Source Security Foundation ([OpenSSF](https://openssf.org/)). My core job is to suggest and implement security changes on widely used open source projects 😊

diogoteles08:
Hey @bwplotka, I noticed that one of the changes made by the PR #[1180](https://github.com/prometheus/client_golang/pull/1180) was reverted, and now one of your workflows lack the minimal permissions. I'm reopening this issue to expose that and try to help understand what happened.

The change was made on [this commit](https://github.com/prometheus/client_golang/commit/51d24f86807c24ce866355d0c82258420f571c17), which seems to be automated and supposedly syncing the file with [this file](https://github.com/prometheus/prometheus/blob/main/scripts/golangci-lint.yml) on prometheus/prometheus. Do you think it makes sense to send a PR to prometheus/prometheus to add the minimal permissions to that original file?

Another think that I could suggest is that you change (if you haven't already) the [repository configuration](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions) to set the default permissions to read-only. In this case, if any workflow gets changed (or created) and forget to explicitly declare the permissions, it would run with read-only permissions and any action that requires write-permissions would fail.

diogoteles08:
I took the liberty and sent a PR to prometheus/prometheus and changed that original file. Now when your workflow imports the file according to the one in prometheus/prometheus, it should keep the minimal permissions.

That said, I'll send a small PR also resetting the minimal permission here, so we won't need to wait for the import from prometheus/prometheus to get it secured.

PR #1417, opened 2023-12-29
CI: Hash pin all GH actions + configure dependabot for those.

diogoteles08:
Closes #1416

As mentioned on the issue, this PR enhances project security by hash-pinning the dependencies that are called under dangerous permissions and configuring dependabot to automatically update them. Additionally, I've also used this PR to contribute with the nit-pick I've mentioned on commentaries at #1294; those changes will ensure all workflows are running with minimal permissions.

I configured dependabot in a way that all of version updates will be collapsed in a single PR sent monthly -- this avoids noisy PRs, which is a common concern haha. Regardless of the frequency chosen, for the case of security updates a PR with the fixed version would be sent right away.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,11 @@ updates:
     directory: "/tutorial/whatsup"
     schedule:
       interval: "monthly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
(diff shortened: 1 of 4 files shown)


ArthurSens (approved):
Sounds like a good improvement!

Do I understand it correctly that dependabot also updates the human-readable comment, alongside the commit SHA?

diogoteles08:
> Do I understand it correctly that dependabot also updates the human-readable comment, alongside the commit SHA?

Yes it does! =D

bwplotka:
Do you mind providing meaningful PR title?

bwplotka (approved):
Thanks!

PR #1422, opened 2024-01-03
Add version package

SuperQ:
Migrate the `version` package from `github.com/prometheus/common` to `github.com/prometheus/client_golang` in order to break the circular dependency.
* Make `version` a top level package because it uses `init()` to populate data.

Related to: https://github.com/prometheus/common/issues/58

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -1,0 +1,154 @@
+// Copyright 2016 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):

bwplotka (approved):
LGTM, just I would put it prometheus/collectors/version to group collectors in some way.

bwplotka:
In our client_golang sync, with @ArthurSens and @kakkoyun we decided at the end to move it to `prometheus/collectors/version` package. Hope @SuperQ that's ok, we want to do it before next release of client_golang. 

Sorry for confusion. 🤗

SuperQ:
I don't think that's the right place for it. The version package is mostly about having a structured version feature, the metrics collector is secondary.

bwplotka:
Yes, I understand, but from the client_golang point of view it's generally metric collector. To use it you have provide that structure, which you might reuse in other places, but for this repo, this is the best place. One could provide this structure on their own and somewhere else, but it's too simple to split version structure AND collector separately, thus collectors directory might be a fair trade-off (: 

I feel it's `prometheus/collectors/version` is still better than exporter_toolkit semantically. 🤔 But up to you! I think we decided to move it to  prometheus/collectors/version for now, it's too confusing and disorganized to keep it in the root @ArthurSens will do the PR later on. 

Sorry I indicated "fair" on Slack on your point around long URL, it's some argument but indeed pretty weak (was late for me) 🙈

PR #1423, opened 2024-01-04
Simplification of middleware example

michurin:
It seems a little overcomplication appears in example


bwplotka (approved):
Nice! Thanks 💪🏽

Issue #1351, opened 2023-09-26
testutil.GatherAndCompare is error-prone


machine424:
After https://github.com/prometheus/client_golang/pull/1143, the function `testutil.GatherAndCompare`:

https://github.com/prometheus/client_golang/blob/1bae6c1e6314f6a20be183a7277059630780232a/prometheus/testutil/testutil.go#L197-L203

https://github.com/prometheus/client_golang/blob/1bae6c1e6314f6a20be183a7277059630780232a/prometheus/testutil/testutil.go#L205-L222

https://github.com/prometheus/client_golang/blob/1bae6c1e6314f6a20be183a7277059630780232a/prometheus/testutil/testutil.go#L236-L245

became error-prone, as we filter the `expected` metrics in `compareMetricFamilies` as well now, if we make a mistake in one of the arguments `expected` or `metricNames`, the function will no longer return an error.

If e.g. we make a mistake in the name of the metric (provide a metric that is not exported because of a typo or re-using a variable containing the wrong metric name), the function will return `nil`.

Or e.g. if we provide an `expected` with the wrong metrics to a faulty code, we could never be aware of that.

I think `expected` lost its safeguard/source of truth utility.

I expect the user/tester to have full control on `expected`, I don't see why one couldn't provide the **exact** series they expect.

Plus having "dynamic" `expected` makes the tests hard to grasp and debug. No one needs tests with too many logic/intelligence that needs to be tested as well :) Way beneficial when they are more explicit even if this result in a little bit more code.

Looking at the code that was justifying the change:

https://github.com/fluxninja/aperture/blob/416d9ceaff9c1b2c4c238210c7c067dcf0a65567/pkg/otelcollector/metricsprocessor/processor_test.go#L148-L168

it's still not making use of the change yet, but I think they can easily split their `expected` and pass it as a map or sth to their test function...

Discovered this while adding a test for kubernetes etcd metrics: https://github.com/kubernetes/kubernetes/pull/120827. I copied/pasted a test function and the test was passing because I forgot to change the metric name.

machine424:
cc @dgrisonnet @wojtek-t

dgrisonnet:
cc @kakkoyun @bwplotka 

Could you please provide some insights here?

bwplotka:
[...]

PR #1424, opened 2024-01-04
testutil compareMetricFamilies: make less error-prone

leonnicolas:
The functions `GatherAndCompare`, `ScrapeAndCompare` and others that use
`compareMetricFamilies` under the hood can return no error if
`metricNames` includes none of the names found in the scraped/gathered
results. To avoid false Positves (an error being the negative case), we
can return an error if there is is at least one name in `metricNames`
that is not in the filtered results.

Fixes: https://github.com/prometheus/client_golang/issues/1351

Signed-off-by: leonnicolas <leonloechner@gmx.de>

### DIFF
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -277,6 +277,9 @@ func compareMetricFamilies(got, expected []*dto.MetricFamily, metricNames ...str
 	if metricNames != nil {
 		got = filterMetrics(got, metricNames)
 		expected = filterMetrics(expected, metricNames)
+		if len(metricNames) > len(got) {
+			return fmt.Errorf("expected metrics name not found")
+		}
 	}
 
 	return compare(got, expected)
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -278,7 +278,20 @@ func compareMetricFamilies(got, expected []*dto.MetricFamily, metricNames ...str
 		got = filterMetrics(got, metricNames)
 		expected = filterMetrics(expected, metricNames)
 		if len(metricNames) > len(got) {
-			return fmt.Errorf("expected metrics name not found")
+			h := make(map[string]struct{})
+			for _, mf := range got {
+				if mf == nil {
+					continue
+				}
+				h[mf.GetName()] = struct{}{}
+			}
+			var missingMetricNames []string
+			for _, name := range metricNames {
+				if _, ok := h[name]; !ok {
+					missingMetricNames = append(missingMetricNames, name)
+				}
+			}
+			return fmt.Errorf("expected metric name(s) not found: %v", missingMetricNames)
 		}
 	}
 
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -278,16 +278,9 @@ func compareMetricFamilies(got, expected []*dto.MetricFamily, metricNames ...str
 		got = filterMetrics(got, metricNames)
 		expected = filterMetrics(expected, metricNames)
 		if len(metricNames) > len(got) {
-			h := make(map[string]struct{})
-			for _, mf := range got {
-				if mf == nil {
-					continue
-				}
-				h[mf.GetName()] = struct{}{}
-			}
 			var missingMetricNames []string
 			for _, name := range metricNames {
-				if _, ok := h[name]; !ok {
+				if ok := hasMetricByName(got, name); !ok {
 					missingMetricNames = append(missingMetricNames, name)
 				}
 			}
@@ -333,4 +326,13 @@ func filterMetrics(metrics []*dto.MetricFamily, names []string) []*dto.MetricFam
 		}
 	}
 	return filtered
+}
+
+func hasMetricByName(metrics []*dto.MetricFamily, name string) bool {
+	for _, mf := range metrics {
+		if mf.GetName() == name {
+			return true
+		}
+	}
+	return false
 }
--- a/prometheus/testutil/testutil_test.go
+++ b/prometheus/testutil/testutil_test.go
@@ -328,27 +328,32 @@ func TestMetricNotFound(t *testing.T) {
 }
 
 func TestScrapeAndCompare(t *testing.T) {
-	const expected = `
+	scenarios := map[string]struct {
+		want        string
+		metricNames []string
+		errPrefix   string
+		fail        bool
+	}{
+		"empty metric Names": {
+			want: `
 		# HELP some_total A value that represents a counter.
 		# TYPE some_total counter
 
 		some_total{ label1 = "value1" } 1
-	`
+	`,
+			metricNames: []string{},
+		},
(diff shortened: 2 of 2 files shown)


prometheus/testutil/testutil.go
```diff
@@ -254,6 +254,9 @@ func compareMetricFamilies(got, expected []*dto.MetricFamily, metricNames ...str
 	if metricNames != nil {
 		got = filterMetrics(got, metricNames)
 		expected = filterMetrics(expected, metricNames)
+		if len(metricNames) > len(got) {
```
leonnicolas, line 257: I was thinking to use something like slices, to find the metricName that is not present in `got`. Would the additional package (slices is not yet in the std lib in go 1.19) and extra complexity of this function be outweighed by the improved usability. You would be able to see the misspelled or missing metricName in the error text.

leonnicolas:
According to Contributing.md I have to address one of the maintainers: @bwplotka

I also tried using table tests as you suggested in #1351, but I found it a bit messy for the test cases where the scraped server is not running or returns a bad request, so I left those tests as there where.

prometheus/testutil/testutil.go
leonnicolas: I added a commit. Finding the difference in 2 slices seems so easy but always requires so much code.

I added the "resulting" slice to the error with `%v`. I wonder if we should somehow allow callers of `compareMetricFamilies` to extract the missing metric Names? I guess that would require an API change :( 
It is just awkward to test if the "correct" metric names are in the error message.

leonnicolas: any feedback @bwplotka ?

prometheus/testutil/testutil.go
```diff
@@ -254,6 +254,22 @@ func compareMetricFamilies(got, expected []*dto.MetricFamily, metricNames ...str
...
+			h := make(map[string]struct{})
+			for _, mf := range got {
+				if mf == nil {
```
bwplotka, line 260: `got` element can be really nil here?

prometheus/testutil/testutil.go
bwplotka: Why exact metric names would be useful? Isn't it enough to return error?

`compareMetricFamilies` has a clear semantics - we are just ensuring the error is more readable and comparison true.

I think small code complexity is fine here - proposed below what can be done.

leonnicolas: > Why exact metric names would be useful? Isn't it enough to return error?

right, error should be enough

leonnicolas:
@bwplotka thanks for your review. I hope I addressed all you comments

leonnicolas:
@bwplotka anything missing from my side to move on with this PR?

prometheus/testutil/testutil_test.go
```diff
@@ -332,27 +332,46 @@ Diff:
 }
 
 func TestScrapeAndCompare(t *testing.T) {
-	const expected = `
+	scenarios := map[string]struct {
+		want        string
+		metricNames []string
+		// expectedErrPrefix if empty, means no fail is expected for the comparison.
+		expectedErrPrefix string
```
bwplotka, line 339: Why prefix and not exact error? Would be nice to test that missing metric semantics.

bwplotka (approved):
Thanks! Looks nice, just one comment, otherwise LGTM

prometheus/testutil/testutil_test.go
leonnicolas: done

bwplotka (approved):
Thanks! ❤️

prometheus/testutil/testutil.go
pckushan: Isn't it better to check the actual metrics names are included in the given **metricsNames**? The reason is if we have a test table including several tests and say it has 3 metics but sometimes it may trigger only two for a specific test scenario. In that execution we check this comparison passing all three metrics names but with this change it breaks the flow from this point `if len(metricNames) > len(got) {`

pckushan: I think `if len(metricNames) < len(got)` it is ok to give an error, but if other way around need to check whether got names are matched with given names.

bwplotka:
FYI as a learning: This broke dozens if not hundreds of downstream users. 

Feedback welcome of how we could improve those kind of changes in the future. This was a potentially "buggy" behaviour of the testing helper (surprising users). However, many users might have not care perhaps and we should just have create another function with the desired logic? 🤔

prometheus/testutil/testutil.go
aknuds1: This change breaks Mimir tests [when upgrading to v1.20.4](https://github.com/grafana/mimir/pull/9307), although I'm not sure exactly why yet.

aknuds1: OK, I can see it breaks Mimir tests because they expect to be able to compare with a greater metrics set than what is present. After this change, that results in an error.

logicalhan:
I think we should revert this. Prometheus should not be breaking a bunch of downstream tests with changes like this.

prometheus/testutil/testutil.go
BupycHuk: This change breaks mongodb_exporter as well, we test there if metric doesn't appear

bwplotka:
Thanks for your feedback @logicalhan. 

### Digression; are all breaking changes bad?

> You can deprecate the existing function and introduce a new function which has the behavior you want. This is NOT the way to change an existing API that hundreds of people are likely using and depending on.

Is it that simple? What if the existing users were not using the API as they should (or abuse it). e.g. why they are filtering by metric names which they don't expect to exist? What if this new behaviour might save you or your project from a critical production bug or major annoyance in the future? What if this breaking change is easily noticeable, only in testing code and might trigger a positive changes around Kubernetes tests.

Interesting enough this breaking change was [proposed and noticed](https://github.com/kubernetes/kubernetes/pull/120827#discussion_r1334445549) by our Kubernetes community. Only thanks to @machine424 due diligence we noticed this and avoid regression (or at least noop test). 

Finally, your suggestion would work normally, but in this case it would mean literally all functions in the `testutil` would be duplicated only to make the test check... actually do its job better.

As a result the proposal to revert the fix and leave the main functions error prone.. might be not rationally better (:

### Proposal

However, I spoke to a few people (e.g. @dashpole @dgrisonnet @kakkoyun @ArthurSens @vesari and your opinion), and it feels it might be easier to revert. 

Created [PR](https://github.com/prometheus/client_golang/pull/1645) -- added rationales there.

I propose we revert this, and work [on a long term solution](https://github.com/prometheus/client_golang/issues/1639). With the team we also discussed a way, where every client_golang release pipeline could propose an upgrade PR to a Kubernetes library that heavily depends on client_golang and check if the CI passes (compatibility check with a big user). 

Thanks!

bwplotka:
This PR was reverted: https://github.com/prometheus/client_golang/releases/tag/v1.20.5

PR #1427, opened 2024-01-09
chore: Move VersionCollector to collectors directory

ArthurSens:
Moving the newly added Version Collector to the collectors' directory. It's being kept in a separate package because, as @SuperQ mentioned, it uses `init()` to collect some information. We don't want to do that unless it is explicit to importers

I also added an example in a separate commit, happy to move it to a separate PR if that's preferred.


SuperQ:
I don't think this package belongs in the collectors directory because the primary use case of this package is to provide a structured version information. The version metric is an extra feature, not the primary feature.

Honestly, I think we were wrong in moving this to client_golang. The version information package probably makes more sense in exporter-toolkit.

SuperQ:
As a counter-proposal, we could keep `version` in common. But move the `NewCollector()` function here.

In order to do this, I've opened https://github.com/prometheus/common/pull/563 to make the helper function public.

ArthurSens:
I've changed the PR to continue using prometheus/common, but still depends on https://github.com/prometheus/common/pull/565

kakkoyun (approved):

bwplotka:
💪🏽

SuperQ:
Nice

PR #1431, opened 2024-01-11
update dep

ywwg:


go.mod
```diff
@@ -29,3 +29,5 @@ require (
 )
 
 exclude github.com/prometheus/client_golang v1.12.1
+
```
ywwg, line 32: will remove once common is updated

PR #1432, opened 2024-01-12
setup OSSF Scorecard workflow

mmorel-35:
Also pin github-actions versions

[OpenSSF Scorecard](https://securityscorecards.dev/viewer/?uri=github.com/prometheus/client_golang)

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,56 @@
+name: Scorecard supply-chain security
+
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: '22 1 * * 0'
+  push:
+    branches: [ "main" ]
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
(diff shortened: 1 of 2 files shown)


ArthurSens:
Hey 👋 -- thanks for the contribution.

Could you provide details about the OSSF scorecard and why we want to maintain it? Please assume I have no knowledge about what it is 😬

mmorel-35:
Ossf is open source security foundation. The workflow is here to create a report that will help maintainers reduce security risk on their project with advices. See the badge I added in the description.

ArthurSens:
I was taking a look at the report provided by the badge, I'm not sure I understood why we got 0 with Token-Permissions.

I don't guarantee that all permissions were configured following the least-privilege principle, but I'm pretty sure most of them are needed. Do we need to configure exceptions somewhere?

ArthurSens:
This PR is also making changes to Dockerfile, which doesn't seem related to the OSSF scorecard, could we split it into a separate PR? It could make the merge process faster, at least for the Dockerfile changes

mmorel-35:
It is related as ossf ask for dependencies to use pined version for docker as dor github-actions.

ArthurSens:
> It is related as ossf ask for dependencies to use pinned version for docker as for github-actions. Please have a look here https://securityscorecards.dev/viewer/?uri=github.com/prometheus/client_golang
> 
> I’m fine seing this in a following pr

Yeah, I imagine that would be the reason :P I sust meant that the changes for the Dockerfile we could merge without problems already, so opening a separate PR would unblock this

For the OSSF scorecard, I'm still struggling to understand why we got a 0 score for github actions that need those permissions to run. I couldn't find ways to add exceptions for those checks 🤔

kakkoyun:
> For the OSSF scorecard, I'm still struggling to understand why we got a 0 score for github actions that need those permissions to run. I couldn't find ways to add exceptions for those checks 🤔

@mmorel-35 Are there any quick wins that we could fix before putting this to the README?

Issue #1442, opened 2024-02-02
Prometheus exp. format: Detect and sort series with label values containing numbers numerically not lexicographically


linuxgcc:
When I accessed node_exporter server from port 9100, there was a problem with the sorting of the data I obtained. This error affected the viewing of CPU、disk and netcard device performance data. Although the data was accurate, it affected readability.
The data obtained by my machine is as follows:
```
# HELP node_network_address_assign_type Network device property: address_assign_type
# TYPE node_network_address_assign_type gauge
node_network_address_assign_type{device="ens3"} 0
node_network_address_assign_type{device="eth1"} 1
node_network_address_assign_type{device="eth10"} 1
node_network_address_assign_type{device="eth11"} 1
node_network_address_assign_type{device="eth12"} 1
node_network_address_assign_type{device="eth13"} 1
node_network_address_assign_type{device="eth14"} 1
node_network_address_assign_type{device="eth15"} 1
node_network_address_assign_type{device="eth16"} 1
node_network_address_assign_type{device="eth2"} 1
node_network_address_assign_type{device="eth3"} 1
node_network_address_assign_type{device="eth4"} 1
node_network_address_assign_type{device="eth5"} 1
node_network_address_assign_type{device="eth6"} 1
node_network_address_assign_type{device="eth7"} 1
node_network_address_assign_type{device="eth8"} 1
node_network_address_assign_type{device="eth9"} 1
node_network_address_assign_type{device="lo"} 0
# HELP node_schedstat_running_seconds_total Number of seconds CPU spent running a process.
# TYPE node_schedstat_running_seconds_total counter
node_schedstat_running_seconds_total{cpu="0"} 330.698211353
node_schedstat_running_seconds_total{cpu="1"} 1656.137375267
node_schedstat_running_seconds_total{cpu="10"} 312.818634313
node_schedstat_running_seconds_total{cpu="11"} 1590.5410981
node_schedstat_running_seconds_total{cpu="12"} 542.222016261
node_schedstat_running_seconds_total{cpu="13"} 539.735986571
node_schedstat_running_seconds_total{cpu="14"} 523.277367904
node_schedstat_running_seconds_total{cpu="15"} 809.123652669
node_schedstat_running_seconds_total{cpu="2"} 1388.031920081
node_schedstat_running_seconds_total{cpu="3"} 989.502352614
node_schedstat_running_seconds_total{cpu="4"} 765.031883684
[...]

PR #1443, opened 2024-02-02
fix metric sort error 

linuxgcc:
#1442

### DIFF
--- a/prometheus/internal/metric.go
+++ b/prometheus/internal/metric.go
@@ -14,7 +14,9 @@
 package internal
 
 import (
+	"regexp"
 	"sort"
+	"strconv"
 
 	dto "github.com/prometheus/client_model/go"
 )
@@ -46,6 +48,21 @@ func (s MetricSorter) Swap(i, j int) {
 	s[i], s[j] = s[j], s[i]
 }
 
+func (s MetricSorter) compare(si, sj string) bool {
+	// When the number of devices exceeds two digits, a sorting error occurs.
+	// dealing with sorting problems, especially CPU sorting errors
+	re := regexp.MustCompile(`\d+$`)
+	if re.Match([]byte(si)) && re.Match([]byte(sj)) {
+		siCode := re.FindStringSubmatch(si)[0]
+		sjCode := re.FindStringSubmatch(sj)[0]
+		siValue, _ := strconv.Atoi(siCode)
+		sjValue, _ := strconv.Atoi(sjCode)
+		return siValue < sjValue
+	}
+	// Non-numeric identification device, use original sorting
+	return si < sj
+}
+
 func (s MetricSorter) Less(i, j int) bool {
 	if len(s[i].Label) != len(s[j].Label) {
 		// This should not happen. The metrics are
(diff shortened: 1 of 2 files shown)


ArthurSens:
Although I agree that sometimes humans debug exposed metrics by manually looking the /metrics endpoint, I believe this endpoint it focused on metrics being scraped by Prometheus.

If we introduce a regex sort, I wonder how much are we sacrificing in performance for this change. Do you mind doing some benchmarking and posting the results here?

linuxgcc:
@ArthurSens 
Sorry, I don't have time recently, benchmark has been completed.

1 . touch such file  prometheus/internal/metric_sort_test.go
2. with out reg sort benchmark
> [root@trace client_golang]# go test -bench="BenchmarkSort" -benchtime=100x -benchmem  ./prometheus/internal                                                                                                         
> goos: linux
> goarch: amd64
> pkg: github.com/prometheus/client_golang/prometheus/internal
> cpu: Intel Xeon Processor (Cascadelake)
> BenchmarkSort-16             100              2020 ns/op             265 B/op          8 allocs/op
> PASS
> ok      github.com/prometheus/client_golang/prometheus/internal 0.012s

4. with reg sort benchmark

> [root@trace client_golang]# go test -bench="BenchmarkSort" -benchtime=100x -benchmem  ./prometheus/internal                                                                                                         
> goos: linux
> goarch: amd64
> pkg: github.com/prometheus/client_golang/prometheus/internal
> cpu: Intel Xeon Processor (Cascadelake)
> BenchmarkSort-16             100             88625 ns/op           52093 B/op        626 allocs/op
> PASS
> ok      github.com/prometheus/client_golang/prometheus/internal 0.022s
> 
> [root@trace client_golang]# 

It can be seen from the above comparison that the comparison of 16-core CPU data before and after using regular expressions takes no more than 0.1ms

ArthurSens:
We can also see a huge increase in memory allocation though 😱

prometheus/internal/metric.go
```diff
@@ -46,6 +48,21 @@ func (s MetricSorter) Swap(i, j int) {
 	s[i], s[j] = s[j], s[i]
 }
 
+func (s MetricSorter) compare(si, sj string) bool {
+	// When the number of devices exceeds two digits, a sorting error occurs.
+	// dealing with sorting problems, especially CPU sorting errors
+	re := regexp.MustCompile(`\d+$`)
```
bwplotka, line 54: Hm I think we can't use regex in such a hot path, we might need to be more smarter here if we ever do this ( let's talk on the issue) - or do it optionally.

bwplotka (changes requested):
Nice, thanks for this!

I added more points to the discussion https://github.com/prometheus/client_golang/issues/1442#issuecomment-2046695361

@ArthurSens is right, as you said it's "only" 0.1ms but if ops is one comparison.. we do this potentially 10 000 times per scrape making it a second. Number of allocs is one thing (a lot more) but the bump from 200 B to 50Kb is not minor (per comparison!). Maybe I am misinterpreting what "ops" is in your benchmark, maybe it's per full sorting (but for how many series?) In future please paste the benchmark code too, so we can verify.

But performance is one thing, the most important part is:
*  if readability hit of NOT having this is big enough to make this change
* if numerical sorting would NOT actually surprise (thus impact readability) of ppl who got used to it

bwplotka:
Sounds like we have to reject this, sorry! This is an intended behaviour.

Issue #1437, opened 2024-01-26
[CI]: Add Concurrency Grouping to GitHub Workflows


Ishani217:
### Requirement

Whenever we push commits to a PR, workflows are triggered for that commit. After that, if we push additional commits to this PR, workflows in Github Actions run on both commit

### Problem

We need to cancel the previous run and run only on the most recent pushed commit. This would help save some GitHub Action Minutes and unexpected workflow failures on subsequent commits.

### Proposal
Use concurrency groups to achieve this

Ishani217:
@ArthurSens @kakkoyun Could you please give me a heads up on this issue. I would love to solve this, to gain experience with this project.

ArthurSens:
Hi @Ishani217, we haven't had any issues with Action minutes as far as I can remember

I don't see a reason to not do this though, so feel free to come up with a PR :)

PR #1444, opened 2024-02-03
[CI]: Add Concurrency Grouping to GitHub Workflows

Ishani217:
### What problem is this PR solving?
fixes https://github.com/prometheus/client_golang/issues/1437

### Description of the changes
- Whenever we push commits to a PR, workflows are triggered for that commit. After that, if we push additional commits to this PR, workflows in Github Actions run on both commits.
- We need to cancel the previous run and run only on the most recent pushed commit. This would help save some GitHub Action Minutes (probably not a problem for this project but its always better to same some resources) and unexpected workflow failures on previous commits.

### DIFF
--- a/.github/workflows/automerge-dependabot.yml
+++ b/.github/workflows/automerge-dependabot.yml
@@ -1,6 +1,10 @@
 name: Dependabot auto-merge
 on: pull_request
 
+concurrency:
+  group: ${{ github.workflow }}-${{ (github.event.pull_request && github.event.pull_request.number) || github.ref || github.run_id }}
+  cancel-in-progress: true
+
 permissions:
   contents: write
   pull-requests: write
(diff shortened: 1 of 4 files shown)


kakkoyun (approved):

ArthurSens (commented):
Thanks for the PR! If I understand it correctly, if we accept the PR as it is, we would be cancelling workflows if we merge 2 or more PRs to main in a short period, is that correct?

Ideally, we never want to cancel workflows on pushes to the `main` branch to make sure it is easy to find commits that break our workflows

Ishani217:
@ArthurSens Nice catch actually :)
If two PRs would be merged in a timeframe when their workflow runs are concurrent, then yes, previous PR's commit push workflow trigger would be cancelled.

I need to close this PR then, since we aren't using merge queues to handle that case

bwplotka (approved):
So surprised this is not by default for PRs. Thanks!

I only hope we DO NOT cancel on main. Let's try this out, we can always revert.

PR #1445, opened 2024-02-13
Add support for Go 1.22

kavu:
Also refactored script for `generate-go-collector-test-files` makefile target, as it was not working for me at all.


kavu:
@ArthurSens @bwplotka would you mind to take a peek on that? Also, wanna ask you straight away — do you think it is possible to publish new version after this PR will be merged? Really want to add new metrics here… Thank you in advance!

generate-go-collector.bash
```diff
@@ -0,0 +1,7 @@
+#!/bin/env bash
+
+set -e
+
+go get github.com/hashicorp/go-version@v1.6.0
+go run prometheus/gen_go_collector_metrics_set.go
+mv -f go_collector_metrics_* prometheus
```
ArthurSens, line 7: I'm confused why this is needed 🤔

ArthurSens (commented):
Hey @kavu, thanks for the PR!

We promise support for only the 3 latest versions of Go, so we won't need to support 1.19 after this PR.

We can remove it in a separate PR though

generate-go-collector.bash
ArthurSens: I notice it doesn't work without it, I'm just struggling to understand what changed in the past that made this necessary :P

kavu:
@ArthurSens thank you, Arthur! Looking forward for new version, yay!

PR #1447, opened 2024-02-23
deps: bump `prometheus/common` version

tdakkota:
Fixes #1446

PR #1449, opened 2024-02-23
Remove support for go 1.19

ArthurSens:
We promise support for the latest 3 major versions of go. With #1445 merged, support for 1.19 is no longer promised.

In addition, the latest release of prometheus/common uses a function only present in go 1.20 (See [workflow logs](https://github.com/prometheus/client_golang/actions/runs/8017443782/job/21901319988?pr=1447) from PR #1447), meaning that we can't upgrade common without bumping our go version.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


kakkoyun (approved):
Awesome 🏆 🥇 💯

PR #1448, opened 2024-02-23
deps: bump prometheus/common version

ywwg:
Fixes https://github.com/prometheus/client_golang/issues/1446

PR #1451, opened 2024-02-26
Add 1.19.0 changelog

ArthurSens:


SuperQ (approved):

SuperQ:
Once this is done, I can mark the Version function in `prometheus/common` deprecated, cut a release, then remove  the circular dependency.

kakkoyun (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## Unreleased
 
+## 1.19.0 / 2023-02-27
+
+The module `prometheus/common v0.48.0` introduced a bug when used together with client_golang. If your project uses client_golang and you want to use `prometheus/common v0.48.0` or higher, please update client_golang to v1.19.0.
```
bboreham, line 5: Personally I find this frustrating, because I can't figure out what the bug is,
even after reading the full list of commits between 1.18 and 1.19.
Could you give a hint?

ArthurSens: A few constants ceased to exist (or were made private) in common 0.48.0, client_golang would break when trying to access those constants.

The fix was done here: https://github.com/prometheus/client_golang/pull/1448

ArthurSens: and now that you mention I totally agree that this text could be clearer, I'm just not sure if users are interested in such explanation "constant values could not be accessed anymore" 🤔

bboreham: Thanks, linking to the fix is enough information to trace it through.
I would say the root of the thing is https://github.com/prometheus/common/pull/576; some constants were deliberately hidden. I would use the word "incompatibility" rather than "bug".

ArthurSens: We still haven't merged `release-1.19` back to `main`, so I've raised https://github.com/prometheus/client_golang/pull/1454 and we can get it rephrased :)

PR #1454, opened 2024-02-28
Rephrase incompatibility with common v0.48.0

ArthurSens:
Stating that it was a bug sounds like the change made upstream was accidental and due to a mistake.

The change was intentional, so let's rephrase our changelog :)


bboreham (approved):
Much better, thanks!

PR #1464, opened 2024-03-10
Integer-only Counters

notrobpike:
Avoid rounding error for large-valued Counters.

@ArthurSens @bwplotka @kakkoyun

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -23,6 +23,9 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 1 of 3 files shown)


prometheus/counter.go
```diff
@@ -23,6 +23,9 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+// 64-bit float mantissa: https://en.wikipedia.org/wiki/Double-precision_floating-point_format
+var float64Mantissa uint64 = 9007199254740992
```
dswarbrick, line 27: Shouldn't this really be a const instead of a var? IMHO it would also be clearer and more legible if it were defined as `1 << 53` rather than a long decimal number.

PR #1471, opened 2024-03-15
add native histogram exemplar support

### LINKED ISSUE #1126: histograms: Support exemplars in pure Native Histograms
This depends on https://github.com/prometheus/client_model/issues/61 .

Exemplars can be added to the buckets of conventional histograms but not to native histograms. So far, the way of sending exemplars is by also rendering a conventional histogram. With the issue above fixed, we have a way of sending exemplars independently of conventional buckets. However, without conventional buckets, we don't have a heuristics which exemplars to pick. Native histograms can have so many buckets that adding an examplar to each is infeasible.

This is P2 as explained in the [milestone's description](https://github.com/prometheus/client_golang/milestone/6).

fatsheep9146:
fix #1126

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -440,7 +440,7 @@ type HistogramOpts struct {
 	// constant (or any negative float value).
 	NativeHistogramZeroThreshold float64
 
-	// The remaining fields define a strategy to limit the number of
+	// The next three fields define a strategy to limit the number of
 	// populated sparse buckets. If NativeHistogramMaxBucketNumber is left
 	// at zero, the number of buckets is not limited. (Note that this might
 	// lead to unbounded memory consumption if the values observed by the
@@ -472,8 +472,22 @@ type HistogramOpts struct {
 	NativeHistogramMaxBucketNumber  uint32
 	NativeHistogramMinResetDuration time.Duration
 	NativeHistogramMaxZeroThreshold float64
-	NativeHistogramMaxExemplarCount uint32
-	NativeHistogramExemplarTTL      time.Duration
+
+	// NativeHistogramMaxExemplars limits the number of exemplars
+	// that are kept in memory for each native histogram. If you leave it at
+	// zero, a default value of 10 is used. If no exemplars should be kept specifically
+	// for native histograms, set it to a negative value. (Scrapers can
+	// still use the exemplars exposed for classic buckets, which are managed
+	// independently.)
+	NativeHistogramMaxExemplars int
+	// NativeHistogramExemplarTTL is only checked once
+	// NativeHistogramMaxExemplars is exceeded. In that case, the
+	// oldest exemplar is removed if it is older than NativeHistogramExemplarTTL.
+	// Otherwise, the older exemplar in the pair of exemplars that are closest
+	// together (on an exponential scale) is removed.
+	// If NativeHistogramExemplarTTL is left at its zero value, a default value of
+	// 5m is used. To always delete the oldest exemplar, set it to a negative value.
+	NativeHistogramExemplarTTL time.Duration
 
 	// now is for testing purposes, by default it's time.Now.
 	now func() time.Time
@@ -534,6 +548,7 @@ func newHistogram(desc *Desc, opts HistogramOpts, labelValues ...string) Histogr
 	if opts.afterFunc == nil {
 		opts.afterFunc = time.AfterFunc
 	}
+
 	h := &histogram{
 		desc:                            desc,
 		upperBounds:                     opts.Buckets,
@@ -558,7 +573,7 @@ func newHistogram(desc *Desc, opts HistogramOpts, labelValues ...string) Histogr
 			h.nativeHistogramZeroThreshold = DefNativeHistogramZeroThreshold
 		} // Leave h.nativeHistogramZeroThreshold at 0 otherwise.
 		h.nativeHistogramSchema = pickSchema(opts.NativeHistogramBucketFactor)
-		h.nativeExemplars = newNativeExemplars(opts.NativeHistogramExemplarTTL, opts.NativeHistogramMaxExemplarCount)
+		h.nativeExemplars = makeNativeExemplars(opts.NativeHistogramExemplarTTL, opts.NativeHistogramMaxExemplars)
 	}
 	for i, upperBound := range h.upperBounds {
 		if i < len(h.upperBounds)-1 {
@@ -728,15 +743,14 @@ type histogram struct {
 	// resetScheduled is protected by mtx. It is true if a reset is
 	// scheduled for a later time (when nativeHistogramMinResetDuration has
 	// passed).
-	resetScheduled bool
+	resetScheduled  bool
+	nativeExemplars nativeExemplars
 
 	// now is for testing purposes, by default it's time.Now.
 	now func() time.Time
 
 	// afterFunc is for testing purposes, by default it's time.AfterFunc.
 	afterFunc func(time.Duration, func()) *time.Timer
-
-	nativeExemplars nativeExemplars
 }
 
 func (h *histogram) Desc() *Desc {
@@ -747,6 +761,8 @@ func (h *histogram) Observe(v float64) {
 	h.observe(v, h.findBucket(v))
 }
 
+// ObserveWithExemplar should not be called in high-frequency settings,
+// since it isn't lock-free for native histograms with configured exemplars.
 func (h *histogram) ObserveWithExemplar(v float64, e Labels) {
 	i := h.findBucket(v)
 	h.observe(v, i)
@@ -827,7 +843,12 @@ func (h *histogram) Write(out *dto.Metric) error {
 			}}
 		}
 
-		his.Exemplars = append(his.Exemplars, h.nativeExemplars.exemplars...)
+		if cap(h.nativeExemplars.exemplars) > 0 {
+			h.nativeExemplars.Lock()
+			his.Exemplars = append(his.Exemplars, h.nativeExemplars.exemplars...)
+			h.nativeExemplars.Unlock()
+		}
+
 	}
 	addAndResetCounts(hotCounts, coldCounts)
 	return nil
@@ -1098,8 +1119,10 @@ func (h *histogram) resetCounts(counts *histogramCounts) {
 	deleteSyncMap(&counts.nativeHistogramBucketsPositive)
 }
 
-// updateExemplar replaces the exemplar for the provided bucket. With empty
-// labels, it's a no-op. It panics if any of the labels is invalid.
+// updateExemplar replaces the exemplar for the provided classic bucket.
+// With empty labels, it's a no-op. It panics if any of the labels is invalid.
+// If histogram is native, the exemplar will be cached into nativeExemplars,
+// which has a limit, and will remove one exemplar when limit is reached.
 func (h *histogram) updateExemplar(v float64, bucket int, l Labels) {
 	if l == nil {
 		return
@@ -1588,56 +1611,140 @@ func addAndResetCounts(hot, cold *histogramCounts) {
 }
 
 type nativeExemplars struct {
-	nativeHistogramExemplarTTL      time.Duration
-	nativeHistogramMaxExemplarCount uint32
+	sync.Mutex
 
+	ttl       time.Duration
 	exemplars []*dto.Exemplar
-
-	lock sync.Mutex
 }
 
-func newNativeExemplars(ttl time.Duration, count uint32) nativeExemplars {
+func makeNativeExemplars(ttl time.Duration, maxCount int) nativeExemplars {
+	if ttl == 0 {
+		ttl = 5 * time.Minute
+	}
+
+	if maxCount == 0 {
+		maxCount = 10
+	}
+
+	if maxCount < 0 {
+		maxCount = 0
+	}
+
 	return nativeExemplars{
-		nativeHistogramExemplarTTL:      ttl,
-		nativeHistogramMaxExemplarCount: count,
-		exemplars:                       make([]*dto.Exemplar, 0),
-		lock:                            sync.Mutex{},
+		ttl:       ttl,
+		exemplars: make([]*dto.Exemplar, 0, maxCount),
 	}
 }
 
 func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
-	n.lock.Lock()
-	defer n.lock.Unlock()
-
-	elogarithm := math.Log(e.GetValue())
-	if len(n.exemplars) == int(n.nativeHistogramMaxExemplarCount) {
-		// check if oldestIndex is beyond TTL,
-		// if so, find the oldest exemplar, and nearest exemplar
-		oldestTimestamp := time.Now()
-		oldestIndex := -1
-		nearestValue := -1.0
-		nearestIndex := -1
-
-		for i, exemplar := range n.exemplars {
-			if exemplar.Timestamp.AsTime().Before(oldestTimestamp) {
-				oldestTimestamp = exemplar.Timestamp.AsTime()
-				oldestIndex = i
-			}
-			logarithm := math.Log(exemplar.GetValue())
-			if nearestValue == -1 || math.Abs(elogarithm-logarithm) < nearestValue {
-				fmt.Printf("gap: %f", math.Abs(elogarithm-logarithm))
-				nearestValue = math.Abs(elogarithm - logarithm)
-				nearestIndex = i
-			}
-		}
-
-		if oldestIndex != -1 && time.Since(oldestTimestamp) > n.nativeHistogramExemplarTTL {
-			n.exemplars[oldestIndex] = e
-		} else {
-			n.exemplars[nearestIndex] = e
-		}
+	if cap(n.exemplars) == 0 {
 		return
 	}
 
-	n.exemplars = append(n.exemplars, e)
+	n.Lock()
+	defer n.Unlock()
+
+	// The index where to insert the new exemplar.
+	var nIdx int = -1
+
+	// When the number of exemplars has not yet exceeded or
+	// is equal to cap(n.exemplars), then
+	// insert the new exemplar directly.
+	if len(n.exemplars) < cap(n.exemplars) {
+		for nIdx = 0; nIdx < len(n.exemplars); nIdx++ {
+			if *e.Value < *n.exemplars[nIdx].Value {
+				break
+			}
+		}
+		n.exemplars = append(n.exemplars[:nIdx], append([]*dto.Exemplar{e}, n.exemplars[nIdx:]...)...)
+		return
+	}
+
+	// When the number of exemplars exceeds the limit, remove one exemplar.
+	var (
+		rIdx int // The index where to remove the old exemplar.
+
+		ot    = time.Now() // Oldest timestamp seen.
+		otIdx = -1         // Index of the exemplar with the oldest timestamp.
+
+		md    = -1.0  // Logarithm of the delta of the closest pair of exemplars.
+		mdIdx = -1    // Index of the older exemplar within the closest pair.
+		cLog  float64 // Logarithm of the current exemplar.
+		pLog  float64 // Logarithm of the previous exemplar.
+	)
+
+	for i, exemplar := range n.exemplars {
+		// Find the exemplar with the oldest timestamp.
+		if otIdx == -1 || exemplar.Timestamp.AsTime().Before(ot) {
+			ot = exemplar.Timestamp.AsTime()
+			otIdx = i
+		}
+
+		// Find the index at which to insert new the exemplar.
+		if *e.Value <= *exemplar.Value && nIdx == -1 {
+			nIdx = i
+		}
+
+		// Find the two closest exemplars and pick the one the with older timestamp.
+		pLog = cLog
+		cLog = math.Log(exemplar.GetValue())
+		if i == 0 {
+			continue
+		}
+		diff := math.Abs(cLog - pLog)
+		if md == -1 || diff < md {
+			md = diff
+			if n.exemplars[i].Timestamp.AsTime().Before(n.exemplars[i-1].Timestamp.AsTime()) {
+				mdIdx = i
+			} else {
+				mdIdx = i - 1
+			}
+		}
+
+	}
+
+	// If all existing exemplar are smaller than new exemplar,
+	// then the exemplar should be inserted at the end.
+	if nIdx == -1 {
+		nIdx = len(n.exemplars)
+	}
+
+	if otIdx != -1 && time.Since(ot) > n.ttl {
+		rIdx = otIdx
+	} else {
+		// In the previous for loop, when calculating the closest pair of exemplars,
+		// we did not take into account the newly inserted exemplar.
+		// So we need to calculate with the newly inserted exemplar again.
+		elog := math.Log(e.GetValue())
+		if nIdx > 0 {
+			diff := math.Abs(elog - math.Log(n.exemplars[nIdx-1].GetValue()))
+			if diff < md {
+				md = diff
+				mdIdx = nIdx
+				if n.exemplars[nIdx-1].Timestamp.AsTime().Before(e.Timestamp.AsTime()) {
+					mdIdx = nIdx - 1
+				}
+			}
+		}
+		if nIdx < len(n.exemplars) {
+			diff := math.Abs(math.Log(n.exemplars[nIdx].GetValue()) - elog)
+			if diff < md {
+				mdIdx = nIdx
+				if n.exemplars[nIdx].Timestamp.AsTime().Before(e.Timestamp.AsTime()) {
+					mdIdx = nIdx
+				}
+			}
+		}
+		rIdx = mdIdx
+	}
+
+	// Adjust the slice according to rIdx and nIdx.
+	switch {
+	case rIdx == nIdx:
+		n.exemplars[nIdx] = e
+	case rIdx < nIdx:
+		n.exemplars = append(n.exemplars[:rIdx], append(n.exemplars[rIdx+1:nIdx], append([]*dto.Exemplar{e}, n.exemplars[nIdx:]...)...)...)
+	case rIdx > nIdx:
+		n.exemplars = append(n.exemplars[:nIdx], append([]*dto.Exemplar{e}, append(n.exemplars[nIdx:rIdx], n.exemplars[rIdx+1:]...)...)...)
+	}
 }
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -761,8 +761,9 @@ func (h *histogram) Observe(v float64) {
 	h.observe(v, h.findBucket(v))
 }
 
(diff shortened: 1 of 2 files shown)


prometheus/histogram.go
```diff
@@ -821,6 +830,11 @@ func (h *histogram) Write(out *dto.Metric) error {
 				Length: proto.Uint32(0),
 			}}
 		}
+
+		his.Exemplars = make([]*dto.Exemplar, 0)
```
ArthurSens, line 834:
```suggestion
		his.Exemplars = make([]*dto.Exemplar, len(h.nativeExemplars.exemplars))
```
We could pre-allocate the array size here :)

prometheus/histogram_test.go
```diff
@@ -1271,3 +1271,91 @@ func TestHistogramVecCreatedTimestampWithDeletes(t *testing.T) {
...
+	}
+
+	for e := range histogram.nativeExemplars.exemplars {
+		if _, ok := expectedValues[e.GetValue()]; !ok {
+			t.Errorf("the value is not in expected value")
+		}
+	}
+
+	time.Sleep(10 * time.Second)
+	histogram.ObserveWithExemplar(6, Labels{"id": "1"})
```
ArthurSens, line 1345: the histogram object has the `now()` function, I imagine we can prevent using sleep here if we just override the histogram now function to something like `h.now = func() time.Time { now.Add(time.Second * 10) }`

ArthurSens (commented):
not a native histogram expert here, just some small comments :)

prometheus/histogram.go
```diff
@@ -1575,3 +1593,152 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+
+type totalItem struct {
+	timeItem      *timeItem
+	logarithmItem *logarithmItem
+}
+
+func newNativeExemplars(ttl time.Duration, count uint32) *nativeExemplars {
+	return &nativeExemplars{
+		nativeHistogramExemplarTTL:      ttl,
+		nativeHistogramMaxExemplarCount: count,
```
beorn7, line 1613: These two need reasonable defaults in case they are left at their zero value (something like 5m for the TTL and 10 for the count?). And the defaults need to be documented in the doc comment for the options, too.

prometheus/histogram.go
```diff
@@ -1575,3 +1593,152 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+		exemplars:                       make(map[*dto.Exemplar]*totalItem),
+		timeList:                        &timeItem{},
+		logarithmList:                   &logarithmItem{},
+	}
+}
+
+func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
```
beorn7, line 1620: So the problem here is that this isn't concurrency safe. Note how the exemplars for classic exemplars are done (with a `[]atomic.Value`).

In the worst case, you could put everything behind a mutex, but that's quite slow. One might argue that `ObserveWithExemplar` shouldn't be called with high frequency anyway, and the mutex is OK for it (but that needs to be documented, should we go down that path).

I'm not sure if there is a feasible way to do something faster than the full mutex protection. For classic buckets, it's easy to know which exemplar to replace, but with the algorithm here, it's a much more complex decision.

Maybe let's do the mutex approach first and then try something more sophisticated if need arises.

prometheus/histogram.go
```diff
@@ -1575,3 +1593,152 @@ func addAndResetCounts(hot, cold *histogramCounts) {
 	atomic.AddUint64(&hot.nativeHistogramZeroBucket, atomic.LoadUint64(&cold.nativeHistogramZeroBucket))
 	atomic.StoreUint64(&cold.nativeHistogramZeroBucket, 0)
 }
+
+type nativeExemplars struct {
+	exemplars                       map[*dto.Exemplar]*totalItem
+	nativeHistogramExemplarTTL      time.Duration
+	nativeHistogramMaxExemplarCount uint32
+	timeList                        *timeItem
+	logarithmList                   *logarithmItem
```
beorn7, line 1602: While the algorithm seems to do the right thing, and your implementation might be the most efficient one in terms of big-O runtime, I do believe that it has a lot of baseline overhead, with two linked list and a map and many many pointers all over the place. Pointers have to be traversed, and they also take a lot of memory, which matters because an instrumented program might create many histograms.

Given that only a moderate number of exemplars make sense fundamentally (because that's the whole idea of exemplars to just present a few example observations), I'm pretty sure that an approach with a single slice will be more efficient both in terms of actual runtime and memory footprint.

So maybe have a single slice of size `nativeHistogramMaxExemplarCount` and sort it by the observed value. You can iterate through all ~10 elements once, searching for the oldest element and for the two closest ones at the same time. That should be super fast in practice. Insertion into the slice requires moving all the elements above by one, which is something computers can do super fast. (See https://go.dev/wiki/SliceTricks for trick to do it fast.)

beorn7 (changes requested):
Thanks for doing this. I think this implements the correct algorithm, but there is the huge problem of concurrency. See comments. A also think a more basic approach with a single slice rather than two linked list and a map will be faster in practice and consume less memory. Also see comments.

prometheus/histogram.go
fatsheep9146: > While the algorithm seems to do the right thing, and your implementation might be the most efficient one in terms of big-O runtime, I do believe that it has a lot of baseline overhead, with two linked list and a map and many many pointers all over the place. Pointers have to be traversed, and they also take a lot of memory, which matters because an instrumented program might create many histograms.
> 
> Given that only a moderate number of exemplars make sense fundamentally (because that's the whole idea of exemplars to just present a few example observations), I'm pretty sure that an approach with a single slice will be more efficient both in terms of actual runtime and memory footprint.
> 
> So maybe have a single slice of size `nativeHistogramMaxExemplarCount` and sort it by the observed value. You can iterate through all ~10 elements once, searching for the oldest element and for the two closest ones at the same time. That should be super fast in practice. Insertion into the slice requires moving all the elements above by one, which is something computers can do super fast. (See https://go.dev/wiki/SliceTricks for trick to do it fast.)

I understand your point, I think this is better since the number of exemplars may be only around 10, I will change the code like you said. @beorn7

prometheus/histogram.go
```diff
@@ -732,6 +737,10 @@ type histogram struct {
 
 	// afterFunc is for testing purposes, by default it's time.AfterFunc.
 	afterFunc func(time.Duration, func()) *time.Timer
+
+	nativeExemplars                 *nativeExemplars
```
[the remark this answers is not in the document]
fatsheep9146: I'm a little confused here, do you mean using pointer will create more overhead than not using pointer? @beorn7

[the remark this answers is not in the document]
beorn7: Yes. It's an additional 8 bytes in the struct, and it has to be dereferenced for each access.

The only upside of a pointer is that you don't need to copy the whole struct the pointer points to if you hand things around. But computers are super fast when it comes to memcopying, and also we are not really handing around the `nativeExemplars` struct a lot here.

prometheus/histogram.go
fatsheep9146: @beorn7 
I already changed the implementation from linked list into slice, but I did not keep the slice sorted as you suggested, here are my thoughts about this  

1. We should always iterate all exemplars to get the exemplar with oldest timestamp, so the time efficient should always be O(n)
2. If we do not keep the slice sorted, we can just change the exemplar in-place with new exemplar, so there is not need to adjust the slice with like  https://go.dev/wiki/SliceTricks does, which makes the code easier.

WDYT?

BTY, I will fix your other reviews when we can agree with this implementation.

beorn7: See my comment below. I think we don't want to find the exemplar closest to the new exemplar, but the pair of exemplars closest to each other (including the new exemplar) and then delete the older of the two in the couple.

prometheus/histogram.go
```diff
@@ -1575,3 +1586,58 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+			if exemplar.Timestamp.AsTime().Before(oldestTimestamp) {
+				oldestTimestamp = exemplar.Timestamp.AsTime()
+				oldestIndex = i
+			}
+			logarithm := math.Log(exemplar.GetValue())
+			if nearestValue == -1 || math.Abs(elogarithm-logarithm) < nearestValue {
+				fmt.Printf("gap: %f", math.Abs(elogarithm-logarithm))
+				nearestValue = math.Abs(elogarithm - logarithm)
+				nearestIndex = i
+			}
```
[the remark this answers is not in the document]
fatsheep9146: > This code selects the existing exemplar that is closest to the new exemplar. However, my idea was to find the two exemplars closest to each other (including the new one, but also all existing one). Finding that pair is O(n²), unless you keep the slice sorted, in which case it is O(n).

Got you, I will change the code.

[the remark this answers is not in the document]
fatsheep9146: The code is refracted, please help review it again when you have time. @beorn7

prometheus/histogram.go
fatsheep9146: added a `sync.Mutex` to `nativeExemplars` struct

prometheus/histogram.go
fatsheep9146: add default value and comments in doc.

prometheus/histogram.go
```diff
@@ -472,6 +472,8 @@ type HistogramOpts struct {
 	NativeHistogramMaxBucketNumber  uint32
 	NativeHistogramMinResetDuration time.Duration
 	NativeHistogramMaxZeroThreshold float64
+	NativeHistogramMaxExemplarCount uint32
```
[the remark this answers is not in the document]
fatsheep9146: Done

beorn7:
I'll try to get to this ASAP, but currently, I'm sick.

fatsheep9146:
> I'll try to get to this ASAP, but currently, I'm sick.

Thanks! Please take good care of yourself and rest well, there is no rush

kakkoyun (commented):
Nice to have this. Thanks for the contribution 🎉

beorn7 (changes requested):
Thanks a lot. I finally have done a detailed review. Things are a bit tricky, but I hope I have explained it well enough.

In different news, the doc comment for the `ExemplarObserver` interface (in file observer.go) needs an update, explaining broadly the behavior for native histograms, which should also include a warning that `ObserveWithExemplar` isn't lock-free if called on a native histogram with configured exemplars so that it should not be called too frequently.

fatsheep9146:
@beorn7 The pr is ready for review again.

beorn7:
Thank you very much. I'm on vacation this week, but this is high on my list once I'm back in action.

prometheus/histogram.go
```diff
@@ -732,6 +750,8 @@ type histogram struct {
 
 	// afterFunc is for testing purposes, by default it's time.AfterFunc.
 	afterFunc func(time.Duration, func()) *time.Timer
+
+	nativeExemplars nativeExemplars
```
beorn7, line 754: Nit: Maybe move this up to just after `resetScheduled` as `now` and `afterFunc` are just for testing purposes, while `nativeExemplars` is important for regular use.

prometheus/histogram.go
```diff
@@ -1102,6 +1129,10 @@ func (h *histogram) updateExemplar(v float64, bucket int, l Labels) {
 		panic(err)
 	}
 	h.exemplars[bucket].Store(e)
+	doSparse := h.nativeHistogramSchema > math.MinInt32 && !math.IsNaN(v)
+	if doSparse {
+		h.nativeExemplars.addExemplar(e)
+	}
```
beorn7, line 1135: The doc comment for this method (lines 1121-1122) needs an update to mention that nativeExemplars are handled, too.

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
 	atomic.AddUint64(&hot.nativeHistogramZeroBucket, atomic.LoadUint64(&cold.nativeHistogramZeroBucket))
 	atomic.StoreUint64(&cold.nativeHistogramZeroBucket, 0)
 }
+
+type nativeExemplars struct {
+	ttl time.Duration
+
+	exemplars []*dto.Exemplar
+
+	sync.Mutex
```
beorn7, line 1615: Nit: I see no reason to use a blank line between exemplars and ttl. With only three members, one might even argue that the mutex shouldn't be separated.

I have a minor preference to list a mutex that protects the whole struct first, but I'm not sure if that's really a convention we generally follow in the code base.

In summary:

```suggestion
	sync.Mutex

	ttl       time.Duration
	exemplars []*dto.Exemplar
```

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+	}
+
+	n.Lock()
+	defer n.Unlock()
+
+	// The index where to insert the new exemplar.
+	var nIdx int = -1
+
+	// When the number of exemplars has not yet exceeded or
+	// is equal to maxCount, then
```
beorn7, line 1649: maxCount → cap(n.exemplars)

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+
+	// When the number of exemplars exceeds the limit, remove one exemplar.
+	var (
+		// The index where to remove the old exemplar.
+		rIdx int
+
+		// The ot stores the oldest timestamp.
+		// The otIdx stores the index of exemplar with oldest timestamp.
+		ot    = time.Now()
+		otIdx = -1
```
beorn7, line 1669: I think those comments are easier to do inline. Suggestion:

```suggestion
		ot    = time.Now() // Oldest timestamp seen.
		otIdx = -1         // Index of the exemplar with the oldest timestamp.
```

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+
+		// The md stores the diff in logarithm of the closest pair of exemplars.
+		// The mdIdx stores the index of the older one of the closest pair of exemplars.
+		md    = -1.0
+		mdIdx = -1
+		// Two temporary variables to store the logarithm of current and previous exemplar.
+		cLog float64
+		pLog float64
```
beorn7, line 1677: In the same spirit:

```suggestion
		md    = -1.0 // Logarithm of the delta of the closest pair of exemplars.
		mdIdx = -1   // Index of the older exemplar within the closest pair.
		cLog float64 // Logarithm of the current exemplar.
		pLog float64 // Logarithm of the previous exemplar.
```

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+	)
+
+	for i, exemplar := range n.exemplars {
+		// Find the exemplar with oldest timestamp.
```
beorn7, line 1681: Language nit:
```suggestion
		// Find the exemplar with the oldest timestamp.
```

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+			}
+		}
+		n.exemplars = append(n.exemplars[:nIdx], append([]*dto.Exemplar{e}, n.exemplars[nIdx:]...)...)
+		return
+	}
+
+	// When the number of exemplars exceeds the limit, remove one exemplar.
+	var (
+		// The index where to remove the old exemplar.
+		rIdx int
```
beorn7, line 1664:
```suggestion
		rIdx int // The index where to remove the old exemplar.
```

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+	)
+
+	for i, exemplar := range n.exemplars {
+		// Find the exemplar with oldest timestamp.
+		if otIdx == -1 || exemplar.Timestamp.AsTime().Before(ot) {
+			ot = exemplar.Timestamp.AsTime()
+			otIdx = i
+		}
+
+		// Find the index to insert new exemplar.
```
beorn7, line 1687: Language nit:
```suggestion
		// Find the index at which to insert new the exemplar.
```

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+		if *e.Value <= *exemplar.Value && nIdx == -1 {
+			nIdx = i
+		}
+
+		// Find the two closest exemplars and pick up the one with older timestamp.
```
beorn7, line 1692:
```suggestion
		// Find the two closest exemplars and pick the one the with older timestamp.
```

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+				mdIdx = i
+			} else {
+				mdIdx = i - 1
+			}
+		}
+
+	}
+
+	// If all existing exemplar are smaller than new exemplar,
+	// then the exemplar should be inserted into the end.
```
beorn7, line 1711:
```suggestion
	// then the exemplar should be inserted at the end.
```

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+	if nIdx == -1 {
+		nIdx = len(n.exemplars)
+	}
+
+	if otIdx != -1 && time.Since(ot) > n.ttl {
+		rIdx = otIdx
+	} else {
+		// In previous for loop, when calculating the closest pair of exemplars,
+		// we do not take into account the newly inserted exemplar.
+		// So we should calculate with newly inserted exemplar again.
```
beorn7, line 1721:
```suggestion
		// In the previous for loop, when calculating the closest pair of exemplars,
		// we did not take into account the newly inserted exemplar.
		// So we need to calculate with the newly inserted exemplar again.
```

prometheus/histogram.go
```diff
@@ -1575,3 +1606,149 @@ func addAndResetCounts(hot, cold *histogramCounts) {
...
+				mdIdx = nIdx
+				if n.exemplars[nIdx].Timestamp.AsTime().Before(e.Timestamp.AsTime()) {
+					mdIdx = nIdx
+				}
+			}
+		}
+		rIdx = mdIdx
+	}
+
+	// Adjust the slice according to rIdx and nIdx
```
beorn7, line 1745:
```suggestion
	// Adjust the slice according to rIdx and nIdx.
```

beorn7 (changes requested):
Very cool.

I think all that remains is wordsmithing and some formatting tweaks. See all my pedantic comments. :)

fatsheep9146:
@beorn7 
The commit history seems too long,  I can squash to one commit if you think there are no more changes needed.

beorn7:
No worries, I can just squash during merge.

Or, if you like a more detailed commit structure with more than one commit, but not the 20 we have now, you could rebase on your own.

beorn7:
One thing fell through the cracks: The doc comment to the ObserveWithExemplar method in histogram.go should contain a warning that it  isn't lock-free for native histograms with configured exemplars and therefore should not be called in high-frequency settings.

fatsheep9146:
@beorn7 I only kept two commits, one for first implementation, the other for reconstitution that you suggested in https://github.com/prometheus/client_golang/pull/1471#discussion_r1559711441

prometheus/histogram.go
```diff
@@ -473,6 +473,22 @@ type HistogramOpts struct {
 	NativeHistogramMinResetDuration time.Duration
 	NativeHistogramMaxZeroThreshold float64
 
+	// NativeHistogramMaxExemplarNumber limits the number of exemplars
+	// that are kept in memory for each native histogram. If you leave it at
+	// zero, a default value of 10 is used. If no exemplars should be kept specifically
+	// for native histograms, set it to a negative value. (Scrapers can
+	// still use the exemplars exposed for classic buckets, which are managed
+	// independently.)
+	NativeHistogramMaxExemplarNumber int
```
bwplotka, line 482:
```suggestion
	NativeHistogramMaxExemplars int
```

bwplotka: Can we simplify? (also ditto name in commentary)

bwplotka (approved):
Happy to get this in, just one suggestion. Thanks!!! Sorry for lag on review.

prometheus/histogram.go
fatsheep9146: Done

bwplotka:
It feels @beorn7 comments are resolved, did you intent @beorn7 to approve it after you requested changes? https://github.com/prometheus/client_golang/pull/1471#pullrequestreview-2036231494

fatsheep9146:
> It feels @beorn7 comments are resolved, did you intent @beorn7 to approve it after you requested changes?

Yes, all comments are resolved, but I think @beorn7 is waiting for other reviewers to raise objections?

> Otherwise, this looks good. I'll give @ArthurSens @bwplotka and @kakkoyun a few more days to raise objections.

prometheus/histogram.go
```diff
@@ -821,6 +842,13 @@ func (h *histogram) Write(out *dto.Metric) error {
 				Length: proto.Uint32(0),
 			}}
 		}
+
+		if cap(h.nativeExemplars.exemplars) > 0 {
```
bwplotka, line 846: This might be taken as a mistake, so let's have a quick commentary on why `cap` is used not `len` here, WDYT?

fatsheep9146: We use `cap` here, because we first define that, if users use a negative `NativeHistogramMaxExemplars` value, then no exemplars will be cached. (You can see this in the comment of `NativeHistogramMaxExemplars`).

So when we call `makeNativeExemplars` to initialize a new `nativeExemplars`, if `NativeHistogramMaxExemplars` is negative, we just create a slice of exemplars with cap 0 to indicates that no exemplars will be cached.

So when `Write` is called, we use cap to check if user wants to cache exemplars. If they wants, then we will output the cached exemplars. 

But I think your idea is better, because when len is 0, although cap maybe greater than 0, we could still skip the following append operation since there are no cached exemplars at that time.

@bwplotka

fatsheep9146:
> Well, but "Change Requested" status is still there. And that from @beorn7 .... one never should mess with the red mark from @beorn7 =DDDD
> 
> On serious note, I wanted to be lazy and trust @beorn7 to approve the most complex part of this PR, so `addExemplar` body properly, I only skimmed through that. Let me check that next week (I have to focus on other work now), if @beorn7 won't have time.
> 
> Is that ok? We can also merge and iterate... it's a new functionality.

I think it's better to do a more detailed review of this pr when you have time next week. I'm definitely ok with that. @bwplotka

prometheus/histogram_test.go
```diff
@@ -1271,3 +1271,168 @@ func TestHistogramVecCreatedTimestampWithDeletes(t *testing.T) {
...
+			if len(h.nativeExemplars.exemplars) != len(tc.expectedValues) {
+				t.Errorf("the count of exemplars is not %d", len(tc.expectedValues))
+			}
+			for i, e := range h.nativeExemplars.exemplars {
+				if e.GetValue() != tc.expectedValues[i] {
+					t.Errorf("the %dth exemplar value %v is not as expected: %v", i, e.GetValue(), tc.expectedValues[i])
+				}
+			}
+		})
+	}
```
ArthurSens, line 1336: This part repeats 3x in this test, maybe we can extract into a separate function just to reduce a few lines?

prometheus/histogram_test.go
```diff
@@ -1271,3 +1271,168 @@ func TestHistogramVecCreatedTimestampWithDeletes(t *testing.T) {
...
+			name: "remove exemplar in closest pair, the removed index is bigger than inserted index",
+			addFunc: func(h *histogram) {
+				h.ObserveWithExemplar(0, Labels{"id": "1"})
+			},
+			expectedValues: []float64{0, 1, 4},
+		},
+		{
+			name: "remove exemplar with oldest timestamp, the removed index is smaller than inserted index",
+			addFunc: func(h *histogram) {
+				time.Sleep(10 * time.Second)
```
ArthurSens, line 1317: histograms have a field called `now`, their intended usage is exactly to mock time and avoid sleeps in tests.

You could do something like this when creating the histogram
```go
now := time.Now()
h.now = func time.Time { return now }
```

in `addFunc` you can change the returned value to:
```go
h.now = func time.Time { return now.Add(time.Second * 10) }
```

this would pretend that 10 seconds have passed, without needing to wait 10 seconds :)

ArthurSens (commented):
I just a quick read through the tests and I already have to leave 😬 

I'll try to take another look at `addExamplar` tomorrow but I'm similar to bartek here, I trust Beorn's review and there's no need to wait for me if we want to merge it already

prometheus/histogram.go
beorn7: If exemplars are configured, and we append the first exemplar, the `cap` is > 1, but the `len` is still 0. So it is correct and actually required to test for `cap` here.

prometheus/histogram.go
```diff
@@ -742,6 +761,8 @@ func (h *histogram) Observe(v float64) {
 	h.observe(v, h.findBucket(v))
 }
 
+// ObserveWithExemplar should not be called in high-frequency settings,
+// since it isn't lock-free for native histograms with configured exemplars.
```
beorn7, line 765: Slight clarification:

ObserveWithExemplar should not be called in a high-frequency setting for a native histogram with configured exemplars. For this case, the implementation isn't lock-free and might suffer from lock contention.

beorn7:
I think @ArthurSens 's comments about the tests are valid. @fatsheep9146 I think you should address them.

Otherwise, I only have a slight clarification for the `ObserveWithExemplar` doc comment (to make clear that it is still lock-free if called for a native histogram without exemplars or for a classic histogram).

And then this is good to go. Thanks for all your efforts and patience. This was quite involved.

prometheus/histogram.go
ArthurSens: From Bartek's comment I believe he agrees with the usage of `cap`, he just requested a comment on top so we can remember the reason in the future

fatsheep9146: @ArthurSens I think you are right :)
I misunderstood @bwplotka ‘s comments.

prometheus/histogram_test.go
fatsheep9146: I convert the compare logic into one common function `compareNativeExemplarValues`

prometheus/histogram_test.go
fatsheep9146: Done

prometheus/histogram.go
fatsheep9146: Done

prometheus/histogram.go
fatsheep9146: I added a quick comment.

beorn7 (approved):

beorn7:
Great. I'll merge on green.

PR #1472, opened 2024-03-16
Add LintDuplicateMetric to promlint

bboreham:
This is from https://github.com/prometheus/prometheus/pull/5312; the code moved to this repo since that PR was opened.

@geekodour is the original author.

### DIFF
--- a/prometheus/testutil/promlint/validations/duplicate_validations.go
+++ b/prometheus/testutil/promlint/validations/duplicate_validations.go
@@ -1,4 +1,4 @@
-// Copyright 2020 The Prometheus Authors
+// Copyright 2024 The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


prometheus/testutil/promlint/validations/duplicate_validations.go
```diff
@@ -0,0 +1,37 @@
...
+
+// LintDuplicateMetric detects duplicate metric.
+func LintDuplicateMetric(mf *dto.MetricFamily) []error {
+	var problems []error
+
+	for i, m := range mf.Metric {
+		for _, k := range mf.Metric[i+1:] {
+			if reflect.DeepEqual(m.Label, k.Label) {
+				problems = append(problems, fmt.Errorf("metric not unique"))
+				break
```
ArthurSens, line 31: If we break here we'll only return the first problem. Since we're returning an array of errors I wonder if we don't want to return all duplicated metrics instead of only the first one.

ArthurSens (approved):
LGTM

I've just made one comment but we can open an issue and tackle it in a separate PR if you prefer

prometheus/testutil/promlint/validations/duplicate_validations.go
ArthurSens: Nevermind, I've just realized we're braking only the first for loop

LGTM!

PR #1478, opened 2024-03-27
Improve: counter overflow and precise lost

xieyuschen:
This is the possible PR for dicsussion #1474


xieyuschen:
closed with the reason stated in the issue.

PR #1480, opened 2024-03-27
Implement Unwrap() for responseWriterDelegator

igor-drozdov:
According to https://pkg.go.dev/net/http#NewResponseController

If the `ResponseWriter` implements any of the following methods, the `ResponseController` will call them as appropriate:

```
Flush()
FlushError() error // alternative Flush returning an error Hijack() (net.Conn, *bufio.ReadWriter, error)
SetReadDeadline(deadline time.Time) error
SetWriteDeadline(deadline time.Time) error
EnableFullDuplex() error
```

If the `ResponseWriter` doesn't implement the methods, the `ResponseController` will call `Unwrap()` method until it finds a ResponseWriter in the chain, according to https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/net/http/responsecontroller.go;l=73

This commit implements `Unwrap()` method to simply return the wrapped `ResponseWriter`

### DIFF
--- a/prometheus/promhttp/delegator.go
+++ b/prometheus/promhttp/delegator.go
@@ -76,6 +76,12 @@ func (r *responseWriterDelegator) Write(b []byte) (int, error) {
 	return n, err
 }
 
+// Unwrap lets http.ResponseController get the underlying http.ResponseWriter,
+// by implementing the [rwUnwrapper](https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/net/http/responsecontroller.go;l=42-44) interface.
+func (r *responseWriterDelegator) Unwrap() http.ResponseWriter {
+	return r.ResponseWriter
+}
+
 type (
 	closeNotifierDelegator struct{ *responseWriterDelegator }
 	flusherDelegator       struct{ *responseWriterDelegator }
--- a/prometheus/promhttp/delegator_test.go
+++ b/prometheus/promhttp/delegator_test.go
@@ -1,0 +1,78 @@
+// Copyright 2024 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


prometheus/promhttp/delegator.go
```diff
@@ -76,6 +76,10 @@ func (r *responseWriterDelegator) Write(b []byte) (int, error) {
 	return n, err
 }
 
+func (r *responseWriterDelegator) Unwrap() http.ResponseWriter {
```
bwplotka, line 79: Do you mind adding quick comment on this public method similar to your PR description?

prometheus/promhttp/delegator_test.go
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+package promhttp
+
+import (
+	"net/http/httptest"
+	"testing"
+)
+
+func TestResponseWriterDelegatorUnwrap(t *testing.T) {
```
bwplotka, line 21: This test is a bit shallow it only tests single variable return 🙈 We could test if the HTTP lib will call Unwrap to get to those Flush methods properly. This would test our method from two sides:
* If the correct thing is passed and returned
* If our signature of the method actually works with http net lib.

What do you think? 🤔

bwplotka (approved):
Thanks!

Curious, how you noticed this? Any practical problematic case you encountered?

I have two potential suggestion, otherwise LGTM 💪🏽

prometheus/promhttp/delegator_test.go
igor-drozdov: Sure, I've implemented a custom ResponseWriter and now it checks whether the methods were propagated correctly by http.ResponseController

I've omitted a test for Flush method, because the library seems to try FlushError instead: https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/net/http/responsecontroller.go;l=51, to actually return an error

igor-drozdov:
> Curious, how you noticed this? Any practical problematic case you encountered?

I was investigating bidirectional streaming for one of the endpoints. When I first used Flush to flush the buffered data, it worked fine with and without [http.ResponseController](https://pkg.go.dev/net/http#ResponseController) because Flush method was implemented in all response writer wrappers of the chain. But when I tried calling EnableFullDuplex, it failed because neither EnableFullDuplex nor Unwrap was implemented in promhttp responsewriter wrapper.

So, it mostly to comply with the newly introduced ResponseController (1.20). I even deleted EnableFullDuplex test because this option was introduced too recently in 1.21 😅 

@bwplotka thank you for the review! Could you please have another look?

bwplotka (approved):
Thanks!

Issue #1435, opened 2024-01-23
Auto increment supported Go version


kakkoyun:
Client_golang promises support for the latest the last 3 major versions of Go. On every new Go release, we need to submit a few PRs updating the metrics collected by Go Collector. We already have scripts that autogenerates those new metrics, but it is not fully automated with our CI/CD.

--- 

This task is about connecting the dots between CI/CD and Go Collector metrics autogeneration

xref: https://github.com/cncf/mentoring/pull/1120

dvjsharma:
Hello @kakkoyun ,

It would be helpful if you could outline the specific segments of the pipeline that haven't been automated yet. Additionally, could you provide detailed insights into the manual tasks required during a new release? If there's any supporting documentation available, please include it.
Thanks!

CC @ArthurSens

kakkoyun:
Hey @dvjsharmam thanks for the question. All these will be discussed as part of the mentorship program. 

IMHO, going through the whole life-cycle of development tasks and helping interns navigate the open-source world is the core idea behind the program. It's not about specifying some issues and just getting them implemented.

I also mentioned what's needed in the issue description. A quick GH search would lend you to the program that generates those metrics https://github.com/prometheus/client_golang/blob/main/prometheus/gen_go_collector_metrics_set.go and an example PR https://github.com/prometheus/client_golang/pull/1105

dvjsharma:
Thanks @kakkoyun for the response, excited for the mentorship program! I'll take a look at the resources.

SachinSahu431:
Relevant PRs:

https://github.com/prometheus/client_golang/pull/1476
https://github.com/prometheus/client_golang/pull/1481
https://github.com/prometheus/client_golang/pull/1489

PR #1481, opened 2024-03-28
feat: metrics generation workflow

SachinSahu431:
This PR adds a bash script that will generate new metric files whenever a new Go version is released. After that, a PR will be raised, containing all these files.

Closes: https://github.com/prometheus/client_golang/issues/1435


.github/workflows/update-go-versions.yml
```diff
@@ -0,0 +1,38 @@
+---
+name: Weekly Go Version Check and Metric Generation
+
+on:
+  schedule:
+    - cron: "0 0 * * 0"  # Every Sunday at midnight UTC
```
ArthurSens, line 6: According to [go release docs](https://go.dev/wiki/Go-Release-Cycle), new releases are created in February and August. Maybe we could limit this workflow to run only during those months?

ArthurSens (commented):
Have you tried this workflow on a fork?

SachinSahu431:
> Have you tried this workflow on a fork?

I tried it on a seperate fork, but was unable to trigger it manually. I added `workflow_dispatch` to do do, but this workflow was not visible in Actions tab. Am I missing something?

SachinSahu431:
The bash script will generate new metric files whenever a new Go version is detected. Then a PR will be raised, containing all these files. 

Example PR: https://github.com/SachinSahu431/client_golang/pull/2

SachinSahu431:
Hey!

Here is one example of autogenerated PR for Go version 1.22: https://github.com/SachinSahu431/client_golang/pull/2

Please let me know if we should modify the format of PR. For example: adding some labels, marking it draft by default, etc.

SachinSahu431:
Note that in [supported_go_versions.txt](https://github.com/SachinSahu431/client_golang/pull/2/files#diff-c99d42eb6b3d2c81f2a74179e4c0909b956fc104da0f19499d28e0129f2be58c), older Go versions are not being removed. Out of all versions listed in this txt file, only the top 3 versions are being used by [update-go-version.bash file](https://github.com/prometheus/client_golang/pull/1481/files#diff-e2c042a423598946ade46fd28f670b306073b19b82cf1b2bcd1ca09050bdc390) even if 4th (and later) versions are supported.

update-go-version.bash
```diff
@@ -0,0 +1,21 @@
...
+# Check for new version of Go, and generate go collector test files
+# New Go version gets appended at top of supported_go_versions.txt, and only top 3 versions are supported
+if [[ ! $current_version =~ $latest_version ]]; then
+  echo "New Go version available: $latest_version"
+  echo "Updating supported_go_versions.txt and generating Go Collector test files"
+  sed -i "1i $latest_version" supported_go_versions.txt
+  make generate-go-collector-test-files
+else
+  echo "No new Go version detected. Current Go version is: $current_version"
+fi
```
kakkoyun, line 21:
```suggestion
fi

```

.github/workflows/update-go-versions.yml
```diff
@@ -0,0 +1,35 @@
+---
+name: Generate Metric files for new Go version
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Runs weekly during month of February and August
+    # Reference: https://go.dev/wiki/Go-Release-Cycle
+    - cron: '0 0 * 2,8 1'
```
kakkoyun, line 9: The document doesn't state the actual day of the release so maybe we should run it on the last day of the month.

Still, this wouldn't be very fail-safe; what happens if the release is stalled or postponed? I don't see any downsides to running the task every month. It's a no-op if there isn't any version released.

SachinSahu431: Agreed 👍
Let's schedule it to run on a monthly basis.

ArthurSens:
> Note that in [supported_go_versions.txt](https://github.com/SachinSahu431/client_golang/pull/2/files#diff-c99d42eb6b3d2c81f2a74179e4c0909b956fc104da0f19499d28e0129f2be58c), older Go versions are not being removed. Out of all versions listed in this txt file, only the top 3 versions are being used by [update-go-version.bash file](https://github.com/prometheus/client_golang/pull/1481/files#diff-e2c042a423598946ade46fd28f670b306073b19b82cf1b2bcd1ca09050bdc390) even if 4th (and later) versions are supported.

I think this will cause confusion in the future 😬. Could we make sure that we only have the supported versions in the `supported_go_versions` file? I.e. the 3 latest go versions

SachinSahu431:
@ArthurSens made the changes, thanks! 🙌

kakkoyun (approved):

PR #1482, opened 2024-03-30
Generate with.*Metrics even if empty

ArthurSens:
I was taking a look at https://github.com/prometheus/client_golang/pull/1389 again and noticed that we'll need to generate the withDebugMetrics even if it returns an empty array.

I've updated the script accordingly :)


ArthurSens:
Oh ok, I'll need to solve everything in a single PR 😅

PR #1489, opened 2024-04-02
ci: Remove hardcoded supported Go versions from go.yml

SachinSahu431:
Currently, go.yml file contains hardcoded list of [supported Go versions](https://github.com/prometheus/client_golang/blob/e38b67e77a6969b0c660691e1a7d29404c24dcfb/.github/workflows/go.yml#L21). This needs to be manually updated when a new Go version is available.

With this PR, this list will be fetched from [supprted_go_versions.txt](https://github.com/prometheus/client_golang/blob/e38b67e77a6969b0c660691e1a7d29404c24dcfb/supported_go_versions.txt) (only the latest 3 Go versions). 

Related PR: https://github.com/prometheus/client_golang/pull/1481

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Read supported_go_versions.txt
         id: matrix
         run: |
-            versions=$(head -n 3 supported_go_versions.txt)
+            versions=$(cat supported_go_versions.txt)
             matrix="[$(echo "$versions" | sed 's/\(.*\)/"\1"/' | paste -s -d,)]"
             echo "supported_versions=$matrix" >> $GITHUB_OUTPUT
   
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -12,13 +12,29 @@ permissions:
   contents: read
 
 jobs:
+  supportedVersions:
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):

.github/workflows/go.yml
```diff
@@ -12,13 +12,29 @@ permissions:
...
+    runs-on: ubuntu-latest
+    outputs:
+        supported_versions: ${{ steps.matrix.outputs.supported_versions }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Read supported_go_versions.txt
+        id: matrix
+        run: |
+            versions=$(head -n 3 supported_go_versions.txt)
```
ArthurSens, line 26:
```suggestion

```

I feel like keeping a long list of versions in the `supported_go_versions` will be confusing in the future, i.e. I would expect support for all versions in that list.

Could we instead remove the latest version from this list when adding support to a new one?

SachinSahu431: Agreed 💯 
I'll make the changes. Thanks for the suggestion!

ArthurSens (approved):
Awesome job :)

PR #1492, opened 2024-04-02
fix: remove redundant go module in middleware example

majolo:
Noticed when the middleware example was added it introduced a go module (presumably by mistake).

`module github.com/jessicalins/instrumentation-practices-examples/middleware`

This caused us an issue so removing it.


majolo:
cc @bwplotka

kakkoyun (approved):

PR #1496, opened 2024-04-10
feat: Support zstd compression

mrueg:
This allows endpoints to respond with zstd compressed metric data, if the requester supports it. For backwards compatibility, gzip compression will take precedence.

I added a benchmark to the http_test.go, that will test with a differently sized synthetic metric series.

```
Running tool: go test -benchmem -run=^$ -bench ^BenchmarkEncoding$ github.com/prometheus/client_golang/prometheus/promhttp

goos: linux
goarch: amd64
pkg: github.com/prometheus/client_golang/prometheus/promhttp
cpu: Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz
BenchmarkEncoding/test_with_gzip_encoding_small-8         	   20670	     53641 ns/op	  111770 B/op	      49 allocs/op
BenchmarkEncoding/test_with_zstd_encoding_small-8         	    7838	    178033 ns/op	 1121493 B/op	      75 allocs/op
BenchmarkEncoding/test_with_no_encoding_small-8           	   56409	     20998 ns/op	   38030 B/op	      43 allocs/op
BenchmarkEncoding/test_with_gzip_encoding_medium-8        	    8493	    143460 ns/op	   78918 B/op	      49 allocs/op
BenchmarkEncoding/test_with_zstd_encoding_medium-8        	    3650	    345559 ns/op	 1155737 B/op	      76 allocs/op
BenchmarkEncoding/test_with_no_encoding_medium-8          	    8816	    122759 ns/op	   72252 B/op	      44 allocs/op
BenchmarkEncoding/test_with_gzip_encoding_large-8         	     944	   1206069 ns/op	  479829 B/op	      52 allocs/op
BenchmarkEncoding/test_with_zstd_encoding_large-8         	     817	   1482854 ns/op	 1429911 B/op	      75 allocs/op
BenchmarkEncoding/test_with_no_encoding_large-8           	     994	   1213586 ns/op	  346837 B/op	      45 allocs/op
BenchmarkEncoding/test_with_gzip_encoding_extra-large-8   	      85	  14008798 ns/op	 2972152 B/op	      56 allocs/op
BenchmarkEncoding/test_with_zstd_encoding_extra-large-8   	      90	  14638520 ns/op	 3731521 B/op	      74 allocs/op
BenchmarkEncoding/test_with_no_encoding_extra-large-8     	      93	  13897780 ns/op	 2648745 B/op	      45 allocs/op
PASS
ok  	github.com/prometheus/client_golang/prometheus/promhttp	17.534s

```

See also: https://github.com/prometheus/prometheus/issues/13866

/kind feature

Release note:
```release-note
promhttp: Support zstd compression	
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,9 @@ go 1.20
 require (
 	github.com/beorn7/perks v1.0.1
 	github.com/cespare/xxhash/v2 v2.3.0
+	github.com/google/go-cmp v0.6.0
 	github.com/json-iterator/go v1.1.12
+	github.com/klauspost/compress v1.17.8
 	github.com/prometheus/client_model v0.6.1
 	github.com/prometheus/common v0.53.0
 	github.com/prometheus/procfs v0.15.1
--- a/internal/github.com/golang/gddo/httputil/negotiate.go
+++ b/internal/github.com/golang/gddo/httputil/negotiate.go
@@ -1,0 +1,36 @@
+// Copyright 2013 The Go Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd.
+
+package httputil
+
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/internal/github.com/golang/gddo/httputil/header"
+)
+
+// NegotiateContentEncoding returns the best offered content encoding for the
+// request's Accept-Encoding header. If two offers match with equal weight and
+// then the offer earlier in the list is preferred. If no offers are
+// acceptable, then "" is returned.
+func NegotiateContentEncoding(r *http.Request, offers []string) string {
+	bestOffer := "identity"
+	bestQ := -1.0
+	specs := header.ParseAccept(r.Header, "Accept-Encoding")
+	for _, offer := range offers {
+		for _, spec := range specs {
+			if spec.Q > bestQ &&
+				(spec.Value == "*" || spec.Value == offer) {
+				bestQ = spec.Q
+				bestOffer = offer
+			}
+		}
+	}
+	if bestQ == 0 {
+		bestOffer = ""
+	}
+	return bestOffer
+}
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -55,6 +55,8 @@ const (
 	processStartTimeHeader = "Process-Start-Time-Unix"
 )
 
+// Compression represents the content encodings handlers support for the HTTP
+// responses.
 type Compression string
 
 const (
@@ -427,7 +429,7 @@ func httpError(rsp http.ResponseWriter, err error) {
 	)
 }
 
-// NegotiateEncodingWriter reads the Accept-Encoding header from a request and
+// negotiateEncodingWriter reads the Accept-Encoding header from a request and
 // selects the right compression based on an allow-list of supported
 // compressions. It returns a writer implementing the compression and an the
 // correct value that the caller can set in the response header.
@@ -451,8 +453,6 @@ func negotiateEncodingWriter(r *http.Request, rw io.Writer, compressions []strin
 		return z, selected, func() { _ = z.Close() }, nil
 	case "gzip":
 		gz := gzipPool.Get().(*gzip.Writer)
-		defer gzipPool.Put(gz)
-
 		gz.Reset(rw)
 		return gz, selected, func() { _ = gz.Close(); gzipPool.Put(gz) }, nil
 	case "identity":
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -361,7 +361,7 @@ type HandlerOpts struct {
 	// DisableCompression disables the response encoding (compression) and
 	// encoding negotiation. If true, the handler will
 	// never compress the response, even if requested
-	// by the client and the OfferedCompressions field is ignored.
+	// by the client and the OfferedCompressions field is set.
 	DisableCompression bool
 	// OfferedCompressions is a set of encodings (compressions) handler will
 	// try to offer when negotiating with the client. This defaults to identity, gzip
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -55,23 +55,13 @@ const (
 	processStartTimeHeader = "Process-Start-Time-Unix"
 )
 
-type Compression int
+type Compression string
 
 const (
-	Identity Compression = iota
-	Gzip
-	Zstd
+	Identity Compression = "identity"
+	Gzip     Compression = "gzip"
+	Zstd     Compression = "zstd"
 )
-
-var compressions = [...]string{
-	"identity",
-	"gzip",
-	"zstd",
-}
-
-func (c Compression) String() string {
-	return compressions[c]
-}
 
 var defaultCompressionFormats = []Compression{Identity, Gzip, Zstd}
 
@@ -143,6 +133,18 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		}
 	}
 
+	// Select all supported compression formats
+	var compressions []string
+	if !opts.DisableCompression {
+		offers := defaultCompressionFormats
+		if len(opts.OfferedCompressions) > 0 {
+			offers = opts.OfferedCompressions
+		}
+		for _, comp := range offers {
+			compressions = append(compressions, string(comp))
+		}
+	}
+
 	h := http.HandlerFunc(func(rsp http.ResponseWriter, req *http.Request) {
 		if !opts.ProcessStartTime.IsZero() {
 			rsp.Header().Set(processStartTimeHeader, strconv.FormatInt(opts.ProcessStartTime.Unix(), 10))
@@ -188,12 +190,19 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		}
 		rsp.Header().Set(contentTypeHeader, string(contentType))
 
-		w, err := GetWriter(req, rsp, opts.DisableCompression, opts.OfferedCompressions)
+		w, encodingHeader, err := NegotiateEncodingWriter(req, rsp, opts.DisableCompression, compressions)
 		if err != nil {
 			if opts.ErrorLog != nil {
 				opts.ErrorLog.Println("error getting writer", err)
 			}
+			// Since the writer received from NegotiateEncodingWriter will be nil, in case there's an error, we set it here
+			w = io.Writer(rsp)
 		}
+
+		if encodingHeader == "" {
+			encodingHeader = string(Identity)
+		}
+		rsp.Header().Set(contentEncodingHeader, encodingHeader)
 
 		enc := expfmt.NewEncoder(w, contentType)
 
@@ -419,48 +428,45 @@ func httpError(rsp http.ResponseWriter, err error) {
 	)
 }
 
-func GetWriter(r *http.Request, rsp http.ResponseWriter, disableCompression bool, offeredCompressions []Compression) (io.Writer, error) {
-	w := io.Writer(rsp)
-	rsp.Header().Set(contentEncodingHeader, "identity")
-	if !disableCompression {
-		offers := defaultCompressionFormats
-		if len(offeredCompressions) > 0 {
-			offers = offeredCompressions
-		}
-		var compressions []string
-		for _, comp := range offers {
-			compressions = append(compressions, comp.String())
-		}
-		// TODO(mrueg): Replace internal/github.com/gddo once https://github.com/golang/go/issues/19307 is implemented.
-		compression := httputil.NegotiateContentEncoding(r, compressions)
-		switch compression {
-		case "zstd":
-			rsp.Header().Set(contentEncodingHeader, "zstd")
-			// TODO(mrueg): Replace klauspost/compress with stdlib implementation once https://github.com/golang/go/issues/62513 is implemented.
-			z, err := zstd.NewWriter(rsp, zstd.WithEncoderLevel(zstd.SpeedFastest))
-			if err != nil {
-				return nil, err
-			}
+// NegotiateEncodingWriter reads the Accept-Encoding header from a request and
+// selects the right compression based on an allow-list of supported
+// compressions. It returns a writer implementing the compression and an the
+// correct value that the caller can set in the response header.
+func NegotiateEncodingWriter(r *http.Request, rw io.Writer, disableCompression bool, compressions []string) (_ io.Writer, encodingHeaderValue string, _ error) {
+	w := rw
 
-			z.Reset(w)
-			defer z.Close()
-
-			w = z
-		case "gzip":
-			rsp.Header().Set(contentEncodingHeader, "gzip")
-			gz := gzipPool.Get().(*gzip.Writer)
-			defer gzipPool.Put(gz)
-
-			gz.Reset(w)
-			defer gz.Close()
-
-			w = gz
-		case "identity":
-			// This means the content is not compressed.
-		default:
-			// The content encoding was not implemented yet.
-			return w, fmt.Errorf("content compression format not recognized: %s. Valid formats are: %s", compression, defaultCompressionFormats)
-		}
+	if disableCompression {
+		return w, string(Identity), nil
 	}
-	return w, nil
+
+	// TODO(mrueg): Replace internal/github.com/gddo once https://github.com/golang/go/issues/19307 is implemented.
+	compression := httputil.NegotiateContentEncoding(r, compressions)
+
+	switch compression {
+	case "zstd":
+		// TODO(mrueg): Replace klauspost/compress with stdlib implementation once https://github.com/golang/go/issues/62513 is implemented.
+		z, err := zstd.NewWriter(rw, zstd.WithEncoderLevel(zstd.SpeedFastest))
+		if err != nil {
+			return nil, "", err
+		}
+
+		z.Reset(w)
+		defer z.Close()
+
+		w = z
+	case "gzip":
+		gz := gzipPool.Get().(*gzip.Writer)
+		defer gzipPool.Put(gz)
+
+		gz.Reset(w)
+		defer gz.Close()
+
+		w = gz
+	case "identity":
+		// This means the content is not compressed.
+	default:
+		// The content encoding was not implemented yet.
+		return nil, "", fmt.Errorf("content compression format not recognized: %s. Valid formats are: %s", compression, defaultCompressionFormats)
+	}
+	return w, compression, nil
 }
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -191,6 +191,13 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		rsp.Header().Set(contentTypeHeader, string(contentType))
 
 		w, encodingHeader, closeWriter, err := NegotiateEncodingWriter(req, rsp, opts.DisableCompression, compressions)
+		if err != nil {
+			if opts.ErrorLog != nil {
+				opts.ErrorLog.Println("error getting writer", err)
+			}
+			w = io.Writer(rsp)
+			encodingHeader = string(Identity)
+		}
 
 		if closeWriter != nil {
 			defer func() {
@@ -202,17 +209,7 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 				}
 			}()
 		}
-		if err != nil {
-			if opts.ErrorLog != nil {
-				opts.ErrorLog.Println("error getting writer", err)
-			}
-			// Since the writer received from NegotiateEncodingWriter will be nil, in case there's an error, we set it here
-			w = io.Writer(rsp)
-		}
 
-		if encodingHeader == "" {
-			encodingHeader = string(Identity)
-		}
 		rsp.Header().Set(contentEncodingHeader, encodingHeader)
 
 		enc := expfmt.NewEncoder(w, contentType)
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -55,7 +55,25 @@ const (
 	processStartTimeHeader = "Process-Start-Time-Unix"
 )
 
-var defaultEncodingOffers = []string{"gzip", "zstd"}
+type Compression int
+
+const (
+	Identity Compression = iota
+	Gzip
+	Zstd
+)
+
+var compressions = [...]string{
+	"identity",
+	"gzip",
+	"zstd",
+}
+
+func (c Compression) String() string {
+	return compressions[c]
+}
+
+var defaultCompressionFormats = []Compression{Identity, Gzip, Zstd}
 
 var gzipPool = sync.Pool{
 	New: func() interface{} {
@@ -168,46 +186,13 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		} else {
 			contentType = expfmt.Negotiate(req.Header)
 		}
-		header := rsp.Header()
-		header.Set(contentTypeHeader, string(contentType))
+		rsp.Header().Set(contentTypeHeader, string(contentType))
 
-		w := io.Writer(rsp)
-		if !opts.DisableCompression {
-			offers := defaultEncodingOffers
-			if len(opts.EncodingOffers) > 0 {
-				offers = opts.EncodingOffers
+		w, err := GetWriter(req, rsp, opts.DisableCompression, opts.OfferedCompressions)
+		if err != nil {
+			if opts.ErrorLog != nil {
+				opts.ErrorLog.Println("error getting writer", err)
 			}
-			// TODO(mrueg): Replace internal/github.com/gddo once https://github.com/golang/go/issues/19307 is implemented.
-			compression := httputil.NegotiateContentEncoding(req, offers)
-			switch compression {
-			case "zstd":
-				header.Set(contentEncodingHeader, "zstd")
-				// TODO(mrueg): Replace klauspost/compress with stdlib implementation once https://github.com/golang/go/issues/62513 is implemented.
-				z, err := zstd.NewWriter(rsp, zstd.WithEncoderLevel(zstd.SpeedFastest))
-				if err != nil {
-					return
-				}
-
-				z.Reset(w)
-				defer z.Close()
-
-				w = z
-			case "gzip":
-				header.Set(contentEncodingHeader, "gzip")
-				gz := gzipPool.Get().(*gzip.Writer)
-				defer gzipPool.Put(gz)
-
-				gz.Reset(w)
-				defer gz.Close()
-
-				w = gz
-			case "identity":
-				// This means the content is not encoded.
-			default:
-				// The content encoding was not implemented yet.
-				return
-			}
-
 		}
 
 		enc := expfmt.NewEncoder(w, contentType)
@@ -373,12 +358,19 @@ type HandlerOpts struct {
 	// no effect on the HTTP status code because ErrorHandling is set to
 	// ContinueOnError.
 	Registry prometheus.Registerer
-	// If DisableCompression is true, the handler will never compress the
-	// response, even if requested by the client.
+	// DisableCompression disables the response encoding (compression) and
+	// encoding negotiation. If true, the handler will
+	// never compress the response, even if requested
+	// by the client and the OfferedCompressions field is ignored.
 	DisableCompression bool
-	// If DisableCompression is false, this option will allow to define the
-	// set of offered encoding algorithms.
-	EncodingOffers []string
+	// OfferedCompressions is a set of encodings (compressions) handler will
+	// try to offer when negotiating with the client. This defaults to zstd,
+	// gzip and identity.
+	// NOTE: If handler can't agree on the encodings with the client or
+	// caller using unsupported or empty encodings in OfferedCompressions,
+	// handler always fallbacks to no compression (identity), for
+	// compatibility reasons. In such cases ErrorLog will be used if set.
+	OfferedCompressions []Compression
 	// The number of concurrent HTTP requests is limited to
 	// MaxRequestsInFlight. Additional requests are responded to with 503
 	// Service Unavailable and a suitable message in the body. If
@@ -425,4 +417,50 @@ func httpError(rsp http.ResponseWriter, err error) {
 		"An error has occurred while serving metrics:\n\n"+err.Error(),
 		http.StatusInternalServerError,
 	)
+}
+
+func GetWriter(r *http.Request, rsp http.ResponseWriter, disableCompression bool, offeredCompressions []Compression) (io.Writer, error) {
+	w := io.Writer(rsp)
+	rsp.Header().Set(contentEncodingHeader, "identity")
+	if !disableCompression {
+		offers := defaultCompressionFormats
+		if len(offeredCompressions) > 0 {
+			offers = offeredCompressions
+		}
+		var compressions []string
+		for _, comp := range offers {
+			compressions = append(compressions, comp.String())
+		}
+		// TODO(mrueg): Replace internal/github.com/gddo once https://github.com/golang/go/issues/19307 is implemented.
+		compression := httputil.NegotiateContentEncoding(r, compressions)
+		switch compression {
+		case "zstd":
+			rsp.Header().Set(contentEncodingHeader, "zstd")
+			// TODO(mrueg): Replace klauspost/compress with stdlib implementation once https://github.com/golang/go/issues/62513 is implemented.
+			z, err := zstd.NewWriter(rsp, zstd.WithEncoderLevel(zstd.SpeedFastest))
+			if err != nil {
+				return nil, err
+			}
+
+			z.Reset(w)
+			defer z.Close()
+
+			w = z
+		case "gzip":
+			rsp.Header().Set(contentEncodingHeader, "gzip")
+			gz := gzipPool.Get().(*gzip.Writer)
+			defer gzipPool.Put(gz)
+
+			gz.Reset(w)
+			defer gz.Close()
+
+			w = gz
+		case "identity":
+			// This means the content is not compressed.
+		default:
+			// The content encoding was not implemented yet.
+			return w, fmt.Errorf("content compression format not recognized: %s. Valid formats are: %s", compression, defaultCompressionFormats)
+		}
+	}
+	return w, nil
 }
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -190,7 +190,18 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		}
 		rsp.Header().Set(contentTypeHeader, string(contentType))
 
-		w, encodingHeader, err := NegotiateEncodingWriter(req, rsp, opts.DisableCompression, compressions)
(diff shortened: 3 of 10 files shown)


prometheus/promhttp/http.go
```diff
@@ -381,13 +398,13 @@ type HandlerOpts struct {
 	ProcessStartTime time.Time
 }
 
-// gzipAccepted returns whether the client will accept gzip-encoded content.
-func gzipAccepted(header http.Header) bool {
+// EncodingAccepted returns whether the client will accept encoded content.
+func EncodingAccepted(header http.Header, encoding string) bool {
```
bwplotka, line 402:
```suggestion
func IsEncodingAccepted(header http.Header, encoding string) bool {
```

prometheus/promhttp/http.go
```diff
@@ -169,15 +170,31 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
...
 
 		w := io.Writer(rsp)
-		if !opts.DisableCompression && gzipAccepted(req.Header) {
-			header.Set(contentEncodingHeader, "gzip")
-			gz := gzipPool.Get().(*gzip.Writer)
-			defer gzipPool.Put(gz)
+		if !opts.DisableCompression {
+			// Gzip takes precedence over zstd
+			// TODO(mrueg): Replace klauspost/compress with stdlib implementation once https://github.com/golang/go/issues/62513 is implemented.
+			if EncodingAccepted(req.Header, "zstd") {
```
bwplotka, line 176: Something is off here. Shouldn't we take whatever is first on "Accept-Encoding" (or what has the biggest q?).

Also as it's implemented here zstd takes precedence I think ;p

prometheus/promhttp/http_test.go
```diff
@@ -331,3 +331,147 @@ func TestHandlerTimeout(t *testing.T) {
...
+func TestEncodingAccepted(t *testing.T) {
+	testCases := []struct {
+		name         string
+		header       http.Header
+		encodingType string
+		expected     bool
+	}{
+		{
+			name:         "test with gzip accept-encoding",
+			header:       http.Header{"Accept-Encoding": {"gzip"}},
```
bwplotka, line 344: Please add more complex cases like:

```
Accept-Encoding: gzip;q=1.0, identity; q=0.5, *;q=0
Accept-Encoding: gzip;zstd
Accept-Encoding: zstd;gzip
```

bwplotka (changes requested):
Nice! I am supportive. I think adding zstd to some places in Prometheus turned out to be sometimes controversial due to higher CPU, but from the client side... we are happy to support more AS LONG as the dependency is not too problematic (zstd should be fine).

Thanks! I have some suggestions to the implementation though, see comments.

mrueg:
Thanks for the feedback!

I haven't thought about enhancing the Accept-Encoding header further and have replaced it now with an internal copy of the archived https://github.com/golang/gddo/blob/master/httputil/negotiate.go#L19 

I see prometheus is using an internal copy of goautoneg: https://github.com/prometheus/common/blob/main/internal/bitbucket.org/ww/goautoneg/autoneg.go  this might be an option to switch it to the gddo/httputil implementation and move all the code into prometheus/common.

Cross-linking these issues from internal TODOs:
https://github.com/golang/go/issues/19307
https://github.com/golang/go/issues/62513

prometheus/promhttp/http.go
mrueg: I've switched it to an actual parser that does the negotiation.

prometheus/promhttp/http.go
```diff
@@ -346,6 +376,9 @@ type HandlerOpts struct {
 	// If DisableCompression is true, the handler will never compress the
 	// response, even if requested by the client.
 	DisableCompression bool
+	// If DisableCompression is false, this option will allow to define the
+	// set of offered encoding algorithms.
+	EncodingOffers []string
```
bwplotka, line 381: I feel we should have a better name than "EncodingOffers". Offers is bit weird name (it makes sense from negotiation library POV), but maybe is better than "Accepted". "Encoding" is weird as we used "Compression" in the first field. There is also potential overlap functionality with DisableCompression. Quickly thinking.. maybe we should clarify all (this also proposes error handling)

```suggestion
	// DisableCompression disables the response encoding (compression) and
	// encoding negotiation. If true, the handler will 
	// never compress the response, even if requested
	// by the client and the OfferedCompressions field is ignored.
	DisableCompression bool
	// OfferedCompressions is a set of encodings (compressions) handler will
	// try to offer when negotiating with the client. This defaults to zstd, gzip, identity
	//
	// NOTE: If handler can't agree on the encodings with the client or caller using unsupported or empty encodings in OfferedCompressions, handler always fallbacks to
	// no compression (identity), for compatibility reasons. In such cases ErrorLog will be used if set.
        OfferedCompressions []encoding
```

prometheus/promhttp/http.go
```diff
@@ -169,15 +172,42 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
...
+				defer gz.Close()
 
-			gz.Reset(w)
-			defer gz.Close()
+				w = gz
+			case "identity":
+				// This means the content is not encoded.
+			default:
+				// The content encoding was not implemented yet.
+				return
```
bwplotka, line 208: What about following what I proposed in the other comment? To use opt.ErrorLog if not nil and continue without compression?

(This is for compatibility and caller error handling)

prometheus/promhttp/http_test.go
```diff
@@ -331,3 +331,95 @@ func TestHandlerTimeout(t *testing.T) {
 
 	close(c.Block) // To not leak a goroutine.
 }
+
+func BenchmarkEncoding(b *testing.B) {
```
bwplotka, line 335: Thanks! Do you mind adding correctness tests too (e.g. for weird OfferedCompressions values and zstd?)

prometheus/promhttp/http.go
```diff
@@ -54,6 +55,8 @@ const (
 	processStartTimeHeader = "Process-Start-Time-Unix"
 )
 
+var defaultEncodingOffers = []string{"gzip", "zstd"}
```
bwplotka, line 58: I propose we use enum here e.g `Encoding`, also add `identity` as a clear options caller can use for no compression as per RFC 9110

bwplotka (commented):
Beautiful work, had to find time to review it

> If the user of the client supplies an not-yet implemented compression, should this be visible via an error (and if so, how)?

Good question. First of all I would use enum so it's less likely it happens. Then.. I propose something in comment, should be good enough. WDYT?

Thanks! Almost there!

prometheus/promhttp/http_test.go
mrueg: I've refactored the code into a GetWriter() function to make it easier to test. I still need to add a few testcases and probably get things right with the response writer.

prometheus/promhttp/http.go
mrueg: GetWriter() throws an error that the handler catches and uses ErrorLog there.

mrueg:
@bwplotka if you have some spare cycles, it would be awesome if you could take another look :)

prometheus/promhttp/http.go
```diff
@@ -165,19 +186,13 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
...
-			gz.Reset(w)
-			defer gz.Close()
+		rsp.Header().Set(contentTypeHeader, string(contentType))
 
-			w = gz
+		w, err := GetWriter(req, rsp, opts.DisableCompression, opts.OfferedCompressions)
+		if err != nil {
+			if opts.ErrorLog != nil {
+				opts.ErrorLog.Println("error getting writer", err)
+			}
```
bwplotka, line 195: It would be nice to explictly do `w = io.Writer(rsp)`  and not rely on GetWriter implementation to do something NOT Go idiomatic (give valid response despite err != nil), WDYT?

prometheus/promhttp/http.go
```diff
@@ -343,9 +358,19 @@ type HandlerOpts struct {
 	// no effect on the HTTP status code because ErrorHandling is set to
 	// ContinueOnError.
 	Registry prometheus.Registerer
-	// If DisableCompression is true, the handler will never compress the
-	// response, even if requested by the client.
+	// DisableCompression disables the response encoding (compression) and
+	// encoding negotiation. If true, the handler will
+	// never compress the response, even if requested
+	// by the client and the OfferedCompressions field is ignored.
```
bwplotka, line 364:
```suggestion
	// by the client and the OfferedCompressions field is set.
```

prometheus/promhttp/http.go
```diff
@@ -343,9 +358,19 @@ type HandlerOpts struct {
...
 	DisableCompression bool
+	// OfferedCompressions is a set of encodings (compressions) handler will
+	// try to offer when negotiating with the client. This defaults to zstd,
+	// gzip and identity.
+	// NOTE: If handler can't agree on the encodings with the client or
+	// caller using unsupported or empty encodings in OfferedCompressions,
+	// handler always fallbacks to no compression (identity), for
+	// compatibility reasons. In such cases ErrorLog will be used if set.
```
bwplotka, line 372:
```suggestion
	// OfferedCompressions is a set of encodings (compressions) handler will
	// try to offer when negotiating with the client. This defaults to identity, gzip
	// and zstd.
	// NOTE: If handler can't agree with the client on the encodings or
	// unsupported or empty encodings are set in OfferedCompressions,
	// handler always fallbacks to no compression (identity), for
	// compatibility reasons. In such cases ErrorLog will be used if set.
```

prometheus/promhttp/http.go
```diff
@@ -406,3 +418,49 @@ func httpError(rsp http.ResponseWriter, err error) {
 		http.StatusInternalServerError,
 	)
 }
+
+func GetWriter(r *http.Request, rsp http.ResponseWriter, disableCompression bool, offeredCompressions []Compression) (io.Writer, error) {
```
bwplotka, line 422:
```suggestion
func NegotiateEncodingWriter(r *http.Request, rsp http.ResponseWriter, disableCompression bool, offeredCompressions []Compression) (io.Writer, error) {
```

prometheus/promhttp/http.go
```diff
@@ -406,3 +418,49 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+	w := io.Writer(rsp)
+	rsp.Header().Set(contentEncodingHeader, "identity")
+	if !disableCompression {
```
bwplotka, line 425: let's return `w` early if disable compression?

(context: https://thanos.io/tip/contributing/coding-style-guide.md/#control-structure-prefer-early-returns-and-avoid-else)

prometheus/promhttp/http.go
```diff
@@ -406,3 +418,49 @@ func httpError(rsp http.ResponseWriter, err error) {
 		http.StatusInternalServerError,
 	)
 }
+
+func GetWriter(r *http.Request, rsp http.ResponseWriter, disableCompression bool, offeredCompressions []Compression) (io.Writer, error) {
+	w := io.Writer(rsp)
+	rsp.Header().Set(contentEncodingHeader, "identity")
```
bwplotka, line 424: This is done by caller, could be deleted?

prometheus/promhttp/http.go
bwplotka, line 422: (non blocking suggestion): To improve reliability of this function I would suggest returning headers callers will set (on success) instead of modifying `http.ResponseWriter` (reduce side effects). To ensure this I would also change rsp to io.Writer, so:

```suggestion
func NegotiateEncodingWriter(r *http.Request, rw io.Writer, disableCompression bool, offeredCompressions []Compression) (_ io.Writer, encodingHeaderValue string, _ error) {
```

prometheus/promhttp/http.go
```diff
@@ -406,3 +418,49 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+	if !disableCompression {
+		offers := defaultCompressionFormats
+		if len(offeredCompressions) > 0 {
+			offers = offeredCompressions
```
bwplotka, line 428: This should likely be done not in handler but when we create it. Somewhere here:

image

(unnecessary "handler" work that could be precomputed)

prometheus/promhttp/http.go
```diff
@@ -406,3 +418,49 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+		}
+		var compressions []string
+		for _, comp := range offers {
+			compressions = append(compressions, comp.String())
+		}
```
bwplotka, line 433: Same as above

prometheus/promhttp/http.go
```diff
@@ -54,6 +55,26 @@ const (
...
+
+var compressions = [...]string{
+	"identity",
+	"gzip",
+	"zstd",
+}
+
+func (c Compression) String() string {
+	return compressions[c]
+}
```
bwplotka, line 74: If we use `string` instead of `iota` we don't need this code, right? (:

prometheus/promhttp/http.go
```diff
@@ -406,3 +418,49 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+		case "zstd":
+			rsp.Header().Set(contentEncodingHeader, "zstd")
+			// TODO(mrueg): Replace klauspost/compress with stdlib implementation once https://github.com/golang/go/issues/62513 is implemented.
+			z, err := zstd.NewWriter(rsp, zstd.WithEncoderLevel(zstd.SpeedFastest))
+			if err != nil {
+				return nil, err
+			}
+
+			z.Reset(w)
+			defer z.Close()
```
bwplotka, line 446: This defer is wrong isn't it? We want to close it on handler finished work? 🤔 

We need another return arg for closer (that's fine). 

But... I am worried our tests didn't catch that. Can we fix tests BEFORE we fix this to fail? (:

prometheus/promhttp/http_test.go
```diff
@@ -331,3 +331,162 @@ func TestHandlerTimeout(t *testing.T) {
 
 	close(c.Block) // To not leak a goroutine.
 }
+
+func TestGetWriter(t *testing.T) {
```
bwplotka, line 335: This is good, but I believe we need to improve bigger tests to at least use gzip on top of this. If we would merge this code it would break ppl I think (wrong defer scope) 🙈

bwplotka (changes requested):
Thanks!!! I have more time this week, so feel free to ping me on Slack/here for next iteration. I think it's better, but added some suggestion. Plus I think this code won't work and our tests didn't catch this bug. (easy to fix both!), so putting "request changes" status. Otherwise LGTM!

Thanks for patience and good work. We need this! 💪🏽

ArthurSens:
Apologies for the noise here @mrueg!

@SachinSahu431, what do you think about not adding comments when everything is fine in the PR description? (Let's discuss this in our sync)

prometheus/promhttp/http.go
mrueg: Good idea, I changed that

prometheus/promhttp/http.go
mrueg: Yes, you're right! Thanks!

prometheus/promhttp/http.go
mrueg: Good point, I moved it to the handler

prometheus/promhttp/http.go
mrueg: I might wanna re-use this function within kube-state-metrics (currently uses the acceptEncoding function but copy'n'pasted). I made it work more nicely using the string enum.

prometheus/promhttp/http.go
mrueg: Renamed it, thanks that's a better name!

prometheus/promhttp/http.go
mrueg: Good point, the function doesn't return a writer anymore in case there's an error.

prometheus/promhttp/http.go
mrueg: Yes, you're right. Let me see how I can add this to the tests.

prometheus/promhttp/http.go
mrueg: Refactored it to make rsp side-effect free. This is not done by the caller. :)

prometheus/promhttp/http_test.go
mrueg: I added more tests that test the whole compression with gzip and zstd, let me know if that is sufficient.

prometheus/promhttp/http.go
```diff
@@ -190,7 +190,18 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		}
 		rsp.Header().Set(contentTypeHeader, string(contentType))
 
-		w, encodingHeader, err := NegotiateEncodingWriter(req, rsp, opts.DisableCompression, compressions)
+		w, encodingHeader, closeWriter, err := NegotiateEncodingWriter(req, rsp, opts.DisableCompression, compressions)
+
+		if closeWriter != nil {
+			defer func() {
```
mrueg, line 196: @bwplotka Test will fail if you comment out this defer. 
Needed to do it that way instead of calling the function directly for golangci-lint's errcheck.

prometheus/promhttp/http.go
mrueg: So I added a test that allowed me to check if it fails: https://github.com/prometheus/client_golang/pull/1496/commits/99037c19d91788aa53b38ace079e6d20d08f1622#r1625073867

mrueg:
@bwplotka I believe I addressed everything, so this is ready for another round of review. :) 
Appreciate the feedback, helped me to ramp up my golang quite a bit.

prometheus/promhttp/http.go
```diff
@@ -122,6 +133,18 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		}
 	}
 
+	// Select all supported compression formats
```
bwplotka, line 136:
```suggestion
	// Select compression formats to offer based on default or user choice.
```

prometheus/promhttp/http.go
bwplotka: I think there is potentially better way (do `_ := ` if we don't care about this err, or pack opt.Err log in helper func), but that's already good enough, Thanks!

prometheus/promhttp/http.go
```diff
@@ -165,20 +188,32 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
...
+			defer func() {
+				err := closeWriter()
+				if err != nil {
+					if opts.ErrorLog != nil {
+						opts.ErrorLog.Println("error closing writer:", err)
+					}
+				}
+			}()
+		}
+		if err != nil {
```
bwplotka, line 205: Technically no defer is needed on this error, no? So we can move closeWriter to after err != nil and skip closerWriter if nil check?

prometheus/promhttp/http.go
```diff
@@ -165,20 +188,32 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
...
+			if opts.ErrorLog != nil {
+				opts.ErrorLog.Println("error getting writer", err)
+			}
+			// Since the writer received from NegotiateEncodingWriter will be nil, in case there's an error, we set it here
```
bwplotka, line 209: No need for this comment TBH

prometheus/promhttp/http.go
```diff
@@ -165,20 +188,32 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
...
+			w = io.Writer(rsp)
+		}
 
-			w = gz
+		if encodingHeader == "" {
```
bwplotka, line 213: After err check we expect this to be always valid, can we ensure in func?

bwplotka (approved):
Amazing, thank you!

I think this is good enough to merge in this state. I think there are some things e.g. number of if statements to handle response from Negotiate yields more code vs if we just inline this function (: But I think there might be ways to simplify this, mostly for readability. 

Happy to merge, and allow others and myself to simplify bit in another PR,  or propose PR to your PR, what do you prefer? Or if you want you could try to simplify too if you know what I mean (added some suggestions), but happy to help here to save your time! 🤗  

Thanks!

prometheus/promhttp/http.go
mrueg: This one isn't the deferred one, this is the one from NegotiateEncodingWriter. I reordered it for better readability.

prometheus/promhttp/http.go
mrueg: Removed it. :)

prometheus/promhttp/http.go
mrueg: Good idea, I moved it into the err check.

mrueg:
> Amazing, thank you!
> 
> I think this is good enough to merge in this state. I think there are some things e.g. number of if statements to handle response from Negotiate yields more code vs if we just inline this function (: But I think there might be ways to simplify this, mostly for readability.
> 
> Happy to merge, and allow others and myself to simplify bit in another PR, or propose PR to your PR, what do you prefer? Or if you want you could try to simplify too if you know what I mean (added some suggestions), but happy to help here to save your time! 🤗
> 
> Thanks!


Can we keep the NegotiateEncodingWriter func separate and not inline it? I want to reuse it in kube-state-metrics to avoid code duplication there (unfortunately we cannot fully use ServeHTTP there).

If there are any other bits that could make the code simpler and easier to read, please let me know, happy to implement them as well.

Do you want me to squash the commits into a single one?

prometheus/promhttp/http.go
bwplotka: Still not resolved, plus we can kill this if != nil check for close Writer I assume (let's ensure `negotiateEncodingWriter` always sets it, even if noop).

prometheus/promhttp/http.go
```diff
@@ -406,3 +435,46 @@ func httpError(rsp http.ResponseWriter, err error) {
 		http.StatusInternalServerError,
 	)
 }
+
+// NegotiateEncodingWriter reads the Accept-Encoding header from a request and
+// selects the right compression based on an allow-list of supported
+// compressions. It returns a writer implementing the compression and an the
+// correct value that the caller can set in the response header.
+func NegotiateEncodingWriter(r *http.Request, rw io.Writer, disableCompression bool, compressions []string) (_ io.Writer, encodingHeaderValue string, closeWriter func() error, _ error) {
```
bwplotka, line 443: This is good, but I think we can't make it public. I know you wanted to reuse this in other projects (e.g. KSM), but this is not ideal for client_golang. Imagine we will have to maintain random HTTP semantic convention logic until v2 for everybody, including that internal vendored package... which is not that internal, because we have exported function that uses it at the end 🙈 

Sorry... For KSM, I would copy for now (A little copying is better than a little dependency](https://go-proverbs.github.io/#:~:text=A%20little%20copying%20is%20better%20than%20a%20little%20dependency)), OR consider adding this to prometheus/common OR separate repo/project.


Second suggestion - we can kill disableCompression and use empty `compressions` for this case.
3rd suggestion - closeWriter does not need to return error (easier to use)

```suggestion
func negotiateEncodingWriter(r *http.Request, rw io.Writer, compressions []string) (_ io.Writer, encodingHeaderValue string, closeWriter func(), _ error) {
```

prometheus/promhttp/http.go
```diff
@@ -406,3 +435,46 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+	w := rw
+
+	if disableCompression {
```
bwplotka, line 446:
```suggestion
	if len(compressions) == 0 {
```

prometheus/promhttp/http.go
```diff
@@ -406,3 +435,46 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+		return w, string(Identity), nil, nil
```
bwplotka, line 447:
```suggestion
		return w, string(Identity), func() {}, nil
```

prometheus/promhttp/http.go
```diff
@@ -406,3 +435,46 @@ func httpError(rsp http.ResponseWriter, err error) {
 		http.StatusInternalServerError,
 	)
 }
+
+// NegotiateEncodingWriter reads the Accept-Encoding header from a request and
+// selects the right compression based on an allow-list of supported
+// compressions. It returns a writer implementing the compression and an the
+// correct value that the caller can set in the response header.
+func NegotiateEncodingWriter(r *http.Request, rw io.Writer, disableCompression bool, compressions []string) (_ io.Writer, encodingHeaderValue string, closeWriter func() error, _ error) {
+	w := rw
```
bwplotka, line 444: Hm.. I feel this can be removed? We can return those `z` and `gz` things?

prometheus/promhttp/http.go
```diff
@@ -406,3 +435,46 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+
+	if disableCompression {
+		return w, string(Identity), nil, nil
+	}
+
+	// TODO(mrueg): Replace internal/github.com/gddo once https://github.com/golang/go/issues/19307 is implemented.
+	compression := httputil.NegotiateContentEncoding(r, compressions)
```
bwplotka, line 451:
```suggestion
	selected := httputil.NegotiateContentEncoding(r, compressions)
```

Perhaps more readable?

prometheus/promhttp/http.go
```diff
@@ -406,3 +435,46 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+		// TODO(mrueg): Replace klauspost/compress with stdlib implementation once https://github.com/golang/go/issues/62513 is implemented.
+		z, err := zstd.NewWriter(rw, zstd.WithEncoderLevel(zstd.SpeedFastest))
+		if err != nil {
+			return nil, "", nil, err
+		}
+
+		z.Reset(w)
+		w = z
+
+		return w, compression, z.Close, nil
```
bwplotka, line 464:
```suggestion
		return z, selected, func() { _ = z.Close() }, nil
```

prometheus/promhttp/http.go
```diff
@@ -406,3 +435,46 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+	case "gzip":
+		gz := gzipPool.Get().(*gzip.Writer)
+		defer gzipPool.Put(gz)
```
bwplotka, line 467: This has to be done on Close. It's a subtle bug, but concurrent Negotiate calls might corrupt gzip writers (hard to repro on tests I guess). "Put" means "I won't use gz anymore", which will happen on return of this function. But that's not true is it?

prometheus/promhttp/http.go
```diff
@@ -406,3 +435,46 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+
+		gz.Reset(w)
+
+		w = gz
+		return w, compression, gz.Close, nil
```
bwplotka, line 472:
```suggestion
		return gz, selected, func() { _ = gz.Close(); gzipPool.Put(gz)}, nil
```

prometheus/promhttp/http.go
```diff
@@ -406,3 +435,46 @@ func httpError(rsp http.ResponseWriter, err error) {
...
+	case "identity":
+		// This means the content is not compressed.
+		return w, compression, nil, nil
```
bwplotka, line 475:
```suggestion
		return w, selected, func() {}, nil
```

bwplotka (commented):
Great work! Let's keep going then ;p 

I think I found one more bug 🙈 And added some suggestions inline. No need to squash, we will do it while merging. Function can be separate, added suggestions, but I think we have to make it private... sorry.  Exporting `NegotiateEncodingWriter` would be not ideal. Commented why.

prometheus/promhttp/http.go
mrueg: Sounds good, I added the empty functions you suggested also if an error is returned.

prometheus/promhttp/http.go
mrueg: Good catch! I was wondering about that as well. Thanks for the suggestions on returning a function here that calls both. TIL :)

prometheus/promhttp/http.go
mrueg: Yes, removed it. :)

prometheus/promhttp/http.go
mrueg: I understand your points and changed it to private
Do you think prometheus/common would be open to include it?

I saw the internalized autoneg dependency https://github.com/prometheus/common/pull/625 got replaced recently and could probably be replaced by gddo/httputil (but that would mean to move it internally as well).

bwplotka:
Are on CNCF Slack @mrueg ? I DM-ed you there (:

prometheus/promhttp/http.go
bwplotka: I think you accidently left this defer here though.

prometheus/promhttp/http.go
```diff
@@ -54,6 +55,16 @@ const (
 	processStartTimeHeader = "Process-Start-Time-Unix"
 )
 
+type Compression string
```
bwplotka, line 58: One last missing thing is commentary here, something like 

```
// Compression represents the content encodings handlers support for the HTTP
// responses.
```

prometheus/promhttp/http.go
```diff
@@ -406,3 +426,40 @@ func httpError(rsp http.ResponseWriter, err error) {
 		http.StatusInternalServerError,
 	)
 }
+
+// NegotiateEncodingWriter reads the Accept-Encoding header from a request and
```
bwplotka, line 430:
```suggestion
// negotiateEncodingWriter reads the Accept-Encoding header from a request and
```

internal/github.com/golang/gddo/httputil/negotiate.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/http"
+
+	"github.com/prometheus/client_golang/internal/github.com/golang/gddo/httputil/header"
+)
+
+// NegotiateContentEncoding returns the best offered content encoding for the
+// request's Accept-Encoding header. If two offers match with equal weight and
+// then the offer earlier in the list is preferred. If no offers are
+// acceptable, then "" is returned.
+func NegotiateContentEncoding(r *http.Request, offers []string) string {
```
bwplotka, line 19: We have another problem.. Seems like this code somehow is exported to outside modules. Just tried using it in another module and managed to succeed... unfortunately. Somehow putting that under internal didn't help here 🤔 

image

image

image

We have to debug this a bit first 🙈

bwplotka (commented):
Nice! Last nits!

internal/github.com/golang/gddo/httputil/negotiate.go
bwplotka: I wonder... is this because we have some more directories under internal? We might want to read deeper how this work? Maybe it's because I use submodule (it should not matter in theory).

bwplotka: Looks like it's expected as https://pkg.go.dev/cmd/go#hdr-Internal_Directories states. (:

bwplotka (approved):
Thanks! Let's try this out 💪🏽 

We had some internal discussion to move vendor and negotiating part to https://github.com/prometheus/common/blob/main/expfmt/encode.go#L65 -- but that can be done without breaking changes. Thanks, good work!

mrueg:
Thanks for taking the time to review the PR and the good suggestions to improve the quality of the code :tada:

go.mod
```diff
@@ -5,7 +5,9 @@ go 1.20
 require (
 	github.com/beorn7/perks v1.0.1
 	github.com/cespare/xxhash/v2 v2.3.0
+	github.com/google/go-cmp v0.6.0
 	github.com/json-iterator/go v1.1.12
+	github.com/klauspost/compress v1.17.8
```
ash2k, line 10: Came here from the release page. Most people don't care about zstd, but now every user will have one more dependency. Would be better to wait for https://github.com/golang/go/issues/62513 and avoid extra dependencies.

Why not let stdlib's http package handle compression negotiation? It's kind of out of scope of this library (from my POV at least).

bwplotka: It's fair point we could outsource compression + negotiation to separate libs, but we wanted default client_golang without extra deps to handle that automatically for everyone. That's some choice which was already made before by adding `gzip`.

> Most people don't care about zstd, but now every user will have one more dependency.

Fair point, hopefully we can remove it once zstd is in std. Luckly, this dep is trusted, proven and very minimal https://github.com/klauspost/compress/blob/master/go.mod - we could vendor that code, but we chosen to add one little dep -- we are open on feedback on how blocking this is for our users.

> stdlib's http package handle compression negotiation

Is there std http flow for compression negotiation? I don't think so.

ash2k: > Is there std http flow for compression negotiation? I don't think so.

Yes, it turns out it doesn't handle that (I didn't know). My point is that there are many things that can be improved and done that will improve how this library works (bigger picture). But this doesn't mean they should be part of this library. Imagine every library that provides e.g. http handlers also tried to take care of compression, compression negotiation, doing http/3, etc, etc, etc. Dependency sprawl quickly gets out of hand, things work differently, become incompatible. I think unix way is a better approach. Provide solution for one problem, limit the scope, reduce dependencies. But also make it possible for your consumers to solve those problems somehow (composition!). With other libraries or with stdlib or with custom code.

I wonder if it's possible to has zstd compression without this PR? Can I do it today with a previous version of this library? It seems it should be possible to have an http handler that wraps the handler of this library (or any other one) and implements compression. Perhaps this should be a separate generic library that anyone could use, contribute too. I found [the chi middleware](https://pkg.go.dev/github.com/go-chi/chi/middleware#Compress) that does gzip compression. Sould be possible to have a zstd middleware? And then anyone who wants it can use it for all/any http handlers they have, not just for the metrics endpoint.

ash2k: > we wanted default client_golang without extra deps to handle that automatically for everyone.

May I ask what is the motivation for this? How many users of this library have asked for this feature?

If something can be done (even if easily), doesn't mean it should be done. My gut feeling is that it is cheaper (cpu/ram/network resources) to pass uncompressed metrics to a scraper that is running in the same local network (99% of people do this?).  If things are running on Kubernetes, the scraper is quite likely on the same host (as a DaemonSet)! So it's "free", virtually/almost no network cost. In that case compression is a pure waste of CPU time. Shall we add some code that detects this situation and optimizes the compression away? 😁 Why not, just a little dependency on the Kubernetes client? Let's make the library handle this case too! Ok, I hope you see my point.

p.s. I'm very grateful to people who contribute and maintain great libraries like this one. Please don't take this personally.

bwplotka: I def see your point. We all learn the hard way and the "golden default/opt-out" pressure is not helping. Thanks!

We could have done this differently, I wonder if we shouldn't vendor this ASAP to mitigate Kubernetes concerns. https://github.com/kubernetes/kubernetes/pull/130569#discussion_r1981355469

ash2k: > I wonder if we shouldn't vendor this ASAP to mitigate Kubernetes concerns

Why not revert the change and let users decide if they want compression by wrapping your handler with a middleware (that lives in a separate library OR a separate Go module in this library so people don't get the dependency if they don't want)? HTTP compression is orthogonal to the purpose of this library, it shouldn't be part of it.

- How many users of this library asked for this feature?
- Servers are connected using 40-100-200-400 gbps networks. It's often cheaper to pass data uncompressed vs spending CPU time on compression and decompression. Metrics endpoint and scraper are almost always in the same AZ/DC (if not the same host!) - what's the point of compression here? Anyone done any real-world benchmarks? How do you know this PR is not a regression of overall resource usage?

My friendly two cents :)

bwplotka: Ideally we don't revert, because we try our best not to break users and we don't know how many rely on this already (release in 1.20).

Vendoring a stripped version feels plausible https://github.com/prometheus/client_golang/pull/1767

Issue #1490, opened 2024-04-02
Remove go-spew dependency


anurag-deshpande:
[Problem Description]
We are consuming `prometheus/client_golang v1.19`, which is latest as of date. This package is scanned for security vulnerabilities and EOLs by **blackduck scanner** at our source.

The blackduck scanner has identified a `Project EOL` component, `go-spew20180930-snapshot-d8f796af`, which is a **transitive dependency** of  `prometheus/client_golang v1.19`. This project is not maintained and thus EOLed much earlier.

[Request]
We wish to consume all the dependencies which are non-EOLed, to maintain good coding practices. Can this EOLed component be updated by client_golang contributors or replaced with some alternative with similar functionality, to reduce the EOL risk?

Let me know if any more information is needed for this issue.

kakkoyun:
Thanks for opening the issue. 

When I checked, I saw that the mentioned dependency is not used anywhere critical. It is only used in tests and tutorial apps. https://github.com/search?q=repo%3Aprometheus%2Fclient_golang%20spew&type=code

That being said, we'd be happy to remove that dependency. 

Contributions are very welcome 😊

dongjiang1989:
@kakkoyun I'll add PR to fix it

PR #1499, opened 2024-04-12
feat(dependency): replace go-spew package 

dongjiang1989:
use [godebug](https://pkg.go.dev/github.com/kylelemons/godebug@v1.1.0/diff) package replace [go-spew](https://github.com/davecgh/go-spew) package 

fix: #1490


dongjiang1989:
@kakkoyun  PTAL

/cc @anurag-deshpande

kakkoyun (commented):
Thanks a lot for handling this.

I wonder how hard this is even without using a transitive dependency.

Could you maybe give it another shot with the standard library testing package?

dongjiang1989:
> Thanks a lot for handling this.
> 
> I wonder how hard this is even without using a transitive dependency.
> 
> Could you maybe give it another shot with the standard library testing package?

Thanks for your review.  Hmmm... Maybe not Diff string with the standard library testing package.
 
Like:
```
-(bytes.Buffer) # HELP some_other_metric A value that represents a counter.
-# TYPE some_other_metric counter
-some_other_metric{label1="value1"} 1
+(bytes.Buffer) # HELP some_total A value that represents a counter.
+# TYPE some_total counter
+some_total{label1="value1"} 1
```

prometheus/testutil/testutil_test.go
```diff
@@ -300,34 +300,20 @@ func TestMetricNotFound(t *testing.T) {
...
-`
-
 	err := CollectAndCompare(c, strings.NewReader(metadata+expected))
 	if err == nil {
 		t.Error("Expected error, got no error.")
 	}
 
-	if err.Error() != expectedError {
-		t.Errorf("Expected\n%#+v\nGot:\n%#+v", expectedError, err.Error())
+	if err.Error() == "" {
```
kakkoyun, line 315: This is basically no-op. You are trying to pretty print an empty string.

prometheus/testutil/testutil_test.go
```diff
@@ -300,34 +300,20 @@ func TestMetricNotFound(t *testing.T) {
...
 		},
 	})
+
 	c.Inc()
 
 	expected := `
 		some_other_metric{label1="value1"} 1
 	`
 
-	expectedError := `
```
kakkoyun, line 309: We need to keep the form of this expected error. There is a reason that the original author put this here. They wanted to check for specific diff here.

kakkoyun (changes requested):
It's great that the test are green now. But we are losing some information.

I'd suggest adding the following file with proper licence headers to the `testutils` package https://github.com/kylelemons/godebug/blob/master/diff/diff.go

https://pkg.go.dev/github.com/kylelemons/godebug@v1.1.0/diff

And updating the tests accordingly.

Or go back the previous version where you used `go-cmp`.

In any case, good job. You're close :)

prometheus/testutil/testutil_test.go
dongjiang1989: Got it. Fixed

dongjiang1989:
> It's great that the test are green now. But we are losing some information.
> 
> I'd suggest adding the following file with proper licence headers to the `testutils` package https://github.com/kylelemons/godebug/blob/master/diff/diff.go
> 
> https://pkg.go.dev/github.com/kylelemons/godebug@v1.1.0/diff
> 
> And updating the tests accordingly.
> 
> Or go back the previous version where you used `go-cmp`.
> 
> In any case, good job. You're close :)

Thanks for your review. Please recheck  :)

kakkoyun (commented):
Great job 👏 

Only one last thing.

> I'd suggest adding the following file with proper licence headers to the testutils package https://github.com/kylelemons/godebug/blob/master/diff/diff.go

In my previous comment, I suggested copying over the `diff.go` file from this repo rather than adding as a dependency. Remember our goal is to reduce the dependencies. Since it's a single file and simple enough. We can just copy the files (with licence headers) into the `testutils` package.

dongjiang1989:
> Great job 👏
> 
> Only one last thing.
> 
> > I'd suggest adding the following file with proper licence headers to the testutils package https://github.com/kylelemons/godebug/blob/master/diff/diff.go
> 
> In my previous comment, I suggested copying over the `diff.go` file from this repo rather than adding as a dependency. Remember our goal is to reduce the dependencies. Since it's a single file and simple enough. We can just copy the files (with licence headers) into the `testutils` package.

Got it. Thanks. Fixed

kakkoyun (approved):

PR #1503, opened 2024-04-29
Add CollectAndFormat to testutil, allowing caller to assert as they want to on the exported metric 

jcass8695:
CollectAndFormat is a helper function that returns the formatted metrics to the caller, allowing them to use it how they want. This is different to CollectAndCompare where the comparison is done strictly on behalf of the caller. Often it is more convenient to perform a simple substring match on the formatted metric rather than the full export. This is especially true testing a duration-like gauge metric, where the exact value of the metric might not be easily determined up front.

### DIFF
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -183,9 +183,8 @@ func ScrapeAndCompare(url string, expected io.Reader, metricNames ...string) err
 	return compareMetricFamilies(scraped, wanted, metricNames...)
 }
 
-// CollectAndCompare registers the provided Collector with a newly created
-// pedantic Registry. It then calls GatherAndCompare with that Registry and with
-// the provided metricNames.
+// CollectAndCompare collects the metrics identified by `metricNames` and compares them in the Prometheus text
+// exposition format to the data read from expected.
 func CollectAndCompare(c prometheus.Collector, expected io.Reader, metricNames ...string) error {
 	reg := prometheus.NewPedanticRegistry()
 	if err := reg.Register(c); err != nil {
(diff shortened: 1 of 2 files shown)


jcass8695:
@ArthurSens @bwplotka @kakkoyun if you have time would you mind taking a look at this please? Do you you think it's something that would be useful?

jcass8695:
@ArthurSens @bwplotka @kakkoyun bumping this PR. Happy to let it fall into your backlog of reviews, but a 👍🏻 or something to show you've seen it would be much appreciated, thank you 🙏🏻 .

bwplotka:
Checking now, sorry for lag!

prometheus/testutil/testutil.go
```diff
@@ -221,6 +221,33 @@ func TransactionalGatherAndCompare(g prometheus.TransactionalGatherer, expected
 	return compareMetricFamilies(got, wanted, metricNames...)
 }
 
+// CollectAndFormat registers the provided Collector with a newly created pedantic Registry. It then calls Gather with
+// that Registry and filters by the provided metricNames. It writes the gathered metrics to a bugger formatted according
+// to the provided format argument
```
bwplotka, line 226: Makes sense. I will be controversial here and suggest we stop doing so verbose comments. This comment explains implementation. I know you copied that for consistency, but we have to start somewhere. What about focusing on contract and usability. Implementation is trivial and self-described by the code. E.g.:

```suggestion
// CollectAndFormat collects the given `metricNames` metrics using pedantic registry and returns in the given format.
```

Do you mind fixing `GatherAndCompare` in the same manner, if that's ok with you?

prometheus/testutil/testutil.go
```diff
@@ -221,6 +221,33 @@ func TransactionalGatherAndCompare(g prometheus.TransactionalGatherer, expected
...
+func CollectAndFormat(c prometheus.Collector, format expfmt.FormatType, metricNames ...string) (*bytes.Buffer, error) {
```
bwplotka, line 227: Any reason to NOT return bytes or string instead? I would vote for string, it does not matter much. Buffer is weird, usually when you want things to be in buffer, we would need to pass io.Writer in args or something.

bwplotka (approved):
Nice work, thanks! Some suggestions, otherwise LGTM!

prometheus/testutil/testutil.go
jcass8695: Yep, I'm OK with that

prometheus/testutil/testutil.go
jcass8695: Nope, no reason I don't think. Although I don't totally understand this:

> we would need to pass io.Writer in args or something

Why is this the case?

jcass8695: Perhaps it should be bytes, as the requested format can be protobuf?

jcass8695:
@bwplotka would it be possible to cut a release that includes this change please?

PR #1511, opened 2024-05-08
improved code more clean

lilijreey:
Change Registry functions check label names dump code style; use two if replace if/else if /else statement； @ArthurSens @bwplotka @kakkoyun


bwplotka (approved):
I like it. Thanks!

PR #1513, opened 2024-05-09
:zap: http client defer CloseIdleConnections

cuisongliu:
when high concurrency occurs, the connection is full and the connection is rejected.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -79,6 +79,10 @@ type Client interface {
 	Do(context.Context, *http.Request) (*http.Response, []byte, error)
 }
 
+type CloseIdler interface {
+	CloseIdleConnections()
+}
+
 // NewClient returns a new Client.
 //
 // It is safe to use the returned Client from multiple goroutines.
@@ -116,6 +120,10 @@ func (c *httpClient) URL(ep string, args map[string]string) *url.URL {
 	u.Path = p
 
 	return &u
+}
+
+func (c *httpClient) CloseIdleConnections() {
+	c.client.CloseIdleConnections()
 }
 
 func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response, []byte, error) {


api/client.go
```diff
@@ -122,6 +122,7 @@ func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response,
 	if ctx != nil {
 		req = req.WithContext(ctx)
 	}
+	defer c.client.CloseIdleConnections()
```
bwplotka, line 125: Are you sure this is a correct solution? 

My understanding is that this closes connections which might be REUSED on the subsequent `Do` request. So you have to establish them again (overhead)?

Source: https://forum.golangbridge.org/t/when-should-i-use-client-closeidleconnections/19254, https://stackoverflow.com/questions/18697290/apache-httpclient-how-to-auto-close-connections-by-servers-keep-alive-time

bwplotka (changes requested):
Hm, my understanding is that this change would only makes things slower, will not help with too many concurrent connections (if you mean that).

Perhaps a reproduction test would help? Perhaps there is some http transport settings you can adjust, or you hit some other limits?

api/client.go
cuisongliu: Currently, doing an HTTP probe requiring allocating a new http.Client and http.Transport, and the underlying connections managed by http.Transport might be ESTABLISHED indefinitely if no parties (client or server) directly close them. As a result, if the probes continuously happen, we can see TCP connections opened by one process
drastically increase.

I understand that there may be such a problem in high concurrency. Even if it is reused, the connection is limited and cannot meet the requirements. In this case, the connection may be occupied and new connections cannot be applied for. As a result, the connection request cannot be continued. 

PS. I appeared in a high-concurrency scenario, with about 10000  of concurrencies.

bwplotka: Nice! What "HTTP probe" do you mean? Is it some OSS code or your own app?

Generally it sounds you might use this client incorrectly if you create new client for single Do, assuming you can share the client.

For cases you cannot share e.g. it's serverless or it's on-off request to adhoc destination, I think you can access [Config.Client](https://github.com/prometheus/client_golang/blob/e74129ab0f4c33b6072bf62b5e3bf550371b5f5d/api/client.go#L46) (which is used to create API client) and invoke this method after `Do` on your own, which would be a correct thing to do 👍🏽 

I am also open to create `**CloseIdleConnections()**` method on our client with a clear commentary when it's useful to make it easier (: 

Doing this on every Do for *everybody* is incorrect.

cuisongliu: in this conde add Close func ? 
```
// Client is the interface for an API client.
type Client interface {
	URL(ep string, args map[string]string) *url.URL
	Do(context.Context, *http.Request) (*http.Response, []byte, error)
}
```

I think it would be more reasonable if the interface or open functions could be adjusted and allowed to be called by users themselves. Very useful to me too.

cuisongliu: @bwplotka  i am add code to interface  using CloseIdleConnections() func . plz review this code ?

bwplotka: Why do we need interface change? Can we remove this defer? Otherwise new method is good 👍🏽

api/client.go
```diff
@@ -75,6 +75,7 @@ func (cfg *Config) validate() error {
 
 // Client is the interface for an API client.
 type Client interface {
+	CloseIdleConnections()
```
bwplotka, line 78: Let's move it down (as 3rd in order)

api/client.go
bwplotka: Ok interface change is needed. Let's kill the defer statement here and LGTM!

api/prometheus/v1/api.go
```diff
@@ -1342,6 +1342,7 @@ type Warnings []string
 // apiClient wraps a regular client and processes successful API responses.
 // Successful also includes responses that errored at the API level.
 type apiClient interface {
```
bwplotka, line 1344: Do we need this change on both interfaces really? Ideally those are very small.

api/prometheus/v1/api.go
```diff
@@ -1342,6 +1342,7 @@ type Warnings []string
...
+	CloseIdleConnections()
```
bwplotka, line 1345: ditto (move it down if this method here is even needed)

api/client.go
```diff
@@ -77,6 +77,7 @@ func (cfg *Config) validate() error {
 type Client interface {
 	URL(ep string, args map[string]string) *url.URL
 	Do(context.Context, *http.Request) (*http.Response, []byte, error)
+	CloseIdleConnections()
```
bwplotka, line 80: I double check and I would propose we remove `CloseIdleConnections()` from both interfaces. We don't want to break compatibility for not too important reason (we kind of can because it's an experimental package, but only if we have to).

This method on interface is not needed because you can totally do 
```

type closeIdler interface {
	CloseIdleConnections()
}

// .. in some function
cl, err := NewClient()
// check err
cl.(closeIdler).CloseIdleConnection()
```

Thus to be safe, I would start with this approach first. It unblocks you without impacting potentially many packages in a very painful way (when upgrading client_golang).

bwplotka (changes requested):
So sorry for blocking this PR again, but I think we should remove interface changes. It is possible to avoid them. WDYT?

Thanks for patience!

api/client.go
```diff
@@ -79,6 +79,10 @@ type Client interface {
 	Do(context.Context, *http.Request) (*http.Response, []byte, error)
 }
 
+type closeIdler interface {
```
bwplotka, line 82:
```suggestion
type CloseIdler interface {
```

bwplotka: This allows us to avoid shallow `ClientCloseIdler` function.

api/client.go
```diff
@@ -99,6 +103,10 @@ func NewClient(cfg Config) (Client, error) {
 	}, nil
 }
 
+func ClientCloseIdler(cl Client) {
+	cl.(closeIdler).CloseIdleConnections()
+}
+
```
bwplotka, line 109:
```suggestion
func ClientCloseIdler(cl Client) {
	cl.(closeIdler).CloseIdleConnections()
}
```

No need for that, anyone can do cl.(CloseIdler).CloseIdleConnection()

bwplotka (approved):
Thanks! Almost there 💪🏽

api/client.go
cuisongliu: done

api/client.go
```diff
@@ -79,6 +79,10 @@ type Client interface {
 	Do(context.Context, *http.Request) (*http.Response, []byte, error)
 }
 
+type CloseIdler interface {
```
bwplotka, line 82: Ideally we can comment what this is for, but we can do in a separare PR, thanks!

bwplotka (approved):
Thanks!

Issue #1434, opened 2024-01-17
Changelog automation


ArthurSens:
On every new release, we need to provide a list of changes that made it into the release. This log of changes is maintained in the [CHANGELOG.md](https://github.com/prometheus/client_golang/blob/main/CHANGELOG.md) file. 

Updating the Changelog file requires manual effort from maintainers, it demands searching through commit history, analyzing which PRs had user-facing changes, and which PRs hadn't, writing clear messages describing the change, etc.

This whole process could receive some improvements through automation.

---

This issue concerns researching, proposing, and implementing the Changelog automation for client_golang. The final goal is not to require manual effort from maintainers to go through a big list of commits and hand-picking changelog entries.

pawarpranav83:
Hi @ArthurSens. I am interested in solving this issue for the upcoming LFX program.

pawarpranav83:
We can use something similar to the knative-prow bot - 
https://github.com/knative/eventing/releases/tag/knative-v1.11.10

But it doesn't follow the keepachangelog format

kartikaysaxena:
Hey @ArthurSens, I am quite interested in working upon this issue in the LFX program.

pawarpranav83:
Hey @ArthurSens, we could use GitHub's Automatic release notes generation feature itself, though we would have to assign labels to PRs and define a yaml file accordingly.

Or do you have any other approach in mind?

ArthurSens:
Hey folks, thanks for the interest :) 

Applications will be done through LFX mentoring website, so we'll have to wait for the official timeline to start working deeply on the proposal. This task alone is also quite small for the LFX program (around 3 months of work), so we're still working on other kinds of automation to add to the project idea (mostly around go version upgrades).

With that said, I believe there's no problem on brainstorming high level ideas like you're doing :) 

---
> we could use GitHub's Automatic release notes generation feature itself, though we would have to assign labels to PRs and define a yaml file accordingly.

The Github's automatic release notes that I know dump the whole commit history between two releases, and that includes non-user-facing changes. Did they change anything recently that takes PR labels into consideration (and possibly a yaml file?)?

kartikaysaxena:
Hey @ArthurSens maybe we can use some shell scripting to get hold of the file changes in the PR (using `git diff` maybe) and set up a yaml file which can contain list of important files which could contain some user-facing-changes, and if they match it would reflect in the changelog.

pawarpranav83:
[...]

PR #1517, opened 2024-05-19
Add PR template for changelog automation

SachinSahu431:
Add PR template to ease the process of changelog automation. Instructions have been updated in CONTRIBUTING.md.

Related to https://github.com/prometheus/client_golang/issues/1434

### DIFF
--- a/.github/pull-request-template.md
+++ b/.github/pull-request-template.md
@@ -1,0 +1,12 @@
+### Describe your PR
+
+
+<!-- Please refer to the [CONTRIBUTING.md](./CONTRIBUTING.md) for detailed instructions on how to fill out this section. -->
+### What type of PR is this?
+
+
+<!-- Add only user-facing changes. Refer [CONTRIBUTING.md](./CONTRIBUTING.md) for detailed instructions. -->
+### Changelog Entry
+```release-note
+
+```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,7 @@
 # Contributing
 
+Thank you for contributing to our project! Here are the steps and guidelines to follow when creating a pull request (PR).
+
 Prometheus uses GitHub to manage reviews of pull requests.
 
 * If you have a trivial fix or improvement, go ahead and create a pull request,
@@ -18,3 +20,29 @@ Prometheus uses GitHub to manage reviews of pull requests.
   Environments](http://peter.bourgon.org/go-in-production/#formatting-and-style).
 
 * Be sure to sign off on the [DCO](https://github.com/probot/dco#how-it-works)
+
+## How to fill the PR template
+
+### Describe your PR
+
+In this section, provide a clear and concise description of what your PR does. This helps reviewers understand the purpose and context of your changes.
+
+### What type of PR is this?
+
+Indicate the type of PR by adding one of the following options:
+
+- /kind chore
+- /kind cleanup
+- /kind fix
+- /kind bugfix
+- /kind enhancement
+- /kind feature
+- /kind feat
+- /kind docs
+- /kind NONE
+
+If this change should not appear in the changelog, use `/kind NONE`.
+
+### Changelog Entry
+
+Include a brief summary of your change for the changelog. This helps users understand what has been modified, added, or fixed in the project. If your change should not appear in the changelog, write `NONE`. Make sure to add only user-facing changes.


.github/pull-request-template.md
```diff
@@ -0,0 +1,12 @@
+### Describe your PR
+
+
+<!-- Please refer to the [CONTRIBUTING.md](./CONTRIBUTING.md) for detailed instructions on how to fill out this section. -->
```
kakkoyun, line 4: I think we should give more instructions and examples here. Most of the people won't bother to copy the link.

And top of that you can embed the section as an anchor. Mylink#my-section-with-hyphens

CONTRIBUTING.md
```diff
@@ -18,3 +20,29 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+
+## How to fill the PR template
+
+### Describe your PR
+
+In this section, provide a clear and concise description of what your PR does. This helps reviewers understand the purpose and context of your changes.
+
+### What type of PR is this?
+
+Indicate the type of PR by adding one of the following options:
```
kakkoyun, line 32: These are concise enough to be part of the template.

kakkoyun (commented):
Thanks. It's on the right direction.

.github/pull-request-template.md
SachinSahu431: Added comments and examples ✅

In my recent [commit](https://github.com/prometheus/client_golang/pull/1518/commits/37203780203ac14c0879ef5aef3da61b3b2130d9), I updated the auto-PR-label configuration file. Previously, the regex was incorrectly capturing labels from commented examples, which has now been corrected.

.github/pull-request-template.md
```diff
@@ -0,0 +1,16 @@
+<!-- Refer to CONTRIBUTING.md for more details and examples.
+https://github.com/prometheus/client_golang/blob/main/CONTRIBUTING.md#how-to-fill-the-pr-template  
+-->
+### Describe your PR
+
+
+### What type of PR is this?
+
+<!-- Format: /kind followed by ONE of the type {fix, bugfix, enhancement, feature, feat, change, release-note-none} -->
```
bwplotka, line 9: Can we give example?

.github/pull-request-template.md
```diff
@@ -0,0 +1,16 @@
...
+
+
+### Changelog Entry
```
bwplotka, line 12: I believe it would be nice to provide format of it e.g. in example

.github/pull-request-template.md
bwplotka: There are examples in contributing guide, but here putting one example would be nice.

bwplotka (approved):
Nice. OK if we want to pursue this design. 

Did we check the https://github.com/googleapis/release-please with conventional commits? In this way users will only care about creating proper commit titles 🤔 here we rely on PR descriptions, which is ok, just pros&cons

SachinSahu431:
Thanks for the review @bwplotka! ❤️

Yes, we also explored [release-please](https://github.com/googleapis/release-please), and it is fantastic! 🔥

Using the PR description offers several advantages over using PR titles. PR titles may not always capture user-facing changes accurately and might need adjustments for different audiences. Moreover, PR descriptions can include multiline changelog entries, which is not possible with single-line PR titles.

According to the [release-please documentation](https://github.com/googleapis/release-please?tab=readme-ov-file#how-should-i-write-my-commits), it works best when conventional commit formats are enforced and a `squash-merge` strategy is used. So, the idea of using PR descriptions looked more flexible, with ease of use for both contributors and maintainers. Without enforcing strict formatting and commit strategies.

.github/pull-request-template.md
ArthurSens: I think people will eventually look at previously merged PRs when looking for examples.

If we make the template too long, it'll get ignored 😅

kakkoyun (approved):

.github/pull-request-template.md
SachinSahu431: We previously discussed about this. Initially, examples were included, but they made the template excessively long. Additionally, the regex used in the [release-notes](https://github.com/kubernetes/release/tree/master/cmd/release-notes) tool, which we are using internally, was incorrectly capturing content from the examples. Therefore, we decided to remove the examples to keep the PR template concise and easy to read.

PR #1518, opened 2024-05-19
Auto label PRs

SachinSahu431:
### Describe your PR

This PR introduces automation for labelling PRs based on their content, aiming to streamline our changelog generation process. For more details, please refer to the [changelog automation proposal document](https://docs.google.com/document/d/1naZG6YvO2Wbvu_LwO4ZvdNp_6JKq_gdjnW2_gByquRQ/edit?usp=sharing).

With this automation, PRs will be tagged automatically with relevant labels like `kind/bug` ,`kind/feature`, or `release-note-none`, facilitating accurate and consistent changelog entries. 

### What type of PR is this?
/kind fix

### Changelog Entry

```release-note
NONE
```

#### Related:
https://github.com/prometheus/client_golang/issues/1434
https://github.com/prometheus/client_golang/pull/1517


ArthurSens (commented):
🥳 Awesome start! But... aren't we supposed to see this workflow acting in your PR already?

It would be awesome if you could elaborate a little bit in your PR description as well! Kemal and I are already familiar with what you're doing but other maintainers and contributors probably aren't.

I think it's a bit our fault (Kemal and I), but your proposal wasn't shared with the community for evaluation. We could've reminded you to share it more often... but better late than never! What do you think of including your proposal in the PR description to make sure other can understand what you're trying to accomplish here and raise concerns/opinions if they have any?

.github/labeler.yml
```diff
@@ -0,0 +1,45 @@
...
+# - keys are labels
+# - values are objects of { include: [ pattern ], exclude: [ pattern ] }
+#    - pattern must be a valid regex, and is applied globally to
+#      title + description of issues and/or prs (see enabled config above)
+#    - 'include' patterns will associate a label if any of these patterns match
+#    - 'exclude' patterns will ignore this label if any of these patterns match
+labels:
+  'kind/CLEANUP':
+    include:
+      - '(/kind (chore|cleanup))|(chore|cleanup):'
```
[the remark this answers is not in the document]
SachinSahu431: Agreed, we should skip these labels. 

I think I added them by looking at [this](https://github.com/prometheus/client_golang/blob/main/CHANGELOG.md#020--2015-02-23). But since our focus is user-facing changes, so we will remove these labels.

.github/labeler.yml
```diff
@@ -0,0 +1,45 @@
...
+    include:
+      - '(/kind (fix|bugfix))|(fix|bugfix):'
+    exclude: []
+  'kind/ENHANCEMENT':
+    include:
+      - '(/kind enhancement)|enhancement:'
+    exclude: []
+  'kind/FEATURE':
+    include:
+      - '(/kind (feat|feature))|(feat|feature):'
```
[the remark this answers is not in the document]
SachinSahu431: The regex was made such that usage of Conventional Commits format in changelog entry will also add labels. So, regex was able to capture `/kind <something>` as well as `<something>:` pattern. 
For example, `/kind feat` and `feat: this is changelog` will add `kind/FEATURE` label.

But it is no longer required now, so I've removed it 👍

.github/labeler.yml
```diff
@@ -0,0 +1,45 @@
...
+    exclude: []
+  'kind/CHANGE':
```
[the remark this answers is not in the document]
SachinSahu431: I tried using lowercase, but it ended up being reflected in the changelog. See `documentation` in this picture:

lowercase


Given that we're utilizing a [custom Go template](https://github.com/SachinSahu431/client_golang/blob/53e02d2e03b7a44cbaf5788719a275785291936b/changelog-template.tpl) without altering the [Kubernetes release-tool](https://github.com/kubernetes/release/tree/master/cmd/release-notes), we're required to use uppercase.

Is there a way to convert a string to uppercase directly within a Go template? I struggled with that 😢

.github/labeler.yml
```diff
@@ -0,0 +1,45 @@
...
+    include:
+      - '(/kind docs)|docs:'
```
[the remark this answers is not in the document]
SachinSahu431: Yep that's confusing, how about using `kind/DOCUMENTATION` instead?

SachinSahu431:
Thanks for the review, @ArthurSens! ❤️

I've included the proposal document in the PR description and left comments where I encountered difficulties.

.github/labeler.yml
```diff
@@ -0,0 +1,34 @@
...
+      - '(^|[^\s])/kind enhancement'
+    exclude: []
+  'kind/FEATURE':
+    include:
+      - '(^|[^\s])/kind (feat|feature)'
+    exclude: []
+  'kind/DOCUMENTATION':
+    include:
+      - '(^|[^\s])/kind (doc|docs|documentation)'
+    exclude: []
```
ArthurSens, line 30: Let's remove `kind/DOCUMENTATION` because we don't add such information to our changelog.

At the same time, we're missing `kind/CHANGE`

.github/pull-request-template.md
```diff
@@ -0,0 +1,62 @@
+### Describe your PR
+
+<!-- 
+Provide a clear and concise description of what your PR does. 
+This helps reviewers understand the purpose and context of your changes. 
+-->
+
+
+### What type of PR is this?
```
ArthurSens, line 9: As discussed in the call, let's remove this from the PR template and update the workflow to create comments on the PR when the workflow fails

.github/workflows/pr-auto-label.yml
```diff
@@ -0,0 +1,27 @@
...
+
+  labeler:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check Labels
+      id: labeler
+      uses: jimschubert/labeler-action@v1
+      with:
+        GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
ArthurSens, line 25: Let's add another step, which adds a comment to the PR if the `labeler` step fails

ArthurSens (commented):
Comments from our pair review today

.github/workflows/pr-auto-label.yml
SachinSahu431: Example PR: https://github.com/SachinSahu431/client_golang/pull/56

Added comment when `labeler` workflow fails:
image

SachinSahu431:
The contributor is notified if PR template is not filled properly:
image

Example PR: https://github.com/SachinSahu431/client_golang/pull/56

.github/workflows/pr-auto-label.yml
```diff
@@ -0,0 +1,49 @@
...
+          ${{ env.BODY }} 
+          EOF) =~ /kind\ (fix|bugfix|feat|feature|enhancement|change|none|NONE|release-note-none|release-notes-none) ]] ; then
+            echo "PR body contains a valid type of change"
+          else
+            echo "PR body does not contain a valid type of change. Please refer to CONTRIBUTING.md for more information."
+            curl -sSL -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+              -X POST https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.number }}/comments \
+              -d '{"body": "PR body does not contain a valid type of change. Please refer to [CONTRIBUTING.md](https://github.com/prometheus/client_golang/blob/main/CONTRIBUTING.md#how-to-fill-the-pr-template) for more information."}'
+            exit 0
+          fi
```
ArthurSens, line 47: Oh wow, I'm impressed that you made this work like this :O 

But I'm not sure if we need to maintain something complex like this, GitHub Actions provide good support for triggering conditional steps.

Maybe we could simplify this a little bit?

ArthurSens: What immediately comes to mind is:

[`continue-on-error` for the labeler step](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error)

Another[ github-action](https://github.com/marketplace/actions/comment-pull-request) that is capable of commenting on PRs, and a [conditional on this step that will only execute if the previous one failed](https://docs.github.com/en/actions/learn-github-actions/contexts#:~:text=steps.%3Cstep_id%3E.conclusion,is%20success.)

ArthurSens (commented):
Awesome job making it work like this, but let's keep things simple if we can 🙂

.github/workflows/pr-auto-label.yml
SachinSahu431: The [labeler action](https://github.com/jimschubert/labeler-action) we are currently using never fails; it always reports success, even if no labels are added. This makes it unsuitable for conditional steps.

I searched for alternative labeler actions that return a failure status but none offer the same level of [regex configuration](https://github.com/jimschubert/labeler-action?tab=readme-ov-file#full-schema) as our current one.

Since `labeler action` can't detect the absence of labels, I had to use GitHub API to fetch the PR description and post a comment 🙂

SachinSahu431: @ArthurSens Thanks for recommending this [GitHub Action](https://github.com/marketplace/actions/label-checker-for-pull-requests)! Our workflow is much cleaner now :)

A comment is added if a label is missing, and it is automatically deleted when the label is added.

.github/workflows/pr-auto-label.yml
```diff
@@ -0,0 +1,57 @@
...
+          message: |
+            PR body contains a valid type of change.
+          comment_tag: labelfailure
+          mode: delete
+          create_if_not_exists: false
+      - name: Exit on Failure
+        run: |
+          if [[ ${{ steps.preview_label_check.outputs.label_check }} == 'failure' ]]; then
+            exit 1
+          fi
```
ArthurSens, line 57:
```suggestion
      - name: Exit on Failure
        if: steps.preview_label_check.outputs.label_check == 'failure'
        run: exit 1
```

PR #1520, opened 2024-05-20
promlint: Add timestamp_seconds suffix restriction

assafad:
### What this PR does
Add a rule to promlint, restricting `timestamp_seconds` suffix for non-gauge metrics. 
`timestamp_seconds` metrics are not cumulative and shouldn't be split to buckets or quantiles. They should be of gauge type. 
For example, in https://kubernetes.io/docs/reference/instrumentation/metrics/#list-of-stable-kubernetes-metrics and https://github.com/kubernetes/kube-state-metrics/tree/main/docs/metrics, all `timestamp_seconds` metrics are gauges.

### DIFF
--- a/prometheus/testutil/promlint/promlint_test.go
+++ b/prometheus/testutil/promlint/promlint_test.go
@@ -499,6 +499,63 @@ x_bytes 10
 	runTests(t, tests)
 }
 
+func TestLintGauge(t *testing.T) {
+	tests := []test{
+		{
+			name: "counter with _timestamp_seconds suffix",
+			in: `
+# HELP _timestamp_seconds Test metric.
+# TYPE _timestamp_seconds counter
+_timestamp_seconds 10
+`,
+			problems: []promlint.Problem{
+				{
+					Metric: "_timestamp_seconds",
+					Text:   `counter metrics should have "_total" suffix`,
+				},
+				{
+					Metric: "_timestamp_seconds",
+					Text:   `non-gauge metrics should not have "_timestamp_seconds" suffix`,
+				},
+			},
+		},
+		{
+			name: "gauge with _timestamp_seconds suffix",
+			in: `
+# HELP _timestamp_seconds Test metric.
+# TYPE _timestamp_seconds gauge
+_timestamp_seconds 10
+`,
+		},
+		{
+			name: "counter without _timestamp_seconds suffix",
+			in: `
+# HELP _total Test metric.
+# TYPE _total counter
+_total 10
+`,
+		},
+		{
+			name: "untyped with _timestamp_seconds suffix",
+			in: `
+# HELP _timestamp_seconds Test metric.
+# TYPE _timestamp_seconds untyped
+_timestamp_seconds 10
+`,
+		},
+		{
+			name: "untyped without _timestamp_seconds suffix",
+			in: `
+# HELP _seconds Test metric.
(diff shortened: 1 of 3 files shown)


assafad:
@ArthurSens Can you please review this PR?

ArthurSens:
Isn't there a linter that already fails if a counter doesn't end with `_total`? Wouldn't it catch this case as well?

assafad:
> Isn't there a linter that already fails if a counter doesn't end with `_total`? Wouldn't it catch this case as well?

@ArthurSens Yes, but it won't catch other metrics types using the `timestamp_seconds` suffix. I don't see how timestamps are relevant for histogram and summaries, so I think this validation makes sense.

prometheus/testutil/promlint/promlint_test.go
```diff
@@ -499,6 +499,63 @@ x_bytes 10
 	runTests(t, tests)
 }
 
+func TestLintGauge(t *testing.T) {
+	tests := []test{
+		{
+			name: "counter with _timestamp_seconds suffix",
+			in: `
+# HELP _timestamp_seconds Test metric.
+# TYPE _timestamp_seconds counter
```
bwplotka, line 508: Side note: Metrics starting with underscore should be a problem too 🙈

prometheus/testutil/promlint/promlint_test.go
```diff
@@ -499,6 +499,63 @@ x_bytes 10
 	runTests(t, tests)
 }
 
+func TestLintGauge(t *testing.T) {
```
bwplotka, line 502: LintGauge tests counters?

prometheus/testutil/promlint/promlint_test.go
```diff
@@ -499,6 +499,63 @@ x_bytes 10
...
+# TYPE _timestamp_seconds counter
+_timestamp_seconds 10
+`,
+			problems: []promlint.Problem{
+				{
+					Metric: "_timestamp_seconds",
+					Text:   `counter metrics should have "_total" suffix`,
+				},
+				{
+					Metric: "_timestamp_seconds",
```
bwplotka, line 517: This does not feel helpful for counters, the above problem duplicates here kind of and is kind of awkward why `timestamp_second` is not allowed. Can I do `ts_seconds`?

prometheus/testutil/promlint/validations/gauge_validations.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"errors"
+	"strings"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+// LintGauge detects issues specific to gauges, as well as patterns that should
+// only be used with gauges.
+func LintGauge(mf *dto.MetricFamily) []error {
```
bwplotka, line 25: Hm, so this feels bit wrong. Name suggests we lint gauges, but looks like we simply lint non-gauges. I would move this to related types one by one.... but is there one?

For counters we don't need this, because _total check would be failed. For histograms and summaries, I think they are always cumulative so this check should be there too (I don't think it's implemented?).

So what we are left with? On what case you got hit by this?

bwplotka (changes requested):
Thanks! Trying to understand the exact problems we want to guard against, but the current PR. is not ideal at the moment I think

assafad:
@bwplotka  thanks for your review :)

PR #1526, opened 2024-05-31
ci: Group all changelog-related CI jobs into single one

ArthurSens:
### Describe your PR

I've noticed that we added the labeling workflow with several different CI jobs that depend on each other to run sequentially.
This dependency upon each other suggests that we might not want to run them as separate jobs.

This PR just groups them together, which makes them run sequentially by default :)

### What type of PR is this?
/kind release-note-none

### Changelog Entry
```release-note
NONE
```

### DIFF
--- a/.github/workflows/pr-auto-label.yml
+++ b/.github/workflows/pr-auto-label.yml
@@ -1,5 +1,5 @@
 ---
-name: PR-auto-labeler
+name: Changelog automation
 
 on:
   pull_request_target:
@@ -12,11 +12,12 @@ permissions:
     checks: write # to generate status
 
 jobs:
-  remove-old-labels:
-    name: Remove old PR labels
+  changelog-automation:
+    name: Changelog automation
     runs-on: ubuntu-latest
     steps:
-      - uses: mondeja/remove-labels-gh-action@v2.0.0
+      - name: Remove old PR labels
+        uses: mondeja/remove-labels-gh-action@v2.0.0
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           labels: |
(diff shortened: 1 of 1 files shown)


ArthurSens:
cc @kakkoyun @SachinSahu431

.github/workflows/pr-auto-label.yml
```diff
@@ -12,11 +12,12 @@ permissions:
     checks: write # to generate status
 
 jobs:
-  remove-old-labels:
-    name: Remove old PR labels
+  changelog-automation:
+    name: Changelog automation
     runs-on: ubuntu-latest
     steps:
```
SachinSahu431, line 18: Let's add this condition to skip all these steps when PR is raised by dependabot

```
runs-on: ubuntu-latest
if: github.actor != 'dependabot[bot]'
steps:
```

SachinSahu431 (approved):
LGTM!🚀🚀

SuperQ (approved):

.github/workflows/pr-auto-label.yml
ArthurSens: Im not sure if this will have the effect that we want 🤔

A skipped job isn't considered successful, so we will keep blocking dependabot PRs this way 😬

SachinSahu431: Oh I see. Any idea if we use [exit with success status](https://docs.github.com/en/actions/creating-actions/setting-exit-codes-for-actions#about-exit-codes), will this work?

ArthurSens: Yeah, we can probably do something similar :)

PR #1537, opened 2024-06-11
Allow creating constant histogram and summary metrics with a created timestamp

### LINKED ISSUE #1535: Ability to set CreatedTimestamp on constant histogram metrics
#1313 added created timestamp to the internal histogram data model:

https://github.com/ArthurSens/client_golang/blob/e5639d1dcb98e3f1027e110d11ab9eda54e36f34/prometheus/histogram.go#L1202-L1218

I don't see a way for clients to set the value for created timestamp? Specifically we are trying to use `prometheus.NewConstHistogram` 

I think this is why we're not getting any results when trying to calculate `histogram_quantile` since our `_bucket` counts don't start at zero without created timestamp injection. It'd also be nice if the `_sum` and `_count` metrics had zero injection.

---

Curious to know if hiding created timestamp from clients is intentional here, like if this part of the feature is WIP.

swar8080:
Closes #1535

### Describe your PR
Currently there's no way to set created timestamp when creating constant histogram and summar metrics. This adds some new functions using the same naming as the existing `NewConstMetricWithCreatedTimestamp`

### What type of PR is this?

/kind enhancement

### Changelog Entry
```release-note
api: Allow setting created timestamp on constant histogram and summary metrics using the new functions `NewConstHistogramWithCreatedTimestamp` and `NewConstSumamryWithCreatedTimestamp`
```

### Testing
Verified locally that zero-injection happens for the _bucket, _sum, and _count series. An example:
image

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -1373,6 +1373,48 @@ func MustNewConstHistogram(
 	return m
 }
 
+// NewConstHistogramWithCreatedTimestamp does the same thing as NewConstHistogram but sets the created timestamp
+func NewConstHistogramWithCreatedTimestamp(
+	desc *Desc,
+	count uint64,
+	sum float64,
+	buckets map[float64]uint64,
+	ct time.Time,
+	labelValues ...string,
+) (Metric, error) {
+	if desc.err != nil {
+		return nil, desc.err
+	}
+	if err := validateLabelValues(labelValues, len(desc.variableLabels.names)); err != nil {
+		return nil, err
+	}
+	return &constHistogram{
+		desc:       desc,
+		count:      count,
+		sum:        sum,
+		buckets:    buckets,
+		labelPairs: MakeLabelPairs(desc, labelValues),
+		createdTs:  timestamppb.New(ct),
+	}, nil
+}
+
+// MustNewConstHistogramWithCreatedTimestamp is a version of NewConstHistogramWithCreatedTimestamp that panics where
+// NewConstHistogramWithCreatedTimestamp would have returned an error.
+func MustNewConstHistogramWithCreatedTimestamp(
+	desc *Desc,
+	count uint64,
+	sum float64,
+	buckets map[float64]uint64,
+	ct time.Time,
+	labelValues ...string,
+) Metric {
+	m, err := NewConstHistogramWithCreatedTimestamp(desc, count, sum, buckets, ct, labelValues...)
+	if err != nil {
+		panic(err)
+	}
+	return m
+}
+
 type buckSort []*dto.Bucket
 
 func (s buckSort) Len() int {
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -1373,7 +1373,7 @@ func MustNewConstHistogram(
 	return m
 }
 
-// NewConstHistogramWithCreatedTimestamp does the same thing as NewConstHistogram but sets the created timestamp
+// NewConstHistogramWithCreatedTimestamp does the same thing as NewConstHistogram but sets the created timestamp.
 func NewConstHistogramWithCreatedTimestamp(
 	desc *Desc,
 	count uint64,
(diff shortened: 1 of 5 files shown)


ArthurSens:
Just wanted to let you know that this is on my list! Hopefully I'll get to it by the end of the week

ArthurSens:
Ok, I've spent more time reading our codebase today and I think I've found some inconsistencies.

I noticed that ConstHistogram doesn't add exemplars to the histogram, but I knew that somehow it was possible. After reading a bit I've found [NewMetricWithExemplars](https://github.com/prometheus/client_golang/blob/bccd68204bf41e9cce27e8561b70e0d1cb350bcb/prometheus/metric.go#L225-L247), that adds exemplars to existing metrics and it doesn't matter if it's a Counter or Histograms.

Now for CreatedTimestamps, we took a different turn 😬. Instead of creating a similar function, we created [NewConstMetricWithCreatedTimestamp](https://github.com/prometheus/client_golang/blob/bccd68204bf41e9cce27e8561b70e0d1cb350bcb/prometheus/value.go#L136-L159),  that only works with Counters.

I'm not sure what's the best solution here, but I'm more inclined to deprecating `NewConstMetricWithCreatedTimestamp` and following the same strategy we had with the exemplars 🤔 

What do you think @swar8080 @bwplotka @kakkoyun ?

swar8080:
Hey @ArthurSens, thanks for looking at this

As a user I remember expecting `NewConstMetricWithCreatedTimestamp` to work with constant histograms and then had to dig into the code to figure out why it didn't

For non-const/stateful metrics, like regular counters, it looks like created timestamp is managed internally and not provided by the caller. So not sure if `NewMetricWithCreatedTimestamp` would be needed for those

ArthurSens:
> Silly question, but can we make `NewConstMetricWithCreatedTimestamp` work with const histograms without breaking change?

I don't think we can, because histograms need different information, i.e. count, sum and bucket boundaries. `NewConstMetric` would need them as extra arguments and that's a breaking change

> Otherwise it might be time for different v2 API here with some builder pattern or something for those side methods, it's getting confusing (:
> 
> Happy with whatever is easier for downstream users here.

Yeah, let's iterate on this later. Although we don't have concrete plans for v2, I think we have enough to start experimenting

ArthurSens (commented):
Alright, I'm happy to continue with `NewConstHistogramWithCreatedTimestamp`, we can experiment with a cleaner API in the v2 package.

Could you also add a small test similar to [this one](https://github.com/prometheus/client_golang/blob/34e02e282dc4a3cb55ca6441b489ec182e654d59/prometheus/value_test.go#L62), but for histograms?

Would you like to add the same thing for summaries as well? (Can be done in another PR if you prefer)

swar8080:
@ArthurSens I added the extra histogram test and the same change for summaries

prometheus/histogram.go
```diff
@@ -1373,6 +1373,48 @@ func MustNewConstHistogram(
 	return m
 }
 
+// NewConstHistogramWithCreatedTimestamp does the same thing as NewConstHistogram but sets the created timestamp
```
ArthurSens, line 1376:
```suggestion
// NewConstHistogramWithCreatedTimestamp does the same thing as NewConstHistogram but sets the created timestamp.
```

prometheus/summary.go
```diff
@@ -783,3 +783,45 @@ func MustNewConstSummary(
 	}
 	return m
 }
+
+// NewConstSummaryWithCreatedTimestamp does the same thing as NewConstSummary but sets the created timestamp
```
ArthurSens, line 787:
```suggestion
// NewConstSummaryWithCreatedTimestamp does the same thing as NewConstSummary but sets the created timestamp.
```

ArthurSens (approved):
LGTM, thanks!

Just two small nits that I'll add in a separate commit

ArthurSens:
Also, I took the liberty to change you PR description to /kind enhancement, since we're not changing any existing behavior :)

PR #1539, opened 2024-06-19
Remove inlined upstream code

SuperQ:
### Describe your PR

Avoid having to maintain inlined code by using Go modules for `github.com/kylelemons/godebug`.

### What type of PR is this?

/kind release-note-none

### Changelog Entry
```release-note
NONE
```

PR #1540, opened 2024-06-19
Switch testutil to go-cmp

SuperQ:
### Describe your PR

Remove a dependency by using go-cmp for `testutil.CollectAndCompare()`. Note that the string output of `cmp.Diff()` is non-stable, so users depending on stable `error` strings will nolonger work.

### What type of PR is this?

/kind change

### Changelog Entry
```release-note
Use non-deterministic diff function for testutil.CollectAndCompare()
```


kakkoyun:
> This is an alternative implementation of #1539

I'd rather accept #1539

SuperQ:
Yea, I figured I would provide both options.

While it's extremely annoying, I find I agree slightly with the opinion that depending on error message exact output is a dangerous.

ArthurSens:
Should we close this one now that #1539 was merged?

SuperQ:
Closing in favor of https://github.com/prometheus/client_golang/pull/1539

PR #1544, opened 2024-06-22
client: Add Option to provide limit query param for APIs that support it

abbyssoul:
### Describe your PR
This PR adds an option to client APIs to specify limit query parameters for APIs that support it.

For example, APIs, such as [Finding series by label matchers](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers), support optional query param:
  - limit=<number>: Maximum number of returned series. 
 
However, the current  [client interface](https://github.com/prometheus/client_golang/blob/main/api/prometheus/v1/api.go#L492) provides no way to pass that parameter.

To minimize the impact on existing users of that client interface, this parameter is implemented as an existing optional argument. 

### What type of PR is this?

/kind enhancement

### Changelog Entry
```release-note
Client APIs support extra arguments to specify limit query parameters for APIs that support it.

```

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -1026,7 +1026,7 @@ func (h *httpAPI) Runtimeinfo(ctx context.Context) (RuntimeinfoResult, error) {
 
 func (h *httpAPI) LabelNames(ctx context.Context, matches []string, startTime, endTime time.Time, opts ...Option) ([]string, Warnings, error) {
 	u := h.client.URL(epLabels, nil)
-	q := formatOptions(u.Query(), opts)
+	q := addOptionalURLParams(u.Query(), opts)
 
 	if !startTime.IsZero() {
 		q.Set("start", formatTime(startTime))
@@ -1049,7 +1049,7 @@ func (h *httpAPI) LabelNames(ctx context.Context, matches []string, startTime, e
 
 func (h *httpAPI) LabelValues(ctx context.Context, label string, matches []string, startTime, endTime time.Time, opts ...Option) (model.LabelValues, Warnings, error) {
 	u := h.client.URL(epLabelValues, map[string]string{"name": label})
-	q := formatOptions(u.Query(), opts)
+	q := addOptionalURLParams(u.Query(), opts)
 
 	if !startTime.IsZero() {
 		q.Set("start", formatTime(startTime))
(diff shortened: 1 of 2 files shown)


abbyssoul:
@bwplotka PTAL :)

api/prometheus/v1/api.go
```diff
@@ -1088,20 +1091,35 @@ func WithTimeout(timeout time.Duration) Option {
 	}
 }
 
-func (h *httpAPI) Query(ctx context.Context, query string, ts time.Time, opts ...Option) (model.Value, Warnings, error) {
-	u := h.client.URL(epQuery, nil)
-	q := u.Query()
+// WithLimit can be used to provide an optional maximum number of returned entries for APIs that support that.
+// https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries
```
bwplotka, line 1095:
```suggestion
// WithLimit provides an optional maximum number of returned entries for APIs that support limit parameter e.g. https://prometheus.io/docs/prometheus/latest/querying/api/#instant-querie:~:text=%3A%20End%20timestamp.-,limit%3D%3Cnumber%3E,-%3A%20Maximum%20number%20of
```

api/prometheus/v1/api.go
```diff
@@ -1088,20 +1091,35 @@ func WithTimeout(timeout time.Duration) Option {
...
+func WithLimit(limit uint64) Option {
+	return func(o *apiOptions) {
+		o.limit = limit
+	}
+}
 
+func formatOptions(q url.Values, opts []Option) url.Values {
```
bwplotka, line 1102:
```suggestion
func addOptionalURLParams(q url.Values, opts []Option) url.Values {
```

bwplotka: Perhaps more explicit?

bwplotka (approved):
LGTM, thanks!

Some readability suggestions only. Unit test would be nice, but we would need to add query/params expected field to `apiTest` and all cases likely, so up to you -- not a blocker.

Thanks!

api/prometheus/v1/api.go
abbyssoul: Arh... [one of the hardest problems in  Computer Science - naming things](https://martinfowler.com/bliki/TwoHardThings.html).

bwplotka (approved):
Thanks!

PR #1559, opened 2024-07-12
Add default Go runtime metrics for /gc/gogc:percent, /gc/gomemlimit:bytes, /sched/gomaxprocs:threads

vesari:
This PR adds 3 of the additional Go runtime metrics to the Prometheus default Go metrics, as recommended in this proposal [#67120](https://github.com/golang/go/issues/67120), namely `/gc/gogc:percent`,  `/gc/gomemlimit:bytes` and  `/sched/gomaxprocs:threads`. As agreed with this repo's maintainers, `/sched/latencies:seconds` will be dealt with in a separate PR.


prometheus/go_collector.go
```diff
@@ -266,6 +285,18 @@ func (c *baseGoCollector) Collect(ch chan<- Metric) {
 	ch <- MustNewConstSummary(c.gcDesc, uint64(stats.NumGC), stats.PauseTotal.Seconds(), quantiles)
 	ch <- MustNewConstMetric(c.gcLastTimeDesc, GaugeValue, float64(stats.LastGC.UnixNano())/1e9)
 	ch <- MustNewConstMetric(c.goInfoDesc, GaugeValue, 1)
+	ch <- MustNewConstMetric(c.goMaxProcs, GaugeValue, float64(runtime.NumCPU()))
+
+	gogcSample := make([]runmetr.Sample, 1)
+	gogcSample[0].Name = "/gc/gogc:percent"
+	runmetr.Read(gogcSample)
```
bwplotka, line 292: I wonder if this would be easier to take from Runtime 🤔

bwplotka: Ok, it's not fetchable, you can only set those variables, nvm (:

prometheus/go_collector.go
```diff
@@ -245,6 +261,9 @@ func (c *baseGoCollector) Describe(ch chan<- *Desc) {
 	ch <- c.gcDesc
 	ch <- c.gcLastTimeDesc
 	ch <- c.goInfoDesc
+	ch <- c.goMaxProcs
```
bwplotka, line 264: I wonder if they could fit better in `goCollector` given it already uses runtime metrics and we could give same time snapshot 🤔

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 
 require (
 	github.com/golang/protobuf v1.5.3 // indirect
+	github.com/hashicorp/go-version v1.7.0
```
bwplotka, line 21: Do we need this?

bwplotka (commented):
Looking good! Some comments 💪🏽

prometheus/collectors/go_collector_latest.go
```diff
@@ -88,6 +88,16 @@ func WithGoCollectorMemStatsMetricsDisabled() func(options *internal.GoCollector
 	}
 }
 
+// WithGoCollectorRuntimeEnvVarsMetricsDisabled disables the following default metrics:
+// go_gogc_percent
+// go_gomemlimit
+// go_gomaxprocs
+func WithGoCollectorRuntimeEnvVarsMetricsDisabled() func(options *internal.GoCollectorOptions) {
```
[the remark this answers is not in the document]
bwplotka: The reason we have WithGoCollectorMemStatsMetricsDisabled is that you can replace them with new runtime metrics. It feels there is no such case here? We could use runtime directly (with name conversion)? 🤔

prometheus/collectors/go_collector_latest_test.go
```diff
@@ -117,6 +118,7 @@ func TestGoCollectorAllowList(t *testing.T) {
 			reg.MustRegister(NewGoCollector(
 				WithGoCollectorMemStatsMetricsDisabled(),
 				WithGoCollectorRuntimeMetrics(test.rules...),
+				WithGoCollectorRuntimeEnvVarsMetricsDisabled(),
```
bwplotka, line 121: Can we test somewhere how it looks with those enabled?

prometheus/go_collector.go
```diff
@@ -279,3 +279,8 @@ type memStatsMetrics []struct {
 	eval    func(*runtime.MemStats) float64
 	valType ValueType
 }
+
+type runtimeEnvVarsMetrics []struct { // I couldn't come up with a better name. Any suggestions?
```
bwplotka, line 283: Nice! I see where you go with this. This way would totally work, but it's extra structures in many places. There is a way to simplify this by reusing generic flow, so by default adding appropriate regexes to `RuntimeMetricRules` in defaultGoCollectorOptions (also you can remove the weird dead code (comment)) - https://github.com/prometheus/client_golang/blob/6478f56935c73f23b38144dc33cae75ed10554ac/prometheus/go_collector_latest.go#L162

I think we can add that by default, then fix tests. We could even test if `WithoutGoCollectorRuntimeMetrics` can turn those off 🤗 WDYT?

bwplotka (changes requested):
Amazing work! It's not easy flow and you managed to get this working! However, there is a bit simpler (lower complexity) way of doing this (unless I missed sth), mentioned in the comment.

prometheus/collectors/go_collector_latest.go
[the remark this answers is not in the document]
vesari: Ok,I wrongly assumed everything that is not in the `BaseCollector` struct should be opted out (or in) explicitly, but as you say it's just 3 extra metrics which are already part of the new runtime metrics, so I should definitely reuse that path. Thanks!

bwplotka:
Another thing, I would use our auto-name conversion and add new metrics as runtime metrics would do so:

```
		"go_gc_gogc_percent",
		"go_gc_gomemlimit_bytes",
"go_sched_gomaxprocs_threads",
```

WDYT?

vesari:
> Another thing, I would use our auto-name conversion and add new metrics as runtime metrics would do so:
> 
> ```
> 		"go_gc_gogc_percent",
> 		"go_gc_gomemlimit_bytes",
> "go_sched_gomaxprocs_threads",
> ```
> 
> WDYT?

Yes, it makes sense, thanks for noticing the discrepancy

vesari:
About `withDefaultRuntimeMetrics` function, since it's just for tests, I didn't bother encapsulating the bool arguments in an `options` struct.

ArthurSens (commented):
Hello hello, getting back from PTO now :) 

@vesari @bwplotka, are you two aware that we generate the go runtime metrics from a script?

https://github.com/prometheus/client_golang/blob/main/prometheus/collectors/gen_go_collector_set.go

I'm confused that we don't see any failures in the CI though 🤔

vesari:
> Hello hello, getting back from PTO now :)
> 
> @vesari @bwplotka, are you two aware that we generate the go runtime metrics from a script?
> 
> https://github.com/prometheus/client_golang/blob/main/prometheus/collectors/gen_go_collector_set.go
> 
> I'm confused that we don't see any failures in the CI though 🤔

@ArthurSens hello and welcome back! Yes, I'm aware of that. This PR doesn't change or influence that. May I ask what failures you were expecting? Thanks! :)

prometheus/go_collector_latest.go
```diff
@@ -153,7 +155,9 @@ func defaultGoCollectorOptions() internal.GoCollectorOptions {
 			"/gc/heap/frees-by-size:bytes":  goGCHeapFreesBytes,
 		},
 		RuntimeMetricRules: []internal.GoCollectorRule{
-			//{Matcher: regexp.MustCompile("")},
+			{
+				Matcher: regexp.MustCompile(`\/gc\/gogc:percent|\/gc\/gomemlimit:bytes|\/sched\/gomaxprocs:threads`),
```
bwplotka, line 159: hm, do we really need to escape `/`? It's not sed, it's RE2 [language](https://github.com/google/re2/wiki/Syntax) 🤔

prometheus/go_collector_latest_go120_test.go
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !go1.21
+// +build !go1.21
+
+package prometheus
+
+var expMetrics = map[string]string{
```
bwplotka, line 19: Nice! Unfortunately this filename does little sense (both latest and 120 in name). We can rename, but I wonder if there's a way to put this into single file with other metrics. 

I think @ArthurSens is right, we might want to check how we could integrate with https://github.com/prometheus/client_golang/blob/main/prometheus/collectors/gen_go_collector_set.go perhaps, as it was created for same reason. Not sure if it's integrated in our CI, worth to check...

prometheus/go_collector_latest_test.go
```diff
@@ -86,8 +93,13 @@ func TestGoCollector_ExposedMetrics(t *testing.T) {
 			expectedFQNameSet: expectedBaseMetrics(),
 		},
 		{
-			// Default, only MemStats.
-			expectedFQNameSet: addExpectedRuntimeMemStats(expectedBaseMetrics()),
+			// Default, only MemStats and default Runtime metrics.
+			opts: internal.GoCollectorOptions{
```
bwplotka, line 97: let's use default function then?

bwplotka (approved):
Nice, small nits around testing only, otherwise good!

prometheus/go_collector_latest_go120_test.go
vesari: Ok, now I understand. Yes, ideally that map should be included in the corresponding Go version autogenerated files.

prometheus/go_collector_latest_test.go
bwplotka: Do you mind using our new internal.GoCollector default etc here?

bwplotka (approved):
One tiny thing to do, then LGTM! thanks!

prometheus/collectors/go_collector_go120_test.go
```diff
@@ -116,3 +118,18 @@ func withSchedulerMetrics() []string {
 func withDebugMetrics() []string {
 	return withBaseMetrics([]string{})
 }
+
+var defaultRuntimeMetrics = []string{
+	"go_sched_gomaxprocs_threads",
+}
+
+func withDefaultRuntimeMetrics(metricNames []string, withoutGC, withoutSched bool) []string {
+	// If withoutSched is true, exclude "go_sched_gomaxprocs_threads".
```
bwplotka, line 127: Nice, so those tests verify if our default metrics can be removed by "without" option, right? Great!

vesari: Basically I had to find a way to compose the expected default Go runtime metrics in the tests according to the various allow/deny lists, in order to mirror the logic's behaviour, which allows removing the default runtime metrics for example and which also avoids duplicated metrics

ArthurSens:
I'll add some work to myself: Understand why CI didn't complain that we changed generated files without the appropriate changes to `prometheus/gen_go_collector_metrics_set.go`!

vesari:
> I'll add some work to myself: Understand why CI didn't complain that we changed generated files without the appropriate changes to `prometheus/gen_go_collector_metrics_set.go`!

Thank you for looking into it! I don't know if that makes any difference, but technically I didn't change any autogenerated files. I just added 2 separate files :)

bwplotka:
Yup, we checked. The CI is NOT running this generative script. We only check if metrics given by runtime does is what we expect in those generated _test files and then complain with instruction to potentially rerun that script.

Let's discuss if that CI step is needed in the ticket you added https://github.com/prometheus/client_golang/issues/1575

bwplotka (approved):
Thanks! Let's go!!! 💪🏽

PR #1566, opened 2024-07-29
deps: Updated to prometheus/common to 0.55

bwplotka:
Rationales - breaking change (see https://github.com/prometheus/client_golang/pull/1550)


kakkoyun (approved):
LGTM.

How disturbing is a change for our downstream users? We will see.

PR #1577, opened 2024-08-13
Remove go_memstat_lookups_total; added runtime/metrics calculation to memstat metric's help.

bwplotka:
I know we ideally should not remove any metric from default list, but this one is always zero, so let's save everyone's money.

### DIFF
--- a/prometheus/go_collector.go
+++ b/prometheus/go_collector.go
@@ -52,7 +52,7 @@ func goRuntimeMemStats() memStatsMetrics {
 		}, {
 			desc: NewDesc(
 				memstatNamespace("mallocs_total"),
-				// TODO(bwplotka): We could add go_memstats_heap_objects, probably useful for discovery. Let's gather more feedback, kind of waste of bytes for everybody for compatibility reason to keep both, and we can't reall rename/remove useful metric.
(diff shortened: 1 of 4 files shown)


prometheus/go_collector.go
```diff
@@ -28,176 +28,168 @@ func goRuntimeMemStats() memStatsMetrics {
...
-				nil, nil,
-			),
-			eval:    func(ms *runtime.MemStats) float64 { return float64(ms.Lookups) },
-			valType: CounterValue,
 		}, {
 			desc: NewDesc(
 				memstatNamespace("mallocs_total"),
-				// TODO(bwplotka): We could add go_memstats_heap_objects, probably useful for discovery. Let's gather more feedback, kind of waste of bytes for everybody for compatibility reason.
-				"Total number of heap objects allocated, both live and gc-ed. Semantically a counter version for go_memstats_heap_objects gauge.",
+				// TODO(bwplotka): We could add go_memstats_heap_objects, probably useful for discovery. Let's gather more feedback, kind of waste of bytes for everybody for compatibility reason to keep both, and we can't reall rename/remove useful metric.
```
ArthurSens, line 55:
```suggestion
				// TODO(bwplotka): We could add go_memstats_heap_objects, probably useful for discovery. Let's gather more feedback, kind of a waste of bytes for everybody for compatibility reasons to keep both, and we can't really rename/remove useful metric.
```

ArthurSens (approved):
Just a tiny comment with grammatical fixes

PR #1588, opened 2024-08-20
ci: bump dagger to the latest version

marcosnils:
this commit also adds an `includes` field to the dagger.json so the initial SDK code generation runs faster as only the Dagger module source is included


bwplotka (approved):
Thanks!

PR #1592, opened 2024-08-20
repro test.

bwplotka:


bwplotka:
Can be repro-ed indeed https://github.com/prometheus/client_golang/actions/runs/10469430197/job/28992461341?pr=1592#step:3:1339, test finished.

PR #1596, opened 2024-08-23
fix: Unset Content-Encoding header when uncompressed

mrueg:
This should fix the bug reported in #1595 as RFC2616 says not to set Content-Encoding header for responses without any encoding.

@ArthurSens @bwplotka


ArthurSens (approved):
Nice, I just tested locally and works as before :) Thanks for working on this

I wonder if we want to have a test that verifies that? Or too much?

bwplotka (approved):
Thanks! I wished we have some safeguard for seeing content-encoding set AND accept-encoding causing us to do some different compression, but this is enough as a quick bug fix. Thanks a lot 💪🏽

bwplotka:
I changed the base to release branch as it has to be fixed there first, do you mind rebasing?

bwplotka:
We also need to add one commit for:
 /VERSION  set to 1.20.2
/CHANGELOG.md section for new release and current bugfix

But I can help do that

bwplotka (approved):

Issue #1590, opened 2024-08-20
processcollector: Add support for Darwin platform.


bwplotka:
Not a high priority but we could implement metrics for MacOS e.g. using [`sysctl`](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/sysctl.3.html). Currently we error out (silently by default).

mharbison72:
https://github.com/prometheus/client_golang/pull/1600 handles most of this, but I didn't add `(fixes #1590)` to the commit because it would be really nice to get the memory stats too.  I'd need to know what's allowable though (3rd party deps, cgo, undocumented APIs), before sinking a lot more time into it though.

bwplotka:
Yea, ideally none of those `3rd party deps, cgo`. Undocumented API might work. 

There is always option for others (or us) to create a separate module for darwin statistics/collector. Here we aim for something standard, good enough for majority of users. Thanks!

mharbison72:
> Yea, ideally none of those `3rd party deps, cgo`. Undocumented API might work.
> 
> There is always option for others (or us) to create a separate module for darwin statistics/collector. Here we aim for something standard, good enough for majority of users. Thanks!

Can you describe this in a little more please?  I think I'm missing something here.  (If there's a better forum for a wider discussion, we can move it there).  To be clear when I said "undocumented API", I meant from Apple, not something in Go.

I think the fundamental problem here is that there's no built-in mechanism in Go to call into random system libraries like there is on Windows (AFAICT), and they've been resistant to adding that support[1][2].  Given that restriction, we can't access the (undocumented) Apple APIs for getting this info without either cgo or a 3rd party mechanism.  (Maybe the `noncgo` module referenced in the second link is another 3rd party option.  It's small and public domain, but well beyond my understanding of Go, so I wouldn't want to try to replicate it.)

Placing this code in another module would just seem to shift the problem, and create additional problems in that `process_virtual_memory_bytes` and `process_resident_memory_bytes` could come from one module on some platforms, and a second on darwin.  (I think I saw a recent issue mention that the `Describe` method isn't per-platform, so I think that would have to be fixed up first to do this.)

[...]

PR #1600, opened 2024-08-29
process_collector: fill in most statistics on macOS

mharbison72:
Unfortunately, the virtual memory, resident memory, and network stats will require access to undocumented C functions.  I was warned off of cgo in IRC because it would then have to be enabled in a bunch of different projects that use this module, but I already was against it because that would break the ability to cross-compile.  There is no interface to `dlopen` built into golang. The `github.com/ebitengine/purego` module looks promising (I can cross-compile and call these methods), but I'm currently getting unexpected results.  I'll follow up with that separately if I can get it working, but hopefully this stuff is pretty uncontroversial.

Tested on macOS 10.14.6 (amd64), macOS 14.6.1 (amd64), and macOS 15.0 (arm64) by spawning `/usr/bin/ulimit -a -S` and `/usr/sbin/lsof -c $my_process` from the test exporter process, and `ps -o lstart,vsize,rss,utime,stime,command` from the shell, and comparing results with the exported metrics.

I can't find documentation for `RLIMIT_AS` on macOS (specifically if it's in bytes or pages).  It's currently being reported back as `RLIM_INFINITY`, which seems reasonable, because I've come across reports that the value is ignored anyway[1].  The bash 3.2 code for the built-in `ulimit` divides the value reported by `getrusage(2)` by 1024 when printing, as it does for `RLIMIT_DATA`, which is documented as being bytes in `getrusage(2)`.  The help for `ulimit` indicates it prints both in kbytes, so it's reasonable to assume this is already in bytes.

[1] https://issues.chromium.org/issues/40581251#comment3


prometheus/process_collector_darwin.go
```diff
@@ -0,0 +1,112 @@
+// Copyright 2015 The Prometheus Authors
```
SuperQ, line 1:
```suggestion
// Copyright 2024 The Prometheus Authors
```

SuperQ (changes requested):
Great, minor nit.

SuperQ (approved):
LGTM, will leave the final review up to the maintainers.

mharbison72:
I guess I forgot to tag a maintainer.

@bwplotka

prometheus/process_collector_darwin.go
```diff
@@ -0,0 +1,110 @@
...
+
+func getOpenFileCount() (float64, error) {
+	// Alternately, the undocumented proc_pidinfo(PROC_PIDLISTFDS) can be used to
+	// return a list of open fds, but that requires a way to call C APIs.  The
+	// benefits, however, include fewer system calls and not failing when at the
+	// open file soft limit.
+
+	if dir, err := os.Open("/dev/fd"); err != nil {
+		return 0.0, err
+	} else {
```
bwplotka, line 46: No need for else here, we just returned above.

prometheus/process_collector_darwin.go
```diff
@@ -0,0 +1,110 @@
...
+		defer dir.Close()
+
+		// Avoid ReadDir(), as it calls stat(2) on each descriptor.  Not only is
+		// that info not used, but KQUEUE descriptors fail stat(2), which causes
+		// the whole method to fail.
+		if names, err := dir.Readdirnames(0); err != nil {
+			return 0.0, err
+		} else {
```
bwplotka, line 54: same here

bwplotka (approved):
Amazing starting point, thanks! Tiny suggestions, but not blocking, up to you. LGTM

prometheus/process_collector_darwin.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if addressSpace, err := getSoftLimit(syscall.RLIMIT_AS); err == nil {
+		ch <- MustNewConstMetric(c.maxVsize, GaugeValue, float64(addressSpace))
+	} else {
+		c.reportError(ch, c.maxVsize, err)
+	}
+
+	// TODO: socket(PF_SYSTEM) to fetch "com.apple.network.statistics" might
+	//  be able to get the per-process network send/receive counts.
```
ArthurSens, line 109: Let's create an issue for this one so we don't forget :)

mharbison72: There's already https://github.com/prometheus/client_golang/issues/1590 (unless you want to close that with this PR and open another one, since the remaining work is likely much harder, and leaving that open implies no darwin support).

ArthurSens: Ah fair point, I'm fine with keeping that open until we finish the whole thing.

It would be nice to update the existing issue then, maybe breaking it down into smaller deliverables that can be implemented separately

prometheus/process_collector_darwin.go
mharbison72: I think the `else` is needed unless the assignment is un-scoped, though this is a tiny function, so scoping doesn't matter much here.  That said, I'd like to keep it as-is because when I pulled this locally, there's now a merge commit and I'm not sure how to handle making additional changes with hg-git.  It'll likely cause more problems than it's worth.

prometheus/process_collector_darwin.go
mharbison72: I don't know enough about GH to do a deliverable list myself, but I'll reference this PR there for visibility if it isn't already.

bwplotka:
Thanks!

PR #1601, opened 2024-08-30
Add structured logging support to promhttp

SuperQ:
In order to better support the standard library `log/slog` add a new interface to the `promhttp` `HandlerOpts`.

### DIFF
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -168,6 +168,9 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 			if opts.ErrorLog != nil {
 				opts.ErrorLog.Println("error gathering metrics:", err)
 			}
+			if opts.StructuredErrorLog != nil {
+				opts.StructuredErrorLog.Error("error gathering metrics", "error", err)
+			}
 			errCnt.WithLabelValues("gathering").Inc()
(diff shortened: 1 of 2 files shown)


SuperQ:
We might want to split out the Go version change into a separate PR.

SuperQ:
Ok, yaks shaven, this is ready to go.

SuperQ:
Seems like dagger is a bit of technical debt for keeping Go versions up to date as well.

ArthurSens:
Hmmm, feels strange that we have 2 loggers in the struct now. I'm assuming this is needed somewhere in a different package/repository, could you clarify how you're planning to use this?

tjhop:
> Hmmm, feels strange that we have 2 loggers in the struct now. I'm assuming this is needed somewhere in a different package/repository, could you clarify how you're planning to use this?

@SuperQ we're only ever logging at `error` level here -- maybe we just lean on slog's built in stdlog adapter and avoid the change here altogether? Something like this:

```go
	// make logger with promslog like normal
	slogger := promslog.New(&promslog.Config{})
	// make stdlog adapter from slogger
	logger := slog.NewLogLogger(slogger.Handler(), slog.LevelError)
```

Bonus points -- log output is still structured

SuperQ:
Yea, I suppose using a stdlib adapter would be fine.

Issue #1607, opened 2024-08-31
Index out of range error in native histogram


krajorama:
Version github.com/prometheus/client_golang v1.20.0

When testing trivial solution for #1605 :+1: 

```
 go test -count 1 -race -timeout 300s -run ^TestNativeHistogramConcurrency$ github.com/prometheus/client_golang/prometheus
panic: runtime error: index out of range [10] with length 10

goroutine 10 [running]:
github.com/prometheus/client_golang/prometheus.(*nativeExemplars).addExemplar(0xc0002307c8, 0xc00055f4f0)
	/home/krajo/go/github.com/krajorama/client_golang/prometheus/histogram.go:1791 +0x1969
github.com/prometheus/client_golang/prometheus.(*histogram).updateExemplar(0xc000230700, 0x3ff22643022060a2, 0x0, 0xc00054fcb0)
	/home/krajo/go/github.com/krajorama/client_golang/prometheus/histogram.go:1140 +0x117
github.com/prometheus/client_golang/prometheus.(*histogram).ObserveWithExemplar(0xb64180?, 0xc00054fcb0?, 0xbe246f?)
	/home/krajo/go/github.com/krajorama/client_golang/prometheus/histogram.go:770 +0x6a
github.com/prometheus/client_golang/prometheus.TestNativeHistogramConcurrency.func1.2({0xc0005d0000, 0x46e4, 0x420820?})
	/home/krajo/go/github.com/krajorama/client_golang/prometheus/histogram_test.go:1058 +0x17b
created by github.com/prometheus/client_golang/prometheus.TestNativeHistogramConcurrency.func1 in goroutine 21
	/home/krajo/go/github.com/krajorama/client_golang/prometheus/histogram_test.go:1050 +0x325
FAIL	github.com/prometheus/client_golang/prometheus	2.189s
```

I think this is probably due to the fact that we have:
https://github.com/prometheus/client_golang/blob/dbf72fc1a20e87bea6e15281eda7ef4d139a01ec/prometheus/histogram.go#L1754

Which instructs the code to insert after the end of the array. Usually this would not end up in rIdx, however in a concurrent environment the timestamp of the last inserted exemplar might be newer than the currently inserted exemplar. So we don't go into this branch: 
https://github.com/prometheus/client_golang/blob/dbf72fc1a20e87bea6e15281eda7ef4d139a01ec/prometheus/histogram.go#L1769

krajorama:
fixed

Issue #1605, opened 2024-08-31
Data race in native histogram


krajorama:
Version github.com/prometheus/client_golang v1.20.0

Capacity for exemplars is called without lock here:
https://github.com/prometheus/client_golang/blob/dbf72fc1a20e87bea6e15281eda7ef4d139a01ec/prometheus/histogram.go#L1685

Which causes the race with 
https://github.com/prometheus/client_golang/blob/dbf72fc1a20e87bea6e15281eda7ef4d139a01ec/prometheus/histogram.go#L1704

See logs from Mimir PR CI run:
```
22:12:09 query-frontend: ==================
22:12:09 query-frontend: WARNING: DATA RACE
22:12:09 query-frontend: Read at 0x00c001bba3d8 by goroutine 78:
22:12:09 query-frontend: github.com/prometheus/client_golang/prometheus.(*nativeExemplars).addExemplar()
22:12:09 query-frontend: /__w/mimir/mimir/vendor/github.com/prometheus/client_golang/prometheus/histogram.go:1685 +0x57
22:12:09 query-frontend: github.com/prometheus/client_golang/prometheus.(*histogram).updateExemplar()
22:12:09 query-frontend: /__w/mimir/mimir/vendor/github.com/prometheus/client_golang/prometheus/histogram.go:1140 +0x116
22:12:09 query-frontend: github.com/prometheus/client_golang/prometheus.(*histogram).ObserveWithExemplar()
22:12:09 query-frontend: /__w/mimir/mimir/vendor/github.com/prometheus/client_golang/prometheus/histogram.go:770 +0x69
22:12:09 query-frontend: github.com/grafana/dskit/instrument.ObserveWithExemplar()
22:12:09 query-frontend: /__w/mimir/mimir/vendor/github.com/grafana/dskit/instrument/instrument.go:76 +0x1a2
22:12:09 query-frontend: github.com/grafana/dskit/middleware.observe()
22:12:09 query-frontend: /__w/mimir/mimir/vendor/github.com/grafana/dskit/middleware/grpc_instrumentation.go:34 +0x205
22:12:09 query-frontend: github.com/grafana/dskit/middleware.StreamServerInstrumentInterceptor.func1()
22:12:09 query-frontend: /__w/mimir/mimir/vendor/github.com/grafana/dskit/middleware/grpc_instrumentation.go:58 +0x18b
22:12:09 query-frontend: google.golang.org/grpc.getChainStreamHandler.func1()
[...]

PR #1608, opened 2024-08-31
bugfix: native histogram: exemplars index out of range

krajorama:
Fixes: #1607 
Fixes: #1605 

fix: native histogram: Simplify and fix addExemplar

mdIdx was redundant when len(exemplars)>1, so got rid of it, rIdx
is enough.

Don't compare timestamp of incoming exemplar to timestamp of
minimal distance exemplar. Most of the time the incoming exemplar
will be newer. And if not, the previous code just replaced an
exemplar one index after the minimal distance exemplar. Which had
an index out of range bug, plus is essentially random.

Contains an unoptimized fix for #1605 , see discussion / fix in #1609 

```
panic: runtime error: index out of range [10] with length 10

goroutine 10 [running]:
github.com/prometheus/client_golang/prometheus.(*nativeExemplars).addExemplar(0xc0002307c8, 0xc00055f4f0)
	/home/krajo/go/github.com/krajorama/client_golang/prometheus/histogram.go:1791 +0x1969
github.com/prometheus/client_golang/prometheus.(*histogram).updateExemplar(0xc000230700, 0x3ff22643022060a2, 0x0, 0xc00054fcb0)
	/home/krajo/go/github.com/krajorama/client_golang/prometheus/histogram.go:1140 +0x117
github.com/prometheus/client_golang/prometheus.(*histogram).ObserveWithExemplar(0xb64180?, 0xc00054fcb0?, 0xbe246f?)
	/home/krajo/go/github.com/krajorama/client_golang/prometheus/histogram.go:770 +0x6a
github.com/prometheus/client_golang/prometheus.TestNativeHistogramConcurrency.func1.2({0xc0005d0000, 0x46e4, 0x420820?})
	/home/krajo/go/github.com/krajorama/client_golang/prometheus/histogram_test.go:1058 +0x17b
created by github.com/prometheus/client_golang/prometheus.TestNativeHistogramConcurrency.func1 in goroutine 21
	/home/krajo/go/github.com/krajorama/client_golang/prometheus/histogram_test.go:1050 +0x325
```

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -1673,6 +1673,7 @@ func makeNativeExemplars(ttl time.Duration, maxCount int) nativeExemplars {
 
 	if maxCount < 0 {
 		maxCount = 0
+		ttl = -1
 	}
 
 	return nativeExemplars{
@@ -1682,12 +1683,12 @@ func makeNativeExemplars(ttl time.Duration, maxCount int) nativeExemplars {
 }
 
 func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
-	n.Lock()
-	defer n.Unlock()
-
-	if cap(n.exemplars) == 0 {
+	if n.ttl == -1 {
 		return
 	}
+
+	n.Lock()
+	defer n.Unlock()
 
 	// The index where to insert the new exemplar.
 	var nIdx int = -1
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -1682,12 +1682,12 @@ func makeNativeExemplars(ttl time.Duration, maxCount int) nativeExemplars {
 }
 
 func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
+	n.Lock()
+	defer n.Unlock()
+
 	if cap(n.exemplars) == 0 {
 		return
 	}
-
-	n.Lock()
-	defer n.Unlock()
 
 	// The index where to insert the new exemplar.
 	var nIdx int = -1
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -1658,7 +1658,10 @@ func addAndResetCounts(hot, cold *histogramCounts) {
 type nativeExemplars struct {
 	sync.Mutex
 
-	ttl       time.Duration
+	// Time-to-live for exemplars, it is set to -1 if exemplars are disabled, that is NativeHistogramMaxExemplars is below 0.
+	// The ttl is used on insertion to remove an exemplar that is older than ttl, if present.
+	ttl time.Duration
+
 	exemplars []*dto.Exemplar
 }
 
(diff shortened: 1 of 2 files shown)


krajorama:
cc @fatsheep9146

prometheus/histogram.go
```diff
@@ -1682,13 +1682,13 @@ func makeNativeExemplars(ttl time.Duration, maxCount int) nativeExemplars {
 }
 
 func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
+	n.Lock()
+	defer n.Unlock()
```
beorn7, line 1686: As said in the commit description, let's discuss the optimization of this in #1609.

beorn7 (approved):
LGTM, but @fatsheep9146 might have an opinion here, too.

Also, leaving it to the maintainers @ArthurSens @kakkoyun @bwplotka to make the final call.

ArthurSens (approved):
LGTM

bwplotka:
Does it needs a minor 1.20 release with this? Or we are ok with 1.21 in a few months or so?

krajorama:
> Does it needs a minor 1.20 release with this? Or we are ok with 1.21 in a few months or so?

I'd love to get a minor release so we can do clean dependency update in Mimir.

ArthurSens:
I'll keep this open for another day or so just in case Bartek or Kemal want to give another review. I create another release, but then we need to point this PR to the `release-1.20` branch

prometheus/histogram.go
krajorama: we agreed on the solution in 1609, which is now in here.

fatsheep9146 (approved):
Thanks for picking this bug! 
This fix is LGTM.

prometheus/histogram.go
```diff
@@ -1764,23 +1776,22 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+				// true, due to concurrency, but it's a good enough approximation.
 				md = diff
-				mdIdx = nIdx
-				if n.exemplars[nIdx-1].Timestamp.AsTime().Before(e.Timestamp.AsTime()) {
-					mdIdx = nIdx - 1
-				}
+				rIdx = nIdx - 1
 			}
 		}
 		if nIdx < len(n.exemplars) {
```
bwplotka, line 1786: I guess this is then guarding the case when nIdx is the last element? Then again, comment 1766 is kind of wrong, or misleading.

bwplotka (commented):
Some readability suggestion, some question likely due to my lack of context for the details. Plus please rebase on `release-1.20` for this to be in minor patch. Otherwise LGTM!

krajorama:
@bwplotka I've updated the comments in the code with more explanations and my understanding of the code :)

bwplotka:
I forced this PR to be against `release-1.20` branch.. which brought new commits from main. Do you mind rebasing your commits to `release-1.20` branch? (:

bwplotka (approved):
Thanks! (just rebase)

PR #1609, opened 2024-08-31
bugfix: native histogram: data race in addExemplar

krajorama:
Fixes: #1605 

Slices `cap` function is not thread safe. Reuse instead the `ttl` to store the ttl atomically with -1 indicating exemplars turned off.

Includes fix for #1607 from #1608 .

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -573,7 +573,7 @@ func newHistogram(desc *Desc, opts HistogramOpts, labelValues ...string) Histogr
 			h.nativeHistogramZeroThreshold = DefNativeHistogramZeroThreshold
 		} // Leave h.nativeHistogramZeroThreshold at 0 otherwise.
 		h.nativeHistogramSchema = pickSchema(opts.NativeHistogramBucketFactor)
-		h.nativeExemplars = makeNativeExemplars(opts.NativeHistogramExemplarTTL, opts.NativeHistogramMaxExemplars)
+		makeNativeExemplars(&h.nativeExemplars, opts.NativeHistogramExemplarTTL, opts.NativeHistogramMaxExemplars)
 	}
 	for i, upperBound := range h.upperBounds {
 		if i < len(h.upperBounds)-1 {
@@ -1658,11 +1658,11 @@ func addAndResetCounts(hot, cold *histogramCounts) {
 type nativeExemplars struct {
 	sync.Mutex
 
-	ttl       time.Duration
+	ttl       atomic.Int64 // It is a duration, but also used to check concurrently if exemplars are enabled.
 	exemplars []*dto.Exemplar
 }
 
-func makeNativeExemplars(ttl time.Duration, maxCount int) nativeExemplars {
+func makeNativeExemplars(exemplars *nativeExemplars, ttl time.Duration, maxCount int) {
 	if ttl == 0 {
 		ttl = 5 * time.Minute
 	}
@@ -1673,16 +1673,16 @@ func makeNativeExemplars(ttl time.Duration, maxCount int) nativeExemplars {
 
 	if maxCount < 0 {
 		maxCount = 0
+		ttl = -1
 	}
 
-	return nativeExemplars{
-		ttl:       ttl,
-		exemplars: make([]*dto.Exemplar, 0, maxCount),
-	}
+	exemplars.ttl.Store(int64(ttl))
+	exemplars.exemplars = make([]*dto.Exemplar, 0, maxCount)
 }
 
 func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
-	if cap(n.exemplars) == 0 {
+	ttl := n.ttl.Load()
+	if ttl == -1 {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


krajorama:
cc @fatsheep9146

beorn7:
Code looks good, but I'm not quite sure if we should do this kind of optimization via ttl as an atomic int. Those atomic operations still have a cost (and it could be significant on systems with many cores). The usual code path is to call `ObserveWithExemplar` on a histogram with configured exemplars, and just call `Observe` on a histogram where exemplars are not configured. The former case becomes even slower with this change here, while the latter case isn't affected at all.
I'm wondering if there is a legitimate use case for calling `ObserveWithExemplar` on a histogram with exemplars disabled. And even if so, if it is common enough to make this PR a net gain.

krajorama:
> Code looks good, but I'm not quite sure if we should do this kind of optimization via ttl as an atomic int. Those atomic operations still have a cost (and it could be significant on systems with many cores). The usual code path is to call `ObserveWithExemplar` on a histogram with configured exemplars, and just call `Observe` on a histogram where exemplars are not configured. The former case becomes even slower with this change here, while the latter case isn't affected at all. I'm wondering if there is a legitimate use case for calling `ObserveWithExemplar` on a histogram with exemplars disabled. And even if so, if it is common enough to make this PR a net gain.

Well, TTL is constant, so in theory we don't need to use atomics. Pushed that version.

prometheus/histogram.go
```diff
@@ -1705,17 +1706,23 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
 		otIdx = -1         // Index of the exemplar with the oldest timestamp.
 
-		md    = -1.0  // Logarithm of the delta of the closest pair of exemplars.
-		mdIdx = -1    // Index of the older exemplar within the closest pair.
-		cLog  float64 // Logarithm of the current exemplar.
-		pLog  float64 // Logarithm of the previous exemplar.
+		md   = -1.0  // Logarithm of the delta of the closest pair of exemplars.
+		rIdx = -1    // Index of the older exemplar within the closest pair and where we need to insert the new exemplar.
+		cLog float64 // Logarithm of the current exemplar.
+		pLog float64 // Logarithm of the previous exemplar.
```
ArthurSens, line 1725: Hmmm, totally out of scope for this PR, but more descriptive variable names would remove the necessity for commenting on their meaning 😕

I'm not sure if I'm missing something, but is there a reason to name these variables `rIdx`, `cLog`, `pLog` instead of `removeIndex`, `currentLog` or `previousLog`?

ArthurSens: Urgh, reading again `replaceIndex` is probably a better name for `rIdx`. Does it make `nIdx` obsolete? (Do I understand correctly that `nIdx` is `newIndex`? 😅

ArthurSens (commented):
Thanks for updating the test in the first commit to show the bug. Helps a lot with the review :)

I do understand that using `ttl` instead of `cap(n.exemplars)` solves the data race condition, so LGTM now that tests pass.

About the index out of range problem. I'm having a hard time understanding the codebase, but I believe in Beorn's expertise here. Out of scope for this PR but I think more descriptive variable names could help

ArthurSens:
Should we close #1608 in favor of this one?

krajorama:
> Should we close #1608 in favor of this one?

I'd like to keep discussion of the data race topic here. If Beorn says ok, I'll close this PR and just udpate the solution in #1608 with the one agreed here and we can discuss the algorithm there.

ArthurSens:
> > Should we close #1608 in favor of this one?
> 
> I'd like to keep discussion of the data race topic here. If Beorn says ok, I'll close this PR and just udpate the solution in #1608 with the one agreed here and we can discuss the algorithm there.

`ttl` doesn't receive any writes after its creation. Look safe and efficient to me :)

krajorama:
> > > Should we close #1608 in favor of this one?
> > 
> > 
> > I'd like to keep discussion of the data race topic here. If Beorn says ok, I'll close this PR and just udpate the solution in #1608 with the one agreed here and we can discuss the algorithm there.
> 
> `ttl` doesn't receive any writes after its creation. Look safe and efficient to me :)

Ack, I'll update the other PR and we can just close this if we merge that.

krajorama:
Discussed on chat with @beorn7 , using ttl==-1 is ok. Closing this as concluded.

prometheus/histogram.go
beorn7: Historically, we had very different "tastes" playing out between various Prometheus developers, all the way from very long and descriptive names to "use one letter names whenever possible" (slightly exaggerated).

My observation is that Go developers tend towards shorter variable names (have a look at the standard library, for example). But we never explicitly stated a preference for the Prometheus project. In lack of a specific guideline, I would go with generally good practices of programming like "the wider the scope of a variable, the more descriptive its name".

In this case, the variables are local to a (fairly long) function. So I wouldn't be too demanding about the descriptiveness of their names. Furthermore, even `previousLog` wouldn't really be self explanatory. (Previous log line?) And a name like `logarithmOfPreviousExemplar` would be clearly too clunky. Given that the comment is needed anyway, I actually like the names as they are. Still very short, but better than just `c`, `p`, and `r` (which others might even prefer).

PR #1621, opened 2024-09-05
registry: drop duplicate checks

lukedirtwalker:
In processMetrics drop duplicate checks. If the metricFamily exists we already have a type check with the call to checkMetricConsistency. The help string is already checked in the checkDescConsistency albeit only if pedantic mode is enabled. However this is probably what is desired anyway.

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -634,54 +634,7 @@ func processMetric(
 		return fmt.Errorf("error collecting metric %v: %w", desc, err)
 	}
 	metricFamily, ok := metricFamiliesByName[desc.fqName]
-	if ok { // Existing name.
-		if metricFamily.GetHelp() != desc.help {
-			return fmt.Errorf(
-				"collected metric %s %s has help %q but should have %q",
-				desc.fqName, dtoMetric, desc.help, metricFamily.GetHelp(),
-			)
-		}
-		// TODO(beorn7): Simplify switch once Desc has type.
-		switch metricFamily.GetType() {
-		case dto.MetricType_COUNTER:
-			if dtoMetric.Counter == nil {
-				return fmt.Errorf(
-					"collected metric %s %s should be a Counter",
-					desc.fqName, dtoMetric,
-				)
-			}
-		case dto.MetricType_GAUGE:
-			if dtoMetric.Gauge == nil {
-				return fmt.Errorf(
-					"collected metric %s %s should be a Gauge",
-					desc.fqName, dtoMetric,
-				)
-			}
-		case dto.MetricType_SUMMARY:
-			if dtoMetric.Summary == nil {
-				return fmt.Errorf(
-					"collected metric %s %s should be a Summary",
-					desc.fqName, dtoMetric,
-				)
-			}
-		case dto.MetricType_UNTYPED:
-			if dtoMetric.Untyped == nil {
-				return fmt.Errorf(
-					"collected metric %s %s should be Untyped",
-					desc.fqName, dtoMetric,
-				)
-			}
-		case dto.MetricType_HISTOGRAM:
-			if dtoMetric.Histogram == nil {
-				return fmt.Errorf(
-					"collected metric %s %s should be a Histogram",
-					desc.fqName, dtoMetric,
-				)
-			}
-		default:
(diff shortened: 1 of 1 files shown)


lukedirtwalker:
I took the freedom to remove the description consistency check in non-pedantic mode.
This was causing issues in opentelemetry-collector prometheus exporter (see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/28617 for more details)

bwplotka (commented):
Thanks!

Type checks were dups. The bigger change is that indeed registry will now not check the help consistency (you propose relaxing it). Let's update PR title, that's the bigger change here (1).

Without this check we need to handle random helps for the same metric family somehow. Quickly looking, I think the logic seems to be the first help from the first metric within a metric family will be used in the collected channel which mean.. random. This can be confusing and even worse, causing flipping help on scrape trashing potentially collector/Prometheus metadata storage.

Can you fix DCO (2) and add test for multiple helps against registry and pedantic registry confirming the behaviour (random/deterministic) (3).

Another issue is that I can see [another check that we missed](https://github.com/prometheus/client_golang/blob/ac114f315e2b5f4bd0fce2830fbb6d28d348eb02/prometheus/registry.go#L310) here, wonder why you didn't hit that first? 🤔 Hopefully test can help us figure out, we might need to update hashing there (4).
Then [another check that is still there](https://github.com/prometheus/client_golang/blob/ac114f315e2b5f4bd0fce2830fbb6d28d348eb02/prometheus/registry.go#L768), it would be weird if we relax in single gatherer, then for other cases we check perhaps (5)?

Then I found various descriptions around logic that will change after this change e.g. https://github.com/prometheus/client_golang/blob/ac114f315e2b5f4bd0fce2830fbb6d28d348eb02/prometheus/registry.go#L127 so we will need to update and test this (6).

Generally it would epic if there was a simple solution for callers like Otel to dedup and decide what help you want to pick. Have you considered that? Then registry can maintain semantics of consistent help for better UX and contract, plus you can see this invariance is tied to various places of this library. Not saying it's impossible to get rid of, but a bit more work is needed.

To sum up, I enumerated 1-6 action items here at the minimum.

cc @beorn7 - do you envision any other surprises we don't see yet here for that decision?

beorn7:
My memories of implementation details are pretty much gone…
I would be surprised if there were too many checks, but who knows…

Note that the Pushgateway has the same problem (of diverging help strings from different sources). It did not need a change of client_golang to cope.

beorn7:
> Generally it would epic if there was a simple solution for callers like Otel to dedup and decide what help you want to pick. Have you considered that? Then registry can maintain semantics of consistent help for better UX and contract, plus you can see this invariance is tied to various places of this library. 

This sounds very reasonable to me and in line with my vague memories (but I might misremember something).

I've just checked what PGW is doing. It is doing the help string check on its own indeed, and then fixes it on its own and logs about it.

lukedirtwalker:
Thank you both for the quick feedback. If I read it right, you both seem to prefer to close this PR and solve the problem on the client side? Works for me as well, it's certainly possible to fix this in the code that uses client_golang (otel in this case).

Let me know if you like to leave it here as is, then I close the PR.

bwplotka:
I mean it would epic to help you in the easiest possible way for everyone. 

I looked a bit, plus we discussed on [Slack too](https://cloud-native.slack.com/archives/C037NMVUAP3/p1725621148160319). I noticed you are using [unchecked collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusexporter/collector.go#L82), which makes sense for otel: aggregating data from various sources. I think it would make sense to do this help check ONLY if `registeredDescIDs == nil` (representing case of unchecked collector), given it's already "unchecked". This might allow us to generally take it lightly and not need to redesign the whole desc consistency model.

However... flapping help issue remains...maybe, probably depending on order of code movings metrics into channel. We could have a simple framework on client_golang side e.g. take the longer one / strings.Compare > 0 or so, but it's pretty important to have it somewhat stable (if possible). 

On client side all of this is possible, just a bit more overhead. Ideas welcome, we are not saying no, but some tests would be mandatory here!

PR #1629, opened 2024-09-17
process_xxx_memory statistics for macOS (without cgo)

mharbison72:
This is the alternative to [PR 1616](https://github.com/prometheus/client_golang/pull/1616), but keeping the cgo usage isolated to tests so that users don't need to fiddle with enabling cgo, or deal with losing cross platform support.

If this is viable, we can probably finish off the rest of #1600 with this technique.  (Currently, RSS, VSIZE, and the network traffic stats are not reported on macOS- this handles the first two items.)

@bwplotka

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 )
 
 require (
+	github.com/ebitengine/purego v0.7.1 // indirect
 	github.com/jpillora/backoff v1.0.0 // indirect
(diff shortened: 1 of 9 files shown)


mharbison72:
The [cgo version](https://github.com/prometheus/client_golang/pull/1616) was landed instead of this.

PR #1637, opened 2024-10-07
promhttp: add support for custom `gzip.Writer` pool

alespour:
`sync.Pool` used for `gzip.Writer`s is not suitable for long(er)-live objects, objects can be evicted by GC (which may not manifest in tight-loop benchmarks). This may occur quite often in memory-constrained environment (eg. with low `GOMEMLIMIT` and/or `GOGC` value), therefore new instances are also allocated again quite often, which further increases pressure on GC.

This PR adds support for providing alternative pool implementation for `gzip.Writers` via new `promhttp.HandlerOpts` field `GzipPool` of `Pool` type:

```go
type Pool interface {
	Get() any
	Put(x any)
}
```

`sync.Pool` implements this interface, and is used as the default implementation as it has been so far.

### DIFF
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -67,11 +67,13 @@ const (
 
 var defaultCompressionFormats = []Compression{Identity, Gzip, Zstd}
 
-var gzipPool = sync.Pool{
-	New: func() interface{} {
-		return gzip.NewWriter(nil)
-	},
+// Pool represents generic object pool. Uses `any` type so that sync.Pool can be used as default implementation.
+type Pool interface {
+	Get() any
+	Put(x any)
 }
+
+var gzipPool Pool
 
 // Handler returns an http.Handler for the prometheus.DefaultGatherer, using
 // default HandlerOpts, i.e. it reports the first error as an HTTP error, it has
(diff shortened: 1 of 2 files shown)


prometheus/promhttp/http.go
```diff
@@ -416,6 +428,8 @@ type HandlerOpts struct {
 	// NOTE: This feature is experimental and not covered by OpenMetrics or Prometheus
 	// exposition format.
 	ProcessStartTime time.Time
+	// GzipPool provides pool of gzip writers. Default is using sync.Pool.
+	GzipPool Pool
```
bwplotka, line 432: Is this really an option or should it be a local variable?

bwplotka (changes requested):
👋🏽 Thanks for this contribution.

The code looks generally ok. However, I wonder if this complexity is needed in client_golang. Scrapes are usually 15-60s apart. In the most extreme cases it's 1s. Even for 1s, even with better pooling, the GC actually SHOULD remove the underlying bytes. Is there are point no point in keeping extra memory for like, in the best case, 900ms between scrapes?

Can you share you use case? Or benchmark/profiles confirming this is an important bottleneck? Thanks!

Issue #1638, opened 2024-10-07
Optionally return error if two metrics would collide under escaping schemes


ywwg:
Hypothetically, if a system defined two metrics that are different in UTF-8 but the same when escaped, the escaped text exposition would show two entries for the "same" metric. (In general users shouldn't do this, but you know what they say about users.) We should help the user avoid doing that.

At registration time, we should check for not only exact duplicates but also escaped duplicates.  Since the default UnderscoreEscaping scheme is the most collision-prone, we will always use that method for testing equivalency. If a collision is found, return an error by default. We can add an optional way to skip this check in case it is intended.

bwplotka:
I also don't get https://github.com/prometheus/client_golang/pull/1641 goal. I thought this issue is to fail on registration, which is generally best-effort (e.g. there are cases of unchecked collectors etc) and that's fine.

However this PR is to fail /metrics call on this? Isn't that... wrong? I would rather have duplicate than your application fail to give any metric? Do we have a precedense? Do we fail handling /metrics in such collisions.. on scrape?

ArthurSens:
> I also don't get #1641 goal. I thought this issue is to fail on registration, which is generally best-effort (e.g. there are cases of unchecked collectors etc) and that's 

This is on me, I've added feedback asking to not to this on registration because, as you said, this is a scrape problem and I wanted to move the solution closer to the scrape.

> I would rather have duplicate than your application fail to give any metric?

I had an understanding the we want to provide a scrape as a transaction. So we provide everything that was registered or we provide nothing, there's no partial result here.

> Do we fail handling /metrics in such collisions.. on scrape?

I believe we started failing since Prometheus 2.52

bwplotka:
Thanks, makes sense!

> I also think maybe we should go back to failing at registration time since that's when the other collisions are caught. The trouble is that creates error messages for people who don't care about these collisions.

Yea, sounds like some option to registry or considering this only for PedenaticRegistry or something.

> At the end of the day maybe we should have skipped escaping and just refused to serve UTF-8 metrics to legacy systems :/.

Yea, I would love to check the impact of not doing this collision work on escaping, e.g. who will actually use escaping mechanism.

ywwg:
> Yea, sounds like some option to registry or considering this only for PedenaticRegistry or something.
> 

I'd like to know more about the pedanticregistry and where it gets used.

tahirghaffar1380-source:
> Hypothetically, if a systen escaped, the escaped text

[...]

PR #1641, opened 2024-10-08
Return an http error during scraping if metrics collide when escaped to underscores

ywwg:
fixes https://github.com/prometheus/client_golang/issues/1638

This change prevents a possible issue when UTF-8 metrics are scraped by a system that does not support UTF-8. The metric names will be escaped, and if two metrics escape to the same legacy name, they will appear in the exposition "twice", causing collision problems. Therefore, during scraping we return an http error if two metrics escape to the same legacy string. 

This change creates a new uint64 per metric and two new maps in the Registry. The new maps are only populated if the hash of the original metric data contains UTF-8 data. This means memory usage will be lower in the near term and will go up in the long term or for OTLP-heavy users. Eventually (i.e. Prom 4.0 or later) we could deprecate the compatibility modes with legacy systems to save this memory, but it's also not clear to me that the memory usage will be that high.

To avoid performance issues checking for collisions on every scrape, the actual collision detection is done at registration time but only reported at scrape time.

### DIFF
--- a/prometheus/desc.go
+++ b/prometheus/desc.go
@@ -57,6 +57,9 @@ type Desc struct {
 	// must be unique among all registered descriptors and can therefore be
 	// used as an identifier of the descriptor.
 	id uint64
+	// escapedID is similar to id, but with the metric and label names escaped
+	// with underscores.
+	escapedID uint64
 	// dimHash is a hash of the label names (preset and variable) and the
 	// Help string. Each Desc with the same fqName must have the same
 	// dimHash.
@@ -142,11 +145,18 @@ func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, const
 	}
 
 	xxh := xxhash.New()
-	for _, val := range labelValues {
+	escapedXXH := xxhash.New()
+	for i, val := range labelValues {
 		xxh.WriteString(val)
 		xxh.Write(separatorByteSlice)
+		if i == 0 {
+			val = model.EscapeName(val, model.UnderscoreEscaping)
+		}
+		escapedXXH.WriteString(val)
+		escapedXXH.Write(separatorByteSlice)
 	}
 	d.id = xxh.Sum64()
+	d.escapedID = escapedXXH.Sum64()
 	// Sort labelNames so that order doesn't matter for the hash.
 	sort.Strings(labelNames)
 	// Now hash together (in this order) the help string and the sorted
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -43,6 +43,7 @@ import (
 
 	"github.com/klauspost/compress/zstd"
 	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/model"
 
 	"github.com/prometheus/client_golang/internal/github.com/golang/gddo/httputil"
 	"github.com/prometheus/client_golang/prometheus"
@@ -121,6 +122,7 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 	if opts.MaxRequestsInFlight > 0 {
 		inFlightSem = make(chan struct{}, opts.MaxRequestsInFlight)
 	}
+	var hasEscapedCollisions bool
 	if opts.Registry != nil {
 		// Initialize all possibilities that can occur below.
 		errCnt.WithLabelValues("gathering")
@@ -133,6 +135,7 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 				panic(err)
 			}
 		}
+		hasEscapedCollisions = opts.Registry.HasEscapedCollision()
 	}
 
 	// Select compression formats to offer based on default or user choice.
@@ -190,6 +193,19 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		} else {
 			contentType = expfmt.Negotiate(req.Header)
 		}
+
+		if hasEscapedCollisions {
+			switch contentType.ToEscapingScheme() {
+			case model.UnderscoreEscaping, model.DotsEscaping:
+				if opts.ErrorLog != nil {
+					opts.ErrorLog.Println("error: one or more metrics collide when escaped")
+				}
+				httpError(rsp, fmt.Errorf("one or more metrics collide when escaped"))
+				return
+			default:
+			}
+		}
+
 		rsp.Header().Set(contentTypeHeader, string(contentType))
 
 		w, encodingHeader, closeWriter, err := negotiateEncodingWriter(req, rsp, compressions)
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -135,8 +135,8 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 				panic(err)
 			}
 		}
-		hasEscapedCollisions = opts.Registry.HasEscapedCollision()
 	}
+	hasEscapedCollisions = reg.HasEscapedCollision()
 
 	// Select compression formats to offer based on default or user choice.
 	var compressions []string
--- a/prometheus/promhttp/http_test.go
+++ b/prometheus/promhttp/http_test.go
@@ -81,6 +81,10 @@ func (g *mockTransactionGatherer) Gather() (_ []*dto.MetricFamily, done func(), 
 	return mfs, func() { g.doneInvoked++ }, err
 }
 
+func (g *mockTransactionGatherer) HasEscapedCollision() bool {
+	return g.g.HasEscapedCollision()
+}
+
 func readCompressedBody(r io.Reader, comp Compression) (string, error) {
 	switch comp {
 	case Gzip:
@@ -567,9 +571,7 @@ func TestEscapedCollisions(t *testing.T) {
 		Help: "A test metric with dots",
 	}))
 
-	handler := HandlerFor(reg, HandlerOpts{
-		Registry: reg,
-	})
+	handler := HandlerFor(reg, HandlerOpts{})
 
 	t.Run("fail case", func(t *testing.T) {
 		writer := httptest.NewRecorder()
--- a/prometheus/promhttp/http_test.go
+++ b/prometheus/promhttp/http_test.go
@@ -28,6 +28,8 @@ import (
 
 	"github.com/klauspost/compress/zstd"
 	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/model"
 
 	"github.com/prometheus/client_golang/prometheus"
 )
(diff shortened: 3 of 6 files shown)


ywwg:
So the edge case here is:

1. collision mode is Compat by default
2. a UTF-8 metric is created in an init function
3. the collision mode is flipped to UTF-8
4. an identical UTF-8 metric is created

This should collide, but because the first metric's hash was Compat mode, it doesn't match the hash of the second metric, and we don't detect the collision.

I think this is another case where we can say "don't do that", but I'm not sure how best to define what "that" is.  In general, I think users should not define metrics in init() functions.  OR, we could say, don't define UTF-8 metrics in init functions.

ywwg:
Another option is to store hashes for both UTF-8 and compat.  This adds 64bits per metric and another collectorsByID map (since we'll need to store both the compat and utf8 collector ids depending on the mode).  don't love that either

ywwg:
I am leaning towards having a second map in the registerer that only checks for compat fqname+labelname collisions. It's a memory increase but it gets around all these issues.

prometheus/desc.go
```diff
@@ -140,15 +142,45 @@ func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, const
 		d.err = fmt.Errorf("duplicate label names in constant and variable labels for metric %q", fqName)
 		return d
 	}
+ 
+	d.id, d.dimHash = makeHashes(labelNames, labelValues, help, true)
+	d.compatID, d.compatDimHash = makeHashes(labelNames, labelValues, help, false)
```
ywwg, line 147: this mutates labelNames so it's not pretty

prometheus/registry.go
```diff
@@ -260,20 +287,27 @@ func (errs MultiError) MaybeUnwrap() error {
...
 	mtx                   sync.RWMutex
 	collectorsByID        map[uint64]Collector // ID is a hash of the descIDs.
+	collectorsByCompatID  map[uint64]Collector
 	descIDs               map[uint64]struct{}
+	compatDescIDs         map[uint64]struct{}
 	dimHashesByName       map[string]uint64
+	compatDimHashesByName map[string]uint64
 	uncheckedCollectors   []Collector
 	pedanticChecksEnabled bool
+	utf8Collision         bool
```
ywwg, line 297: So we're adding a uint64, empty struct, and Collector interface obj for every metric registered.  This could be a lot of memory with many millions of metrics.  Perhaps we could only store the compat IDs if they are *different* from the default?

prometheus/registry.go
```diff
@@ -391,6 +481,10 @@ func (r *Registry) Unregister(c Collector) bool {
 	defer r.mtx.Unlock()
 
 	delete(r.collectorsByID, collectorID)
+	if _, exists := r.collectorsByCompatID[compatID]; !exists {
```
ywwg, line 484: there is other cleanup needed

prometheus/registry.go
ywwg: (I have now implemented this)

ywwg:
I think this new approach is a decent solution. We can flip UTF-8 mode on and off at will and we don't pay a huge memory or performance penalty (that I can see)

prometheus/desc.go
```diff
@@ -57,6 +57,8 @@ type Desc struct {
 	// must be unique among all registered descriptors and can therefore be
 	// used as an identifier of the descriptor.
 	id uint64
+	// compatID is similar to id, but is a hash of all the relevant names escaped with underscores.
+	compatID uint64
```
ArthurSens, line 61: When I first saw this variable I thought it was a typo for compactID, but I see you're are using compact is other places as well.

What does compat means? 😅

ArthurSens: And still related to this variable, have we thought about using id to store the hash of the already normalized descriptor?

ywwg: "compatible." as in, "compatible with legacy systems." I need to pick better names for sure

ywwg: > have we thought about using id to store the hash of the already normalized descriptor?

do you mean only storing the normalized descriptor? We need both in case the user switches the mode to UTF-8 comparison with no legacy compatibility. Or do you mean something else?

ArthurSens: Yes, I meant just storing the normalized descriptor. Good point that's it's possible to change the validation in-flight!

ArthurSens: Not sure why someone would like to do that though 🤔

ywwg: for the record (as discussed in slack), it's because metrics are often created in `init()` functions before the user has a chance to actually configure the correct escaping scheme. So we have to be ready for both scenarios.

prometheus/promhttp/http.go
```diff
@@ -190,6 +193,17 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 		} else {
 			contentType = expfmt.Negotiate(req.Header)
 		}
+
+		if hasEscapedCollisions {
+			switch contentType.ToEscapingScheme() {
+			case model.UnderscoreEscaping, model.DotsEscaping:
+				opts.ErrorLog.Println("error: one or more metrics collide when escaped")
```
ywwg, line 200: need a better error message?

ArthurSens (commented):
Looking good!

Since we always allow registration regardless of collision, I would just move the test closer to the HTTP package. We can keep the same test cases you've added, but the test would look similar to this:

```go
func TestEscapedCollisions(t *testing.T) {
	reg.MustRegister(prometheus.NewCounter(prometheus.CounterOpts{
		Name: "test_metric",
		Help: "A test metric with underscores",
	}))
	reg.MustRegister(prometheus.NewCounter(prometheus.CounterOpts{
		Name: "test.metric",
		Help: "A test metric with dots",
	}))


	
	handler := HandlerFor(reg, HandlerOpts{})
	writer := httptest.NewRecorder()
	request, _ := http.NewRequest("GET", "/metrics", nil)
	request.Header.Add(acceptEncodingHeader, "the header that requires escaping")
	handler.ServeHTTP(writer, request)
	require.Equal(t, "some http error", writer.Code)

	request, _ := http.NewRequest("GET", "/metrics", nil)
	request.Header.Add(acceptEncodingHeader, "the header that allows utf8")
	handler.ServeHTTP(writer, request)
	require.Equal(t, "no erro error", writer.Code)
}
```

prometheus/promhttp/http.go
ArthurSens: maybe something like 

"One or more metrics collide when UTF-8 characters in name/label are translated to underscores"?

prometheus/registry_test.go
```diff
@@ -1181,6 +1182,183 @@ func TestAlreadyRegisteredCollision(t *testing.T) {
...
+			err = reg.Register(tc.counterB())
+			if !tc.expectErr {
+				if err != nil {
+					t.Errorf("expected no error, got %T: %v", err, err)
+				}
+			} else {
+				if err == nil {
+					t.Error("expected an error, got none")
+				}
+			}
```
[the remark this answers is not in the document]
bwplotka: Nope. Please don't add it (: We can built some testutil helper but we went far to not add this dep.

ywwg:
I'll squash this down one we're done with notes

prometheus/promhttp/http_test.go
```diff
@@ -548,6 +551,45 @@ func TestNegotiateEncodingWriter(t *testing.T) {
...
+	reg.MustRegister(prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "test_metric",
+		Help: "A test metric with underscores",
+	}))
+	reg.MustRegister(prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "test.metric",
+		Help: "A test metric with dots",
+	}))
+
+	handler := HandlerFor(reg, HandlerOpts{
```
ywwg, line 571: this is the sucky part, it's necessary to specify the registry here or else we don't detect the collision case (because Gatherer does not know about collisions)

prometheus/registry_test.go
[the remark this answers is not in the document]
ArthurSens: Yeah good point about dependencies, I had the impression we already used it somewhere else

prometheus/registry.go
```diff
@@ -351,23 +402,36 @@ func (r *Registry) Register(c Collector) error {
...
 
 // Unregister implements Registerer.
 func (r *Registry) Unregister(c Collector) bool {
 	var (
-		descChan    = make(chan *Desc, capDescChan)
-		descIDs     = map[uint64]struct{}{}
-		collectorID uint64 // All desc IDs XOR'd together.
+		descChan           = make(chan *Desc, capDescChan)
+		descIDs            = map[uint64]struct{}{}
+		escpaedDescIDs     = map[uint64]struct{}{}
```
ArthurSens, line 432:
```suggestion
		escapedDescIDs     = map[uint64]struct{}{}
```

prometheus/registry_test.go
```diff
@@ -1181,6 +1184,173 @@ func TestAlreadyRegisteredCollision(t *testing.T) {
...
+			reg := prometheus.NewRegistry()
+			err := reg.Register(tc.counterA())
+			require.NoError(t, err)
+			err = reg.Register(tc.counterB())
+			if !tc.expectErr {
+				require.NoError(t, err)
+			} else {
+				require.Error(t, err)
+			}
+			require.Equal(t, tc.expectLegacyCollision, reg.HasEscapedCollision())
```
ArthurSens, line 1349: Sorry for creating the confusion here, that's my bad! [Bartek made a good point for not using require here](https://github.com/prometheus/client_golang/pull/1641#discussion_r1803258350), would you mind reverting the change? 🙈 

We always have problems when we introduce new dependencies, let's avoid it if we can

ArthurSens (approved):
LGTM!

Just one comment about adding another dependency, sorry about that 😬

prometheus/registry.go
ywwg: lol, thanks (applying manually)

prometheus/registry_test.go
ywwg: reverted

prometheus/promhttp/http.go
```diff
@@ -134,6 +136,7 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 			}
 		}
 	}
+	hasEscapedCollisions = reg.HasEscapedCollision()
```
ywwg, line 139: hmm... should I use the singular or the plural? :)

prometheus/promhttp/http_test.go
```diff
@@ -548,6 +554,50 @@ func TestNegotiateEncodingWriter(t *testing.T) {
...
+	reg.MustRegister(prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "test_metric",
+		Help: "A test metric with underscores",
+	}))
+	reg.MustRegister(prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "test.metric",
+		Help: "A test metric with dots",
+	}))
+
+	handler := HandlerFor(reg, HandlerOpts{})
```
ywwg, line 574: tada

prometheus/registry.go
```diff
@@ -194,6 +205,10 @@ func (gf GathererFunc) Gather() ([]*dto.MetricFamily, error) {
 	return gf()
 }
 
+func (gf GathererFunc) HasEscapedCollision() bool {
```
ywwg, line 208: this is the most risky change. both mimir and avalanche use gathererfunc

ywwg: nevermind, avalanche doesn't use it

https://github.com/grafana/mimir/blob/main/pkg/api/handlers.go#L279

prometheus/desc.go
```diff
@@ -57,6 +57,9 @@ type Desc struct {
 	// must be unique among all registered descriptors and can therefore be
 	// used as an identifier of the descriptor.
 	id uint64
+	// escapedID is similar to id, but with the metric and label names escaped
+	// with underscores.
+	escapedID uint64
```
bwplotka, line 62: Why not use `id` for escaped ID?

prometheus/desc.go
```diff
@@ -142,11 +145,18 @@ func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, const
...
 		xxh.WriteString(val)
 		xxh.Write(separatorByteSlice)
+		if i == 0 {
+			val = model.EscapeName(val, model.UnderscoreEscaping)
+		}
+		escapedXXH.WriteString(val)
+		escapedXXH.Write(separatorByteSlice)
 	}
 	d.id = xxh.Sum64()
+	d.escapedID = escapedXXH.Sum64()
```
bwplotka, line 159: Again, why not use id as escaped ID always?

bwplotka (changes requested):
Thanks for this! 

I think I would love to understand this change more before committing here. Asked some likely stupid questions on the issue https://github.com/prometheus/client_golang/issues/1638#issuecomment-2421704717

We never did those kind of checks on HTTP handler only, only on registration layer, I wonder if it's worth to keep that. Having your app working fine and suddenly /metrics are erroring only because 2 series collide is bit unexpected (especially by default) and wasteful, scrape could just ignore one problematic metric.

If anything we could start with a special Gatherer that checks that, without changes to descriptor ever, and see who would actually use it (who would like to opt-in for extra safety here). I would also be happy to consider adding this logic on-registry check/error when escaped metric name is colliding OR only check escaped ones (is there a case when escaped don't collide, but non-escaped collide?)

Is it worth to double check those first @ArthurSens ?

ywwg:
We are going to put this on hold and have a meeting to discuss the various approaches and their pros and cons

ywwg:
https://docs.google.com/document/d/1KS7RuhHZZbxPLoNmRwxCG1csPGPcO5FwDnmgTwv9x9M/edit?tab=t.0

@bwplotka @ArthurSens @beorn7

Issue #1639, opened 2024-10-08
testutil: Better registry state assertion framework/API


bwplotka:
## Context

With our [breaking (for a good reason) testutil `(Gather|Scrape)AndCompare` function change](https://github.com/prometheus/client_golang/pull/1424#issuecomment-2395412321) a good discussions e.g. [here](https://cloud-native.slack.com/archives/C037NMVUAP3/p1728216119838339) emerged where we chatted about what broke and how and what those downstream users (e.g. Mimir, Kubernetes and [others](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil?tab=importedby)) needs in terms of registry testing capabilities.

Historically, `testutil` was meant to be an experimental, internal utils for testing, mostly for client_golang use. Then Prometheus started to use it then everywhere else. Funny enough, we even did [some tech talks with @kakkoyun around why it's useful](https://www.youtube.com/watch?v=LU6D5cNeHks&list=PLUXT7uzX81byVuvjGmF-OopGKZo_wxcmq). Now here we are -- many downstream project use it, primarily for: 

* Close-box testing (asserting state of a component/structure/binary by polling certain metrics), both on [code](https://github.com/kubernetes/component-base/blob/c0cd2c6692d9fc4f5c729eb493696894499cc66a/metrics/prometheus/slis/metrics_test.go#L79C23-L79C39) and [process/container level](https://github.com/efficientgo/e2e/blob/97db25a0c6c073b68a9e746cc247c6e3c2cef097/monitoring/instrumented_test.go#L116).
* Instrumentation testing (are my metrics correctly instrumented).

Notably, when chatting with @dashpole on Kubernetes changes there are two cases currently impossible with client_golang functions:
* Asserting certain metrics is NOT existing in registry at the given moment.
* Mix of existing with certain value or not existing.

One could implement everything on their own using `ToFloat64` etc, but maybe it's time to redesign `testutil`, perhaps as another package or even module?

Let's discuss some ideas here, help wanted to propose some design.

Ideally we verify the design on some downstream user e.g. https://github.com/kubernetes/component-base before releasing (: 

## Yolo Ideas

I guess we could create a redesigned package with a **stable API** and ways to:

* Get float/histogram values for testing purposes (similar to `ToFloat`, just more flexible, more robust)
* Use text format as an expectation with a certain **configurable** semantics around filtering, each metric values or custom conditions?
[...]

PR #1645, opened 2024-10-14
Cut 1.20.5 with the revert of #1424 (compareMetricFamilies)

bwplotka:
I spoke to a few people (e.g. @dashpole @dgrisonnet @kakkoyun @ArthurSens @vesari @logicalhan), and it feels it might be easier for everyone to revert.

cc @machine424 

#### Rationales:

* We feel the pain of the subsequent changes in [Kubernetes](https://github.com/kubernetes/kubernetes/pull/127941) and other places (e.g. Google). **We want to make sure downstream users trust client_golang as a stable and minimal dependency.**
* Reverting this change should NOT break (re-break) users who fixed their tests and switched to a new behaviour.
* We broke users who used this API to test for a metric to be absent without replacement (thanks @dgrisonnet for [innovating here in a PR](https://github.com/prometheus/client_golang/pull/1643)).
* Based on @dashpole work on [the upgrade](https://github.com/kubernetes/kubernetes/pull/127941), this change didn't provide any major "improvement" in Kubernetes tests so far.

I propose we revert this, and work [on a long term solution](https://github.com/prometheus/client_golang/issues/1639). With the team we also discussed a way, where every client_golang release pipeline could propose an upgrade PR to a Kubernetes library that heavily depends on client_golang and check if the CI passes (compatibility check with a big user) -- to avoid similar issues in the future.


ArthurSens (approved):
> Reverting this change should NOT break (re-break) users who fixed their tests and switched to a new behaviour.

LGTM as long as the above was properly tested 😅!

ArthurSens:
Impressive that we've got a fifth patch release this time. As discussed in our call today, creating release candidates would improve everybody's life by a ton! 

Issue opened here: https://github.com/prometheus/client_golang/issues/1646

dashpole (approved):

bwplotka:
> LGTM as long as the above was properly tested 😅!

Yea, we only remove a check, so except this being more error-prone (we will reopen https://github.com/prometheus/client_golang/issues/1351), those who already did a work to adjust their test should be not impacted. However, for empirical test purposes, I kindly asked @krajorama to create a quick PR on Mimir with PR to double check (:

PR #1650, opened 2024-10-15
examples: Follow best practices and established naming conventions

lilic:
This is a nitpick, so feel free to ignore, but from my experience and understanding the best practice for label key naming is to use one word, otherwise using an underscore. Since this is an example users tend to copy, I think correcting it might be a good idea.


examples/customlabels/main.go
```diff
@@ -33,7 +33,7 @@ func main() {
 
 	// Create a new registry.
 	reg := prometheus.NewRegistry()
-	prometheus.WrapRegistererWith(prometheus.Labels{"serviceName": "my-service-name"}, reg).MustRegister(
```
lilic, line 36: Also not sure but it might be worth including the note and link that is mentioned in the `WrapRegistererWith` description, thoughts?

```
// WrapRegistererWith provides a way to add fixed labels to a subset of
// Collectors. It should not be used to add fixed labels to all metrics
// exposed. See also
// https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels
```

bwplotka: Hm good point. One way of doing this is to use a bit more realistic label name and situation. Service name could mean like a gRPC service (ok practice?) or application/service name (anti-pattern, should be a SD role). We clearly wrap GLOBAL process metrics with this label which feels like an anti pattern.

I would be keen on switching to e.g. `component` (quite often representing code component) e.g. something like

```go
func main() {
	flag.Parse()

	// Create a new registry.
	reg.MustRegister(
		collectors.NewGoCollector(),
		collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}),
	)
	
	startComponentA(prometheus.WrapRegistererWith(prometheus.Labels{"component": "A"}, reg))

	// Expose the registered metrics via HTTP.
	// We should see my_counter_objects_total with an extra label.
	// 	# HELP my_counter_objects_total Counter of objects
	//	# TYPE my_counter_objects_total counter
	//	my_counter_objects_total{component="A"} 1
	http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{}))
	log.Fatal(http.ListenAndServe(*addr, nil))
}

func startComponentA(reg prometheus.Registerer) {
	ct := promauto.With(reg).NewCounter(prometheus.CounterOpts{
		Name: "my_counter_objects_total",
		Help: "Counter of objects",
	})

	// ...
	ct.Inc()
}
```

Would like to update this example here, or should we do it in other PRs?

bwplotka (approved):
Yes! Thanks, great catch. All our examples would use some refresh 🤔 added https://github.com/prometheus/client_golang/issues/1652

examples/customlabels/main.go
lilic: > Hm good point. One way of doing this is to use a bit more realistic label name and situation. 

I think thats a good suggestion. It illustrates best practice with an example! 👍  

> Would you like to update this example here, or should we do it in other PRs?

I can update it, will try to get to it this week still.

lilic: @bwplotka I might have gone overboard and adjusted the example to the point where it doesn't make sense or maybe it does? 🫠 Please take a look thanks! Otherwise will adjust to a more simple example. 😄

kakkoyun (approved):
LGTM ❤️

kakkoyun:
@lilic I'm not sure if the tests' failure are related. I'm happy to merge it when the CI is green. 

I'll check whether it's something that a maintainer should take care of soonish 😅 Sorry for the inconvenience if it's the case.

ArthurSens:
> @lilic I'm not sure if the tests' failure are related. I'm happy to merge it when the CI is green.
> 
> I'll check whether it's something that a maintainer should take care of soonish 😅 Sorry for the inconvenience if it's the case.

They are not, it is failing in several PRs 😬

Issue #1617, opened 2024-09-04
Consider supporting constant native histograms


dashpole:
For OpenTelemetry's prometheus exporters, we would like to be able to export OTel exponential histograms: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33703, https://github.com/open-telemetry/opentelemetry-go/issues/5777.

It would be helpful to have a NewConstNativeHistogram constructor, which mirrors the NewConstHistogram constructor.  Unclear if we need all of the variants (Vec, with CT, etc).

Alternatively, we could directly implement the prometheus.Metric interface.  I don't think it would be _that_ hard to do, but I haven't tried.  Feel free to close this issue if we should go that route instead.

ArthurSens:
I'm supportive :)

Maniktherana:
I can give this a shot if thats okay

beorn7:
#1686 fixed this.

PR #1654, opened 2024-10-16
[1617] Add ConstnativeHistogram

shivanthzen:
Fixes: #1617 
Add exported ConstNativeHistogram - this can be used to build native histograms if one already has the relevant data without calling the observe function. A viable use case is while converting exponential histograms to nativehistograms.

### DIFF
--- a/prometheus/histogram.go
+++ b/prometheus/histogram.go
@@ -14,6 +14,7 @@
 package prometheus
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"runtime"
@@ -26,6 +27,11 @@ import (
 
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const (
+	nativeHistogramSchemaMaximum = 8
+	nativeHistogramSchemaMinimum = -4
 )
 
 // nativeHistogramBounds for the frac of observed values. Only relevant for
@@ -1460,9 +1466,9 @@ func pickSchema(bucketFactor float64) int32 {
 	floor := math.Floor(math.Log2(math.Log2(bucketFactor)))
 	switch {
 	case floor <= -8:
-		return 8
+		return nativeHistogramSchemaMaximum
 	case floor >= 4:
-		return -4
+		return nativeHistogramSchemaMinimum
 	default:
 		return -int32(floor)
 	}
@@ -1850,4 +1856,197 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
 	case rIdx > nIdx:
 		n.exemplars = append(n.exemplars[:nIdx], append([]*dto.Exemplar{e}, append(n.exemplars[nIdx:rIdx], n.exemplars[rIdx+1:]...)...)...)
 	}
+}
+
+type constNativeHistogram struct {
+	desc *Desc
+	dto.Histogram
+	labelPairs []*dto.LabelPair
+}
+
+func validateCount(sum float64, count uint64, negativeBuckets, positiveBuckets map[int]int64, zeroBucket uint64) error {
+	var bucketPopulationSum int64
+	for _, v := range positiveBuckets {
+		bucketPopulationSum += v
+	}
+	for _, v := range negativeBuckets {
+		bucketPopulationSum += v
+	}
+	bucketPopulationSum += int64(zeroBucket)
+
+	// If the sum of observations is NaN, the number of observations must be greater or equal to the sum of all bucket counts.
+	// Otherwise, the number of observations must be equal to the sum of all bucket counts .
+
+	if math.IsNaN(sum) && bucketPopulationSum > int64(count) ||
+		!math.IsNaN(sum) && bucketPopulationSum != int64(count) {
+		return errors.New("the sum of all bucket populations exceeds the count of observations")
+	}
+	return nil
+}
+
+// NewConstNativeHistogram returns a metric representing a Prometheus native histogram with
+// fixed values for the count, sum, and positive/negative/zero bucket counts. As those parameters
+// cannot be changed, the returned value does not implement the Histogram
+// interface (but only the Metric interface). Users of this package will not
+// have much use for it in regular operations. However, when implementing custom
+// OpenTelemetry Collectors, it is useful as a throw-away metric that is generated on the fly
+// to send it to Prometheus in the Collect method.
+//
+// zeroBucket counts all (positive and negative)
+// observations in the zero bucket (with an absolute value less or equal
+// the current threshold).
+// positiveBuckets and negativeBuckets are separate maps for negative and positive
+// observations. The map's value is an int64, counting observations in
+// that bucket. The map's key is the
+// index of the bucket according to the used
+// Schema. Index 0 is for an upper bound of 1 in positive buckets and for a lower bound of -1 in negative buckets.
+// NewConstNativeHistogram returns an error if
+//   - the length of labelValues is not consistent with the variable labels in Desc or if Desc is invalid.
+//   - the schema passed is not between 8 and -4
+//   - the sum of counts in all buckets including the zero bucket does not equal the count if sum is not NaN (or exceeds the count if sum is NaN)
+//
+// See https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#exponential-histograms for more details about the conversion from OTel to Prometheus.
+func NewConstNativeHistogram(
+	desc *Desc,
+	count uint64,
+	sum float64,
+	positiveBuckets, negativeBuckets map[int]int64,
+	zeroBucket uint64,
+	schema int32,
+	zeroThreshold float64,
+	createdTimestamp time.Time,
+	labelValues ...string,
+) (Metric, error) {
+	if desc.err != nil {
+		return nil, desc.err
+	}
+	if err := validateLabelValues(labelValues, len(desc.variableLabels.names)); err != nil {
+		return nil, err
+	}
+	if schema > nativeHistogramSchemaMaximum || schema < nativeHistogramSchemaMinimum {
+		return nil, errors.New("invalid native histogram schema")
+	}
+	if err := validateCount(sum, count, negativeBuckets, positiveBuckets, zeroBucket); err != nil {
+		return nil, err
+	}
+
+	NegativeSpan, NegativeDelta := makeBucketsFromMap(negativeBuckets)
+	PositiveSpan, PositiveDelta := makeBucketsFromMap(positiveBuckets)
+	ret := &constNativeHistogram{
+		desc: desc,
+		Histogram: dto.Histogram{
+			CreatedTimestamp: timestamppb.New(createdTimestamp),
+			Schema:           &schema,
+			ZeroThreshold:    &zeroThreshold,
+			SampleCount:      &count,
+			SampleSum:        &sum,
+
+			NegativeSpan:  NegativeSpan,
+			NegativeDelta: NegativeDelta,
+
+			PositiveSpan:  PositiveSpan,
+			PositiveDelta: PositiveDelta,
+
+			ZeroCount: proto.Uint64(zeroBucket),
+		},
+		labelPairs: MakeLabelPairs(desc, labelValues),
+	}
+	if *ret.ZeroThreshold == 0 && *ret.ZeroCount == 0 && len(ret.PositiveSpan) == 0 && len(ret.NegativeSpan) == 0 {
+		ret.PositiveSpan = []*dto.BucketSpan{{
+			Offset: proto.Int32(0),
+			Length: proto.Uint32(0),
+		}}
+	}
+	return ret, nil
+}
+
+// MustNewConstNativeHistogram is a version of NewConstNativeHistogram that panics where
+// NewConstNativeHistogram would have returned an error.
+func MustNewConstNativeHistogram(
+	desc *Desc,
+	count uint64,
+	sum float64,
+	positiveBuckets, negativeBuckets map[int]int64,
+	zeroBucket uint64,
+	nativeHistogramSchema int32,
+	nativeHistogramZeroThreshold float64,
+	createdTimestamp time.Time,
+	labelValues ...string,
+) Metric {
+	nativehistogram, err := NewConstNativeHistogram(desc,
+		count,
+		sum,
+		positiveBuckets,
+		negativeBuckets,
+		zeroBucket,
+		nativeHistogramSchema,
+		nativeHistogramZeroThreshold,
+		createdTimestamp,
+		labelValues...)
+	if err != nil {
+		panic(err)
+	}
+	return nativehistogram
+}
+
+func (h *constNativeHistogram) Desc() *Desc {
+	return h.desc
+}
+
+func (h *constNativeHistogram) Write(out *dto.Metric) error {
+	out.Histogram = &h.Histogram
+	out.Label = h.labelPairs
+	return nil
+}
+
+func makeBucketsFromMap(buckets map[int]int64) ([]*dto.BucketSpan, []int64) {
+	if len(buckets) == 0 {
+		return nil, nil
+	}
+	var ii []int
+	for k := range buckets {
+		ii = append(ii, k)
+	}
+	sort.Ints(ii)
+
+	var (
+		spans     []*dto.BucketSpan
+		deltas    []int64
+		prevCount int64
+		nextI     int
+	)
+
+	appendDelta := func(count int64) {
+		*spans[len(spans)-1].Length++
+		deltas = append(deltas, count-prevCount)
+		prevCount = count
+	}
+
+	for n, i := range ii {
+		count := buckets[i]
+		// Multiple spans with only small gaps in between are probably
+		// encoded more efficiently as one larger span with a few empty
+		// buckets. Needs some research to find the sweet spot. For now,
+		// we assume that gaps of one or two buckets should not create
+		// a new span.
+		iDelta := int32(i - nextI)
+		if n == 0 || iDelta > 2 {
+			// We have to create a new span, either because we are
+			// at the very beginning, or because we have found a gap
+			// of more than two buckets.
+			spans = append(spans, &dto.BucketSpan{
+				Offset: proto.Int32(iDelta),
+				Length: proto.Uint32(0),
+			})
+		} else {
+			// We have found a small gap (or no gap at all).
+			// Insert empty buckets as needed.
+			for j := int32(0); j < iDelta; j++ {
+				appendDelta(0)
+			}
+		}
+		appendDelta(count)
+		nextI = i + 1
+	}
+	return spans, deltas
 }
--- a/prometheus/histogram_test.go
+++ b/prometheus/histogram_test.go
@@ -25,11 +25,11 @@ import (
 	"testing/quick"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus/internal"
-
 	dto "github.com/prometheus/client_model/go"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/prometheus/client_golang/prometheus/internal"
 )
 
 func benchmarkHistogramObserve(w int, b *testing.B) {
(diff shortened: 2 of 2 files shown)


shivanthzen:
@ArthurSens Maybe you can take a look at this PR.

prometheus/histogram.go
```diff
@@ -1835,3 +1835,129 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	nativeHistogramMinResetDuration time.Duration
+	timeStamp                       time.Time
+	nativeExemplars                 []*dto.Exemplar
+
+	positiveBuckets map[int]int64
+	negativeBuckets map[int]int64
+	zeroBucket      uint64
+}
+
+func NewconstNativeHistogram(desc *Desc, count uint64, sum float64, postiveBuckets, negativeBuckets map[int]int64, zeroBucket uint64,
```
ArthurSens, line 1857:
```suggestion
func NewConstNativeHistogram(desc *Desc, count uint64, sum float64, postiveBuckets, negativeBuckets map[int]int64, zeroBucket uint64,
```

ArthurSens: Could you also add.MustNewConstNativeHistogram that re uses the one you just added and panics if there's an error?

This would be consistent with all the other NewCost... functions that we have :)

prometheus/histogram_test.go
```diff
@@ -1455,3 +1455,546 @@ func compareNativeExemplarValues(t *testing.T, exps []*dto.Exemplar, values []fl
...
+func SyncMaptomap(syncmap *sync.Map) (m map[int]int64) {
+	m = map[int]int64{}
+	syncmap.Range(func(key, value any) bool {
+		m[key.(int)] = *value.(*int64)
+		return true
+	})
+	return m
+}
+
+func TestConstNativeHistogram(t *testing.T) {
```
ArthurSens, line 1468: I'm struggling a bit to understand what is being tested here 🤔 

A new constant metric is a fairly simple operation and constant metrics do not have any methods that change their internal fields such as `Observe()`, `Add()` or `Inc()`.  It's so simple that we don't even have tests for `NewConstHistogram` 🤔 

I'm not saying tests aren't useful here, I think it's worth verifying valid labels. I'm not a native histogram expert but maybe there are other things we need to make sure are correct?

prometheus/histogram.go
```diff
@@ -1835,3 +1835,129 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+type constNativeHistogram struct {
+	desc                            *Desc
+	count                           uint64
+	sum                             float64
+	labelPairs                      []*dto.LabelPair
+	nativeHistogramSchema           int32
+	nativeHistogramZeroThreshold    float64
+	nativeHistogramMaxZeroThreshold float64
+	nativeHistogramMaxBuckets       uint32
+	nativeHistogramMinResetDuration time.Duration
```
ArthurSens, line 1848:
```suggestion
```

They are constant and we can't add extra observations, which means we don't need to care about bucket increase or resets

prometheus/histogram.go
```diff
@@ -1835,3 +1835,129 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	timeStamp                       time.Time
```
ArthurSens, line 1849:
```suggestion
	createdTimestamp                       time.Time
```

prometheus/histogram.go
```diff
@@ -1835,3 +1835,129 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	positiveBuckets map[int]int64
+	negativeBuckets map[int]int64
+	zeroBucket      uint64
+}
+
+func NewconstNativeHistogram(desc *Desc, count uint64, sum float64, postiveBuckets, negativeBuckets map[int]int64, zeroBucket uint64,
+	labelPairs []*dto.LabelPair, nativeHistogramSchema int32, nativeHistogramZeroThreshold float64,
+	nativeHistogramMaxZeroThreshold float64, nativeHistogramMaxBuckets uint32,
+	nativeHistogramMinResetDuration time.Duration,
+	timeStamp time.Time,
```
ArthurSens, line 1861:
```suggestion
	createdTimestamp time.Time,
```

or 

```suggestion
	ct time.Time,
```

prometheus/histogram.go
```diff
@@ -1835,3 +1835,129 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	nativeExemplars []*dto.Exemplar,
+) constNativeHistogram {
```
ArthurSens, line 1863:
```suggestion
) (constNativeHistogram, error) {
```

I believe we need to do some validation before returning the object. The first thing I can remember is validating labels, but maybe there are more stuff we need to validate?

ArthurSens (commented):
Thanks for working on this! I've added a few comments but I'm not a native histogram expert, it would be nice if @beorn7 or @krajorama could also take a look here.

question: Have you tried using your commit on OTel's side to make sure you can translate their format into ours?

beorn7:
Will have a look ASAP (which won't be before next week, sorry).

prometheus/histogram_test.go
shivanthzen: THis test is an exact copy of existing test cases for native histograms in the same file.  https://github.com/prometheus/client_golang/blob/1eea8641d00e125557cf50c8123840608c1cba89/prometheus/histogram_test.go#L471 .  The tests ensure that native histograms serialised from an `observe()`ed nativehistogram is the same as the one created from constNativeHistograms - Or that serialisation of constNativehistogram is as it exists today .  Agree that this isn't really testing all the scenarios, I'll add more tests.

prometheus/histogram.go
```diff
@@ -1835,3 +1835,131 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	nativeHistogramMaxZeroThreshold float64
+	createdTimestamp                time.Time
+	nativeExemplars                 []*dto.Exemplar
+
+	positiveBuckets map[int]int64
+	negativeBuckets map[int]int64
+	zeroBucket      uint64
+}
+
+func NewConstNativeHistogram(desc *Desc, count uint64, sum float64, postiveBuckets, negativeBuckets map[int]int64, zeroBucket uint64,
```
krajorama, line 1855: Needs doc string to explain the arguments and the fact that this is exposed as a counter type histogram, treated as a counter on scrape.

krajorama (commented):
Hi, I understand the use case of wanting to expose OTEL exponential histograms from the Prometheus exporter in the collector. How would you use it actually? Constantly register and unregister the metric to update the exposed values? Or just use it's `Write()` method to get to the data?

In any case, the naming is going to be weird as you're not actually using it as a const :)

If I understand correctly, currently it will only work for counter histograms, not gauge histograms due to limitations in client_golang.

I'm a little afraid of abuse of this interface, but don't see true stoppers for now, will consult @beorn7 .

beorn7 (changes requested):
Thanks @shivanthzen for tackling this, and thanks for the review so far @krajorama and @ArthurSens . If I haven't commented on a previous review comment, I agree with it. :)

prometheus/histogram.go
```diff
@@ -1835,3 +1835,131 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+			// Insert empty buckets as needed.
+			for j := int32(0); j < iDelta; j++ {
+				appendDelta(0)
+			}
+		}
+		appendDelta(count)
+		nextI = i + 1
+	}
+	return spans, deltas
+}
```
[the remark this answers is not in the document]
krajorama: ack

shivanthzen:
Thanks for the comments. I'll resolve them this weekend

prometheus/histogram.go
shivanthzen: Tried this.  Wouldn't this blow up the argument list? Is it better to create a seperate a constNaticveGuageHistogram ?

prometheus/histogram.go
```diff
@@ -1835,3 +1835,131 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	nativeExemplars []*dto.Exemplar,
+	labelValues ...string,
+) (Metric, error) {
+	if desc.err != nil {
+		return nil, desc.err
+	}
+	if err := validateLabelValues(labelValues, len(desc.variableLabels.names)); err != nil {
+		return nil, err
+	}
+	return &constNativeHistogram{
```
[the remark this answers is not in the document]
shivanthzen: Just noticed [this](https://github.com/prometheus/client_golang/blob/02b8952e721c7dc7725c4ecba742f74e980f8270/prometheus/histogram_test.go#L678) test case. Will the validation -`Count should match the sum of bucket counts` be valid in this case as well ? It seems Nan observation increments the total count, but doesn't fall into any of the buckets.

prometheus/histogram.go
```diff
@@ -1835,3 +1835,143 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	labelValues ...string,
+) (Metric, error) {
+	if desc.err != nil {
+		return nil, desc.err
+	}
+	if err := validateLabelValues(labelValues, len(desc.variableLabels.names)); err != nil {
+		return nil, err
+	}
+	if nativeHistogramSchema > 8 || nativeHistogramSchema < -4 {
+		return nil, fmt.Errorf("invalid native histogram schema ")
```
ArthurSens, line 1880:
```suggestion
		return nil, fmt.Errorf("invalid native histogram schema")
```

prometheus/histogram.go
```diff
@@ -1835,3 +1835,143 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	}
+	if err := validateCount(count, negativeBuckets, positiveBuckets, zeroBucket); err != nil {
+		return nil, err
+	}
```
ArthurSens, line 1884: Do we need to validate Sum as well?

prometheus/histogram.go
```diff
@@ -1835,3 +1835,143 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	return h.desc
+}
+
+func (h *constNativeHistogram) Write(out *dto.Metric) error {
+	out.Histogram = &h.Histogram
+	out.Label = h.labelPairs
+	return nil
+}
+
+func makeBucketsAny(buckets map[int]int64) ([]*dto.BucketSpan, []int64) {
```
ArthurSens, line 1927: hey, could you clarify what `Any` means here?

ArthurSens (commented):
I had just a quick look today :)

prometheus/histogram.go
```diff
@@ -1835,3 +1835,143 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+		bucktCountSum += v
+	}
+	bucktCountSum += int64(zeroBucket)
+	if bucktCountSum != int64(count) {
+		return fmt.Errorf("bucket sums do not add up to total count")
+	}
+	return nil
+}
+
+func NewConstNativeHistogram(
```
beorn7, line 1860: Needs doc comment.

prometheus/histogram.go
```diff
@@ -1835,3 +1835,143 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	var bucktCountSum int64
+	for _, v := range positiveBuckets {
+		bucktCountSum += v
+	}
+	for _, v := range negativeBuckets {
+		bucktCountSum += v
+	}
+	bucktCountSum += int64(zeroBucket)
+	if bucktCountSum != int64(count) {
+		return fmt.Errorf("bucket sums do not add up to total count")
```
beorn7, line 1855: The count of observations can actually be greater than the sum of all bucket counts because NaN observations go into no bucket (but then the sum of observations must be NaN).

So these are the conditions to test:
- If the sum of observations is NaN, the number of observations must be greater or equal to the sum of all bucket counts.
- Otherwise, the number of observations must be equal to the sum of all bucket counts (what you are validating right now).

As a side note, let's be more explicit in the error message. "bucket sums do not add up to total count" could mean many things. How about "the sum of all bucket populations exceeds the count of observations"?

prometheus/histogram.go
```diff
@@ -1835,3 +1835,143 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
 }
+
+type constNativeHistogram struct {
+	desc *Desc
+	dto.Histogram
+	labelPairs []*dto.LabelPair
+}
+
+func validateCount(count uint64, negativeBuckets, positiveBuckets map[int]int64, zeroBucket uint64) error {
+	var bucktCountSum int64
```
beorn7, line 1846: Maybe `bucketPopulationSum` or just `bucketSum`? `bucketCountSum` could also mean the total number of buckets. (I assume `bucktCountSum` is a typo. I wouldn't make a variable name harder to read by saving only one letter in an abbreviation attempt.)

prometheus/histogram.go
```diff
@@ -1835,3 +1835,143 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+
+func NewConstNativeHistogram(
+	desc *Desc,
+	count uint64,
+	sum float64,
+	positiveBuckets,
+	negativeBuckets map[int]int64,
+	zeroBucket uint64,
+	nativeHistogramSchema int32,
+	nativeHistogramZeroThreshold float64,
```
beorn7, line 1868: The function name makes it clear that we are dealing with native histograms. Let's use shorter names here:

```suggestion
	schema int32,
	zeroThreshold float64,
```

prometheus/histogram.go
```diff
@@ -1835,3 +1835,143 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	createdTimestamp time.Time,
+	nativeExemplars []*dto.Exemplar,
```
beorn7, line 1870: As above, let's just call this `exemplars`.

prometheus/histogram.go
```diff
@@ -1835,3 +1835,143 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	labelValues ...string,
+) (Metric, error) {
+	if desc.err != nil {
+		return nil, desc.err
+	}
+	if err := validateLabelValues(labelValues, len(desc.variableLabels.names)); err != nil {
+		return nil, err
+	}
+	if nativeHistogramSchema > 8 || nativeHistogramSchema < -4 {
```
beorn7, line 1879: I'm wondering if we should finally create constants for the max and min schema. This PR adds yet another occurrence.

prometheus/histogram.go
beorn7: Not really. Sum could by anything. I mean, there could be consistency checks like the sum cannot be negative if there are only positive observations, or the sum has to be zero without any observations, and many more. But I don't know if that would be very helpful. We should only check for conditions that would make Prometheus reject the histogram upon ingestion.

prometheus/histogram.go
beorn7: Probably "positive or negative".
I would just call it `makeBuckets`.

prometheus/histogram.go
```diff
@@ -1835,3 +1835,143 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+
+	NegativeSpan, NegativeDelta := makeBucketsAny(negativeBuckets)
+	PositiveSpan, PositiveDelta := makeBucketsAny(positiveBuckets)
+	ret := &constNativeHistogram{
+		desc: desc,
+		Histogram: dto.Histogram{
+			CreatedTimestamp: timestamppb.New(createdTimestamp),
+			Schema:           &nativeHistogramSchema,
+			ZeroThreshold:    &nativeHistogramZeroThreshold,
+			Exemplars:        nativeExemplars,
```
beorn7, line 1894: Exemplars for native histograms MUST have a timestamp. We should validate that.

prometheus/histogram.go
beorn7: We should also have a version `MustNewConstNativeHistogram` (which wraps `NewConstNativeHistogram`, panicking where `NewConstNativeHistogram` would have returned an error).

beorn7 (changes requested):
Looks good in general.

Despite the frequent nit pickings about naming etc., I haven't yet checked the technical correctness very deeply yet (although it looks fine as far as I can see right now).

prometheus/histogram.go
ArthurSens: +1 for makeBuckets

shivanthzen: Hmm, makebuckets alreadyexists in the namespace. I could do `makeBucketfrommap`

prometheus/histogram.go
shivanthzen: Added the docstring

ArthurSens:
Oh, looks like some commits weren't signed and there's one test failing 🤔 

The test seems to be flaky, it's also failing in other PRs

prometheus/histogram.go
shivanthzen: I'm still unclear about including a ` gauge histogram` . As per my understanding , we will need to add  postiveBuckets, negativeBuckets of type  map[int]float64 as an argument (or use a generic parameter).

shivanthzen:
I can sign the commits once the PR is approved, It seems  signing will need a force push

prometheus/histogram.go
beorn7: It sounds like you are thinking about a float histogram. That's something we might want to add later (ideally together with the option of creating classic float histograms). However, the use case for float histograms in instrumentation is very rare.

I am in fact talking about a gauge histogram. It merely requires to set a different `MetricType` in the `MetricFamily`. However, I've just realized that we currently have no way of communicating the difference of counter histogram (i.e. the "normal" histogram) and gauge histogram during metrics collection. For that, we needed `Desc` to have a metric type field (which was the plan for quite some time).

So I'm pretty sure we cannot solve this problem in this PR, because it needed too much re-architecting.

But if we assume that the difference between counter histogram and gauge histogram will be in the `Desc`, we actually don't need a bool parameter here, but just have to later allow the option of creating "typed" `Desc`s.

In conclusion, let's not do gauge histograms right now.

prometheus/histogram.go
```diff
@@ -1835,3 +1840,208 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+func NewConstNativeHistogram(
+	desc *Desc,
+	count uint64,
+	sum float64,
+	positiveBuckets, negativeBuckets map[int]int64,
+	zeroBucket uint64,
+	schema int32,
+	zeroThreshold float64,
+	createdTimestamp time.Time,
+	exemplars []*dto.Exemplar,
```
beorn7, line 1908: We shouldn't use naked `dto.Exemplar` here. Check out how it is done for `NewMetricWithExemplars`, i.e. there is a `prometheus.Exemplar` and a `newExemplar` constructor to create the `dto.Exemplar`.

Looking at `NewMetricWithExemplars`, however, I'm wondering if we should even allow an exemplar argument here. We could also implement native histogram handling in `withExemplarsMetric` (strictly speaking, we have to do that anyway), and then all `NewConst...` functions would be consistent (as in: not having exemplars, but you use the `NewMetricWithExemplars` to wrap them).

@ArthurSens @bwplotka WDYT?

prometheus/histogram_test.go
```diff
@@ -1455,3 +1455,545 @@ func compareNativeExemplarValues(t *testing.T, exps []*dto.Exemplar, values []fl
 		}
 	}
 }
+
+func syncMaptoMap(syncmap *sync.Map) (m map[int]int64) {
```
beorn7, line 1459: Nit:
```suggestion
func syncMapToMap(syncmap *sync.Map) (m map[int]int64) {
```

beorn7 (changes requested):
Thanks for addressing the feedback.

I have no realized something about exemplars we had not discussed before.

I have also realized we cannot do gauge histograms right now. Needs more discussion and planning, which is out of scope of this PR.

prometheus/histogram.go
```diff
@@ -1835,3 +1840,208 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	if math.IsNaN(sum) && bucketPopulationSum > int64(count) ||
+		!math.IsNaN(sum) && bucketPopulationSum != int64(count) {
+		return fmt.Errorf("the sum of all bucket populations exceeds the count of observations")
+	}
+	return nil
+}
+
+func validateExemplars(exemplars []*dto.Exemplar) error {
+	for _, e := range exemplars {
+		if !e.GetTimestamp().IsValid() {
```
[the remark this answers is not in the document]
shivanthzen: About Nil timestamp:
From https://pkg.go.dev/google.golang.org/protobuf/types/known/timestamppb#Timestamp.CheckValid , 
```
CheckValid returns an error if the timestamp is invalid. In particular, it checks whether the value represents a date that is in the range of 0001-01-01T00:00:00Z to 9999-12-31T23:59:59Z inclusive. An error is reported for a nil Timestamp.
```
More  here: https://github.com/protocolbuffers/protobuf-go/blob/158d2b331a354322bceddf905a52b129d1a740d7/types/known/timestamppb/timestamp.pb.go#L215
In short,
`e.GetTimestamp().IsValid()` will return false for a nil timestamp. No explicit check required.

prometheus/histogram.go
shivanthzen: That's true. I was indeed confused about a float histogram and gauge histogram.

shivanthzen:
Force pushed with signed commits.

prometheus/histogram.go
```diff
@@ -1835,3 +1840,208 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+// the current threshold)
+// positiveBuckets and negativeBuckets are separate maps for negative and positive
+// observations. The map's value is an int64, counting observations in
+// that bucket. The map's key is the
+// index of the bucket according to the used
+// Schema. Index 0 is for an upper bound of 1.
+// NewConstNativeHistogram returns an error if
+// 1.the length of labelValues is not consistent with the variable labels in Desc or if Desc is invalid.
+// 2.the schema passed is not between 8 and -4
+// 3.the sum of counts in the bucket do not equal count
```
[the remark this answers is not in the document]
shivanthzen: updated to `the sum of counts in the bucket do not equal count when sum is not Nan`

prometheus/histogram.go
shivanthzen: I'll wait for a consensus here.

ArthurSens: Sounds reasonable to me :), let's keep this pattern.

ArthurSens:
@shivanthzen, looks like you accidentally added unrelated commits to your PR, could you keep only the ones that you've authored here?

prometheus/histogram.go
ArthurSens: To unblock this PR and make it simpler, maybe here we could leave exemplars out. We deal with exemplars in a follow-up PR that implements `.withExemplars(e []*dto.Exemplar)`. Would that work for both of you @beorn7 @shivanthzen?

beorn7: Yes, that's essentially what I sketched up as plan B above. As said, we have to make `NewMetricWithExemplars` work with native histograms anyway, and then we don't really need it derictly in `NewConstNativeHistogram` anymore.

shivanthzen:
Removed other commits from the branch

krajorama (commented):
couple of minor comments

kakkoyun:
@beorn7 could you have another look at it?

prometheus/histogram.go
```diff
@@ -1835,3 +1840,192 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+// to send it to Prometheus in the Collect method.
+//
+// zeroBucket counts all (positive and negative)
+// observations in the zero bucket (with an absolute value less or equal
+// the current threshold).
+// positiveBuckets and negativeBuckets are separate maps for negative and positive
+// observations. The map's value is an int64, counting observations in
+// that bucket. The map's key is the
+// index of the bucket according to the used
+// Schema. Index 0 is for an upper bound of 1 in positive buckets and for a lower bound of -1 in negative buckets.
```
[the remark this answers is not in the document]
shivanthzen: Is the suggestion to add 

```Note that Prometheus Native Histograms buckets are indexed by upper boundary while OTEL Exponential Histograms are indexed by lower boundary, the result being that the Offset fields are different-by-one.`` ?

**Comment by beorn7** [comment]
This is very close to the top of my review queue now. Will have a look soon, hopefully today.

File: prometheus/histogram.go (hunk 24 above)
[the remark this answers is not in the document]
Comment (beorn7) [reply prometheus/histogram.go:1885]: I think we should just add the link @krajorama suggested. Something like: "See https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#exponential-histograms for more details about the conversion from OTel to Prometheus."

File: prometheus/histogram.go [hunk 25]
```diff
@@ -1835,3 +1840,192 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+// NewConstNativeHistogram returns an error if
+// - the length of labelValues is not consistent with the variable labels in Desc or if Desc is invalid.
+// - the schema passed is not between 8 and -4
+// - the sum of counts in all buckets including the zero bucket does not equal the count when sum  is not Nan
```
beorn7, line 1889: Some wordsmithing, but also note the indentation so that Godoc will actually render this as a list:

```suggestion
//  - the length of labelValues is not consistent with the variable labels in Desc or if Desc is invalid.
//  - the schema passed is not between 8 and -4
//  - the sum of counts in all buckets including the zero bucket does not equal the count if sum is not NaN (or exceeds the count if sum is NaN)
```

prometheus/histogram.go
```diff
@@ -1835,3 +1840,192 @@ func (n *nativeExemplars) addExemplar(e *dto.Exemplar) {
...
+	if *ret.ZeroThreshold == 0 && *ret.ZeroCount == 0 && len(ret.PositiveSpan) == 0 && len(ret.NegativeSpan) == 0 {
+		ret.PositiveSpan = []*dto.BucketSpan{{
+			Offset: proto.Int32(0),
+			Length: proto.Uint32(0),
+		}}
+	}
+	return ret, nil
+}
+
+func MustNewConstNativeHistogram(
```
beorn7, line 1944: This needs a doc comment. (A simple one, see the other `MustNewConst...` functions al reference.)

beorn7 (changes requested):
Just some minor things about doc comments left.

beorn7:
Looks good to me, but there are lint warnings now. (They are coming from a recently added check, but they are still relevant.) Also, DCO signing is missing. @shivanthzen could you address the lint warnings and sign the commits? (Feel free to squash commits into fewer or just one before doing that.)

beorn7 (approved):
Thank you very much. I leave it to the maintainers to do the final approval and merge.

ArthurSens (approved):
LGTM, thank you all for the work here :)

Follow-up here would be to tackle the addition of exemplars!

beorn7:
> Follow-up here would be to tackle the addition of exemplars!

To clarify: This means making [withExemplarsMetric.Write](https://github.com/prometheus/client_golang/blob/291b0b0c426b6fb4025d8a3c2c8e7385026ba1d8/prometheus/metric.go#L180) aware of native histogram exemplars. Then, we only need to update the doc comment of [NewMetricWithExemplar](https://github.com/prometheus/client_golang/blob/291b0b0c426b6fb4025d8a3c2c8e7385026ba1d8/prometheus/metric.go#L233) and everything should work fine.

PR #1656, opened 2024-10-18
api: Create a new Go module for API; with remote write client and handler.

bwplotka:
I wanted to propose`remotewrite` module for generic PRW client and handler code. During PromCon audienced asked exactly about this.

Lot's of assumptions here:

* We want separate module (we don't have processes on how to version it and release).
* We want to expose generated protos, without gogo.
* Do we need v1 PRW?

TODO:
* [ ] Check if we want to reuse this "weird" httpClient, is it useful?
* [ ] Add examples, documentation
* [ ] Try using it in Prometheus before merging.


ArthurSens:
Interesting idea :) 

I like that we're doing this in a separate module, but we could do this move in stages maybe? What do you think of moving our HTTP API (the one that calls Prometheus API for instant_query, query_range, etc) and see how versioning and releases would work with that move?

The rationale I'm having here is that worrying with new functionality AND how to handle versioning in the same PR is probably to much...?

bwplotka:
I think I might then put it in the current module. If I scope out vtprotobuf helpers (but still allow it for advanced users) and vendor backoff we might be able to NOT increase deps here with this 🤔

bwplotka:
But then it's not clear where to put it. `api.go` was built in [very non-flexible way](https://github.com/prometheus/client_golang/blob/34e02e282dc4a3cb55ca6441b489ec182e654d59/api/prometheus/v1/api.go#L464). We do say it's experimental versioning, but as we discussed, this impacts users nevertheless.

Then for remote write we also want to offer receiving handler, which does not fit within `v1 api` really 🤔 

cc @jmichalek132 as it might help you in the implementation of client in Otel?

bwplotka:
Closing this, in favor of https://github.com/prometheus/client_golang/pull/1658

PR #1658, opened 2024-10-21
api: Add remote API with write client; add remote handler.

bwplotka:
This proposes Remote Write API directly in client_golang main module. No dependency added (some new small packages are added under the internal vendored code).

Old attempt https://github.com/prometheus/client_golang/pull/1656

TODO:
* [ ] Add examples, documentation
* [ ] Open PR that uses it on Prometheus


api/prometheus/v1/remote/remote_api.go
```diff
@@ -0,0 +1,376 @@
...
+func (n nopSlogHandler) Enabled(context.Context, slog.Level) bool  { return false }
+func (n nopSlogHandler) Handle(context.Context, slog.Record) error { return nil }
+func (n nopSlogHandler) WithAttrs([]slog.Attr) slog.Handler        { return n }
+func (n nopSlogHandler) WithGroup(string) slog.Handler             { return n }
+
+// NewAPI returns a new API for the clients of Remote Write Protocol.
+//
+// It is not safe to use the returned API from multiple goroutines, create a
+// separate *API for each goroutine.
+func NewAPI(c api.Client, opts ...APIOption) (*API, error) {
```
bwplotka, line 88: We could consider removing error from here and from `APIOption` as we don't have any option with validation or so. However I added it for future compatibility (we won't be able to easily extend it without breaking compatibility). Note: Currently api/ is experimenta... 🙈  I think I will remove error here and hope for the best, YAGNI.

```suggestion
func NewAPI(c api.Client, opts ...APIOption) *API {
```

api/prometheus/v1/remote/genproto/v2/symbols.go
```diff
@@ -0,0 +1,97 @@
...
+	for k := range t.symbolsMap {
+		if k == "" {
+			continue
+		}
+		delete(t.symbolsMap, k)
+	}
+}
+
+// DesymbolizeLabels decodes label references, with given symbols to labels.
+func DesymbolizeLabels(labelRefs []uint32, symbols, buf []string) []string {
```
jmichalek132, line 91: yeah I just ran into DesymbolizeLabels not being public in prometheus codebase here https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35751#discussion_r1807920602

bwplotka: Do you use labels.Labels or some other struct (for labels in Otel)

api/prometheus/v1/remote/remote_api.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+type handler struct {
+	logger *slog.Logger
+	store  writeStorage
+}
+
+// NewRemoteWriteHandler returns HTTP handler that receives Remote Write 2.0
+// protocol https://prometheus.io/docs/specs/remote_write_spec_2_0/.
+func NewRemoteWriteHandler(logger *slog.Logger, store writeStorage) http.Handler {
```
bwplotka, line 289: It's bit inconsistent with NewRemoteAPI in terms of options, let's unify one to one or another 🤔

api/prometheus/v1/remote/remote_headers.go
```diff
@@ -0,0 +1,178 @@
...
+
+const (
+	// SnappyBlockCompression represents https://github.com/google/snappy/blob/2c94e11145f0b7b184b831577c93e5a41c4c0346/format_description.txt
+	SnappyBlockCompression Compression = "snappy"
+)
+
+// WriteProtoFullName represents the fully qualified name of the protobuf message
+// to use in Remote write 1.0 and 2.0 protocols.
+// See https://prometheus.io/docs/specs/remote_write_spec_2_0/#protocol.
+type WriteProtoFullName protoreflect.FullName
```
bwplotka, line 46: nit: Change to string, bit too hard to use

api/prometheus/v1/remote/remote_headers.go
```diff
@@ -0,0 +1,178 @@
...
+
+const (
+	// WriteProtoFullNameV1 represents the `prometheus.WriteRequest` protobuf
+	// message introduced in the https://prometheus.io/docs/specs/remote_write_spec/.
+	// DEPRECATED: Use WriteProtoFullNameV2 instead.
```
bwplotka, line 51: For receivers who wants to support it for longer, deprecation might cause some friction, but maybe worth it? 🤔

bwplotka:
Example use of handler in `sink` project: https://github.com/bwplotka/sink/blob/main/go/sink/main.go#L51

api/prometheus/v1/remote/genproto/v2/symbols.go
jmichalek132: In the translation layer we have `createAttributes` https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/prometheusremotewrite/helper.go#L99 which returns `[]prompb.Label` and here we store them https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/prometheusremotewrite/metrics_to_prw_v2.go#L119

api/prometheus/v1/remote/remote_api.go
```diff
@@ -0,0 +1,376 @@
...
+			*tmpbuf = make([]byte, n)
+		}
+		return compressed, nil
+	default:
+		return compressed, fmt.Errorf("unknown compression scheme [%v]", enc)
+	}
+}
+
+func (r *API) attemptWrite(ctx context.Context, compr Compression, proto WriteProtoFullName, payload []byte, attempt int) (WriteResponseStats, error) {
+	u := r.client.URL("api/v1/write", nil)
```
cstyan, line 217: this should be configurable imo, not all remote write receivers use the same endpoint

api/prometheus/v1/remote/remote_api.go
```diff
@@ -0,0 +1,376 @@
...
+	resp, body, err := r.client.Do(ctx, req)
+	if err != nil {
+		// Errors from Client.Do are likely network errors, so recoverable.
+		return WriteResponseStats{}, retryableError{err, 0}
+	}
+
+	rs, err := parseWriteResponseStats(resp)
+	if err != nil {
+		r.opts.logger.Warn("parsing rw write statistics failed; partial or no stats", "err", err)
+	}
```
cstyan, line 247: should we only try to get write response stats/log a warning if the request type was v2?

api/prometheus/v1/remote/remote_api.go
```diff
@@ -0,0 +1,376 @@
...
+	store  writeStorage
+}
+
+// NewRemoteWriteHandler returns HTTP handler that receives Remote Write 2.0
+// protocol https://prometheus.io/docs/specs/remote_write_spec_2_0/.
+func NewRemoteWriteHandler(logger *slog.Logger, store writeStorage) http.Handler {
+	return &handler{logger: logger, store: store}
+}
+
+func parseProtoMsg(contentType string) (WriteProtoFullName, error) {
```
cstyan, line 293: a comment with an example of what the string might look like would be useful for people who haven't read or fully understood the v2 spec

internal/github.com/efficientgo/core/backoff/backoff.go
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) The EfficientGo Authors.
+// Licensed under the Apache License 2.0.
+
+// Initially copied from Cortex project.
+
+// Package backoff implements backoff timers which increases wait time on every retry, incredibly useful
+// in distributed system timeout functionalities.
```
cstyan, line 7: should we be pulling this via `go.mod` as opposed to having copies of the files? or is this the way we include dependencies for the client library?

cstyan (commented):
> Both API (client) and handler supports 1.0 and 2.0 Proto messages, but I intentionally only host writev2.Request generated Go code as we plan to deprecate 1.0 and users should use writev2 now. Both API and handler also will work against protobuf message types that use custom generators e.g. with gogoproto or anything else.

Given that the spec is only just finalized recently and (afaik) still no receivers like Mimir support v2, we should include v1 request code imo.

internal/github.com/efficientgo/core/backoff/backoff.go
ArthurSens: We try as hard as we can to avoid dependencies. I'm in favor of copying other libraries code (respecting copyright) if possible

api/prometheus/v1/remote/remote_api.go
yeya24, line 293: Is this something we can expose?

api/prometheus/v1/remote/remote_api.go
```diff
@@ -0,0 +1,376 @@
...
+		return
+	}
+
+	decompressed, err := snappy.Decode(nil, body)
+	if err != nil {
+		// TODO(bwplotka): Add more context to responded error?
+		h.logger.Error("Error decompressing remote write request", "err", err.Error())
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
```
yeya24, line 358: I wonder if we can expose a hook for the logic of decoding and decompressing part so that downstream project can just use their existing code without many adjustment. 

```
func decodeRequestBody(r *http.Request) (body []byte, err error)
```

The main logic that can benefit us is handling v1 and v2 proto and response headers. It would be nice to still reuse the other code.

api/prometheus/v1/remote/remote_api.go
```diff
@@ -0,0 +1,376 @@
...
+	u := r.client.URL("api/v1/write", nil)
+	req, err := http.NewRequest(http.MethodPost, u.String(), bytes.NewReader(payload))
+	if err != nil {
+		// Errors from NewRequest are from unparsable URLs, so are not
+		// recoverable.
+		return WriteResponseStats{}, err
+	}
+
+	req.Header.Add("Content-Encoding", string(compr))
+	req.Header.Set("Content-Type", contentTypeHeader(proto))
```
saswatamcode, line 226: Maybe there should be an option to specify custom headers here too?

api/prometheus/v1/remote/remote_api.go
```diff
@@ -0,0 +1,376 @@
...
+		r.opts.logger.Warn("parsing rw write statistics failed; partial or no stats", "err", err)
+	}
+
+	if resp.StatusCode/100 == 2 {
+		return rs, nil
+	}
+
+	err = fmt.Errorf("server returned HTTP status %s: %s", resp.Status, body)
+	if resp.StatusCode/100 == 5 ||
+		(r.opts.retryOnRateLimit && resp.StatusCode == http.StatusTooManyRequests) {
```
saswatamcode, line 255: Should retryOn429 be configurable here?

api/prometheus/v1/remote/remote_api.go
bwplotka: You can do that with tripperware though 🤔

api/prometheus/v1/remote/remote_api.go
bwplotka: sure!

api/prometheus/v1/remote/remote_api.go
```diff
@@ -0,0 +1,376 @@
...
+
+type vtProtoEnabled interface {
+	SizeVT() int
+	MarshalToSizedBufferVT(dAtA []byte) (int, error)
+}
+
+// Write writes given, non-empty, protobuf message to a remote storage.
+// The https://github.com/planetscale/vtprotobuf methods will be used if your msg
+// supports those (e.g. SizeVT() and MarshalToSizedBufferVT(...)), for efficiency.
+func (r *API) Write(ctx context.Context, msg proto.Message) (_ WriteResponseStats, err error) {
```
saswatamcode, line 123: Ok I sort of hit a snag with this. `proto.Message` implements a `protoreflect.ProtoMessage` interface, which basically means the type would need a `ProtoReflect() protoreflect.Message` method.

This is absent in the case of both [v1](https://github.com/prometheus/prometheus/blob/main/prompb/remote.pb.go#L76) and [v2](https://github.com/prometheus/prometheus/blob/main/prompb/io/prometheus/write/v2/types.pb.go#L130) in Prometheus prompb, which are generated with `protoc-gen-gogofast` I think. 
The method is added when you generate code with buf here (which afaiu uses `protoc-gen-go`).

However, since for v1, we still need to import WriteRequest from prometheus, it can't be passed into this.

To make this method usable for both v1/v2, maybe we could also import v1 proto with buf here? 🙂

bwplotka: Nice, good finding! I think the solution here would be to use `any` or generics that try ProtoReflect, VT Unmarshal or old gogo unmarshal. This way we support all common deserialization methods. Alternatively we could do our own method so anyone can implement.

In your case v1 won't help you because Thanos likely needs gogo (or want to upgrade slowly here). Any if v1 help you can generate yourself too. Still, more relaxed interface here would be way to go. Wanna try it?

saswatamcode: Yup! Will try to add something

saswatamcode: Trying here: https://github.com/prometheus/client_golang/pull/1710

saswatamcode: Thanks for merging. Follow up on moving to exp https://github.com/prometheus/client_golang/pull/1711

exp/api/remote/remote_api.go
```diff
@@ -0,0 +1,530 @@
...
+func (n nopSlogHandler) Enabled(context.Context, slog.Level) bool  { return false }
+func (n nopSlogHandler) Handle(context.Context, slog.Record) error { return nil }
+func (n nopSlogHandler) WithAttrs([]slog.Attr) slog.Handler        { return n }
+func (n nopSlogHandler) WithGroup(string) slog.Handler             { return n }
+
+// NewAPI returns a new API for the clients of Remote Write Protocol.
+//
+// It is not safe to use the returned API from multiple goroutines, create a
+// separate *API for each goroutine.
+func NewAPI(client *http.Client, baseURL string, opts ...APIOption) (*API, error) {
```
bwplotka, line 117: healthy default for client?

exp/api/remote/remote_api.go
```diff
@@ -0,0 +1,530 @@
...
+// Write writes given, non-empty, protobuf message to a remote storage.
+//
+// Depending on serialization methods,
+//   - https://github.com/planetscale/vtprotobuf methods will be used if your msg
+//     supports those (e.g. SizeVT() and MarshalToSizedBufferVT(...)), for efficiency
+//   - Otherwise https://github.com/gogo/protobuf methods (e.g. Size() and MarshalToSizedBuffer(...))
+//     will be used
+//   - If neither is supported, it will marshaled using generic google.golang.org/protobuf methods and
+//     error out on unknown scheme.
+func (r *API) Write(ctx context.Context, msg any) (_ WriteResponseStats, err error) {
```
bwplotka, line 182: It might be more future proof to allow setting content type here (with our constant).

exp/api/remote/remote_api.go
```diff
@@ -0,0 +1,530 @@
...
+	// Reset the buffer.
+	r.reqBuf = r.reqBuf[:0]
```
bwplotka, line 184: This is good for single thread, but people might want to use the same http client etc for multiple concurrent streams, so we probably can't do this. I would remove and in future add (without breaking variadic options for write, we could do it now too it's a common with `WithWriteBuffer`)

exp/api/remote/remote_api.go
```diff
@@ -0,0 +1,530 @@
...
+			// Replace the body with decompressed data
+			r.Body = io.NopCloser(bytes.NewReader(decompressed))
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+// NewRemoteWriteHandler returns HTTP handler that receives Remote Write 2.0
+// protocol https://prometheus.io/docs/specs/remote_write_spec_2_0/.
+func NewRemoteWriteHandler(store writeStorage, opts ...HandlerOption) http.Handler {
```
bwplotka, line 437:
```suggestion
func NewHandler(store writeStorage, opts ...HandlerOption) http.Handler {
```

exp/api/remote/remote_api.go
```diff
@@ -0,0 +1,530 @@
...
+	if err != nil {
+		return 0
+	}
+	return time.Duration(d) * time.Second
+}
+
+// writeStorage represents the storage for RemoteWriteHandler.
+// This interface is intentionally private due its experimental state.
+type writeStorage interface {
+	Store(ctx context.Context, proto WriteProtoFullName, serializedRequest []byte) (_ WriteResponseStats, code int, _ error)
```
bwplotka, line 364: Discussion, should it by `any` and people provide a type for `NewHandler`? This feels tricky as user would need to give us a map of content type -> type and the exacty method to use for parsing, kind of sily.

However, if we don't do this, should we just give http.Request?

bwplotka: Let's focus on value we bring vs write your own ServeHTTP:

* We write write response stats for you (limiting error surface) with correct success response code
* We validate headers and respond accordingly
* We decompress for you

After that we can't do much.

Disadventages of http.Request

* user don't know what we validated, we can write down we validated proto full name, but should we give it in a separate parameter or double write with both param and header, or only header
* should we remove proto and compression (encoding) headers from request? What if user want to have a metric that instruments what compression was used (then, they could use middleware).
* We kind of have to put decompressed content into io.Reader of request to avoid duplicate data.
* We have to give options to limit known ctypes to X.

```suggestion
        // Store stores remote write metrics encoded in the given WriteFullName content type. Provided http.Request contains the encoded bytes in the req.Body with all the HTTP information, except "Content-Type" header which is provided in a separate, validated ctype. Other headers might be trimmed, 
        // depending on the configured middlewares e.g. a default SnappyMiddleware trims "Content-Encoding".
        // 
        // Store blocks an HTTP request and waits for WriteResponseStats to provide to the client.
	Store(ctx context.Context, ctype WriteFullName, req *http.Request) (_ WriteResponseStats, code int, _ error)
```

Questions:
* How to attach extra response headers e.g. Retry-After.

bwplotka: For questions, we might want to create more complex `WriteResponse` type that gives a way to return stats (force user ideally) and header. However, should we HTTP style input param?

bwplotka:
```suggestion
	// Store stores remote write metrics encoded in the given WriteFullName content type. Provided http.Request contains the encoded bytes in the req.Body with all the HTTP information, except "Content-Type" header which is provided in a separate, validated ctype. Other headers might be trimmed, 
        // depending on the configured middlewares e.g. a default SnappyMiddleware trims "Content-Encoding".
        // 
        // Store blocks an HTTP request and waits for WriteResponseStats to provide to the client.
	Store(ctx context.Context, ctype WriteFullName, req *http.Request) (_ *WriteResponse, _ error)
```

```
type Stats struct {
	// Samples represents X-Prometheus-Remote-Write-Written-Samples
	Samples int
	// Histograms represents X-Prometheus-Remote-Write-Written-Histograms
	Histograms int
	// Exemplars represents X-Prometheus-Remote-Write-Written-Exemplars
	Exemplars int

	// Confirmed means we can trust those statistics from the point of view
	// of the PRW 2.0 spec. When parsed from headers, it means we got at least one
	// response header from the Receiver to confirm those numbers, meaning it must
	// be at least 2.0 Receiver. See ParseWriteResponseStats for details.
	confirmed bool
}

func StatsFromV2() Stats

func NewWriteResponse(s Stats) *WriteResponse {
    ....
}

func (r *WriteResponse) SetHeaders

func (r *WriteResponse) SetCode
```

exp/api/remote/remote_headers.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return ret
+}
+
+func (m protoMsgs) String() string {
+	return strings.Join(m.Strings(), ", ")
+}
+
+var contentTypeHeaders = map[WriteProtoFullName]string{
+	WriteProtoFullNameV1: appProtoContentType, // Also application/x-protobuf;proto=prometheus.WriteRequest but simplified for compatibility with 1.x spec.
```
bwplotka, line 81: Let's make it `WriteV1ContentType`

bwplotka (commented):
Thanks @saswatamcode for helping a lot. Let's get those last things to this branch + quick README and let's get it merged in the experimental module!

exp/api/remote/remote_api.go
```diff
@@ -0,0 +1,530 @@
...
+			enc := r.Header.Get("Content-Encoding")
+			if enc != "" && enc != string(SnappyBlockCompression) {
+				err := fmt.Errorf("%v encoding (compression) is not accepted by this server; only %v is acceptable", enc, SnappyBlockCompression)
+				logger.Error("Error decoding remote write request", "err", err)
+				http.Error(w, err.Error(), http.StatusUnsupportedMediaType)
+				return
+			}
+
+			// Read the request body.
+			bodyBytes, err := io.ReadAll(r.Body)
```
GiedriusS, line 413: I think there's an opportunity to use a sync.Pool here.

exp/api/remote/remote_api.go
```diff
@@ -0,0 +1,530 @@
...
+
+	msgType, err := ParseProtoMsg(contentType)
+	if err != nil {
+		h.opts.logger.Error("Error decoding remote write request", "err", err)
+		http.Error(w, err.Error(), http.StatusUnsupportedMediaType)
+		return
+	}
+
+	// Read the already decompressed body
+	body, err := io.ReadAll(r.Body)
```
GiedriusS, line 507: By using a `bytes.Buffer` instead of a `bytes.Reader` we could access the original bytes directly through `Bytes()` and avoid a copy/allocation here.

PR #1673, opened 2024-11-04
Duplicate of #1662

imorph:
This changes were supposed to be merged in #1662 but after some git manipulations it closed itself.
Here clean version of it.


ArthurSens (approved):
LGTM, thank you!

I'll wait a couple of days in case another maintainer wants to take a look before merging

PR #1675, opened 2024-11-07
process_collector: avoid a compiler warning on macOS (fixes #1660)

### LINKED ISSUE #1660: darwin: Building with CGO produces warning
```
# github.com/prometheus/client_golang/prometheus
In file included from process_collector_cgo_darwin.c:21:
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/mach/shared_memory_server.h:48:2: warning: "<mach/shared_memory_server.h> is deprecated.  Please use <mach/shared_region.h> instead." [-W#warnings]
```

This is on the latest client_golang (main) version, likely after https://github.com/prometheus/client_golang/pull/1616

Does not look serious, but something to fix cc @mharbison72

mharbison72:
The header has a warning when included, with no way to shut it off, and no alternative to obtain these symbols.  They're technically architecture specific values, but they aren't different between amd64 and arm64, so combine the definitions.

@bwplotka

### DIFF
--- a/prometheus/process_collector_cgo_darwin.c
+++ b/prometheus/process_collector_cgo_darwin.c
@@ -15,11 +15,15 @@
 
 #include <mach/mach_init.h>
 #include <mach/task.h>
-// Compiler warns that shared_memory_server.h is deprecated, use this instead.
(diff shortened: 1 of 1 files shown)


bwplotka:
Thanks!

PR #1676, opened 2024-11-08
[chore]: enable perfsprint linter

mmorel-35:
#### Description

[perfsprint](https://golangci-lint.run/usage/linters/#perfsprint) checks that fmt.Sprintf can be replaced with a faster alternative.


kakkoyun (approved):
LGTM. Thanks for noticing and taking care of it.

PR #1677, opened 2024-11-08
chore: use testify instead of t.Fatal

mmorel-35:
This uses testify instead of testing for t.Fatal calls

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,21 +12,25 @@ require (
 	github.com/prometheus/client_model v0.6.1
 	github.com/prometheus/common v0.60.1
 	github.com/prometheus/procfs v0.15.1
+	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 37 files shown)


go.mod
```diff
@@ -12,21 +12,25 @@ require (
 	github.com/prometheus/client_model v0.6.1
 	github.com/prometheus/common v0.60.1
 	github.com/prometheus/procfs v0.15.1
+	github.com/stretchr/testify v1.9.0
```
bwplotka, line 15: We don't want to add extra dependency on purpose in client_golang for little value.

bwplotka (changes requested):
Thanks for this, but the last maintainer's decision is to not include extra dependencies which are not mandatory for the project. While a bit verbose, we kept client_golang without testify for long and it would be nice to keep that.

PR #1685, opened 2024-11-13
feat: add nocache url parameter option

scott-quinlan:


scott-quinlan:
@kakkoyun could I please have this reviewed

vesari (commented):
Hello! Could you please provide some additional context? Currently the `nocache` parameter doesn't seem to be listed among the supported query parameters for any of the Prometheus server endpoints https://prometheus.io/docs/prometheus/latest/querying/api/ .

bwplotka:
👋🏽  Thanks @scott-quinlan for the PR, but as @vesari mentioned it would be nice to know the rationales as it's not a standard Prometheus URL option. As a result I would rather not add that in client_golang, since you can add that parameter (or anything to http.Request) with your own "tripperware" similar to the example [here](https://github.com/prometheus/client_golang/blob/16ca7d8770f0ea31927f546a4717d705da76ce40/api/prometheus/v1/example_test.go#L88).

WDYT? (:

kakkoyun:
@scott-quinlan Friendly ping.

PR #1686, opened 2024-11-17
Add exemplars for native histograms

shivanthzen:
Support exemplars in native histograms:
https://github.com/prometheus/client_golang/issues/1617


shivanthzen:
@ArthurSens @beorn7 Part-2 for [native histogram support](https://github.com/prometheus/client_golang/issues/1617)

beorn7:
This has been on my review list for a while. It's just hard to catch up through all my backlog… :face_exhaling:

prometheus/metric.go
```diff
@@ -187,6 +187,12 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
 		pb.Counter.Exemplar = m.exemplars[len(m.exemplars)-1]
 	case pb.Histogram != nil:
 		for _, e := range m.exemplars {
+			if pb.Histogram.Schema != nil {
+				if *pb.Histogram.Schema > math.MinInt32 && e.GetTimestamp() != nil {
+					pb.Histogram.Exemplars = append(pb.Histogram.Exemplars, e)
+				}
+				continue
```
beorn7, line 194: If a classic and a native histogram is exposed at the same time (for the scraper to pick), we should have the exemplars on both. Maybe the following works for that:

```suggestion
				if len(pb.Histogram.Bucket) == 0 {
					// Don't proceed to classic buckets if there are none.
					continue
				}
```

We do have to check if there are already buckets defined, because otherwise there will be an Inf bucket created just for the histograms.

prometheus/metric_test.go
```diff
@@ -90,3 +92,29 @@ func TestWithExemplarsMetric(t *testing.T) {
...
+
+func TestWithExemplarsNativeHistogramMetric(t *testing.T) {
+	t.Run("histogram", func(t *testing.T) {
+		// Create a constant histogram from values we got from a 3rd party telemetry system.
+		h := MustNewConstNativeHistogram(
+			NewDesc("http_request_duration_seconds", "A histogram of the HTTP request durations.", nil, nil),
+			10, 12.1, map[int]int64{1: 7, 2: 1, 3: 2}, map[int]int64{}, 0, 2, 0.2, time.Date(
+				2009, 11, 17, 20, 34, 58, 651387237, time.UTC))
+		m := &withExemplarsMetric{Metric: h, exemplars: []*dto.Exemplar{
+			{Value: proto.Float64(2000.0), Timestamp: timestamppb.New(time.Date(2009, 11, 17, 20, 34, 58, 3243244, time.UTC))},
```
beorn7, line 104: We should also test the following cases:

- More than one exemplar.
- An exemplar without timestamp (to verify that a timestamp is added automatically then, because exemplars without timestamps are not allowed for native histograms).
- A histogram with both classic and native buckets. (I believe for that you might have to create an ad-hoc implementation of `Metric` that spits out a custom-made `dto.Metric`.)

prometheus/metric.go
```diff
@@ -227,6 +233,7 @@ type Exemplar struct {
 // Only last applicable exemplar is injected from the list.
 // For example for Counter it means last exemplar is injected.
 // For Histogram, it means last applicable exemplar for each bucket is injected.
+// Note that for a nativeHistogram, all valid exemplars are injected.
```
beorn7, line 236:
```suggestion
// For a Native Histogram, all valid exemplars are injected.
```

beorn7 (changes requested):
I think this implements the intended idea. Thank you very much. However, I think we should also be able to handle a histogram with both native and classic buckets (even though we have no `MustNewConst...` constructor for it, but people could create it on their own using `dto` primitives). And we need to test that. See comments.

beorn7:
@shivanthzen let us know when it is ready for the next round of review.

prometheus/metric_test.go
```diff
@@ -90,3 +93,263 @@ func TestWithExemplarsMetric(t *testing.T) {
...
+	})
+}
+
+func PointOf[T any](value T) *T {
+	return &value
+}
+
+// This is a dummy Histogram which has both buckets and nativehistogram
+// to test metrics with exemplars
+func NewDummyConstNativeHistogram(
```
beorn7, line 298: Let's not export this, so that nobody will use this for real. Also, how about a more descriptive name like "newNativeHistogramWithClassicBuckets"? And finally, end a sentence with a period. :)

```suggestion
// newNativeHistogramWithClassicBuckets returns a Metric representing
// a native histogram that also has classic buckets. This is for testing purposes.
func newNativeHistogramWithClassicBuckets(
```

prometheus/metric.go
```diff
@@ -187,6 +187,15 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
 		pb.Counter.Exemplar = m.exemplars[len(m.exemplars)-1]
 	case pb.Histogram != nil:
 		for _, e := range m.exemplars {
+			if pb.Histogram.Schema != nil {
+				if *pb.Histogram.Schema > math.MinInt32 && e.GetTimestamp() != nil {
```
beorn7, line 191: While we use `math.MinInt32` as a schema within the `histogram` struct to mark the absence of sparse buckets, we do not use the same marker in the protobuf. In the protobuf, we know it represents a native histogram if the zero threshold or the zero count or the number of negative spans or the number of positive spans are not zero. (If need be, we add a no-op span.)

So what we have to check is something like the following:
```Go
h := pb.Histogram
if (*h.ZeroThreshold != 0 || *h.ZeroCount != 0 || len(h.PositiveSpan) != 0 || len(h.NegativeSpan) != 0) && e.GetTimestamp() != nil {
```

beorn7 (changes requested):
Thank you. I have just a naming nit and also just discovered that we need one small actual code change after all.

shivanthzen:
@beorn7

beorn7 (changes requested):
Sorry for the review delay.

A small improvement, see comment.

prometheus/metric.go
```diff
@@ -186,21 +186,31 @@ func (m *withExemplarsMetric) Write(pb *dto.Metric) error {
 	case pb.Counter != nil:
 		pb.Counter.Exemplar = m.exemplars[len(m.exemplars)-1]
 	case pb.Histogram != nil:
+		h := pb.Histogram
 		for _, e := range m.exemplars {
-			// pb.Histogram.Bucket are sorted by UpperBound.
-			i := sort.Search(len(pb.Histogram.Bucket), func(i int) bool {
-				return pb.Histogram.Bucket[i].GetUpperBound() >= e.GetValue()
+			if (h.GetZeroThreshold() != 0 ||
+				(h.ZeroCount != nil && *h.ZeroCount != 0) || len(h.PositiveSpan) != 0 ||
```
beorn7, line 192: As said, the same trick works here:

```suggestion
				(h.GetZeroCount() != 0) || len(h.PositiveSpan) != 0 ||
```

With this in, you should probably rearrange the line breaks so the grouping makes more sense.

beorn7 (changes requested):
Could you also resolve the merge conflicts?

beorn7 (approved):
Looks good now. Thank you.

PR #1687, opened 2024-11-19
Fix: handle nil variableLabels in Desc.String() method and add tests for nil label values

### LINKED ISSUE #1684: prometheus.InvalidDesc.String() is formatted incorrectly
On v1.20.5, if String() method of a prometheus.Desc returned from NewInvalidDesc called, it will cause a nil pointer derefence, which will result in an incorrectly formatted string

```golang
type Desc struct {
	//...
	variableLabels *compiledLabels
        //...
}

func NewInvalidDesc(err error) *Desc {
	return &Desc{
		err: err,
	}
}

func (d *Desc) String() string {
	//...
	vlStrings := make([]string, 0, len(d.variableLabels.names))
	for _, vl := range d.variableLabels.names {
        //...
}
```
 NewInvalidDesc will create a Desc with `variableLabels` equal to nil, therefore accessing its field `names` will cause a nil-pointer dereference.

The result is: `descriptor %!s(PANIC=String method: runtime error: invalid memory address or nil pointer dereference) is invalid: test`

The consequence of that is an unpleasant error message in prometheus.Register

kakkoyun:
Fixes #1684


bwplotka (approved):
Thanks!

PR #1711, opened 2025-01-17
Move remote write API to client_golang/exp

saswatamcode:
Moving to exp.

Do we want to create new client in this remote write API addition as well? Or do future iteration on it? :)

I might add the decompression middleware stuff here as well

cc: @bwplotka


bwplotka:
I would straight away try to avoid old client shenigans and rethink this from scratch.

exp/api/remote/remote_api.go
```diff
@@ -22,23 +22,26 @@ import (
...
 )
 
 // API is a client for Prometheus Remote Protocols.
 // NOTE(bwplotka): Only https://prometheus.io/docs/specs/remote_write_spec_2_0/ is currently implemented,
 // read protocols to be implemented if there will be a demand.
 type API struct {
-	client api.Client
-	opts   apiOpts
+	baseURL *url.URL
+	client  *http.Client
```
saswatamcode, line 42: Passing http.Client directly here instead of api.Client structs. Seems like the most flexible way, without doing custom configs.

exp/api/remote/remote_api.go
```diff
@@ -360,22 +374,71 @@ func WithHandlerLogger(logger *slog.Logger) HandlerOption {
...
 	return func(o *handlerOpts) {
-		o.decompressor = decompressor
+		o.middlewares = middlewares
+	}
+}
+
+// SnappyDecompressorMiddleware returns a middleware that checks if the request body is snappy-encoded and decompresses it.
+// If the request body is not snappy-encoded, it returns an error.
+// Used by default in NewRemoteWriteHandler.
+func SnappyDecompressorMiddleware(logger *slog.Logger) func(http.Handler) http.Handler {
```
saswatamcode, line 390: Tried to convert this into a middleware with the ability for downstream users to override and add their own middlewares if needed.

Let me know what you think, or if this is too much

bwplotka:
Can we create a go mod file? It has to be a separate Go module if it will be experimental.

kakkoyun (commented):
Looking good.

We need documentation:
- We need a README. And we need to make the intention crystal clear.
- Also, we have to be clear, we don't give any guarantees that the experimental APIs won't introduce breaking changes, or they will eventually be promoted to stable.
- How do we release? This suppose to have different release cycles and maybe tags.

go.work
saswatamcode, line 1: Added a workspace file as well

go.work
```diff
@@ -0,0 +1,6 @@
+go 1.21
+
+use (
+	.
```
bwplotka, line 4: Nice, thanks!

bwplotka (approved):
Epic, let's try this how it looks on the main PR 💪🏽 
Thanks!

PR #1712, opened 2025-01-20
Upgrade to prometheus/common 0.62.0 with breaking change

bwplotka:
common 0.62.0 switches NameValidationScheme global variable to UTF8Validation. While relaxing validation this will likely yield test failures downstream.

It's unfortunate, but the alternatives were worse: https://github.com/prometheus/common/pull/724

Good news is that client_golang should work with older common deps and callers can easily change model.NameValidationScheme value to old value to solve any compatibility issues.


kakkoyun (approved):
LGTM.

We need a similar warning for the release notes: https://github.com/prometheus/common/releases/tag/v0.62.0

Issue #1698, opened 2024-12-12
Client does not respect context canceled


ameyaach:
The client does not respect timeouts set on the context. We set a context timeout of 5s and the call blocked for 3 minutes.

Following the code, we realized that the client will always block on the done channel being closed (see [here](https://github.com/prometheus/client_golang/blob/main/api/client.go#L154)). The client waits for all data to be streamed regardless of whether the context is canceled or timed out. Once the context deadline has passed, there is no point in blocking because the underlying function (QueryRange) ultimately returns an error when the context is canceled, so we blocked for an additional 2min 55seconds when we knew that after 5s QueryRange would return an error.

SHA: dcf944d6035ee45bcb9165ec081541da6b5b1ead

bwplotka:
Nice, good point.

It feels we have to make `buf.ReadFrom` go routine context aware:

```
                var buf bytes.Buffer
		_, err = buf.ReadFrom(resp.Body)
		body = buf.Bytes()
		close(done)
```

Help wanted!

PR #1729, opened 2025-02-03
fix: client prompt return on context cancellation

umegbewe:
This fixes the issue where the client blocks for the entire duration of the response stream, even when the provided context had already been canceled or times out

Relevant issue #1698

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"context"
 	"errors"
-	"io"
 	"net"
 	"net/http"
 	"net/url"
--- a/api/client.go
+++ b/api/client.go
@@ -132,36 +132,26 @@ func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response,
 		req = req.WithContext(ctx)
 	}
 	resp, err := c.client.Do(req)
-	defer func() {
-		if resp != nil {
-			_, _ = io.Copy(io.Discard, resp.Body)
-			_ = resp.Body.Close()
-		}
-	}()
-
 	if err != nil {
 		return nil, nil, err
 	}
 
 	var body []byte
-	done := make(chan struct{})
+	done := make(chan error, 1)
 	go func() {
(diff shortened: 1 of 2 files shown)


api/client.go
```diff
@@ -131,34 +131,26 @@ func (c *httpClient) Do(ctx context.Context, req *http.Request) (*http.Response,
...
-		_, err = buf.ReadFrom(resp.Body)
+		_, err := buf.ReadFrom(resp.Body)
 		body = buf.Bytes()
-		close(done)
+		done <- err
 	}()
 
 	select {
 	case <-ctx.Done():
+		resp.Body.Close()
```
bwplotka, line 149: This assumes that close will somehow cancel `_, err := buf.ReadFrom(resp.Body)`. Is it true? How?

Can we have a unit tests that perhaps shows this? 🤔

bwplotka (commented):
Thanks! Some questions, not sure if this will work. My understanding was that we have to manually chunk reads from the resp.Body.

api/client.go
umegbewe: @bwplotka could you take a look 👀

kakkoyun:
@umegbewe The PR need to rebase. 
I have also pinged @bwplotka to have another look at it.

api/client.go
bwplotka: Sorry for lag! 

Yea, looks like this behaviour is documented also in HTTP docs:

```
	// Body must allow Read to be called concurrently with Close.
	// In particular, calling Close should unblock a Read waiting
	// for input. <--- this one.
	Body io.ReadCloser
```

It would be epic to have this test yes! However instead of 500ms delay, I would block it with` <-t.Context.Done()` or something. This will ensure we get `Do` returning without Go test timeout. 

Thanks!

umegbewe: Resolved in new commit

umegbewe:
@bwplotka @kakkoyun updated, PR is good to go

PR #1730, opened 2025-02-06
NewDefaultRegistry function

dnovikoff:
I suggest to add a function for creating a defaulty configured Registry.

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -52,15 +52,10 @@ const (
 // consequences. Users who want to avoid global state altogether should not use
 // the convenience functions and act on custom instances instead.
 var (
-	defaultRegistry              = NewRegistry()
+	defaultRegistry              = NewDefaultRegistry()
 	DefaultRegisterer Registerer = defaultRegistry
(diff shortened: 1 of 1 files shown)


kakkoyun:
Thanks a lot for opening the PR. 

Could you please add some context what is the reason behind the change? What would be the potential implications of it? Any benefits?

I understand using `init` is not a best practice but we also wouldn't want to change things for sake of changing it. client_golang is a dependency for a lot of packages and programs out there.

dnovikoff:
> Could you please add some context what is the reason behind the change? What would be the potential implications of it? Any benefits?

I don't like the idea of using global Registry in my projects. I prefer to explicitly define the object. Still I want to have the ability to create it exactly the same way as the default is created.

> I understand using `init` is not a best practice but we also wouldn't want to change things for sake of changing it. client_golang is a dependency for a lot of packages and programs out there.

I think, that the behaviour seems not to change here for the existing code. Only new function is added. The initialization seems to be better controlled from the constuctor, than from the init function

PR #1738, opened 2025-02-17
Upgrade Golang version v1.24

dongjiang1989:
This library supports the three most recent major releases of Go.  Move golang v1.21


.github/workflows/golangci-lint.yml
```diff
@@ -28,12 +28,12 @@ jobs:
...
+          go-version: 1.24.x
       - name: Install snmp_exporter/generator dependencies
         run: sudo apt-get update && sudo apt-get -y install libsnmp-dev
         if: github.repository == 'prometheus/snmp_exporter'
       - name: Lint
         uses: golangci/golangci-lint-action@ec5d18412c0aeab7936cb16880d708ba2a64e1ae # v6.2.0
         with:
           args: --verbose
-          version: v1.63.4
+          version: v1.64.5
```
dongjiang1989, line 39:
```suggestion
          version: v1.63.4
```

dongjiang1989: Remove to golangci-lint version v1.63.4
Update golangci-lint version by Next PR

prometheus/collectors/go_collector_latest_test.go
```diff
@@ -258,7 +258,7 @@ func TestGoCollectorDenyList(t *testing.T) {
 	}
 }
 
-func ExampleGoCollector() {
+func ExampleNewGoCollector() {
```
kakkoyun, line 261: Why do we have these name changes?

Could you provide some context?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/prometheus/client_golang
 
-go 1.21
+go 1.22
```
kakkoyun, line 3: We don't need to bump this if we don't use any new APIs from 1.22.

We officially support only the last three versions but if we don't use new APIs no need to break compatibly.

kakkoyun (changes requested):
Thanks for the PR. I have some comments.

go.mod
dongjiang1989: Thanks @kakkoyun
I got it
Revert golang version to 1.21

prometheus/collectors/go_collector_latest_test.go
dongjiang1989: Thanks @kakkoyun 
In Golang v1.24 Vet Change:
> The new tests analyzer reports common mistakes in declarations of tests, fuzzers, benchmarks, and examples in test packages, such as malformed names, incorrect signatures, or examples that document non-existent identifiers. Some of these mistakes may cause tests not to run. This analyzer is among the subset of analyzers that are run by go test.
ref: https://tip.golang.org/doc/go1.24#vet

go.mod
SuperQ: Except we do need to change it if we remove 1.21 from the supported version test matrix.

kakkoyun: You’re right—I’m still undecided on this. We could continue testing all versions down to the minimum required API level (1.21) since there’s no downside to doing so.

That said, we could also make the case for upgrading the Go toolchain, given its strong backward compatibility.

Better to stick what we have right now.

kakkoyun (approved):

go.mod
dongjiang1989: > Except we do need to change it if we remove 1.21 from the supported version test matrix.

Thanks @SuperQ 
I got it.

bwplotka (approved):
Thanks!

PR #1750, opened 2025-02-27
exp/api: Add accepted msg type validation to handler

saswatamcode:
This PR adds an accepted message arg to NewHandler, to allow callers to limit what remote write messages their servers can handle.

### DIFF
--- a/exp/api/remote/remote_api.go
+++ b/exp/api/remote/remote_api.go
@@ -379,7 +379,6 @@ type writeStorage interface {
 type handler struct {
 	store                writeStorage
 	acceptedMessageTypes MessageTypes
-	acceptedMap          map[WriteMessageType]struct{}
 	opts                 handlerOpts
 }
 
@@ -470,7 +469,6 @@ func NewHandler(store writeStorage, acceptedMessageTypes MessageTypes, opts ...H
 		opts:                 o,
 		store:                store,
 		acceptedMessageTypes: acceptedMessageTypes,
-		acceptedMap:          acceptedMessageTypes.Map(),
 	}
 
 	// Apply all middlewares in order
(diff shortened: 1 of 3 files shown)


exp/api/remote/remote_api.go
```diff
@@ -455,15 +457,21 @@ func SnappyDecompressorMiddleware(logger *slog.Logger) func(http.Handler) http.H
...
 	for _, opt := range opts {
 		opt(&o)
 	}
-	h := &handler{opts: o, store: store}
+
+	h := &handler{
+		opts:                 o,
+		store:                store,
+		acceptedMessageTypes: acceptedMessageTypes,
+		acceptedMap:          acceptedMessageTypes.Map(),
```
bwplotka, line 473: I would just use slice and do the a quick loop lookup, but not a blocker.

exp/api/remote/remote_headers.go
```diff
@@ -59,24 +59,32 @@ func (n WriteMessageType) Validate() error {
...
 	}
 	return ret
 }
 
-func (m messageTypes) String() string {
+func (m MessageTypes) String() string {
 	return strings.Join(m.Strings(), ", ")
 }
 
+func (m MessageTypes) Map() map[WriteMessageType]struct{} {
```
bwplotka, line 80: Should we do `Contains` instead?

bwplotka (approved):
Thanks, one suggestion only, map might be premature optimizations (for like 1 or 2 elements) (:

exp/api/remote/remote_api.go
saswatamcode: Ah I kinda copied this from prometheus handler. Should be fine to do loop

saswatamcode: Ah I kinda copied this from prometheus handler. Should be fine to do loop

PR #1754, opened 2025-03-01
exp: Add dependabot config

saswatamcode:
Allow dependabot to update modules for exp.

However, dependabot doesn't seem to work with go.work files as of yet https://github.com/dependabot/dependabot-core/issues/6012

Maybe we can remove the go.work stuff for now? 🤔


kakkoyun (approved):

kakkoyun (commented):
I don't even remember why we have the go.work, maybe a remnant from an old feature development. It's ok to remove it.

saswatamcode:
Yup I added this previously whilst introducing exp. It seems like convenient thing to keep all submodules in sync. But dependabot not recognizing it means it'll be a pain. Will remove it in this PR

kakkoyun (approved):
🚀

Issue #1749, opened 2025-02-24
Version 1.21 does not compile for the iOS platform


fat-fellow:
```
# github.com/prometheus/client_golang/prometheus
In file included from process_collector_cgo_darwin.c:18:
/Applications/Xcode-16.2.0.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator18.2.sdk/usr/include/mach/mach_vm.h:1:2: error: mach_vm.h unsupported.
process_collector_cgo_darwin.c:66:17: error: call to undeclared function 'mach_vm_region'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```

bwplotka:
Noted we need to exclude this during build, we might do it for 1.21.1 actually

PR #1758, opened 2025-03-04
Fix ios build for 1.21.1

bwplotka:
Fixes https://github.com/prometheus/client_golang/issues/1749


kakkoyun (approved):
LGTM.

TIL client_golang is used for iOS apps.

PR #1766, opened 2025-03-06
Add `WrapCollectorWith` and `WrapCollectorWithPrefix`

colega:
I found myself in a situation where I had to instantiate multiple instances of a third party library that registered some metrics and never un-registered them.

In my app's lifecycle I needed to un-register them and then register them again, but I couldn't achieve that by wrapping the Registerer.

I decided to register that library's metrics in a separate Registry and register that Registry, but in order to handle multiple instances of the library, I also needed to wrap it with labels, and while such functionality existed already as part of the Registerer wrapping, it wasn't exposed.

This PR just exposes Collector wrapping as a public function.


ArthurSens (approved):
Thanks for the test and example, makes the review process pretty smooth and showcase why we'd ever want something like this :) 

LGTM, but if possible I'd like to see an approval from someone else who has been more active than I've been during the last couple of months 😅. Maybe @bwplotka or @kakkoyun ?

kakkoyun (approved):
LGTM.

Thanks for the detailed tests and examples.

@bwplotka any objections?

kakkoyun:
I'm merging it. We will still have time to have subsequent PRs until the next release if we want to change anything.

PR #1767, opened 2025-03-06
promhttp: vendor the minimal, noasm version of klauspost/compress/zstd.

bwplotka:
This proposes alternative [to PR](https://github.com/prometheus/client_golang/pull/1765#issue-2897705027
) solving [Kubernetes concerns around klauspost/compress dependency](https://github.com/kubernetes/kubernetes/pull/130569#discussion_r1981503174).

I propose vendoring a stripped zstd pkg (with required deps like "le", "fse" and "huff") that has only code needed for FastSpeed mode for encoding (writing). Removed tests, ASM, unsafe code (trivial - klauspost/compress  allows noasm, nounsafe tags).  I propose we do 1.20.x, 1.21.x with this. 1.20.x specifically to unblock Kubernetes ASAP.

NOTE: We still have klauspost/compress in go.mod because we use it in tests (for testing promhttp zstd encoding). All importers will NOT have it as indirect dependency.

Pros:
* We don't break users, we can easily patch 1.20 too.
* It's only tmp; there is a path forward -- zstd should be supported in std lib soon.
* We already have tmp vendored code, so we have precedence.

Cons:
* In case of security vulns. maintainability here will be more difficult.
* Low test coverage for the vendor -- I removed all tests, we only have one test for zstd in promhttp.

cc @liggitt @kakkoyun

### DIFF
--- a/prometheus/promhttp/http_test.go
+++ b/prometheus/promhttp/http_test.go
@@ -26,7 +26,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/klauspost/compress/zstd"
+	orignalzstd "github.com/klauspost/compress/zstd"
 	dto "github.com/prometheus/client_model/go"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -90,7 +90,7 @@ func readCompressedBody(r io.Reader, comp Compression) (string, error) {
 		got, err := io.ReadAll(reader)
 		return string(got), err
 	case Zstd:
-		reader, err := zstd.NewReader(r)
+		reader, err := orignalzstd.NewReader(r)
 		if err != nil {
 			return "", err
 		}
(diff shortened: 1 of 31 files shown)


bwplotka:
Tests are dead, because Dagger is still on this branch. If we need to backport this, we need to remove dagger too, will do later...

prometheus/promhttp/http_test.go
```diff
@@ -26,7 +26,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/klauspost/compress/zstd"
+	orignalzstd "github.com/klauspost/compress/zstd"
```
kakkoyun, line 29: Why don't we use the vendored package here as well? 🤔

bwplotka: Great question! I mentioned in the description, but to reiterate:

1. To reduce the vendored code by at least 50% (we only need encoding). I am pulling the original dep for test purposes only to decode for validation.
2. Extra benefit is the validation against original, stable decoder so we know we did not mess something in our vendored code that makes the result encodable only for us :p

bwplotka:
Seems like we are leaning towards https://github.com/prometheus/client_golang/pull/1765#pullrequestreview-2665713667 - closing this for now.

PR #1773, opened 2025-03-11
promhttp: Ignore informational response status codes when handling response

### LINKED ISSUE #1772: `responseWriterDelegator` incorrectly handles `100 Continue` response
### Description

### Steps To Reproduce

1. Use following handler:
    ```go
    		
    		body, _ := io.ReadAll(r.Body)
    		w.Write(body)
    	})
    ```
1. Run the code.

### Expected behavior

VirrageS:
Closes #1772

### DIFF
--- a/prometheus/promhttp/delegator.go
+++ b/prometheus/promhttp/delegator.go
@@ -53,12 +53,21 @@ func (r *responseWriterDelegator) Written() int64 {
 }
 
 func (r *responseWriterDelegator) WriteHeader(code int) {
-	if r.observeWriteHeader != nil && !r.wroteHeader {
-		// Only call observeWriteHeader for the 1st time. It's a bug if
-		// WriteHeader is called more than once, but we want to protect
-		// against it here. Note that we still delegate the WriteHeader
-		// to the original ResponseWriter to not mask the bug from it.
-		r.observeWriteHeader(code)
+	if !r.wroteHeader {
+		// Ignore informational response status codes.
+		// Based on https://github.com/golang/go/blob/go1.24.1/src/net/http/server.go#L1216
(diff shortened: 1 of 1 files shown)


prometheus/promhttp/delegator.go
```diff
@@ -53,12 +53,21 @@ func (r *responseWriterDelegator) Written() int64 {
...
-	if r.observeWriteHeader != nil && !r.wroteHeader {
-		// Only call observeWriteHeader for the 1st time. It's a bug if
-		// WriteHeader is called more than once, but we want to protect
-		// against it here. Note that we still delegate the WriteHeader
-		// to the original ResponseWriter to not mask the bug from it.
-		r.observeWriteHeader(code)
+	if !r.wroteHeader {
+		// Ignore informational response status codes.
+		// Based on https://github.com/golang/go/blob/go1.24.1/src/net/http/server.go#L1216
+		if code >= 100 && code <= 199 && code != http.StatusSwitchingProtocols {
```
bwplotka, line 59: What about something like:

```suggestion
		if code / 100 == 1 && code != http.StatusSwitchingProtocols {
```

bwplotka (commented):
Amazing! Any chances for a small unit test to ensure this works correctly long term? 🤗

Otherwise LGTM!

kakkoyun:
@VirrageS Do you want to take care of the comments? We will have a release soon and we can include this.

prometheus/promhttp/delegator.go
VirrageS: As mentioned, this code is based on the `server.go` implementation and I wanted to preserve the whole syntax so it is clear.

I can change that but I don't see a reason to do that since it is clear what this code does now. And `code / 100 == 1` doesn't make it faster or more clear.

PR #1787, opened 2025-04-02
Add the GOOS to CI test

### LINKED ISSUE #1761: Add CI tests that builds/tests client_golang for all supported GOOS
As per title, we had too often cases of a broken build for a less popular operating systems e.g. ios recently https://github.com/prometheus/client_golang/issues/1749

The risky code is for the prometheus/process_collector_* that has to use C API for darwin to obtain memory OS metrics.

khanhtc1202:
From the KubeCon ContribFest 👋 

This PR added GOOS as variable to the go test CI

Fix #1761

Please check @ArthurSens


kakkoyun (changes requested):
LGTM. Thanks for the contribution.

However CI fails, so it needs some more work.

kakkoyun:
When we are at it maybe we should add GOARCH as well. Now that GitHub actions have the arm64 machines.

PR #1792, opened 2025-04-07
fix: do not use new metric instances every Go memstats collection

alespour:
This PR changes how Go memstats are collected. Instead of creating new `ConstMetric` during every collection, it (re)uses instances of `Gauge` or `Counter` in order to reduce needless memory allocations.


kakkoyun (changes requested):
This is looking good.

Could you please add some benchmarks? So that we can actually see if we have tangible improvements?

prometheus/go_collector.go
```diff
@@ -261,14 +157,28 @@ func (c *baseGoCollector) Collect(ch chan<- Metric) {
...
-	return "go_memstats_" + s
+// memStatsMetric provide description, evaluator, runtime/metrics name, and
+// value type for memstat metric.
+type memStatsMetric struct {
+	last   float64
+	eval   func(*runtime.MemStats) float64
+	metric Metric
+}
+
+func (m *memStatsMetric) update(memStats *runtime.MemStats) Metric {
```
beorn7, line 168: This is not concurrency-safe. Metrics collection can happen concurrently, and then all kind of weirdness will happen here.

beorn7:
Updating the metrics primitives that are meant for direct instrumentation just before a scrape is an anti-pattern. (The "reverse engineering" needed to calculate the counter increment is a good sign that you are working against the intention of the API here.)

Scrapes can happen at any time and in particular concurrently. This PR introduces races.

If you try to work around the concurrency problem by using mutexes, you might easily have a slower metrics collection than before this change.

I also doubt that the allocation savings here really make a dent in the big picture. Encoding the metrics later to be sent out via the network is probably an order of magnitude more effort anyway.

PR #1797, opened 2025-04-13
test(registry): Add goleak-based goroutine leak detection

surinkim:
test(registry): Add goleak-based goroutine leak detection test for Registry.Gather

This commit adds a new test case `TestGatherDoesNotLeakGoroutines` which verifies that repeated calls to `Registry.Gather()` do not leak goroutines.

- 100 collectors are registered to a custom registry
- `Gather()` is called 1000 times
- `goleak.VerifyNone` ensures that no unexpected goroutines remain

This test serves as a future safeguard to detect goroutine leaks under heavy usage.


bwplotka (approved):
Generally LGTM -- was there any historical leak that trigger this PR? 

One thing to ensure -- does this upset client_golang importers in terms of extra dependency. I assume NO given we only import this in test, importers should not see it (in theory).

bwplotka:
Thanks!

PR #1822, opened 2025-06-24
feat(prometheus): explicit label/metric name validation scheme

juliusmh:


prometheus/desc.go
```diff
@@ -95,7 +95,7 @@ func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, const
 		help:           help,
 		variableLabels: variableLabels.compile(),
 	}
-	if !model.IsValidMetricName(model.LabelValue(fqName)) {
+	if !model.IsValidMetricName(model.LabelValue(fqName), model.UTF8Validation) {
```
aknuds1: Why are you hard coding `model.UTF8Validation` here?

aknuds1 (commented):
Why are you hard coding `model.UTF8Validation` in certain places?

prometheus/desc.go
juliusmh: `NewDesc` is controlled by us, e.g. we never call it with a customer's metric name. Changing the function signature touches 94 places in Mimir.

aknuds1: @juliusmh `client_golang` isn't just used by Mimir though. Also, hard coding the name validation scheme here is a backwards incompatible change, albeit _implicit_. I personally think that either the validation scheme should be an explicit argument, or alternatively `NewDesc` could take variadic options and default to `model.UTF8Validation` (this would also be backwards incompatible, but at least we avoid hard coding). @ywwg What's your take?

juliusmh: I added an optional `WithValidationScheme` to this method.

PR #1825, opened 2025-06-30
feat: support arbitrary query params in request

alecmerdler:
Adds `WithCustomOption()` func which allows
setting arbitrary key/value pairs to the URL
query parameters of the request.

The primary use case for this is to work with
https://github.com/prometheus-community/prom-label-proxy.


alecmerdler:
This is a naive approach, I will be happy to update the pull request with any feedback. But as it stands today, the official Prometheus client cannot be used with `prom-label-proxy`.

kakkoyun:
@alecmerdler Let's open an issue and discuss this further before jumping into any conclusions.

I would love to learn more about your use cases, especially.

bwplotka:
Thanks!

> But as it stands today, the official Prometheus client cannot be used with prom-label-proxy.

Are we sure? Would custom roundtripper be possible for those use cases? We could add example how to write one - should be trivial to write one that appends parameters to the query for all requests. Similar to https://github.com/prometheus/client_golang/blob/93a67cb28444c1d068b891c2c642d371d61f19b7/api/prometheus/v1/example_test.go#L90

Perhaps hacky, but for advanced use cases should be enough.

Otherwise adding WithCustomOption would mean:
* There are many ways of doing the same thing (WithStats vs WithCustomOptions("stats", ...) vs roundtripper )
* In the next PR someone proposes WithCustomOptions to every other method.. which is not simple (Option has other stuff).

Generally as we repeat on every API contribution.. it's NOT a perfect API interface. We learned a lot. It's time for a new one, ideally auto-generated (https://github.com/prometheus/prometheus/issues/2392#issuecomment-2690211999). Given that, I would say roundtripper should suffice, WDYT?

alecmerdler:
@bwplotka Thanks for your suggestion to use a custom `RoundTripper`, I believe that will solve my use case. I'm going to go ahead and close this.

PR #1835, opened 2025-07-14
Fix `errNotImplemented` reference

aknuds1:
Replace `notImplementedErr` reference with `errNotImplemented` in prometheus/process_collector_mem_nocgo_darwin.go, since the former doesn't exist any longer.


kakkoyun (approved):
Thanks for quick action.

PR #1841, opened 2025-07-15
retract v1.23.0-rc.0

vesari:
A mistake was made in making v1.23.0-rc.0, retracting as agreed with @kakkoyun

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,3 +31,5 @@ require (
 )
 
 exclude github.com/prometheus/client_golang v1.12.1
+
+retract v1.23.0-rc.0


SuperQ:
Do we need to retract an RC?

vesari:
> Do we need to retract an RC?

Unfortunately yes, because I made mistake and recreated the tag, sorry.

kakkoyun (approved):

SuperQ:
Using `go get` won't fetch RCs, you would have to explicitly fetch it. I'm not sure this is necessary.

kakkoyun:
Good call. Thanks @SuperQ

PR #1843, opened 2025-07-28
Support matchers in rules API

jotak:
Similar to the older change on Labels API (https://github.com/prometheus/client_golang/pull/828), this adds `matches` to the `Rules` API.

This is a breaking change. If sending breaking changes is an issue, we could create another function signature - but that would be less consistent with the existing code base.

**breaking change migration:**
Compilation will fail for client callers who use `api.Rules(ctx)`.
Simply change it to `api.Rules(ctx, nil)` to keep the same behaviour.

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -494,7 +494,7 @@ type API interface {
 	// under the TSDB's data directory and returns the directory as response.
 	Snapshot(ctx context.Context, skipHead bool) (SnapshotResult, error)
 	// Rules returns a list of alerting and recording rules that are currently loaded.
-	Rules(ctx context.Context) (RulesResult, error)
+	Rules(ctx context.Context, matches []string) (RulesResult, error)
 	// Targets returns an overview of the current state of the Prometheus target discovery.
 	Targets(ctx context.Context) (TargetsResult, error)
 	// TargetsMetadata returns metadata about metrics currently scraped by the target.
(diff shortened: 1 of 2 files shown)


api/prometheus/v1/api.go
```diff
@@ -494,7 +494,7 @@ type API interface {
 	// under the TSDB's data directory and returns the directory as response.
 	Snapshot(ctx context.Context, skipHead bool) (SnapshotResult, error)
 	// Rules returns a list of alerting and recording rules that are currently loaded.
-	Rules(ctx context.Context) (RulesResult, error)
+	Rules(ctx context.Context, matches []string) (RulesResult, error)
```
kakkoyun, line 497: Not a blocker, how about making `matches` a variadic argument? It would make it neater

jotak: I can do it if desired, but I also see some cons:
- it's less consistent with other API functions that also use `matches []string`, such as `LabelNames`
- if the signature has to change again later to add more arguments, depending on what those new arguments are, variadic may have to be reconsidered (e.g. if another `[]string`-like argument is added)

I don't really have an opinion here, let me know what's best

jotak: Also, I only proposed to add `matches` because it's the use-case I'm interested in, but there are a bunch of other arguments that could be relevant, [according to the doc](https://prometheus.io/docs/prometheus/latest/querying/api/#rules). I guess I'm trying to avoid a bigger refactoring, if that makes sense. But a more future-proof change would be to accept a parameters struct.

kakkoyun: Yes, never mind. I was lazy, and I hadn't checked the other parameters.
Even though we consider the API package as experimental, let's stick to the existing conventions.

We can add `Option`s if needed.

When it comes to making it future-proof, maybe we can have the `type` as a first-order parameter as well and make the rest optional.

Any thoughts? cc @bwplotka

bwplotka: I would vote for sticking to consistency and breaking ppl here, as we historically did on this experimental package. We should likely deprecate and move to exp or api module to make it clearer that it's experimental... (or make OpenAPI/swagger flow possible finally)

I would say LGTM for now

PR #1850, opened 2025-07-31
Refactor LabelNames to return model.LabelNames type for consistency

yshngg:


bwplotka (approved):
This is breaking change, but API is still experimental, so LGTM, thanks!

PR #1846, opened 2025-07-30
feat(api/prometheus): add format_query endpoint for query formatting

yshngg:
The changes introduce a new `/format_query` API endpoint to handle query formatting in the Prometheus API.

This includes implementing the `FormatQuery` function in `api.go` and adding corresponding test cases in `api_test.go` to ensure proper functionality.

Fix #1845

PR #1856, opened 2025-08-20
feat(api): add FormatQuery method to Prometheus v1 API

yshngg:
As a follow-up to PR #1846 .


yshngg:
@bwplotka PTAL

bwplotka:
Thanks!

PR #1858, opened 2025-08-23
Refactor *Vec types

cl-bvl:
Hello.
While working on https://github.com/prometheus/client_golang/pull/1854 I ran into a few typing issues that got in the way:

1. Most types like `Counter`, `Gauge`, `Histogram`, and `Summary` have both an interface and a concrete private implementation (`counter`, `gauge`, etc.) that implements it. But for the vector types (`CounterVec`, `GaugeVec`) there’s no interface — they’re implemented directly. I introduced interfaces for these vector types and moved the implementations into private types (`counterVec`, `gaugeVec`).

2. The `ObserverVec` interface for `HistogramVec` and `SummaryVec` implements all its methods and also satisfies `Collector`. However, for `Histogram` and `Summary`, the `Observer` interface only defines `Observe(float64)`. This prevents creating a generic interface for `Histogram` and `Summary` in places that require a `Collector` implementation. I added the missing methods to `Observer` and introduced a new `ObserverMetod` interface.

3. Added the missing methods to `ObserverVec`.

4. All *Vec types use `MetricVec` under the hood. I also turned `MetricVec` into an interface.

5. I tried to maintain maximum backward compatibility to minimize the likelihood of changes in existing projects.
In most cases, everything will work without changes or with minimal changes (for example, replacing type from pointer to the interface -- `*CounerVec` => `CounerVec`)

### DIFF
--- a/prometheus/benchmark_test.go
+++ b/prometheus/benchmark_test.go
@@ -19,42 +19,42 @@ import (
 )
 
 func BenchmarkCounter(b *testing.B) {
-	type fns []func(*CounterVec) Counter
+	type fns []func(CounterVec) Counter
 
 	twoConstraint := func(_ string) string {
 		return "two"
 	}
 
-	deLV := func(m *CounterVec) Counter {
+	deLV := func(m CounterVec) Counter {
 		return m.WithLabelValues("eins", "zwei", "drei")
 	}
-	frLV := func(m *CounterVec) Counter {
+	frLV := func(m CounterVec) Counter {
 		return m.WithLabelValues("une", "deux", "trois")
 	}
-	nlLV := func(m *CounterVec) Counter {
+	nlLV := func(m CounterVec) Counter {
 		return m.WithLabelValues("een", "twee", "drie")
 	}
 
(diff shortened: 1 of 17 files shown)


bwplotka (changes requested):
Thanks!

This is great in general, but unfortunately, given a high adoption of this library, we need to embrace an extremely low tolerance bar for breaking changes. Those minimal changes, multiplied by [hundreds of thousands](https://github.com/prometheus/client_golang/network/dependents) importers mean combined effort of likely a multiple SWEyears without notice (next automatic pull of the release of client_golang). There are things we could do without breaking changed, but even replacing struct with interfaces etc is hard to accept now.

However, we were always thinking about v2, you can find many [issues/ideas that require v2](https://github.com/prometheus/client_golang/issues?q=is%3Aissue%20state%3Aopen) on our board, so perhaps it's time to start something together on a branch of directory? With generics now, things might be bit easier. V2 will be a long process to get, and it will need to be 10x easier to use to get ppl on it... but we need to start it one day (:

PR #1860, opened 2025-08-25
WIP: collectors/version: Allow custom labels

mrueg:
We have a use case here, where it would be nice to support custom labels in the build info: https://github.com/kubernetes/kube-state-metrics/pull/2739


prometheus/collectors/version/version.go
```diff
@@ -24,6 +25,23 @@ import (
 // NewCollector returns a collector that exports metrics about current version
 // information.
 func NewCollector(program string) prometheus.Collector {
+	return NewCollectorWithLabels(program, nil)
+}
+
+// NewCollectorWithLabels returns a collector that exports metrics about current
+// version information and allows to set additional custom labels.
+func NewCollectorWithLabels(program string, labels prometheus.Labels) prometheus.Collector {
```
bwplotka: Wanna spend extra 10m and add proper options to this type? Will help in future changes (e.g. filtering some labels etc)

```
type Option func(*options)

type options struct {
    extraConstLabels prometheus.Labels
}

func WithExtraConstLabels(l prometheus.Labels) Option {
  return func(o *options) {
     o.extraConstLabels = l
  }
}


func NewCollector(program string, opts ...Option) prometheus.Collector {
   o := option{}
   for _, opt := range opts {
       opt(&o)
   }
//...
```

bwplotka (commented):
Thanks! Makes sense, just one suggestion to make it flexible for future options, WDYT?

prometheus/collectors/version/version.go
mrueg: Do you want me to rename the func to NewCollectorWithOptions then or should I add the opts arg to NewCollector?

mrueg: Ah, understood, it's a variadic function now.

mrueg:
@bwplotka Thanks for the suggestion, I applied you change.

kakkoyun (approved):

PR #1862, opened 2025-08-25
chore: Drop support for <go1.22

mrueg:
Drops a bunch of files for outdated go versions that are not supported anymore.


kakkoyun:
> The idea was to keep a "soft" support for older versions, until we actually break things in somewhere common for older version.
> 
> That being said I would be happy to drop anything older than 1.22

What do you think? @bwplotka @ArthurSens @vesari

mrueg:
Works for me, I readded the go 1.22 bits.

kakkoyun:
> Thanks!
> 
> My vote would be to drop everything that is not explicitly build in our CI, otherwise we **don't know** if what we maintain is really usable 🙈 (so why maintaining that then). Not a super strong opinion, not a biggie to keep 1.22 -- thanks!

Fair! We can do this to make it straight-forward.

PR #1863, opened 2025-08-29
promhttp: implement WithXFromContext in terms of WithXFromRequest

tie:
This change adds WithLabelFromRequest and WithExemplarFromRequest options and updates FromContext counterparts to be a convenience wrappers for these options.

For example, without this change, setting a label based on http.Request.Pattern requires some juggling with context:

	var ctxHTTPRequestKey = httpRequestContext{}

	type httpRequestContext struct {
		*http.Request
	}

	func httpPatternFromContext(ctx context.Context) string {
		r := ctx.Value(ctxHTTPRequestKey).(*httpRequestContext)
		return r.Pattern
	}

	func instrumentHTTPHandler(h http.Handler) http.Handler {
		h = promhttp.InstrumentHandlerCounter(httpRequestsTotal, h,
			promhttp.WithLabelFromCtx("handler", httpPatternFromContext),
		)
		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			var c httpRequestContext
			ctx := context.WithValue(r.Context(), ctxHTTPRequestKey, &c)
			r = r.WithContext(ctx)
			c = httpRequestContext{r}
			h.ServeHTTP(w, r)
		})
	}

promhttp.WithLabelFromRequest allows to access http.Request directly:

	func instrumentHTTPHandler(h http.Handler) http.Handler {
		return promhttp.InstrumentHandlerCounter(httpRequestsTotal, h,
			promhttp.WithLabelFromRequest("handler", func(r *http.Request) string {
				return r.Pattern
			}),
		)
	}


tie:
@bwplotka can you take a look at this please?

yves-tutti:
Thank you for preparing this PR, it would be really nice if it can be merged.
I see no problem with the changes to accept the *http.Request other than that it theoretically allows to do really nasty manipulations to the request data-structure with side-effects that are completely out of scope of monitoring. While ctx access only is too limiting, it also prevent some chances to shoot ourself in the foot.

As example where it's required to have access to http.Request is to access Pattern in the standard library:
https://github.com/golang/go/blob/master/src/net/http/request.go#L327
or in gorilla mux:
https://github.com/gorilla/mux/blob/db9d1d0073d27a0a2d9a8c1bc52aa0af4374d265/mux.go#L477

kakkoyun:
> Thank you for preparing this PR, it would be really nice if it can be merged. I see no problem with the changes to accept the *http.Request other than that it theoretically allows to do really nasty manipulations to the request data-structure with side-effects that are completely out of scope of monitoring. While ctx access only is too limiting, it also prevent some chances to shoot ourself in the foot.
> 
> As example where it's required to have access to http.Request is to access Pattern in the standard library: [golang/go@`master`/src/net/http/request.go#L327](https://github.com/golang/go/blob/master/src/net/http/request.go?rgh-link-date=2026-04-21T12%3A10%3A24Z#L327) or in gorilla mux: [gorilla/mux@`db9d1d0`/mux.go#L477](https://github.com/gorilla/mux/blob/db9d1d0073d27a0a2d9a8c1bc52aa0af4374d265/mux.go#L477)

This pattern is what also default handler uses https://pkg.go.dev/net/http#Handler so pointer passing shouldn't be a problem. The same foot gun already exists everywhere in Go.

kakkoyun (approved):
LGTM!

PR #1884, opened 2025-10-03
Public ParseWriteResponseStats method and Confirmed to reuse in Prometheus

pipiland2612:
#### Notes
- This PR is to supports https://github.com/prometheus/prometheus/issues/17192

#### Changes
- Public ParseWriteResponseStats to reuse in https://github.com/prometheus/prometheus/blob/f4b8840f51bee8afa258e0d4e27fc7fc03358760/storage/remote/client.go#L307
- Public Confirmed to reuse in https://github.com/prometheus/prometheus/blob/f4b8840f51bee8afa258e0d4e27fc7fc03358760/storage/remote/queue_manager.go#L1817

### DIFF
--- a/exp/api/remote/remote_api.go
+++ b/exp/api/remote/remote_api.go
@@ -241,7 +241,7 @@ func (r *API) Write(ctx context.Context, msgType WriteMessageType, msg any) (_ W
 		if err == nil {
 			// Check the case mentioned in PRW 2.0.
 			// https://prometheus.io/docs/specs/remote_write_spec_2_0/#required-written-response-headers.
-			if msgType == WriteV2MessageType && !accumulatedStats.confirmed && accumulatedStats.NoDataWritten() {
+			if msgType == WriteV2MessageType && !accumulatedStats.Confirmed && accumulatedStats.NoDataWritten() {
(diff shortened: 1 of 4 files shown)


pipiland2612:
cc @bwplotka

bwplotka:
Thanks. I checked and those things are only used because we didn't fully switch to the remote API from this library.

If we would switch sending parts in Prometheus as well, we would not need to make those methods public. This means that once we switch to Prometheus this change would not be needed and as a result we would need to either leave an unclean API (leaving things that should not be used outside of the library) or revert, right? Is that really a correct approach?  Or should we either, on Prometheus code:

1. Not remove  write response structure, so it can be used on sender (for now, we will remove later)
2. Move fully in one atomic PR to this lib including sending (some might say it's risky, but maybe it's fine)

I'd vote for (1)

WDYT?  @pipiland2612 @saswatamcode

bwplotka (changes requested):
See comment above, let's see if we could avoid this.

saswatamcode:
Yup, I would vote for 1 to do it in pieces.

pipiland2612:
Hi @bwplotka, thanks for suggesting. The choice (1) seems reasonable to me.

PR #1888, opened 2025-10-07
Add RetryCallBack to remote_api.go 

pipiland2612:
#### Changes
- Add RetryCallBack to remote_api.go to support for providing observability while Write function are being called
- This change is needed because we want to use it in storage/remote/queue_manager.go (prometheus code) see: https://github.com/prometheus/prometheus/blob/a4da440dad4be13da121cce79caa74b1c76ec9f5/storage/remote/queue_manager.go#L1801

### DIFF
--- a/exp/api/remote/remote_api.go
+++ b/exp/api/remote/remote_api.go
@@ -48,6 +48,9 @@ type API struct {
 // APIOption represents a remote API option.
 type APIOption func(o *apiOpts) error
 
+// RetryCallback is called each time Write() retries a request.
+type RetryCallback func()
+
 // TODO(bwplotka): Add "too old sample" handling one day.
 type apiOpts struct {
 	logger           *slog.Logger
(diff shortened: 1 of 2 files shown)


pipiland2612:
cc @bwplotka @saswatamcode

exp/api/remote/remote_api.go
```diff
@@ -267,6 +280,12 @@ func (r *API) Write(ctx context.Context, msgType WriteMessageType, msg any) (_ W
 		}
 
 		backoffDelay := b.NextDelay() + retryableErr.RetryAfter()
+
+		// Invoke retry callback if provided (after NextDelay which increments the retry counter).
+		if r.opts.retryCallback != nil {
```
bwplotka, line 284: If we pass `err` we can remove my TODOs from line 273 and 278 perhaps - and it will be more useful for users (e.g. why retrying). WDYT?

pipiland2612: Hi @bwplotka  Thanks for suggesting. it makes sense to do this: add error to the params gives more information to the user and thus improve their experience. Solved at https://github.com/prometheus/client_golang/pull/1888/commits/f5abd52c959088d6b31e4033be99ae320740f154

bwplotka (approved):

PR #1889, opened 2025-10-07
bug(remote_write): Fix compression buffer pooling

fpetkovski:
The compressed buffer is released before the request is made, which leads to corrupted remote-write payloads being sent to the target.

The compression function uses the pattern where the destination slice can be passed in, and the function returns a reference to it, adjusted for length.


saswatamcode (approved):
Thanks for catching!

bwplotka:
Thanks!

bwplotka:
Although.. some unit test would be nice so we don't repeat this mistake (:

PR #1890, opened 2025-10-07
Change RetryCallBack initialized

pipiland2612:
#### Changes
- Change RetryCallBack to pass in as a params for Write function instead of from API option to be more flexible.


pipiland2612:
cc @bwplotka, after a while drafting some changes with queue_manager.go, I decided to rewrite a bit :)

PR #1891, opened 2025-10-08
Implement filter for remote write 

pipiland2612:
#### Changes
- Add MessageFilter to remote_api.go to support for filtering message while call Write

PR #1892, opened 2025-10-09
Fix CI bug

pipiland2612:
The issues is https://github.com/prometheus/client_golang/pull/1891#issuecomment-3381275277 and the PR related to this bugs https://github.com/prometheus/client_golang/pull/1864


bwplotka:
Do we know why this happened? We have a routine that generates those metrics. Why we suddenly missed one metric and literally in 1.24 as well?

pipiland2612:
Hmm, let me take a closer look why it is the case here.

bwplotka:
ping (:

bwplotka:
Also can you provide more meaningful PR title (and commit msg)

bwplotka:
Ok digged deeper. The reason is that Go team added this metric in minor version of 1.25.2 and [1.24.8](https://github.com/golang/go/commit/c6b04dd33b0215f5deb83724661921842bf67607). Fair (:

PR #1893, opened 2025-10-10
Public backoff directory

pipiland2612:
Expose backoff directory so that we could use the function https://github.com/prometheus/client_golang/blob/c5900a16e995dd13df14cbe89a543dbb41e9f43b/exp/api/remote/remote_api.go#L111 without the error: "Use of the internal package is not allowed"


pipiland2612:
Thanks this changes is not ideal with our intention to make less dependency and divergence from our purpose of library for remote write. I will make another change soon!

PR #1894, opened 2025-10-17
Clarify exp library stability

pipiland2612:
Related to: https://github.com/prometheus/prometheus/pull/17197#issuecomment-3376254657 and https://github.com/prometheus/prometheus/issues/17192

### DIFF
--- a/exp/api/remote/remote_api.go
+++ b/exp/api/remote/remote_api.go
@@ -35,6 +35,13 @@ import (
 	"github.com/prometheus/client_golang/exp/internal/github.com/efficientgo/core/backoff"
 )
 
+// BackoffConfig configures exponential backoff with jitter for retry operations.
+type BackoffConfig struct {
+	Min        time.Duration `yaml:"min_period"`  // Start backoff at this level
+	Max        time.Duration `yaml:"max_period"`  // Increase exponentially to this level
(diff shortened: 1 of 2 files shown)


exp/README.md
bwplotka, line 8:
```suggestion

This module contains production quality code with explicitly unstable API. Any code in this module can change its API or be removed; use with care.

The intention is that, with maturity some of the packages would graduate to stable version of client_golang module.

Packages within this module are listed below.

## Remote

NOTE: The `api/remote` package is used by `prometheus/prometheus`. Any changes to this package should ensure Prometheus is not affected (e.g. does not break Prometheus on upgrade or if it breaks, the Prometheus gets updated soon.
```

pipiland2612:
Hi @bwplotka, thanks for suggesting. PTAL again when you are free, thanks!

exp/README.md
```diff
@@ -1,12 +1,21 @@
 # client_golang experimental module
 
 Contains experimental utilities and APIs for Prometheus.
-The module may be contain breaking changes or be removed in the future.
+This code is used by released versions of Prometheus, but should not be relied upon outside the Prometheus project.
```
bwplotka: Something went wrong? I think this suppose to be removed? (everything from top header to first `## remote`

pipiland2612: Ok I have removed it

PR #1895, opened 2025-10-17
Public backoff config to allow usage of WithAPIBackoff

pipiland2612:
#### What Issue is this fixing ? 
- Expose backoff directory so that we could use the function ` func WithAPIBackoff(cfg BackoffConfig) ` without the error: "Use of the internal package is not allowed"

#### Changes
- Create a BackoffConfig public struct that serve as a public layer to internal backoff.Config

### DIFF
--- a/exp/api/remote/remote_api.go
+++ b/exp/api/remote/remote_api.go
@@ -114,8 +114,9 @@ func WithAPINoRetryOnRateLimit() APIOption {
 	}
 }
 
-// WithAPIBackoffConfig returns APIOption that allows overriding backoff configuration.
-func WithAPIBackoffConfig(cfg BackoffConfig) APIOption {
+// WithAPIBackoff returns APIOption that allows configuring backoff.
+// By default, exponential backoff with jitter is used (see defaultAPIOpts).
+func WithAPIBackoff(cfg BackoffConfig) APIOption {
 	return func(o *apiOpts) error {
 		o.backoffConfig = cfg
 		return nil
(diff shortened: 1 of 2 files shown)


pipiland2612:
Another option is to define an interface that allows passing in the WithBackoff method. However, this approach feels less clean and adds unnecessary complexity for the client, since our goal is only to modify the configuration—not the underlying implementation.

bwplotka:
Can you update title and description (and commit message)? It's out of date (description) and title is not human readable.

exp/api/remote/remote_api.go
```diff
@@ -107,10 +114,10 @@ func WithAPINoRetryOnRateLimit() APIOption {
 	}
 }
 
-// WithAPIBackoff returns APIOption that allows overriding backoff configuration.
-func WithAPIBackoff(backoff backoff.Config) APIOption {
+// WithAPIBackoffConfig returns APIOption that allows overriding backoff configuration.
+func WithAPIBackoffConfig(cfg BackoffConfig) APIOption {
```
bwplotka:
```suggestion
// WithAPIBackoff returns APIOption that allows configuring backoff.
// By default, X backoff is used.
func WithAPIBackoff(cfg BackoffConfig) APIOption {
```

bwplotka: And let's show how we can mention X (is there a default)?

bwplotka: The rationale for NOT having Config for this option (but still have config suffix on BackoffConfig type) is simplicity - it's clear we are "configuring" backoff here, so just backoff  in this context is enough?

pipiland2612: Hi @bwplotka thanks for suggesting. Solved at https://github.com/prometheus/client_golang/pull/1895/commits/68ea185eaeae89ed085cd2f11a886c3b22b7e01b

PR #1896, opened 2025-10-20
feat: add support for `/status/tsdb/blocks` endpoint

tjhop:
Adds a new method to the API interface to support the new TSDB blocks
endpoint that was added in prometheus 3.6.0:

https://github.com/prometheus/prometheus/releases/tag/v3.6.0
prometheus/prometheus#16695

Signed-off-by: TJ Hoplock <t.hoplock@gmail.com>


bwplotka (approved):
It's great, thank you!

PR #1898, opened 2025-10-24
chore(ci): Add CRLF detection and fix targets to prevent CRLF contamination

kakkoyun:
Add two new Makefile targets to detect and fix carriage return (CRLF)
line endings that can contaminate the codebase from Windows environments:

- check-crlf: Scans all text files (excluding hidden dirs and vendor)
  for CRLF line endings and fails if any are found
- fix-crlf: Automatically converts CRLF to LF using the tr command

Add CRLF check to CI workflow to catch line ending issues early in PRs.
The check runs immediately after checkout, before any other validation.

Also rename go.yml to validate.yml to better reflect the workflow's
purpose, and fix job name casing (supportedVersions -> supported_versions).

Signed-off-by: Kemal Akkoyun <kemal.akkoyun@datadoghq.com>

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -36,12 +36,12 @@ jobs:
   test:
     name: Tests (${{ matrix.go_version }})
     runs-on: ubuntu-latest
-    needs: supportedVersions
+    needs: supported_versions
     # Set fail-fast to false to ensure all Go versions are tested regardless of failures
     strategy:
       fail-fast: false
       matrix:
-        go_version: ${{ fromJSON(needs.supportedVersions.outputs.supported_versions) }}
+        go_version: ${{ fromJSON(needs.supported_versions.outputs.supported_versions) }}
     # Define concurrency at the job level for matrix jobs
     concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/validate.yml
```diff
@@ -1,68 +1,71 @@
...
-jobs:
-  supportedVersions:
-    name: Fetch supported Go versions
-    runs-on: ubuntu-latest
-    outputs:
-        supported_versions: ${{ steps.matrix.outputs.supported_versions }}
-    steps:
-      - name: Checkout code
+---
+name: Validate
```
bwplotka, line 2: Can we keep Go name (also in file)? Otherwise we need to change GH CI required checks and generally change the experience of contributing to client_golang. Is it worth the name change?

kakkoyun: The "Go" is not in the required checks. It is just the "Tests *". I will make sure we don't break any required checks.

kakkoyun: FWIW, it is step name based

PR #1912, opened 2025-11-03
chore: Comment auth examples to reduce common pkg deps

JorTurFer:
As we have discussed [via slack](https://cloud-native.slack.com/archives/C01AUBA4PFE/p1762169496875849), to reduce `client_golang` dependencies from `common` pkg and its deps just used for examples, the places where they are used are commented with a note explaining how to run the examples.

This reduces the direct and indirect dependencies of `client_golang`

Related to https://github.com/prometheus/common/pull/862


bwplotka (changes requested):
Thanks! 

I am afraid this switch to non-buildable documentation while reusing exemple for the purpose of allowing one less "indirect testing dep" kind of defies the point of using example framework at this point.

We have a few options:
a) Leave the import on this example (close this PR) and live with new "testing indirect dep". I double checked with @dashpole and our tests confirms - we don't see a leak on importer side.
b) Move examples to commentary (e.g. doc.go, hard to ensure they are up to date.)
c) Move examples to different module (hard to discover)

Understanding the whole problem space, I think I'm ok with "a" =D

WDYT?

JorTurFer:
Let me try to solve it moving that example to other folder. Maybe if I move that file to the current examples folder, I can create a submodule there.

JorTurFer:
I close this as we discussed via slack

PR #1917, opened 2025-11-20
Prevent OOM from malformed snappy payloads by validating decoded length

makasim:
A specially crafted remote-write request can declare an extremely large decoded length while providing only a small encoded payload. Prometheus allocates memory based on the declared decoded size, so a single request can trigger an allocation of ~2.5 GB. A few such requests are enough to crash the process with OOM.

Here's the script that can be used to reproduce the issue:

```
echo "97eab4890a170a085f5f6e616d655f5f120b746573745f6d6574726963121009000000000000f03f10d48fc9b2a333" \
  | xxd -r -p \
  | curl -X POST \
      "http://127.0.0.1:9090/api/v1/write" \
      -H "Content-Type: application/x-protobuf" \
      -H "Content-Encoding: snappy" \
      -H "X-Prometheus-Remote-Write-Version: 0.1.0" \
      --data-binary @-
```

This change adds a hard limit: the requested decoded length must be less than 32 MB. Requests exceeding the limit are rejected with HTTP 400 before any allocation occurs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
 ## Unreleased
 
+* [FEATURE] HTTP handlers created by `promhttp` package now support metrics filtering by providing one or more `name[]` query parameters. The default behavior when none are provided remains the same, returning all metrics. #1925
 
 ## Unreleased `exp` module
 
 * [BUGFIX] exp/api: Reject malformed snappy payloads declaring huge decoded sizes. Enforce a 32MB decoded-size limit to prevent OOM from oversized remote-write requests. #1917.
+
 ## 1.23.2 / 2025-09-05
 
 This release is made to upgrade to prometheus/common v0.66.1, which drops the dependencies github.com/grafana/regexp and go.uber.org/atomic and replaces gopkg.in/yaml.v2 with go.yaml.in/yaml/v2 (a drop-in replacement).
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -89,6 +89,10 @@ var gzipPool = sync.Pool{
 // metrics used for instrumentation will be shared between them, providing
 // global scrape counts.
 //
+// The handler supports filtering metrics by name using the `name[]` query parameter.
+// Multiple metric names can be specified by providing the parameter multiple times.
+// When no name[] parameters are provided, all metrics are returned.
+//
 // This function is meant to cover the bulk of basic use cases. If you are doing
 // anything that requires more customization (including using a non-default
 // Gatherer, different instrumentation, and non-default HandlerOpts), use the
@@ -105,6 +109,10 @@ func Handler() http.Handler {
 // Gatherers, with non-default HandlerOpts, and/or with custom (or no)
 // instrumentation. Use the InstrumentMetricHandler function to apply the same
 // kind of instrumentation as it is used by the Handler function.
+//
+// The handler supports filtering metrics by name using the `name[]` query parameter.
+// Multiple metric names can be specified by providing the parameter multiple times.
+// When no name[] parameters are provided, all metrics are returned.
 func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 	return HandlerForTransactional(prometheus.ToTransactionalGatherer(reg), opts)
 }
@@ -112,6 +120,10 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 // HandlerForTransactional is like HandlerFor, but it uses transactional gather, which
 // can safely change in-place returned *dto.MetricFamily before call to `Gather` and after
 // call to `done` of that `Gather`.
+//
+// The handler supports filtering metrics by name using the `name[]` query parameter.
+// Multiple metric names can be specified by providing the parameter multiple times.
+// When no name[] parameters are provided, all metrics are returned.
 func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerOpts) http.Handler {
 	var (
 		inFlightSem chan struct{}
@@ -245,7 +257,21 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 			return false
 		}
 
+		// Build metric name filter set from query params (if any)
+		var metricFilter map[string]struct{}
+		if metricNames := req.URL.Query()["name[]"]; len(metricNames) > 0 {
+			metricFilter = make(map[string]struct{}, len(metricNames))
+			for _, name := range metricNames {
+				metricFilter[name] = struct{}{}
+			}
+		}
+
 		for _, mf := range mfs {
+			if metricFilter != nil {
+				if _, ok := metricFilter[mf.GetName()]; !ok {
+					continue
+				}
+			}
 			if handleError(enc.Encode(mf)) {
 				return
 			}
--- a/prometheus/promhttp/http_test.go
+++ b/prometheus/promhttp/http_test.go
@@ -640,3 +640,110 @@ func BenchmarkCompression(b *testing.B) {
 		}
 	}
 }
+
+func TestHandlerWithMetricFilter(t *testing.T) {
+	reg := prometheus.NewRegistry()
+
+	counter := prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "test_counter",
+		Help: "A test counter.",
+	})
+	gauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "test_gauge",
+		Help: "A test gauge.",
+	})
+	histogram := prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name: "test_histogram",
+		Help: "A test histogram.",
+	})
+
+	reg.MustRegister(counter, gauge, histogram)
+	counter.Inc()
+	gauge.Set(42)
+	histogram.Observe(3.14)
+
+	testCases := []struct {
+		name             string
+		url              string
+		shouldContain    []string
+		shouldNotContain []string
+	}{
+		{
+			name:             "single metric filter",
+			url:              "/?name[]=test_counter",
+			shouldContain:    []string{"test_counter"},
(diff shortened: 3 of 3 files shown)


exp/api/remote/remote_api.go
```diff
@@ -448,6 +448,8 @@ func WithWriteHandlerMiddlewares(middlewares ...func(http.Handler) http.Handler)
 	}
 }
 
+var maxDecodedSize = 32 * 1024 * 1024
```
kakkoyun: Should we document how did we come up with this number?
Is there another source of truth we need to be in sync with?

kakkoyun:
```suggestion
const maxDecodedSize = 32 * 1024 * 1024
```

makasim: I used [the max remote-write size that VictoriaMetrics](https://github.com/VictoriaMetrics/VictoriaMetrics/blob/24fac64875391a6ec9999ec2255f32fada39a6f3/lib/protoparser/promremotewrite/stream/streamparser.go#L19) as const value. Users can override it via a flag `-maxInsertRequestSize` if needed. Prometheus previously had no limit, so enforcing 32 MB might break some setups. 

Need some guidance here, should we raise or lower the value, or make it explicitly configurable?

makasim:
> We need some documentation.

I can take a look. Can you please point out where I can find the docs' sources?

kakkoyun:
> I think @kakkoyun meant just a clear commentary.

Sorry, if it is not clear. Yes, that was what I meant.

makasim:
Added const description and mw tests. Please review

makasim:
Added changelog, rebased&squashed.

makasim:
@kakkoyun kind reminder

kakkoyun:
@bwplotka @ArthurSens @vesari any objections? Or concerns?

exp/api/remote/remote_api.go
```diff
@@ -479,6 +487,18 @@ func SnappyDecodeMiddleware(logger *slog.Logger) func(http.Handler) http.Handler
 				return
 			}
 
+			decodedSize, err := snappy.DecodedLen(bodyBytes)
+			if err != nil {
+				logger.Error("Error snappy decoding remote write request", "err", err)
```
bwplotka: nit

```suggestion
				logger.Error("Error snappy decoding request body length", "err", err)
```

exp/api/remote/remote_api.go
```diff
@@ -479,6 +487,18 @@ func SnappyDecodeMiddleware(logger *slog.Logger) func(http.Handler) http.Handler
...
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			}
+			if decodedSize > maxDecodedSize {
+				logger.Error("Error snappy decoding remote write request: decoded size exceeds maximum allowed", "decodedSize", decodedSize, "maxDecodedSize", maxDecodedSize)
```
bwplotka:
```suggestion
				logger.Error("Snappy decoded size exceeds the limit", "sizeBytes", decodedSize, "limitBytes", maxDecodedSize)
```

exp/api/remote/remote_api.go
```diff
@@ -479,6 +487,18 @@ func SnappyDecodeMiddleware(logger *slog.Logger) func(http.Handler) http.Handler
...
+				http.Error(w, "decoded size exceeds maximum allowed", http.StatusBadRequest)
```
bwplotka:
```suggestion
				http.Error(w, fmt.Sprintf("decoded size exceeds the %v bytes limit", maxDecodedSize), http.StatusBadRequest)
```

exp/api/remote/remote_api_test.go
```diff
@@ -295,3 +299,136 @@ func TestRemoteAPI_Write_WithHandler(t *testing.T) {
 		}
 	})
 }
+
+func TestSnappyDecodeMiddleware(t *testing.T) {
+	tLogger := slog.Default()
+
+	var actReq *writev2.Request
+	successH := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
bwplotka: `successH` might be too cryptic 🤔 I initially thought it's a typo.

```suggestion
	successHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```

exp/api/remote/remote_api_test.go
```diff
@@ -295,3 +299,136 @@ func TestRemoteAPI_Write_WithHandler(t *testing.T) {
 		}
 	})
 }
+
+func TestSnappyDecodeMiddleware(t *testing.T) {
+	tLogger := slog.Default()
+
+	var actReq *writev2.Request
```
bwplotka: This would perhaps allow us to remove comments "populated by ... "? (It's clear it's something we "get" from test.

```suggestion
	var gotRequest *writev2.Request
```

exp/api/remote/remote_api_test.go
```diff
@@ -295,3 +299,136 @@ func TestRemoteAPI_Write_WithHandler(t *testing.T) {
...
+	successH := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		b, err := io.ReadAll(r.Body)
+		if err != nil {
+			t.Fatalf("failed to read body: %v", err)
+		}
+		actReq = &writev2.Request{}
+		if err := proto.Unmarshal(b, actReq); err != nil {
```
bwplotka: We have this beautiful `remote.API` for client and server, yet we don't test/use it in this test and we manually unmarshal, send remote write.

Shouldn't we use it?

exp/api/remote/remote_api_test.go
```diff
@@ -295,3 +299,136 @@ func TestRemoteAPI_Write_WithHandler(t *testing.T) {
...
+
+		if rw.Code != http.StatusOK {
+			t.Fatalf("expected status 200, got %d: %s", rw.Code, rw.Body.String())
+		}
+		if diff := cmp.Diff(expReq, actReq, protocmp.Transform()); diff != "" {
+			t.Fatalf("unexpected request after decoding: %s", diff)
+		}
+	})
+
+	t.Run("too_big", func(t *testing.T) {
```
bwplotka: I'd keep one. Less code to maintain.

 It does not matter from the test perspective what those bytes represent. If we want to keep doing low level HTTP server/client test, I'd stick to bytes only.

If we choose remote.API / remote.Handler, I'd use proto requests.

WDYT?

exp/api/remote/remote_api_test.go
```diff
@@ -295,3 +299,136 @@ func TestRemoteAPI_Write_WithHandler(t *testing.T) {
...
+			t.Fatalf("expected status 400, got %d", rw.Code)
+		}
+
+		body := rw.Body.String()
+		if !strings.Contains(body, "decoded size exceeds maximum allowed") {
+			t.Fatalf("expected error message about size limit, got: %s", body)
+		}
+	})
+
+	t.Run("invalid", func(t *testing.T) {
```
bwplotka, line 371: 👍🏽

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+:warning: This release introduces a 32MB hard limit on decoded snappy payloads to prevent potential OOM attacks. Requests with decoded payloads exceeding this limit will be rejected with HTTP 400 status code. See #1917 for details. :warning:
+
+* [BUGFIX] exp/api: Reject malformed snappy payloads declaring huge decoded sizes. Enforce a 32MB decoded-size limit to prevent OOM from oversized remote-write requests. #1917.
+
```
bwplotka: Thanks for the changelog item! I'd remove warning. It's an experimental module, totally outside of client_golang main module actually, so we need perhaps section, but no need for warning.

```suggestion

## Unreleased `exp` module

* [BUGFIX] exp/api: Reject malformed snappy payloads declaring huge decoded sizes. Enforce a 32MB decoded-size limit to prevent OOM from oversized remote-write requests. #1917.
```

exp/api/remote/remote_api.go
```diff
@@ -448,6 +448,14 @@ func WithWriteHandlerMiddlewares(middlewares ...func(http.Handler) http.Handler)
 	}
 }
 
+// maxDecodedSize limits the maximum allowed size of decompressed snappy payloads.
```
bwplotka:
```suggestion
// maxDecodedSize limits the maximum allowed bytes of decompressed snappy payloads.
```

exp/api/remote/remote_api.go
```diff
@@ -448,6 +448,14 @@ func WithWriteHandlerMiddlewares(middlewares ...func(http.Handler) http.Handler)
...
+// This protects against maliciously crafted payloads that could cause excessive memory
+// allocation and potentially lead to out-of-memory (OOM) conditions.
+// All usual payloads should be much smaller than this limit and pass without any problems.
+//
+// See more in https://github.com/prometheus/client_golang/pull/1917
```
bwplotka: I think this comment is clear enough, perhaps no need for anyone to dig old PR here. It's a great change and clear enough to understand 👍🏽 

```suggestion
```

exp/api/remote/remote_api_test.go
makasim: fixed in 21da12f

exp/api/remote/remote_api_test.go
makasim: fixed in 21da12f

exp/api/remote/remote_api_test.go
makasim: To me, they highligh different scenarios. If one has to go, I’d drop this one and keep `crafted_decode_len`. If you prefer a different direction, let me know and I’ll adjust.

makasim:
@bwplotka kind reminder

exp/api/remote/remote_api_test.go
makasim: too_big has been removed

makasim:
DCO fixed, commits squashed, `too_big` test removed.

PR #1925, opened 2025-12-02
Allow `/metrics` handler output filtering via `name[]` query param

colega:
## Context

We have a huge application with thousands of metrics per pod, and we have a Prometheus instance that need to scrape a small subset of those metrics (5 to 10 metrics)

I've set up some relabeling rules, and it works great, but profiles say that this instance spends more than 50% of its CPU time just parsing openmetrics responses and dropping the metrics.

We could save some trees if there was a way to signal Prometheus that we just want a subset of them. I considered running a dumb nginx proxy that would strip out the unneeded metrics, but with native histograms being exposed in proto that won't be possible anymore, plus we'd still be spending precious CPU cycles encoding that response.

I have considered exposing those metrics on a different registry/handler, but:
- It adds complexity to the application, those metrics are deep in the business logic
- It would leak the details of how I operate the application to the application itself: the app shouldn't know whether I'm scraping with one Prometheus, two, or N.

## What this PR does

This PR adds support for filtering metrics by name using the `name[]` query parameter in `HandlerForTransactional` (and hence in `HandlerFor` and `promhttp.DefaultHandler`). The default behavior is unchanged.

## Related issue

I see this has been discussed already in https://github.com/prometheus/client_golang/issues/135
I don't see much consensus there, however I found out that this is something already implemented in `client_python` [ref](https://pypi.org/project/prometheus-client/0.18.0/) and `client_java` [ref](https://prometheus.github.io/client_java/exporters/filter/) through the `name[]` query param, so I hope there isn't much discussion about whether this is a good fit here.

## Example
```bash
# Get all metrics (existing behavior)
curl http://localhost:8080/metrics

# Get specific metrics
curl 'http://localhost:8080/metrics?name[]=http_requests_total&name[]=process_cpu_seconds_total'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+* [FEATURE] HTTP handlers created by `promhttp` package now support metrics filtering by providing one or more `name[]` query parameters. The default behavior when none are provided remains the same, returning all metrics. #1925
+
 ## 1.23.2 / 2025-09-05
 
 This release is made to upgrade to prometheus/common v0.66.1, which drops the dependencies github.com/grafana/regexp and go.uber.org/atomic and replaces gopkg.in/yaml.v2 with go.yaml.in/yaml/v2 (a drop-in replacement).


prometheus/promhttp/http.go
```diff
@@ -112,6 +112,10 @@ func HandlerFor(reg prometheus.Gatherer, opts HandlerOpts) http.Handler {
 // HandlerForTransactional is like HandlerFor, but it uses transactional gather, which
 // can safely change in-place returned *dto.MetricFamily before call to `Gather` and after
 // call to `done` of that `Gather`.
+//
+// The handler supports filtering metrics by name using the `name[]` query parameter.
+// Multiple metric names can be specified by providing the parameter multiple times.
+// When no name[] parameters are provided, all metrics are returned.
```
aknuds1, line 126: Could you add this paragraph to the more high-level functions `HandlerFor` and `Handler` as well? I think that would make sense, as the new functionality applies to them too and they are probably more often used by API clients.

prometheus/promhttp/http_test.go
```diff
@@ -640,3 +640,125 @@ func BenchmarkCompression(b *testing.B) {
...
+			url:              "/?name[]=test_counter&name[]=test_counter",
+			shouldContain:    []string{"test_counter"},
+			shouldNotContain: []string{"test_gauge", "test_histogram"},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			writer := httptest.NewRecorder()
+			request, _ := http.NewRequest(http.MethodGet, tc.url, nil)
```
aknuds1: I would personally prefer not ignoring errors in tests, however unlikely they may be, because of the chaos missed errors could cause:
```suggestion
			request, err := http.NewRequest(http.MethodGet, tc.url, nil)
			if err != nil {
				t.Errorf("unexpected error: %s", err)
			}
```

prometheus/promhttp/http_test.go
```diff
@@ -640,3 +640,125 @@ func BenchmarkCompression(b *testing.B) {
...
+			for _, notExpected := range tc.shouldNotContain {
+				if strings.Contains(body, notExpected) {
+					t.Errorf("expected body to NOT contain %q, got: %s", notExpected, body)
+				}
+			}
+		})
+	}
+}
+
+func TestHandlerWithMetricFilterTransactionalCalls(t *testing.T) {
```
aknuds1: On reading this test, I'm thinking this isn't a use case per se, but additional assertions. Could you instead merge them into the sub-test in `TestHandlerWithMetricFilter`, so they are checked per use case?

aknuds1 (commented):
I lack experience with the code in question, but the new functionality looks sensible to me. Leaving some suggestions. Also, shouldn't there be a changelog entry?

prometheus/promhttp/http.go
colega: Sure, done in https://github.com/prometheus/client_golang/pull/1925/commits/13f3eb9cd1853ee590315e643ecfb26b61df7f1b

prometheus/promhttp/http_test.go
colega: Done in https://github.com/prometheus/client_golang/pull/1925/commits/499e6f4be4737c3fac9d3630d7b0cb61d6edd0ff

prometheus/promhttp/http_test.go
colega: Done in https://github.com/prometheus/client_golang/pull/1925/commits/0fbe04e4fd1f10f46d29da8e5e8ef815f40c5ad1

colega:
Oh, right, sorry, I forgot about the CHANGELOG.md, updated now.

kakkoyun:
@roidelapluie @aknuds1 for taking care of it. I guess we (client_golang maintainers) were not fast enough.

bwplotka:
Thanks for this!

Please avoid merging PRs on a hot path without [maintainers](https://github.com/prometheus/client_golang/blob/main/MAINTAINERS.md) acknowledgement. Especially as it looks like it was only done/reviewed by a single company (plus one [community member](https://github.com/mohsanabbas)), when this library is used by thousands. If you want to maintain this project, respond on issues, fix bugs, do releases we would welcome new active members (:

Nevertheless, I reviewed and change is great, LGTM. If it was not, I hope you wouldn't mind if we would revert to iterate.

Re []name: This has been discussed a lot in the past. Users definitely need it, and the only controversy is that there is no STANDARD that would define clean API. E.g. users want to match on regexes, labels etc. This is important as once adopted, scrapers will need to adjust etc. I wished it was defined on OpenMetrics side etc (cc @dashpole @krajorama @ywwg should we?).

dashpole:
I'm not sure it belongs in the protocol specification.  We do have a section on protocol negotiation that covers TLS and version negotiation, but I would prefer to keep the spec as limited to the text format as possible.

bwplotka:
Then what else we can do to ensure all SDKs have standard API for matching/URL of `/metrics` and avoid issue of scrape having to have custom flow (likely manually configurable) per specific application? cc @dashpole

dashpole:
I would probably write a simple proposal, and then document it here https://prometheus.io/docs/instrumenting/writing_clientlibs/?  Or maybe it can be its own page, similar to how UTF-8 is documented. I would hope we can apply this to more than just OpenMetrics 2.0

juliusv:
> Re []name: This has been discussed a lot in the past. Users definitely need it, and the only controversy is that there is no STANDARD that would define clean API. E.g. users want to match on regexes, labels etc.

Besides the call for a standard, I also think that being able to filter by regexes is something I would expect from this kind of filtering functionality. If I wanted all `container_.*` metrics, I would have to list a lot of `name[]` params, and I might not always even know all the metric names that exist with a given prefix (if I want to just get them all).

colega:
I think that would require a better feature definition, and also what's the use case for that? Our use case for this was to bring only the minimum required metrics to evaluate some recording rules, I think that bringing arbitrary series matching a regexp defies that purpose.

I'm not against it, I just think that we shouldn't try to build it if there's no real use case.

juliusv:
The main use case I can think of would be what people are already doing with metric relabeling in the Prometheus server to only scrape a subset of metrics from a target (for example, because they care about one topic area of metrics that all start with a given prefix), but doing so more efficiently because the filtering happens earlier on in the chain. So a metric relabeling rule like:

```yaml
metric_relabel_configs:
  - action: keep
    source_labels: [__name__]
    regex: '(api_|http_|container_).*'
```

Relabeling is powerful because it also allows matching on other labels, rewriting them, extracting parts of one label into another, and so on. That's probably more than we want here, but some form of pattern-based matching would still be nice for this use case.

roidelapluie:
`?keep[]=(api_|http_|container_).*`

PR #1928, opened 2025-12-15
fix: updates for tsdb blocks handling

tjhop:
Two small fixes:

- I forgot the implementation for `*apiClientImpl.Do()` already extracts
  the response content from the enveloped `data` field in the response,
so the wrapper struct isn't needed here.
- The `compaction` section of a block has an optional `parents` field
  for tracking block lineage in deeper compaction levels. This slipped
my mind during implementation because I was looking at the example API
response on the API docs, which use a mock response with a single block
that is uncompacted.

I noticed these while working on tjhop/prometheus-mcp-server#77

Signed-off-by: TJ Hoplock <t.hoplock@gmail.com>


kakkoyun (changes requested):
Considering these issues slipped through the cracks, I would really appreciate if you could add more comprehensive tests so that we prevent similar future issues.

Thanks in advance 🙏

tjhop:
That's fair, I can expand tests. 

Ideally, it would be simpler to just import prometheus itself as a library and use the types [already defined for these in prometheus](https://github.com/prometheus/prometheus/blob/main/tsdb/block.go#L164-L217). That would obviously help to keep things in sync and benefit from existing testing there. It becomes a bit awkward though, since prometheus itself also imports the client_golang library (mostly for implementing metrics) and so we'd be importing ourselves and nearing a circular import. It would also blow up dependencies, and I know this library is already heavily used/strictly watched for deps. 

Perhaps this is a situation where a little bit of copying is better than a little bit of dependency?

tjhop:
Both of the checks that are failing are for `TestGoCollectorAllowList` which is unrelated to this PR, so I think we should be good? 

@kakkoyun sorry it's been a bit -- if you get some more time, I'd appreciate some fresh eyes please!

tjhop:
Both of the test failures in CI are for collectors/ and are unrelated to the changes here :+1:

kakkoyun (approved):

tjhop:
@kakkoyun @bwplotka @ArthurSens would you mind taking a fresh look at this? Now that 1.24.x is out, this will be a breaking change. 

To the point about improving tests, in addition to the expanded test suite and edge cases covered I also used claude to help create a parity test -- its a separate test suite in a separate go mod that uses prometheus as a library with our upstream prom version's code and compares it against client_golang's counterparts. It also wires up a separate CI job to bump prometheus versions and fail loudly if the parity test fails. Thoughts?

api/prometheus/v1/paritytest/tsdb_blocks_parity_test.go
```diff
@@ -0,0 +1,257 @@
...
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+)
+
+// baseT is an arbitrary fixed sample timestamp; it keeps the fixture
+// deterministic.
+const baseT int64 = 1750860000000
+
+// writeBlock writes a block with float and native histogram samples to
+// dir and returns its ULID.
+func writeBlock(t *testing.T, dir string, seed int64) ulid.ULID {
```
bwplotka: This is likely an overkill -- tons of complexity that can change anytime (Prom often adds breaking changes to those e.g. TSDB internal methods). 
This will also not catch the other bug around "data" parsing.

Here it would be enough to compare if structs match (eg. .reflection) or even regenerate struct with some go generator and put that into a test, but instead I wonder - should we invest in [using OpenAPI spec](https://github.com/prometheus/client_golang/issues/1998) ([PoC](https://github.com/prometheus/client_golang/pull/2080/)) instead?

tutorials/whatsup/internal/acceptance_test.go
```diff
@@ -78,5 +78,4 @@ func TestAcceptance(t *testing.T) {
 	if gotErr {
 		fmt.Println("Got this response from ", fmt.Sprintf("http://localhost:%v", WhatsupPort), ":", metrics)
```
bwplotka, line 79: Let's minimize unrelated changes

api/prometheus/v1/api_test.go
```diff
@@ -1202,52 +1202,553 @@ func TestAPIs(t *testing.T) {
...
+							},
+						},
+						Failed: true,
+						Hints:  []string{"deduplication", "downsampling"},
+					},
+				}},
+			},
+		},
+
+		// Multiple blocks at different compaction levels.
```
bwplotka: Do we care about this case? Nothing changes from the API standpoint here - less tests to maintain the better

api/prometheus/v1/api_test.go
```diff
@@ -1202,52 +1202,553 @@ func TestAPIs(t *testing.T) {
...
+									MaxTime: 1750867200000,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+
+		// Null blocks field (server returns null instead of empty array).
```
bwplotka: Do we need to test Go unmarshalling here? We don't do this for other API endpoints.

api/prometheus/v1/api_test.go
```diff
@@ -1202,52 +1202,553 @@ func TestAPIs(t *testing.T) {
...
+			},
+		},
+
+		// Block with all optional fields populated. The stats values
+		// deliberately exceed 2^32: narrowing any of the uint64 stats
+		// fields (which real blocks do overflow past 32 bits) makes the
+		// expected literal below a compile error and the decode a
+		// truncation failure.
+		{
+			do:        doTSDBBlocks(),
```
bwplotka, line 1233: Can we have a single case that have all of those quirks? (CI time is important)

api/prometheus/v1/api_test.go
```diff
@@ -1202,52 +1202,553 @@ func TestAPIs(t *testing.T) {
...
+						"minTime": 1750860000000,
+						"maxTime": 1750867200000,
+						"compaction": map[string]any{
+							"level": 1,
+							"sources": []any{
+								"01CCCCCCCCCCCCCCCCCCCCCC01",
+							},
+						},
+						"version": 1,
+						// Note: no "stats" field
```
bwplotka: Again, could be in one case if we want

api/prometheus/v1/api_test.go
```diff
@@ -1202,52 +1202,553 @@ func TestAPIs(t *testing.T) {
...
+					Stats:   TSDBBlockStats{}, // zero values
+					Compaction: TSDBBlockMetaCompaction{
+						Level:   1,
+						Sources: []string{"01CCCCCCCCCCCCCCCCCCCCCC01"},
+					},
+				}},
+			},
+		},
+
+		// Block with the compaction hint Prometheus's OOO feature emits
```
bwplotka: ditto

api/prometheus/v1/api_test.go
```diff
@@ -1202,52 +1202,553 @@ func TestAPIs(t *testing.T) {
...
+					Compaction: TSDBBlockMetaCompaction{
+						Level:   1,
+						Sources: []string{"01DDDDDDDDDDDDDDDDDDDDDD01"},
+						Hints:   []string{"from-out-of-order"},
+					},
+				}},
+			},
+		},
+
+		// Block with zero-value time boundaries (edge case).
```
bwplotka: Perhaps too much? Could be removed?

api/prometheus/v1/api_test.go
```diff
@@ -1202,52 +1202,553 @@ func TestAPIs(t *testing.T) {
...
+					},
+					Compaction: TSDBBlockMetaCompaction{
+						Level:   1,
+						Sources: []string{"01EEEEEEEEEEEEEEEEEEEEEE01"},
+					},
+				}},
+			},
+		},
+
+		// Block with different version values.
```
bwplotka: Again, not really changing anything - should be part of a single case

bwplotka (changes requested):
Thanks! 

Maybe I am not used to not worrying about extra code in the AI era, but I would prefer we don't reimplement block creation or use internal Prometheus DB for niche API like TSDB block API that we didn't notice it was broken (: 

I wonder if we can focus on long term testing for those things - for now I am ok to fix it and merge and figure out e2e testing later. There is also https://github.com/prometheus/client_golang/issues/1998 which is an interesting alternative.

If we really need testing, then I'd vote for reflection based solution only (diff of structs).

tutorials/whatsup/internal/acceptance_test.go
tjhop: Totally fair. this was added purely to get  CI to pass and kept in a separate commit for the `go fmt`. I'll drop the commit 👍

api/prometheus/v1/paritytest/tsdb_blocks_parity_test.go
tjhop: Agreed it's above and beyond, the goal was to more functionally exercise parity, but I hear where you're coming from. I can drop the commit and rework to use reflection for now -- I didn't know about the openapi PoC and would vote to focus on that as the better long term effort 👍

tjhop:
Thanks @bwplotka, all feedback should be addressed, mind taking another look when time allows?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * [FEATURE] testutil: Add GatherAndFormat to encode a subset of metrics from a Gatherer. #2091
+* [CHANGE] **breaking** api/prometheus/v1: Fix `TSDBBlocks` so it can decode real server responses; fixes issues with data enveloping and aligns the TSDB Blocks struct types to match the upstream Prometheus counterpart. #1928
```
bwplotka:
```suggestion
* [BUGFIX] **breaking** api/prometheus/v1: Fix `TSDBBlocks` so it can decode real server responses; fixes issues with data enveloping and aligns the TSDB Blocks struct types to match the upstream Prometheus counterpart. #1928
```

bwplotka (approved):
Thanks!

bwplotka:
Hm, I did break something by changing CHANGELOG? 🤔

PR #1929, opened 2025-12-20
promhttp/zstd: add unit tests for zstd writer registration

90ashish:
### What does this PR do?

Adds unit tests for the promhttp/zstd package to validate that importing the
package correctly enables zstd support and that the registered writer
produces valid, decompressible output.

### Why is this needed?

The promhttp/zstd package relies on an init()-time side effect to register
zstd compression support via internal.NewZstdWriter. Prior to this change,
this behavior was untested.

These tests:
- Assert that zstd support is registered when the package is imported
- Verify that the writer returned by NewZstdWriter produces valid zstd output

### Impact

- No production code changes
- Improves test coverage and confidence in zstd support
- No API or behavior changes

### Notes for reviewers

The tests intentionally focus on the package’s side-effect contract
(blank import enables zstd support) and avoid any changes to existing logic.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -46,11 +46,11 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@fe4161a26a8629af62121b670040955b330f9af2 # v3.29.5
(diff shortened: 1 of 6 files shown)


prometheus/promhttp/zstd/zstd_test.go
```diff
@@ -0,0 +1,65 @@
...
+	if _, err := w.Write(in); err != nil {
+		t.Fatalf("write failed: %v", err)
+	}
+	closeFn()
+
+	r, err := kpzstd.NewReader(bytes.NewReader(buf.Bytes()))
+	if err != nil {
+		t.Fatalf("zstd NewReader failed: %v", err)
+	}
+	defer r.Close()
```
kakkoyun: nit:
```suggestion
	t.Cleanup(func() { r.Close() })
```

kakkoyun (approved):

prometheus/promhttp/zstd/zstd_test.go
90ashish: Updated as suggested — replaced `defer r.Close()` with `t.Cleanup(func() { r.Close() })`.
Thanks for the review!

kakkoyun (approved):

Issue #1900, opened 2025-10-25
Registry.Gather should turn panics into errors


bboreham:
This code:

https://github.com/prometheus/client_golang/blob/9358c16714047a9913e13285f8d0f8b21e2c80b1/prometheus/registry.go#L452-L456

will end the whole program if the collector panics, since it is run on a background goroutine. This may be an unpleasant lesson.

As a comparison, the `net/http` package handles panics anywhere in the serving code, although the author [says they should have made this opt-in](https://groups.google.com/g/golang-dev/c/rXs4TG1gdXw/m/7BQ29S4NPrgJ). 

I note that `ErrGroup` added then removed this feature; reading [the thread](https://github.com/golang/go/issues/53757) I understand that the reasons were (1) it could take a long time to reveal the error and (2) the stack trace was lost. `Gather` can fail immediately on panic, and can supply the stack trace in the error it returns.

KaranPradhan266:
Hi @kakkoyun, I’d love to take this on! If it's open for external contributors.

Could you please share a bit more information on the expected behavior? Any pointers or context would be really helpful before I get started. Thanks!

kakkoyun:
> Hi [@kakkoyun](https://github.com/kakkoyun), I’d love to take this on! If it's open for external contributors.
> 
> Could you please share a bit more information on the expected behavior? Any pointers or context would be really helpful before I get started. Thanks!

Everything is open of external contributors. I think @bboreham did a great job on explaining the issue. I would suggest opening a draft PR and discussing any details that is not quite clear for you. We can also use https://cloud-native.slack.com/archives/C037NMVUAP3 for discussions if you think GH comments are not enough.

Just go for it!

Saflaski:
Hi, I noticed this issue has not been touched on for a while and PR #1936 appears to not include the tests mentioned. I've gone ahead and opened a slightly cleaner implementation PR with a test. My approach differs in that it uses a reusable named function wrapping around `Collector.Collect`, which returns err/nil which then gets appended to Gather's `errs`.

raeperd:
Isn't this issue can be closed with #1936 ?

kakkoyun:
It's actually handled in #1961.

PR #1936, opened 2026-01-02
fix(prometheus/registry): recover from collector panic in collectWorker

stlb57:
Issue : Registry.Gather should turn panics into errors #1900 

This PR adds panic recovery around **collector.Collect()** calls in collectWorker. Previously, a panic inside a collector would crash the entire program.

With this change, panics are caught, a stack trace is captured, and an invalid metric is emitted instead. This prevents crashes and surfaces the error to Prometheus.

Added a test with a fake collector that panics to verify recovery.


kakkoyun:
Superseded by #1961

PR #1939, opened 2026-01-09
docs: document using client_golang with OpenTelemetry (OTLP)

### LINKED ISSUE #1495: Document the bridge that allows for client_golang instrumented code to push with OpenTelemetry OTLP
Thanks to amazing community behind OpenTelemetry Go SDK (special kudos to @dashpole), users of client_golang can optionally import a new Go module, a special bridge [[1]](https://pkg.go.dev/go.opentelemetry.io/contrib/bridges/prometheus)[[2]](https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/bridges/prometheus) and push data from `prometheus.Registry` using OTLP push.

This is useful if you have your application metrics instrumented with Prometheus client_golang, but wish to use OTLP push instead of (or additionally to) `promhttp` /metrics pull method, because perhaps it's a batch job or you want to use push based collection mechanisms with e.g. OTel collector.  Notably, initial benchmarks shows that, currently, this way is more efficient than e.g. pulling metrics using Otel collector Prometheus receiver and converting to OTLP data model. This situation, however, can change as the source of the efficiency issues can be, in theory, optimized away.

**Opinionated disclaimer: Using OTLP push (or generally push) has its own disadvantages. Converting Prometheus metrics to OTLP format has another set of disadvantages. Our recommendation is to prefer pull mechanisms for metric pipelines for reliability, centralised discovery, efficiency and simple client failover scenarios. Standard `promhttp` /metrics pull mechanism remains the most efficient and reliable way to directly get metrics from your application to Prometheus or Prometheus compatible backend.**

Nevertheless for those who wants to leverage pure OTLP, this issue is to track adding relevant explanation to our docs. This is also on part with [Java Prometheus client offering OTLP](https://prometheus.github.io/client_java/otel/otlp/)

Saihgax:
This PR documents how applications instrumented with `client_golang` can be integrated into OpenTelemetry pipelines using the OpenTelemetry Collector. 

Fixes #1495.


jeonwilliam727-code:
Saihgax:docs-otel-otlp-bridge

kakkoyun:
@Saflaski this does not exactly belong to client_golang. It should be part of official OpenTelemetry Collector documentation (if it is not there already). This is very generic any process that exposes Prometheus metrics could use this.

PR #1940, opened 2026-01-09
ci: disable unused container_description workflow

### LINKED ISSUE #1938: container_description.yml workflow failing
The "Push README to Docker Hub" workflow (`container_description.yml`) has been failing on every run since at least November 2024, with the following error:

  level=debug msg="namespace: prom"
  level=debug msg="repo: "
  level=error msg="Invalid [IMAGE argument] - bad characters or empty value"

Apparently this is because of a missing name, in that the workflow calls `make docker-repo-name`, but this target doesn't exist in the Makefile. The `Makefile.common` has `common-docker-repo-name`, but:
  1. There's no `docker-repo-name` alias
  2. `DOCKER_IMAGE_NAME` is never defined for this repository

The only Dockerfile in this repo is for building local examples, so I wonder whether we can keep on ignoring this failure or remove this flow or disable it.

sibasispadhi:
Fixes #1938 

This workflow was added via automated sync (PrometheusBot) in May 2024 but has never worked because client_golang is a library, not a containerized service that publishes container images.

The workflow fails because:
- DOCKER_IMAGE_NAME is not defined in the Makefile
- The 'docker-repo-name' Makefile target doesn't exist
- The only Dockerfile is for building local examples, not distribution

This change disables the workflow (rather than deleting) to:
- Stop failing CI runs that have occurred since May 2024
- Document why client_golang doesn't need container image publishing
- Prevent future automated syncs from re-enabling it
- Preserve the file for historical context
- Allow manual trigger if ever needed in the future

The workflow trigger is changed from 'push' to 'workflow_dispatch' (manual only) with detailed comments explaining the situation and jobs that exit with an error if accidentally triggered.

### DIFF
--- a/.github/workflows/container_description.yml
+++ b/.github/workflows/container_description.yml
@@ -1,12 +1,32 @@
 ---
 name: Push README to Docker Hub
+
+# DISABLED: This workflow was added via automated sync (PrometheusBot) in May 2024
+# but client_golang is a library, not a containerized service, and doesn't publish
+# container images to Docker Hub or Quay.io.
+#
+# The workflow has never worked because:
+# - DOCKER_IMAGE_NAME is not defined in the Makefile
+# - The `docker-repo-name` target doesn't exist
+# - The only Dockerfile is for building local examples, not for distribution
+#
+# This workflow is disabled to:
+# - Stop failing CI runs
+# - Document that client_golang doesn't need container image publishing
+# - Prevent future automated syncs from re-enabling it
+#
+# See: https://github.com/prometheus/client_golang/issues/1938
+
 on:
-  push:
-    paths:
-      - "README.md"
-      - "README-containers.md"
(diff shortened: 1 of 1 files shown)


sibasispadhi:
@vesari per my analysis, I thought its best to disable the feature. Kindly take a look on the changes, and let me know what is your thought process on same or any further changes needed.

kakkoyun:
@vesari could you take a look at it?

@sibasispadhi This needs a rebase.

bwplotka:
This will not work - it would be synced by Prometheus [anyway](https://github.com/prometheus/prometheus/blob/19273691bb2a1f7584c12134dd23cb6a31de8d52/scripts/sync_repo_files.sh#L209C35-L209C77). I proposed https://github.com/prometheus/client_golang/pull/2044 instead.

Thanks!

PR #1941, opened 2026-01-10
docs: add dependency management guidelines to CONTRIBUTING.md

### LINKED ISSUE #1913: docs: Add guidance on adding new indirect/direct imports.
As per https://cloud-native.slack.com/archives/C01AUBA4PFE/p1762272462094259?thread_ts=1762169496.875849&cid=C01AUBA4PFE there is a lot to understand when it comes to adding new dependencies to client_golang (any stable, highly adopted lib).

Things to cover:
* Vendoring vs new dep
* Using unstable dep
* Topic of indirect deps
* Deps in tests & examples vs prod code

sibasispadhi:
Fixes #1913

Adds comprehensive guidance on managing dependencies in client_golang, based on learnings from the Slack discussion and recent dependency-related PRs. Covers the four key topics requested:

1. Vendoring vs. direct dependencies:
   - Prefer direct dependencies with Go modules
   - Vendoring only as last resort for critical cases
   - Go modules provide reproducible builds via go.sum

2. Using unstable dependencies:
   - Avoid pre-1.0, alpha/beta, and commit-based versions in production
   - Exceptions require strong justification and maintainer approval
   - Includes practical examples of good vs. bad dependency versions

3. Indirect dependencies:
   - KEY INSIGHT: Indirect deps don't always propagate to users!
   - Go modules only propagate deps that are actually used in import chain
   - Example: common's kingpin dependency doesn't leak to client_golang users
   - Practical commands: 'go mod why' and 'go mod graph' for analysis

4. Dependencies in different contexts:
   - Production code: Strictest (affects all users, requires approval)
   - Test code: Flexible (test deps don't leak to importers)
   - Examples: Most flexible (but imports CAN cause dep leakage)
   - Real example: Removing example_test.go imports cleaned up many deps

The guidelines emphasize that client_golang is part of the Kubernetes dependency chain, making our dependency decisions impact thousands of projects. Includes the context that adding dependencies creates significant work for Kubernetes maintainers.

Also covers:
- Why we're strict (Kubernetes factor, transitive dep problem)
- Dependency evaluation criteria
- Update procedures
- Red flags for rejecting dependencies
- Practical examples and real cases from our history


sibasispadhi:
@bwplotka tried to add the documentation as per slack discussion.Kindly take a look.

CONTRIBUTING.md
```diff
@@ -20,3 +20,246 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+### Understanding Indirect Dependencies
+
+**Key Insight:** Not all indirect dependencies propagate to our users!
+
+Go modules are smarter than you might think. An indirect dependency in our `go.mod` doesn't
+automatically become an indirect dependency for projects that import `client_golang`.
+
+**How it works:**
+- If we depend on package A, and A depends on package B, but we never import/use anything from B,
+  then B will **not** propagate to projects that import `client_golang`
```
bwplotka:
```suggestion
  then B will **not** propagate to projects that import `client_golang`.
```

CONTRIBUTING.md
```diff
@@ -20,3 +20,246 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+- Go modules only propagate dependencies that are actually used in the import chain
```
bwplotka: This point is the same and the first one no? Can we remove it.


```suggestion
```

CONTRIBUTING.md
```diff
@@ -20,3 +20,246 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+
+**Example from our codebase:**
+- `prometheus/common` has many dependencies (e.g., `kingpin` for CLI parsing)
+- `client_golang` depends on `common`
+- But `client_golang` users don't get `kingpin` as an indirect dependency because `client_golang`
+  doesn't import the parts of `common` that use `kingpin`
```
bwplotka: Can we fix formatting (trailing period). Also can we be more specific e.g. give permanent links to packages. It is worth noting that kingpin is only imported by package X (let's find that X and link it) in common and client_golang does not use X anywhere.

CONTRIBUTING.md
```diff
@@ -20,3 +20,246 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+
+**Testing this:**
```
bwplotka, line 61: This is great, but either we add more specific info or even script for this or we remove this. It's a bit fuzzy e.g. 

`# See why a package is in our dependencies` - is this from client_golang repo? Or from repo that imports client_golang?

> Test if a dependency propagates by creating a test module that imports client_golang
> and checking if the dependency appears in its go.mod

Kind of out of place? Can we give more specific output? 

It feels a script would help?

CONTRIBUTING.md
```diff
@@ -20,3 +20,246 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+go mod why github.com/some/package
+
+# See the full dependency graph
+go mod graph
+
+# Test if a dependency propagates by creating a test module that imports client_golang
+# and checking if the dependency appears in its go.mod
+```
+
+**Important exception - Examples can leak dependencies:**
```
bwplotka: Why this is exception? 🤔 This is kind of out of place? Should we remove it?

CONTRIBUTING.md
```diff
@@ -20,3 +20,246 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+**Key Concerns:**
+
+When evaluating new dependencies, critical factors include:
+
+- **Security:** Vulnerabilities in our dependency chain affect all users
+- **Transitive dependencies:** How many indirect dependencies does it bring?
+- **Version conflicts:** Can impact downstream projects (like KEDA, Kubernetes, and other users)
+- **Licensing:** Must be Apache 2.0 compatible (avoid GPL, LGPL, or copyleft licenses)
+
+**Process for production code dependencies:**
```
bwplotka:
```suggestion
**Process for adding production code dependencies:**
```

CONTRIBUTING.md
```diff
@@ -20,3 +20,246 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+- Can use newer versions to demonstrate features
+- Still prefer stable dependencies when possible
+- **Warning:** Example imports CAN cause indirect deps to appear in `go.mod`
+
+**Real example discovered in dependency analysis:**
+
+The file `api/prometheus/v1/example_test.go` imports `prometheus/common/config`. Testing
+showed that this single import causes many transitive dependencies to appear in our `go.mod`.
+Removing such imports would clean up numerous indirect dependencies because those packages
+aren't actually used in the production code path.
```
bwplotka: This is maybe too much of the detail. We don't want people to stop reading because CONTRIBUTING.md is too long and repetitive (or explain things that Go documentation explains).

CONTRIBUTING.md
```diff
@@ -20,3 +20,246 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+**Before committing:**
+- Run the full test suite
+- Review the dependency's CHANGELOG for breaking changes
+- Verify `go.sum` changes are reasonable
+
+**Consider the impact:**
+- Will the update affect downstream projects (like Kubernetes)?
+- Does it introduce new transitive dependencies?
+- Are there any security advisories for the old or new version?
+
```
bwplotka: Perhaps we should remove? Can be learned from Go tutorials/docs

CONTRIBUTING.md
```diff
@@ -20,3 +20,246 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+**Alternatives:**
+- What alternatives were considered?
+- Why are they not suitable?
+
+**Recommendation:**
+- Your assessment with justification
+
+Maintainers will review and provide feedback before you proceed with implementation.
+
```
bwplotka: Again, bit repetitive and AI watery for humans to read.

CONTRIBUTING.md
```diff
@@ -20,3 +20,246 @@ Prometheus uses GitHub to manage reviews of pull requests.
...
+- [Semantic Versioning](https://semver.org/)
+- [Go Security Best Practices](https://go.dev/security/best-practices)
+
+### Key Takeaways
+
+1. **Indirect dependencies don't always propagate** - Go modules are smart about this
+2. **Test dependencies don't leak** to projects that import us
+3. **Example code can leak dependencies** - be careful with example imports
+4. **We're part of the Kubernetes dependency chain** - our choices have wide impact
+5. **When in doubt, discuss first** - open an issue before adding dependencies
```
bwplotka: This is not a blog post, let's remove it?

bwplotka (commented):
Thanks! Some comments to make it more readable.

Reviewed during our maintainers sync.

NOTE: It feels this was AI generated, which is fine as long as you double checked this closely. Next time please disclose the use of AI and the level of confidence you have in this change. We will likely add this to our CoC or policy one day clearly. (:

CONTRIBUTING.md
sibasispadhi: removed.

CONTRIBUTING.md
sibasispadhi: Updated inline. Let me know if you meant to be in a separate script ?

CONTRIBUTING.md
sibasispadhi: removed.

CONTRIBUTING.md
sibasispadhi: done

CONTRIBUTING.md
sibasispadhi: removed

CONTRIBUTING.md
sibasispadhi: removed

CONTRIBUTING.md
sibasispadhi: removed.

CONTRIBUTING.md
sibasispadhi: removed

CONTRIBUTING.md
sibasispadhi: done

sibasispadhi:
Hi @bwplotka — thank you for the detailed review.

This was my first Go contribution, so I started with a documentation update. I did use GitHub Copilot for the changes, as AI-assisted workflows are becoming a common part of modern development. I was not aware that explicit disclosure was required.

Regarding the AI-generated language in the documentation, that was my oversight. I did a high-level review but didn’t sufficiently refine or de-duplicate the content. I will be more careful about this in future contributions.

On your point about **potentially adding this to the CoC or policy in the future**;  it’s an interesting area. As AI-assisted tools become more common, which would generate more comments and documentation and it will be an important challenge to address.

Thanks again for the feedback.

Issue #1949, opened 2026-02-11
chore: Migrate to go.dev/blog/synctest in the tests and remove time.Sleep


kakkoyun:
Opportunity for https://go.dev/blog/synctest, eventually (:

_Originally posted by @bwplotka in https://github.com/prometheus/client_golang/pull/1942#discussion_r2794465999_

This requires a bump in `go.mod` version

WZH8898:
Hi, I'd like to work on this issue. Can you assign it to me?

manute:
Hey 👋🏽  

I just created this [PR](https://github.com/prometheus/client_golang/pull/1950) to this issue, please let me know your thoughts. thanks!

kakkoyun:
Done in #1950

PR #1950, opened 2026-02-15
chore(test): bump 1.25, tests with synctest and check not panic

manute:
- bum go version 1.25, 
- Add synctest for the timer tests 
- Check not panic for other test 

ref https://github.com/prometheus/client_golang/issues/1949

### DIFF
--- a/api/prometheus/v1/api_bench_test.go
+++ b/api/prometheus/v1/api_bench_test.go
@@ -100,7 +100,7 @@ func generateData(timeseries, datapoints int) (floatMatrix, histogramMatrix mode
 		floatMatrix = append(floatMatrix, fss)
 		histogramMatrix = append(histogramMatrix, hss)
 	}
-	return
+	return floatMatrix, histogramMatrix
 }
 
 func BenchmarkSamplesJsonSerialization(b *testing.B) {
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -28,6 +28,9 @@ linters:
           - nolintlint
           - structcheck
         path: _test.go
+      - linters:
+          - govet
+        text: "buildtag: \\+build line is no longer needed"
     paths:
       - ^.*\.(pb|y)\.go$
     warn-unused: true
--- a/internal/github.com/golang/gddo/httputil/header/header.go
+++ b/internal/github.com/golang/gddo/httputil/header/header.go
@@ -90,7 +90,7 @@ loop:
 			s = skipSpace(s[1:])
 		}
 	}
-	return
+	return specs
 }
 
 func skipSpace(s string) (rest string) {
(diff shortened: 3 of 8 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -28,14 +28,14 @@ jobs:
         with:
           persist-credentials: false
       - name: Install Go
-        uses: actions/setup-go@c0137caad775660c0844396c52da96e560aba63d # v5.5.0
+        uses: actions/setup-go@v6
```
kakkoyun: We still need to pin these.

.github/workflows/golangci-lint.yml
```diff
@@ -28,14 +28,14 @@ jobs:
...
         with:
-          go-version: 1.24.x
+          go-version: 1.25.x
       - name: Install snmp_exporter/generator dependencies
         run: sudo apt-get update && sudo apt-get -y install libsnmp-dev
         if: github.repository == 'prometheus/snmp_exporter'
       - name: Lint
-        uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
+        uses: golangci/golangci-lint-action@v9
```
kakkoyun: ditto.

api/prometheus/v1/api_bench_test.go
```diff
@@ -100,7 +100,7 @@ func generateData(timeseries, datapoints int) (floatMatrix, histogramMatrix mode
 		floatMatrix = append(floatMatrix, fss)
 		histogramMatrix = append(histogramMatrix, hss)
 	}
-	return
+	return //nolint:gofumpt
```
kakkoyun: I would rather fix the issue instead of putting `nolints`s

.github/workflows/golangci-lint.yml
manute: What is the reason ?

api/prometheus/v1/api_bench_test.go
manute: To fix the issue I would need to remove those `return`s at the end of the funcs that are not returning anything , is that ok ?

.github/workflows/golangci-lint.yml
kakkoyun: Pinning a dependency would prevent any implicit updates. It's more predictable, and secure. We wouldn't me exposed to any unwanted update to a compromised version. It's a best practice. (further reading: https://dev.to/nicolasv/github-actions-why-you-should-pin-your-actions-to-a-specific-version-1513)

api/prometheus/v1/api_bench_test.go
kakkoyun: I don't think it is the case. Here the new `gofumpt` rule is either replaces naked returns or removes the unnecessary returns. I'm sure if you run the tool it would tell you.

manute:
@kakkoyun Done, thanks for the suggestion

PR #1963, opened 2026-03-16
Return info annotations in addition to warnings in the API

zenador:
Now that Prometheus API returns infos as well as warnings, see [here](github.com/prometheus/prometheus/pull/14327/changes#diff-315f251cdd7e93fcec1e7e9505744da1d1828f30d2b61d1f4ce963fa26bf1909), we should add that here as well. This is the non-breaking change version, please let me know if you'd prefer to change the API here as well to have infos as a separate field.

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -477,13 +477,13 @@ type API interface {
 	// Flags returns the flag values that Prometheus was launched with.
 	Flags(ctx context.Context) (FlagsResult, error)
 	// LabelNames returns the unique label names present in the block in sorted order by given time range and matchers.
-	LabelNames(ctx context.Context, matches []string, startTime, endTime time.Time, opts ...Option) (model.LabelNames, Warnings, error)
+	LabelNames(ctx context.Context, matches []string, startTime, endTime time.Time, opts ...Option) (model.LabelNames, Warnings, Infos, error)
 	// LabelValues performs a query for the values of the given label, time range and matchers.
-	LabelValues(ctx context.Context, label string, matches []string, startTime, endTime time.Time, opts ...Option) (model.LabelValues, Warnings, error)
+	LabelValues(ctx context.Context, label string, matches []string, startTime, endTime time.Time, opts ...Option) (model.LabelValues, Warnings, Infos, error)
 	// Query performs a query for the given time.
-	Query(ctx context.Context, query string, ts time.Time, opts ...Option) (model.Value, Warnings, error)
+	Query(ctx context.Context, query string, ts time.Time, opts ...Option) (model.Value, Warnings, Infos, error)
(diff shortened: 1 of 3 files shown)


api/prometheus/v1/api.go
```diff
@@ -1542,7 +1543,7 @@ func (h *apiClientImpl) Do(ctx context.Context, req *http.Request) (*http.Respon
 		}
 	}
 
-	return resp, []byte(result.Data), result.Warnings, err
+	return resp, []byte(result.Data), append(result.Warnings, result.Infos...), err
 }
```
charleskorn, line 1546: Should this return the info and warning annotations separately?

It'd be a breaking change, but separating them would be clearer.

api/prometheus/v1/api.go
```diff
@@ -477,21 +477,21 @@ type API interface {
 	// Flags returns the flag values that Prometheus was launched with.
 	Flags(ctx context.Context) (FlagsResult, error)
 	// LabelNames returns the unique label names present in the block in sorted order by given time range and matchers.
-	LabelNames(ctx context.Context, matches []string, startTime, endTime time.Time, opts ...Option) (model.LabelNames, Warnings, error)
+	LabelNames(ctx context.Context, matches []string, startTime, endTime time.Time, opts ...Option) (model.LabelNames, Warnings, Infos, error)
```
kakkoyun, line 480: I think we have to introduce a new type here to wrap `model.Value`, `model.Label`, `Warnings`, `Infos` here. It's getting out of control. Something like:

```go
type Response[T] struct {
	val T // model.LabelNames or model.Value or model.LabelSet
	Warnings
	Infos
}
```
```suggestion
	LabelNames(ctx context.Context, matches []string, startTime, endTime time.Time, opts ...Option) (Response[model.LabelNames], error)
```

What do you think? @bwplotka

kakkoyun (changes requested):
Thanks for the PR.

I think we need to make some design decisions for the API here.

api/prometheus/v1/api.go
bwplotka: I don't mind both options, 4 return params is not that unclear in this context - one could say it's easier to spot you could have more info.

bwplotka (approved):
Hi! Sorry for lags!

Do you mind rebasing?

I don't disagree with @kakkoyun but I would be happy with 4 params too

zenador:
Hi np! I'm happy to change it if that's the better option, so let me know which is preferred and I'll update it if necessary and rebase!

api/prometheus/v1/api.go
bwplotka: We decided on slack simple version is fine.

bwplotka (approved):
Thanks!

PR #1969, opened 2026-03-25
feat(promhttp): add CoalesceGather option to deduplicate concurrent Gather calls

### LINKED ISSUE #1477: A potential goroutine memory leak
Hello, I found a potential bug when I used the project, I'm not sure I'm right, we can discuss the problem to avoid a potential trouble possibly. 

Blocking position:
https://github.com/prometheus/client_golang/blob/26e3055e5133a9d64e8e5a07a7cf026875d5f55d/prometheus/registry.go#L472
At line 449 call the wg.Add(goroutineBudget), and wg.Wait() is blocked until wg.Done() is called the number of goroutineBudget times to awaken the wg.Wait(). 
But if the select statement chooses the default path, return the function, maybe the wg.Done() has not executed enough times to awaken the wg.Wait(), which can result in a goroutine leak
https://github.com/prometheus/client_golang/blob/26e3055e5133a9d64e8e5a07a7cf026875d5f55d/prometheus/registry.go#L451-L463
Complete codes of the part
https://github.com/prometheus/client_golang/blob/26e3055e5133a9d64e8e5a07a7cf026875d5f55d/prometheus/registry.go#L449-L475

kakkoyun:
## What problem does this solve?

Issue #1477 reports apparent goroutine leaks when using `prometheus/client_golang`. The root cause is **not** a WaitGroup bug — `Registry.Gather()` logic is correct. The real problem:

When a collector's `Collect()` is slower than the scrape interval, each incoming HTTP scrape triggers an independent `Gather()`. Each call spawns its own goroutine pipeline. With no upper bound, concurrent pipelines grow proportionally to `(scrape rate / collection time)` — a goroutine pile-up that looks like a leak but is unbounded concurrent work.

This was originally investigated by @initialed85 in [this comment](https://github.com/prometheus/client_golang/issues/1477#issuecomment-4043521036), who validated the issue and prototyped a "piggybacking" (singleflight) approach. This PR implements that idea cleanly inside the HTTP handler, with zero new public types and zero overhead when disabled.

## Solution

Add `HandlerOpts.CoalesceGather bool`. When `true`, `HandlerForTransactional` wraps the `TransactionalGatherer` in an unexported `coalescingGatherer` that allows only one `Gather()` to run at a time. Concurrent HTTP requests arriving while a gather is in-flight join the existing cycle and receive the same result.

### Usage

```go
http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{
    CoalesceGather: true,
}))
```

## Design decisions

| Decision | Rationale |
|---|---|
| `HandlerOpts` field, not a public wrapper type | Zero new public types; users opt in with one field |
| Per-cycle `gatherCycle` object | Prevents a race where a new cycle overwrites result fields while prior waiters are still reading them |
| Mutex-based ref counting (not atomic) | Ensures `c.cycle = nil` is cleared before `cy.done()` is called, ruling out double-done on a stale pointer |
| `close(cy.ready)` as the visibility barrier | Go memory model guarantees `cy.mfs/err/done` are safely readable after `<-cy.ready` without additional locking |
| Default `false` | Zero overhead; identical code path to today when not enabled |

## Changes

| File | Change |
|---|---|
| `prometheus/promhttp/http.go` | Add unexported `coalescingGatherer` + `gatherCycle` types; add `HandlerOpts.CoalesceGather`; wire into `HandlerForTransactional` |
| `prometheus/promhttp/http_test.go` | Add 4 tests: sequential invariant, done-called-exactly-once, goroutine-leak-free (goleak), fresh-cycle-after-release |

## Test plan

- [x] `go test -race -count=1 ./prometheus/promhttp/...` — all pass, no data races
- [x] `go vet ./prometheus/...` — clean
- [x] `TestCoalesceGatherGoroutineLeakFree` uses `goleak.VerifyNone` to assert no leaked goroutines under concurrent slow-collector load
- [x] `TestCoalesceGatherDoneCalledExactlyOnce` verifies the `TransactionalGatherer` contract: `done()` calls == `Gather()` calls regardless of concurrency

Closes #1477

### DIFF
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -37,6 +37,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"slices"
 	"strconv"
 	"sync"
 	"time"
@@ -96,7 +97,7 @@ type coalescingGatherer struct {
 // gatherCycle tracks a single in-flight Gather and all HTTP handlers sharing it.
 type gatherCycle struct {
 	ready chan struct{}       // closed when Gather completes; happens-before reads of mfs/err/done
-	mfs   []*dto.MetricFamily // set before ready is closed; shared across handlers, must not be mutated
+	mfs   []*dto.MetricFamily // canonical result, set before ready is closed; callers get a slices.Clone, the element values stay shared and must not be mutated
 	err   error               // set before ready is closed
 	done  func()              // underlying done callback; set before ready is closed
 	refs  int                 // number of handlers using this cycle; protected by coalescingGatherer.mu
@@ -107,7 +108,7 @@ var _ prometheus.TransactionalGatherer = (*coalescingGatherer)(nil) // compile-t
 // errGatherPanicked is returned to callers that joined an in-flight coalesced
 // Gather whose underlying gatherer panicked. See the panic guard in Gather for
 // why joiners receive this error instead of the panic itself.
-var errGatherPanicked = errors.New("promhttp: coalesced gather panicked")
+var errGatherPanicked = errors.New("coalesced gather panicked")
 
 func (c *coalescingGatherer) Gather() ([]*dto.MetricFamily, func(), error) {
 	c.mu.Lock()
@@ -116,7 +117,10 @@ func (c *coalescingGatherer) Gather() ([]*dto.MetricFamily, func(), error) {
 		cy.refs++
 		c.mu.Unlock()
 		<-cy.ready
-		return cy.mfs, c.releaseFunc(cy), cy.err
+		// Each caller gets its own slice header so it can filter or reorder
+		// without racing other callers sharing this cycle. The *dto.MetricFamily
+		// values remain shared and must not be mutated in place.
+		return slices.Clone(cy.mfs), c.releaseFunc(cy), cy.err
 	}
 	cy := &gatherCycle{
 		ready: make(chan struct{}),
@@ -136,6 +140,12 @@ func (c *coalescingGatherer) Gather() ([]*dto.MetricFamily, func(), error) {
 	// set cy.err before closing cy.ready so joiners waiting on <-cy.ready fail
 	// with that error instead of silently returning an empty, successful
 	// response, and we clear c.cycle so the next Gather starts a fresh cycle.
+	//
+	// The leader never runs its own releaseFunc on this path, so its ref is
+	// not decremented; that is harmless because the cycle is detached (c.cycle
+	// = nil) and cy.done is still the no-op set at construction (c.g.Gather
+	// panicked before assigning a real done). If cy.done is ever made non-nil
+	// before c.g.Gather runs, this path would need to release it.
 	panicked := true
 	defer func() {
 		if panicked {
@@ -152,7 +162,10 @@ func (c *coalescingGatherer) Gather() ([]*dto.MetricFamily, func(), error) {
 	panicked = false
 	close(cy.ready) // happens-before joiners' reads of cy.mfs/err/done
 
-	return cy.mfs, c.releaseFunc(cy), cy.err
+	// Clone here too so cy.mfs stays the write-once canonical slice: joiners
+	// read it concurrently via slices.Clone, so the leader must not hand out
+	// (and potentially reorder) the same backing array.
+	return slices.Clone(cy.mfs), c.releaseFunc(cy), cy.err
 }
 
 // releaseFunc returns the done callback for one caller sharing cy.
@@ -535,14 +548,23 @@ type HandlerOpts struct {
 	// faster than the time collectors need to produce metrics.
 	//
 	// When enabled, concurrent scrapers share a single metric snapshot per
-	// collection cycle. The returned MetricFamily values are shared and must
-	// not be mutated in place. The built-in handler only reads them, so this
-	// is safe in practice, but custom TransactionalGatherer implementations
-	// that modify the returned families after Gather returns must not use
-	// this option.
+	// collection cycle. Each request receives its own copy of the returned
+	// slice, so filtering or reordering it (for example via name[] query
+	// parameters) is safe. The pointed-to MetricFamily values are still
+	// shared: the built-in handler only reads them, so this is safe in
+	// practice, but a custom TransactionalGatherer that mutates the returned
+	// families in place after Gather returns must not use this option.
 	//
-	// Consider using CoalesceGather together with Timeout to bound both the
-	// scrape response time and the number of concurrent background Gathers.
+	// Because the snapshot is shared, a request that arrives while a cycle is
+	// in flight receives that cycle's result even though collection began
+	// before the request; two scrapers joined to one cycle observe the same
+	// timestamps rather than independently gathered data.
+	//
+	// Consider using CoalesceGather together with Timeout. Timeout bounds the
+	// client-facing response time and keeps at most one collection running at
+	// a time, but it does not cancel the underlying Gather: a joined request
+	// that times out still holds a MaxRequestsInFlight slot until the shared
+	// collection completes.
 	//
 	// Panic handling: a panicking Collector is already turned into an error by
 	// the registry, so joiners receive that error like any other. In the rare
--- a/prometheus/promhttp/http.go
+++ b/prometheus/promhttp/http.go
@@ -104,6 +104,11 @@ type gatherCycle struct {
 
 var _ prometheus.TransactionalGatherer = (*coalescingGatherer)(nil) // compile-time interface check
 
+// errGatherPanicked is returned to callers that joined an in-flight coalesced
+// Gather whose underlying gatherer panicked. See the panic guard in Gather for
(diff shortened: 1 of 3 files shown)


initialed85:
@kakkoyun LGTM- I tested with the roughly same approach as for my hack implementation (ref.: https://github.com/ncabatoff/process-exporter/compare/master...initialed85:process-exporter:kakkoyun/goroutine_leak)

Worked as expected- with a fake 1s sleep added to the scrape method in process-exporter:

- A request loop with a 600ms timeout would would return a result every other request
- 4 x concurrent request loops with no timeout would get a result for every request (after 1s)
- `process_open_fds` reported by the process-exporter stayed pretty steady at 11 or 12 (impacted by whether or not the fast request loop was active at the time)
- `process_open_fds` dropped down to 9 if I killed 2 of the 4 concurrent request loops

So I think we're in good shape- lots of timing out of requests, no climbing FDs; works nicely.

Awesome work!

prometheus/promhttp/http.go
```diff
@@ -79,6 +80,88 @@ var gzipPool = sync.Pool{
...
+		if panicked {
+			c.mu.Lock()
+			if c.cycle == cy {
+				c.cycle = nil
+			}
+			c.mu.Unlock()
+			close(cy.ready)
+		}
+	}()
+	cy.mfs, cy.done, cy.err = c.g.Gather()
```
vesari, line 138: One question: if c.g.Gather() panics, do the joiners actually see the panic, or do they just unblock and return as if everything was ok?

kakkoyun: This is a bit of defensive coding here. They just unblock and clean up. If there is no `recover` in the chain, it will be printed as any other panic. 

Feel free to challenge and propose better alternative. Would you prefer us to capture the error?

prometheus/promhttp/http.go
```diff
@@ -79,6 +80,88 @@ var gzipPool = sync.Pool{
...
+	// deadlocked, and future Gather calls start a fresh cycle.
+	panicked := true
+	defer func() {
+		if panicked {
+			c.mu.Lock()
+			if c.cycle == cy {
+				c.cycle = nil
+			}
+			c.mu.Unlock()
+			close(cy.ready)
```
vesari, line 135:
```suggestion
            cy.err = fmt.Errorf("gather panicked")
			close(cy.ready)
```

prometheus/promhttp/http.go
vesari: I made a suggestion above to capture the error. This is because I was thinking that `HandlerForTransactional` basically returns an `http.Handler` and if I’m reading this correctly  https://github.com/golang/go/blob/e2fa8596cf7016d080c51d772c3012f2533e2219/src/net/http/server.go#L84, then the panics are recovered in any case.

vesari (approved):
I left just a small suggestion.

kakkoyun:
@bwplotka Could you have a look at this before the next release?

bwplotka:
I did start looking, I wonder if it shouldn't be a default behaviour 🤔 

I also don't get yet the panic behaviour given we don't want directly use recover. I also don't see any test for it at the moment.

prometheus/promhttp/http.go
kakkoyun: I set a sentinel error (`errGatherPanicked`) before closing the cycle, so joiners fail with it rather than returning an empty 200. As you spotted, net/http recovers the leader's panic, so no recover is needed here. Added a test.

prometheus/promhttp/http.go
```diff
@@ -79,6 +80,88 @@ var gzipPool = sync.Pool{
...
+		}
+	}()
+	cy.mfs, cy.done, cy.err = c.g.Gather()
+	panicked = false
+	close(cy.ready) // happens-before joiners' reads of cy.mfs/err/done
+
```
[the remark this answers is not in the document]
kakkoyun: Joiners now receive a sentinel error, set before the cycle unblocks, so they no longer return an empty 200 on a gatherer panic. The leader is intentionally not recovered; its panic propagates to net/http as in the non-coalesced path. Added `TestCoalesceGatherPanicUnblocksJoinersWithError`.

prometheus/promhttp/http_test.go
```diff
@@ -875,3 +879,155 @@ func TestHandlerWithMetricFilter(t *testing.T) {
...
+	// Key invariant: done() must be called exactly once per Gather cycle.
+	gathers := counter.gatherCalled.Load()
+	dones := counter.doneCalled.Load()
+	if gathers != dones {
+		t.Errorf("Gather called %d times but done called %d times; invariant violated", gathers, dones)
+	}
+	// Coalescing should keep gather count below the number of requests.
+	if gathers > 2 {
+		t.Errorf("Gather called %d times for 2 requests; expected ≤ 2 with coalescing", gathers)
+	}
```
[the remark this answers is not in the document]
kakkoyun: The test now waits until the second request has joined the in-flight cycle (`refs == 2`) and asserts exactly one Gather and one done, instead of the previous `≤ 2` bound. I replaced the `runtime.Gosched()` timing hack with a deterministic wait.

prometheus/promhttp/http_test.go
```diff
@@ -875,3 +879,155 @@ func TestHandlerWithMetricFilter(t *testing.T) {
...
+	)
+	req, _ := http.NewRequest(http.MethodGet, "/", nil)
+	req.Header.Add(acceptHeader, acceptTextPlain)
+
+	var wg sync.WaitGroup
+	for range 5 {
+		wg.Go(func() {
+			handler.ServeHTTP(httptest.NewRecorder(), req)
+		})
+	}
```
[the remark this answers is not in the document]
kakkoyun: `wg.Go` was added in Go 1.25 and is the current idiom for this; `wg.Add(1)` + `defer wg.Done()` is the older pattern. This module targets Go 1.25 (see go.mod), so I'm keeping `wg.Go`.

kakkoyun:
Thanks for looking. Two parts.

On default vs opt-in: I'd keep it opt-in. Turning it on by default changes semantics for everyone. Concurrent scrapers would share one snapshot per cycle, and the "returned MetricFamily must not be mutated" rule becomes mandatory, which breaks custom gatherers that mutate families after Gather returns. Opt-in keeps zero overhead and today's exact behaviour when off. If we do want it on by default, I'd rather do that as a separate change with its own discussion.

On the panic behaviour: you're right it wasn't clear. The common case never hits special handling. A panicking Collector is already recovered by the registry (safeCollect) and returned as an error, so joiners get that error like any other. The guard only fires if the wrapped gatherer itself panics. Before, joiners were unblocked with a nil error and returned an empty 200, which is the bug. Now the leader's panic still propagates to net/http (no recover, as you preferred), and joiners get a sentinel error so they fail with 500 instead of an empty body. I added `TestCoalesceGatherPanicUnblocksJoinersWithError` and expanded the comments. If you'd rather match safeCollect and recover, then wrap the panic into an error for all callers, that's a small change and I'm happy to switch.

kakkoyun:
Pushed the review follow-ups: joiners now fail with a sentinel error on a gatherer panic instead of an empty 200; the done-once test is deterministic (exactly one Gather/done); and each coalesced caller now gets its own `slices.Clone` of the metric-family slice, so per-request `name[]` filtering or reordering can't race another caller sharing the cycle. The `CoalesceGather` doc now spells out the shared-snapshot staleness, the Timeout interaction, and slice-vs-contents mutation. CI is green.

prometheus/promhttp/http.go
```diff
@@ -428,15 +541,47 @@ type HandlerOpts struct {
 	// Service Unavailable and a suitable message in the body. If
 	// MaxRequestsInFlight is 0 or negative, no limit is applied.
 	MaxRequestsInFlight int
+	// CoalesceGather, if true, deduplicates concurrent Gather calls so that
```
bwplotka, line 544: Is there a value on marking it experimental?

bwplotka (approved):
Thanks for this! I checked the code, it's a solid implementation of a singleflight.

I'm mostly curious about the general usage intend. Essentially.. would we go in Prometheus /metrics or Kubernetes and enable coalesce? When it's recommended? Is it only when you have risk of slow collectors? It's not clear, both solutions have pros and cons.

There are some alternatives too:
* context propagation to collectors so this problem is fixable on slow collectors.
* allow this collector as a composable middleware outside promhttp 
* considering having this in `exp` module for more self-service. 
 
Anyway, this was waiting for quite long, so I don't want to block this or release any longer. Perhaps it's good to experiment.

I kindly asked @dashpole for opinion too. Interestingly similar failure mode (although on push) is being fixed on Otel side: https://github.com/open-telemetry/opentelemetry-specification/issues/5184#issuecomment-4973702489

prometheus/promhttp/http.go
kakkoyun: I have added a comment.

bwplotka (approved):

PR #1982, opened 2026-04-13
promhttp: skip 1xx informational responses in responseWriterDelegator

### LINKED ISSUE #1772: `responseWriterDelegator` incorrectly handles `100 Continue` response

alliasgher:
## What does this do?

`WriteHeader` recorded the first status code it received, so a `100 Continue` sent before the actual response caused the delegator to report `100` as the final status and never invoke `observeWriteHeader` with the real code.

`net/http` applies the same rule — informational (1xx) responses are forwarded to the underlying `ResponseWriter` for correct wire behaviour (e.g. the `Expect: 100-continue` handshake) but do not update the recorded status or the `wroteHeader` flag. See https://github.com/golang/go/blob/go1.24.1/src/net/http/server.go#L1212-L1228.

**Fix:** if `code` is in the 1xx range, forward it to the underlying `ResponseWriter` and return immediately without touching `status` or `wroteHeader`.

Fixes #1772

## Checklist

- [x] `go test ./prometheus/promhttp/`
- [x] `gofmt` clean
- [x] New `TestResponseWriterDelegatorIgnores1xxStatus` covering the bug-report scenario (100 followed by implicit 200 via `Write`) and explicit `WriteHeader(200)` after `WriteHeader(100)`.

Signed-off-by: Ali <alliasgher123@gmail.com>

### DIFF
--- a/prometheus/promhttp/delegator.go
+++ b/prometheus/promhttp/delegator.go
@@ -53,6 +53,17 @@ func (r *responseWriterDelegator) Written() int64 {
 }
 
 func (r *responseWriterDelegator) WriteHeader(code int) {
+	// Informational (1xx) responses are interim and must not be recorded as
+	// the final status code. net/http itself applies the same rule — see
+	// https://github.com/golang/go/blob/go1.24.1/src/net/http/server.go#L1212-L1228.
+	// We still forward them to the underlying ResponseWriter so the client
+	// receives the correct wire behaviour (e.g. 100 Continue for Expect
+	// headers), but we do not update wroteHeader or status so that the
+	// eventual non-informational response is recorded correctly.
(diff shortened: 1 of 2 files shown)


VirrageS:
There is already MR to fix this issue https://github.com/prometheus/client_golang/pull/1773

alliasgher:
Thanks for pointing that out @VirrageS! #1773 has been open since March 2025 and the reviewer @bwplotka requested a unit test before merging — this PR adds exactly that missing test.

Issue #1983, opened 2026-04-14
Proposal: built-in TTL


jpinsonneau:
Hello team,

Several projects in the Prometheus ecosystem need to expire stale metric series from `*Vec` metrics (high-cardinality, ephemeral label sets / think per-pod, per-connection, etc.).

Today the only option is to maintain an **external cache** that shadows `metricMap` to track last-access times and periodically call `Delete()`. This duplicates memory (label sets stored twice), duplicates hashing and leads to a lot of reimplemented boilerplate across projects.

Projects that have independently built this:
- [prometheus/statsd_exporter](https://github.com/prometheus/statsd_exporter/pull/164) — per-mapping `ttl`
- [OTel Collector prometheus exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/24179) — `metric_expiration` + `lastSeen`
- [netobserv/flowlogs-pipeline](https://github.com/netobserv/flowlogs-pipeline) — `TimedCache` LRU shadowing every Vec
- [go.withmatt.com/metrics](https://github.com/mattrobenolt/go-metrics) — alternative Go lib with built-in TTL
- [prom-client-net (.NET)](https://github.com/prom-client-net/prom-client/pull/343) — TTL merged in March 2025
- [prometheus-net (.NET)](https://github.com/prometheus-net/prometheus-net#deleting-metrics) — "managed lifetime" metrics

I know [#1397](https://github.com/prometheus/client_golang/issues/1397) was closed with "use `Delete`/`Reset`", but the number of projects reimplementing the same pattern suggests this would be valuable as a first-class feature. The core idea is simple: an atomic `lastAccessed` timestamp per child (+8 bytes, updated on `GetMetricWith`), and a way to clean up expired entries — either via an explicit method, during `Collect()`, or both.

Happy to discuss the design further and contribute a PR if there's interest. Would something like this be welcome ?

jpinsonneau:
I drafted some code changes using claude code if you are interested: https://github.com/prometheus/client_golang/commit/2444fefa0bfd48c34bfec2257b744eb785a388a0

kakkoyun:
> I drafted some code changes using claude code if you are interested: [2444fef](https://github.com/prometheus/client_golang/commit/2444fefa0bfd48c34bfec2257b744eb785a388a0)

@jpinsonneau Please go ahead and open a PR and let's start discussing details in there.

Some of the examples that you shared not official. The following seems closer to what we can get inspiration on though.

[...]

PR #1986, opened 2026-04-21
Add optional TTL for *Vec children (MetricVec)

jpinsonneau:
### Context

High-cardinality and ephemeral label sets (for example per-pod or per-connection series) are painful to manage with today’s client: the usual approach is an **external shadow cache** plus periodic `Delete`, which duplicates label storage and repeats the same pattern across projects. This is discussed in [**#1983 – Proposal: built-in TTL**](https://github.com/prometheus/client_golang/issues/1983).

### What this PR does

- Adds an optional **`TTL`** field on **`Opts`** (used by `CounterVec`, `GaugeVec`, and `HistogramVec`) and threads it into **`MetricVec`** via **`NewMetricVecWithTTL`**.
- Each child in the vector keeps an atomic **`lastAccessed`** timestamp (Unix ms). **Zero `TTL`** keeps behavior identical to **`NewMetricVec`**.
- **`Collect`** only emits children whose last access is within the TTL window; older children are **omitted** from scrape output.
- **`(*MetricVec).CleanupExpired()`** walks the internal map and **removes** expired children, returning how many were deleted. Callers that care about **memory** (not only hiding series from scrapes) should invoke this on a schedule they control—similar in spirit to how [**statsd_exporter #164**](https://github.com/prometheus/statsd_exporter/pull/164) expires series at the process boundary with a periodic pass.

**“Access”** means resolving the child through the vector API (**`GetMetricWith`**, **`GetMetricWithLabelValues`**, **`With`**, **`WithLabelValues`**, curried variants, etc.). Holding a child handle and only calling **`Add`** / **`Set`** / **`Observe`** on it does **not** refresh TTL unless the instrumentation goes through the **`Vec`** again.

**`SummaryVec`** is unchanged in this PR (still uses **`NewMetricVec`** without TTL). Custom **`MetricVec`** types can opt in by constructing with **`NewMetricVecWithTTL`** and exposing **`CleanupExpired`** if desired.

### Tests

Adds **`vec_test.go`** coverage for counter/gauge/histogram TTL, partial refresh, zero TTL, and **`GetMetricWith`**.

---

Assisted-by: `claude-4.6-opus-high`

### DIFF
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -211,16 +211,17 @@ func (v2) NewCounterVec(opts CounterVecOpts) *CounterVec {
 	if opts.now == nil {
 		opts.now = time.Now
 	}
+	newMetric := func(lvs ...string) Metric {
+		if len(lvs) != len(desc.variableLabels.names) {
+			panic(makeInconsistentCardinalityError(desc.fqName, desc.variableLabels.names, lvs))
+		}
+		result := &counter{desc: desc, labelPairs: MakeLabelPairs(desc, lvs), now: opts.now}
+		result.init(result) // Init self-collection.
+		result.createdTs = timestamppb.New(opts.now())
+		return result
+	}
 	return &CounterVec{
-		MetricVec: NewMetricVec(desc, func(lvs ...string) Metric {
-			if len(lvs) != len(desc.variableLabels.names) {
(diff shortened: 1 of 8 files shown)


kakkoyun (commented):
Please see https://github.com/prometheus/client_golang/issues/1983#issuecomment-4295318325

jpinsonneau:
Closing in favor of https://github.com/prometheus/client_golang/pull/1989

PR #1989, opened 2026-04-23
Add opt-in Vec TTL via MetricVec and TTLRegistry

jpinsonneau:
## Summary
- Opt-in Vec TTL via `*VecOpts.TTL` (Counter/Gauge/Histogram/Summary + MetricVecOpts)
- Decorator wrappers refresh lastAccessed on cached child mutators
- Unified `metricMap` (no `TTLRegistry`, no duplicated `ttlMetricMap`)
- `Registry.Gather` runs `CleanupExpired` only for TTL-enabled collectors
- Documented orphan-after-expiry (same as Delete); no background cleaner

Relates to https://github.com/prometheus/client_golang/issues/1983
Inspired by https://gist.github.com/bwplotka/5c4d8fc904398909d7c4dd10122230f1

Signed-off-by: Julien Pinsonneau <jpinsonn@redhat.com>
Made-with: Cursor


prometheus/ttl_registry_test.go
```diff
@@ -0,0 +1,52 @@
...
+		Name: "ttl_reg_gather",
+		Help: "test",
+	}, []string{"code"})
+
+	vec.WithLabelValues("200").Add(1)
+	if n := collectCount(vec); n != 1 {
+		t.Fatalf("expected 1 metric before sleep, got %d", n)
+	}
+
+	time.Sleep(ttl + 40*time.Millisecond)
```
kakkoyun: We should use `synctest` in this test.

prometheus/vec_test.go
```diff
@@ -1004,3 +1005,160 @@ func benchmarkMetricVecWithLabelValues(b *testing.B, labels map[string][]string)
 		vec.WithLabelValues(values...)
 	}
 }
+
+func collectCount(c Collector) int {
```
kakkoyun, line 1010: Similar to the above test. We should `synctest` where ever, we slept.

prometheus/vec.go
```diff
@@ -304,8 +330,9 @@ func (m *MetricVec) hashLabels(labels Labels) (uint64, error) {
 // metricWithLabelValues provides the metric and its label values for
 // disambiguation on hash collision.
 type metricWithLabelValues struct {
```
kakkoyun, line 368: Even though, it's a lot of code repetition. We should completely separate the path for TTL metrics. I really don't want to touch what is working and introduce additional memory usage even it's as small as an int timestamp value.

client_golang very-well adopted and this is core functionality, anything we introduce to this path will be noticed.

kakkoyun (commented):
Thanks for the PR. This is close to the finish line.

Please see my comments.

prometheus/ttl_registry_test.go
jpinsonneau: Done in https://github.com/prometheus/client_golang/pull/1989/commits/cf4b9e9d79f4542ae16d5b6dfe5d5db5565893c0

prometheus/vec_test.go
jpinsonneau: Done in https://github.com/prometheus/client_golang/pull/1989/commits/cf4b9e9d79f4542ae16d5b6dfe5d5db5565893c0

prometheus/vec.go
jpinsonneau: Sure, I separated it in https://github.com/prometheus/client_golang/pull/1989/commits/cf4b9e9d79f4542ae16d5b6dfe5d5db5565893c0

Thanks for the feedback !

prometheus/vec.go
```diff
@@ -56,6 +58,40 @@ func NewMetricVec(desc *Desc, newMetric func(lvs ...string) Metric) *MetricVec {
 	}
 }
 
+// NewMetricVecWithTTL returns an initialized MetricVec with TTL-based expiration.
+// Children that have not been accessed (via GetMetricWith or GetMetricWithLabelValues)
+// for longer than ttl will be excluded from Collect and can be cleaned up via
+// CleanupExpired. If ttl is 0, this behaves identically to NewMetricVec.
+func NewMetricVecWithTTL(desc *Desc, newMetric func(lvs ...string) Metric, ttl time.Duration) *MetricVec {
```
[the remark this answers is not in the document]
jpinsonneau: https://github.com/prometheus/client_golang/pull/1989/commits/e8d0bdf260625461ba2e11154b73b26699e18866

prometheus/ttl_vec.go
```diff
@@ -0,0 +1,323 @@
...
+	deadline := time.Now().Add(-m.ttl).UnixMilli()
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	var numDeleted int
+	for h, metrics := range m.metrics {
+		remaining := metrics[:0]
+		for i := range metrics {
+			if metrics[i].lastAccessed.Load() >= deadline {
+				remaining = append(remaining, metrics[i])
```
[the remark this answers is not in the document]
jpinsonneau: https://github.com/prometheus/client_golang/pull/1989/commits/e8d0bdf260625461ba2e11154b73b26699e18866

prometheus/ttl_registry.go
```diff
@@ -0,0 +1,124 @@
...
+		Registry: NewRegistry(),
+		ttl:      ttl,
+	}
+}
+
+func (r *TTLRegistry) track(mv *MetricVec) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.vecs = append(r.vecs, mv)
+}
```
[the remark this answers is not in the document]
jpinsonneau: https://github.com/prometheus/client_golang/pull/1989/commits/e8d0bdf260625461ba2e11154b73b26699e18866

jpinsonneau:
@kakkoyun any update here ?

kakkoyun (approved):
LGTM.


I leave the decision to move forward with it to @bwplotka. He is the release shepherd for the next release.

prometheus/counter.go
```diff
@@ -201,6 +201,12 @@ func NewCounterVec(opts CounterOpts, labelNames []string) *CounterVec {
 
 // NewCounterVec creates a new CounterVec based on the provided CounterVecOpts.
 func (v2) NewCounterVec(opts CounterVecOpts) *CounterVec {
+	return newCounterVecWithTTL(opts, 0)
+}
+
+// newCounterVecWithTTL creates a CounterVec. ttl must be >= 0; ttl == 0 disables
+// TTL behavior (identical to NewMetricVec).
+func newCounterVecWithTTL(opts CounterVecOpts, ttl time.Duration) *CounterVec {
```
bwplotka: Are we sure this is what we want? How this will scale when we will have another option, `newCounterVecWithTTLWithX`?

If we have opts, why not using it with TTL for this purpose? Would it make sense to add `ttl` for those cases?

prometheus/vec.go
```diff
@@ -56,6 +58,47 @@ func NewMetricVec(desc *Desc, newMetric func(lvs ...string) Metric) *MetricVec {
...
+		hashAdd:     hashAdd,
+		hashAddByte: hashAddByte,
+	}
+}
+
+// CleanupExpired removes all children that have not been accessed within the
+// configured TTL. It returns the number of children removed. If TTL is not
+// configured (zero), this is a no-op and returns 0.
+func (m *MetricVec) CleanupExpired() int {
+	if m.ttlMap == nil {
```
bwplotka: This feels like an anti-pattern. We have a single type, but we workaround it by having a completely different data-structure inside to mask completely different, mostly duplicated code path (ttlMap vs map). 🤔

prometheus/ttl_registry.go
```diff
@@ -0,0 +1,165 @@
...
+// time-to-live behavior on *Vec children. It embeds a plain *Registry and adds:
+//   - Vec constructors that enable per-child TTL (same semantics as MetricVec
+//     built with NewMetricVecWithTTL).
+//   - Automatic CleanupExpired on all Vecs created through this registry before
+//     each Gather, so memory can be reclaimed even when scrapes are infrequent
+//     (see discussion in https://github.com/prometheus/client_golang/issues/1983).
+//
+// Default prometheus.NewRegistry, NewCounterVec, and Opts are unchanged; use
+// TTLRegistry only when you explicitly opt in to Vec TTL.
+type TTLRegistry struct {
```
bwplotka: Registry is for registering metrics and then gathering registered ones. This API breaks this by having a registry with tons of New* constructors. Are we sure this will not confuse users?

LLM review also captures this well in this point: https://gist.github.com/bwplotka/694d6c4a9be64e629d5c7b2f3541319b#a-coupling-constructors-to-registry

bwplotka: There's indeed also false impression that MustRegister would suddenly start TTL tracking: https://gist.github.com/bwplotka/694d6c4a9be64e629d5c7b2f3541319b#b-incomplete--error-prone-registration-surface

prometheus/vec_test.go
```diff
@@ -1004,3 +1006,187 @@ func benchmarkMetricVecWithLabelValues(b *testing.B, labels map[string][]string)
...
+func TestTTLHistogramVec(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
+		ttl := 100 * time.Millisecond
+		reg := NewTTLRegistry(ttl)
+		vec := reg.NewHistogramVec(HistogramOpts{
+			Name: "test_ttl_histo",
+			Help: "test",
+		}, []string{"status"})
+
+		vec.WithLabelValues("ok").Observe(0.5)
```
bwplotka, line 1090: A common approach is to cache `vec.WithLabelValues("xyz")` and do `cachedVec.Observe(xyz)`.  In the current code we don't 

If we want TTL to work correctly, it `Observe` and `Add` must update last access time, otherwise this feature won't work for typical cases.

See: https://gist.github.com/bwplotka/694d6c4a9be64e629d5c7b2f3541319b#2-correctness-issue-cached-child-metrics-cachedvecobservexyz

prometheus/ttl_registry.go
```diff
@@ -0,0 +1,165 @@
+// Copyright 2014 The Prometheus Authors
```
bwplotka: 2014?

prometheus/vec_test.go
bwplotka: Even worse it creates bugs in this case: https://gist.github.com/bwplotka/694d6c4a9be64e629d5c7b2f3541319b#b-catastrophic-consequence

bwplotka: Although this is already the case DeleteSeries - it must be documented though.

bwplotka (changes requested):
Thanks for this!

I hate to block it but this creates tons of duplicated code that will be hard to maintain. Added comments inline. Especially registry API here is not ideal - registries in client_golang are for registering metrics not creating them.

I tried to play with LLM to play with alternative implementations e.g. https://gist.github.com/bwplotka/5c4d8fc904398909d7c4dd10122230f1 - this is not ideal but perhaps more acceptable.

Notably we could decorate/wrap gauge/counters etc with TTL logic to support a common "cached" metric refs flow. This also does not impact default path which is what @kakkoyun aimed for. 

**I am also worried about orphaned refs here,** common with TTL. While this problem exists with Delete API it's more explicit to users when delete is invoked (user has to do it). Here TTL can kill the metrics any time. LLM gave some ideas to recover but it's a hot path (Inc).

There is also aspect of releasing memory. If someone has TTL to their metrics and expect it to be cleaned, but no-one scrapes then we have leak too. Have we considered time-based TTL cleanup?

prometheus/counter.go
jpinsonneau: Agreed. Dropped the *WithTTL helpers.

TTL now lives on CounterVecOpts / GaugeVecOpts / HistogramVecOpts / SummaryVecOpts / MetricVecOpts, e.g.:

```
  prometheus.V2.NewCounterVec(prometheus.CounterVecOpts{
      CounterOpts:    ...,
      VariableLabels: ...,
      TTL:            time.Minute,
  })
```

See 8d33a81 (and SummaryVec in 649400b).

prometheus/vec.go
jpinsonneau: Agreed. Removed the parallel ttlMetricMap path.

TTL is a field on the existing metricMap; Collect skips expired ttlMetric children and CleanupExpired reclaims them. Default TTL == 0 keeps the previous Vec path (no wrapper, no expiration work).

See 49d10c3 (drop ttl_vec / TTLRegistry) and f366f3e (unified metricMap + MetricVecOpts).

prometheus/ttl_registry.go
jpinsonneau: Agreed. Removed TTLRegistry entirely.

Construction stays on V2.New*Vec(...Opts{TTL: ...}); registration/gathering stay on a normal Registry.

See 49d10c3.

prometheus/vec_test.go
jpinsonneau: Fixed via decorator wrappers (ttlCounter / ttlGauge / ttlHistogram / ttlSummary): mutating methods refresh lastAccessed.

Added TestTTLCachedChildKeepsAlive (and Summary coverage) for the cached-child hot path.

Active Observe/Add/Set on a cached child keeps the series alive, so TTL no longer deletes a hot cached handle out from under the caller.

Idle/orphaned handles after CleanupExpired still detach (same class of issue as Delete); documented and covered by TestTTLOrphanedCachedHandleAfterCleanup.

See 8d33a81 and 649400b.

prometheus/ttl_registry.go
jpinsonneau: Oops, bad copy paste. The new ttl.go uses the proper copyright header :+1:

jpinsonneau:
Addressed feedback and refactored the PR. See PR summary. PTAL

prometheus/histogram.go
```diff
@@ -970,7 +974,7 @@ func (h *histogram) maybeReset(
 	// We are using the possibly mocked h.now() rather than
 	// time.Since(h.lastResetTime) to enable testing.
 	if h.nativeHistogramMinResetDuration == 0 || // No reset configured.
-		h.resetScheduled || // Do not interfere if a reset is already scheduled.
+		h.resetScheduled || // Do not interefere if a reset is already scheduled.
```
jotak, line 977: nit: typo

PR #1996, opened 2026-05-06
exp/api/remote: limit request body size in SnappyDecodeMiddleware

roidelapluie:
Add io.LimitReader to cap compressed body reads at maxDecodedSize bytes, preventing unbounded memory allocation before the decoded size check. Remove the unused bufPool/TeeReader that was copying body bytes into a buffer that was never read.


kakkoyun (approved):

roidelapluie:
@kakkoyun @bwplotka @vesari 

When releasing this, can we credit @hibrian827 (by github handler) for repoting this? thanks!

PR #1999, opened 2026-05-08
examples: improve simple main.go example

dhanudhanushree:
## What

Improve the description comment in `examples/simple/main.go`
to better clarify that the example demonstrates exposing
Prometheus metrics from a Go application.

## Why

The updated wording is more explicit and easier to understand
for readers discovering Prometheus instrumentation examples.


bwplotka (approved):
Thanks!

PR #2000, opened 2026-05-08
promhttp: add regression test for concurrent map writes (#1274)

### LINKED ISSUE #1274: emptyLabels used in instrument_server.go cause concurrent map writes
When using dynamic labels from context on metrics that have no labels the same map is used for all handlers. To illustrate here is the `InstrumentHandlerDuration` function:
```go
func InstrumentHandlerDuration(obs prometheus.ObserverVec, next http.Handler, opts ...Option) http.HandlerFunc {
	hOpts := defaultOptions()
	for _, o := range opts {
		o.apply(hOpts)
	}

	// Curry the observer with dynamic labels before checking the remaining labels.
	code, method := checkLabels(obs.MustCurryWith(hOpts.emptyDynamicLabels()))

	if code {
		return func(w http.ResponseWriter, r *http.Request) {
			now := time.Now()
			d := newDelegator(w, nil)
			next.ServeHTTP(d, r)

			l := labels(code, method, r.Method, d.Status(), hOpts.extraMethods...)
			for label, resolve := range hOpts.extraLabelsFromCtx {
				l[label] = resolve(r.Context())
			}
			observeWithExemplar(obs.With(l), time.Since(now).Seconds(), hOpts.getExemplarFn(r.Context()))
		}
	}

	return func(w http.ResponseWriter, r *http.Request) {
		now := time.Now()
		next.ServeHTTP(w, r)
		l := labels(code, method, r.Method, 0, hOpts.extraMethods...)
		for label, resolve := range hOpts.extraLabelsFromCtx {
			l[label] = resolve(r.Context())
		}
		observeWithExemplar(obs.With(l), time.Since(now).Seconds(), hOpts.getExemplarFn(r.Context()))
	}
}
```
The issue arises when both `code` and `method` are false. Within the `labels` function:
```go
// emptyLabels is a one-time allocation for non-partitioned metrics to avoid
// unnecessary allocations on each request.
var emptyLabels = prometheus.Labels{}

func labels(code, method bool, reqMethod string, status int, extraMethods ...string) prometheus.Labels {
	if !(code || method) {
		return emptyLabels
	}
	labels := prometheus.Labels{}

	if code {
		labels["code"] = sanitizeCode(status)
	}
	if method {
		labels["method"] = sanitizeMethod(reqMethod, extraMethods...)
	}

	return labels
}
```

if both `code` and `method` are false the `emptyLabels` map is returned.
the `emptyLabels` map is only allocated once as can seen by the comment.

After it is returned, the map is then filled with dynamic labels in the following code segment:
```go
		l := labels(code, method, r.Method, 0, hOpts.extraMethods...)
		for label, resolve := range hOpts.extraLabelsFromCtx {
			l[label] = resolve(r.Context())
		}
```

This fills the `emptyLabels` map with values, which is a bug in and of itself, but the real issue arises when there are two handlers that reach this code. When two handlers both try to write to the same map it results in `panic: concurrent map writes`.

Suggested fix (simplest): Let's allocate the empty map every time instead of once, i.e.:
```go
func labels(code, method bool, reqMethod string, status int, extraMethods ...string) prometheus.Labels {
	labels := prometheus.Labels{}

	if code {
		labels["code"] = sanitizeCode(status)
	}
	if method {
		labels["method"] = sanitizeMethod(reqMethod, extraMethods...)
	}

	return labels
}
```

pedrampdd:
Closes #1274.

Without the fix in #1318, instrumenting a metric vec partitioned only by labels supplied via `WithLabelFromCtx` (i.e. with no `code` or `method` label) would cause concurrent requests to share and mutate the single package-level `emptyLabels` map returned by `labels()`, producing `fatal error: concurrent map writes` under load.

#1318 made `labels()` allocate a fresh map on every call, but as @bwplotka noted on that PR (`> Can we add regression test? This will make sure we don't reintroduce same bug.`) no regression test was added. This adds it.

The test wires up `InstrumentHandlerCounter(InstrumentHandlerDuration(...))` around metrics whose only label is supplied via `WithLabelFromCtx`, then drives the chain from 32 goroutines × 200 requests = 6,400 concurrent `ServeHTTP` calls. Either `go test -race` or the runtime's concurrent-map-writes detector will fire if the shared-empty-map regression returns.

### Verification

```
$ go test -v -count=1 -race -run TestInstrumentHandlerLabelFromCtxConcurrent ./prometheus/promhttp
=== RUN   TestInstrumentHandlerLabelFromCtxConcurrent
--- PASS: TestInstrumentHandlerLabelFromCtxConcurrent (0.05s)
PASS
ok  github.com/prometheus/client_golang/prometheus/promhttp  1.077s
```

I also locally re-introduced the pre-#1318 `labels()` (returning the package-level `emptyLabels` map) and confirmed the test fails as expected — the race detector reports a data race on the shared `prometheus.Labels` map and the run aborts. Reverting that one-file change makes the test pass again.

The full package suite stays green with `-race`.

### DIFF
--- a/prometheus/promhttp/instrument_server_test.go
+++ b/prometheus/promhttp/instrument_server_test.go
@@ -19,6 +19,7 @@ import (
 	"log"
 	"net/http"
 	"net/http/httptest"
+	"sync"
 	"testing"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -521,6 +522,59 @@ func TestInterfaceUpgrade(t *testing.T) {
 	if _, ok := d.(http.Hijacker); ok {
 		t.Error("delegator unexpectedly implements http.Hijacker")
 	}
+}
+
+func TestInstrumentHandlerLabelFromCtxConcurrent(t *testing.T) {
+	const (
+		workers           = 32
+		requestsPerWorker = 200
+	)
+
+	counter := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "test_requests_total",
+			Help: "A counter for requests to the wrapped handler.",
+		},
+		[]string{"dyn"},
+	)
+	histogram := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "test_request_duration_seconds",
+			Help: "A histogram of latencies for requests to the wrapped handler.",
+		},
+		[]string{"dyn"},
+	)
+
+	reg := prometheus.NewRegistry()
+	reg.MustRegister(counter, histogram)
+
+	next := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+
+	dyn := WithLabelFromCtx("dyn", func(_ context.Context) string {
+		return "v"
+	})
+
+	chain := InstrumentHandlerCounter(
+		counter,
+		InstrumentHandlerDuration(histogram, next, dyn),
+		dyn,
+	)
+
+	var wg sync.WaitGroup
+	wg.Add(workers)
+	for i := 0; i < workers; i++ {
+		go func() {
+			defer wg.Done()
+			for j := 0; j < requestsPerWorker; j++ {
+				req := httptest.NewRequest(http.MethodGet, "http://example.com/", nil)
+				rec := httptest.NewRecorder()
+				chain.ServeHTTP(rec, req)
+			}
+		}()
+	}
+	wg.Wait()
 }
 
 func ExampleInstrumentHandlerDuration() {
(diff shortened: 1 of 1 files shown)


prometheus/promhttp/instrument_server_test.go
```diff
@@ -523,6 +524,59 @@ func TestInterfaceUpgrade(t *testing.T) {
 	}
 }
 
+func TestInstrumentHandlerLabelFromCtxConcurrent(t *testing.T) {
```
bwplotka, line 528:
```suggestion
// Regression test against https://github.com/prometheus/client_golang/pull/1318
func TestInstrumentHandlerLabelFromCtxConcurrent(t *testing.T) {
```

bwplotka (approved):
Thanks!

pedrampdd:
No problem!

bwplotka (approved):

PR #2002, opened 2026-05-09
promhttp: ignore 1xx informational status codes in WriteHeader

### LINKED ISSUE #1772: `responseWriterDelegator` incorrectly handles `100 Continue` response

nghiack7:
## Problem

Fixes #1772

`responseWriterDelegator.WriteHeader` records **any** status code, including `100 Continue`, as the final response status and calls `observeWriteHeader` for it.

A handler that explicitly signals `100 Continue` before writing the body ends up reporting `100` as its final status code instead of `200`, corrupting Prometheus metrics.

## Root cause

The delegator did not mirror what `net/http`'s own `responseWriter` does: [server.go](https://github.com/golang/go/blob/go1.24.1/src/net/http/server.go#L1212-L1228) only records the status and sets `wroteHeader` for non-informational responses.

## Fix

Return early for 1xx codes: delegate to the inner `ResponseWriter` (so the upstream still receives the informational response) but do **not** update `wroteHeader`, `status`, or call `observeWriteHeader`.

```go
// Before
    if r.observeWriteHeader != nil && !r.wroteHeader {
        r.observeWriteHeader(code)   // called for 100 Continue ← bug
    }
    r.status = code                  // set to 100 ← bug
    r.wroteHeader = true
    r.ResponseWriter.WriteHeader(code)
}

// After
    if code >= 100 && code < 200 {
        r.ResponseWriter.WriteHeader(code)  // forward but don't record
        return
    }
    // ... existing logic unchanged
}
```

## Tests

Added `TestResponseWriterDelegatorInformationalStatusCode` that:
- Sends `100 Continue` and asserts `wroteHeader` is not set, `status` is not 100, and `observeWriteHeader` is not called
- Then sends `200 OK` and asserts all three are correctly recorded

### DIFF
--- a/prometheus/promhttp/delegator.go
+++ b/prometheus/promhttp/delegator.go
@@ -53,6 +53,14 @@ func (r *responseWriterDelegator) Written() int64 {
 }
 
 func (r *responseWriterDelegator) WriteHeader(code int) {
+	// 1xx informational responses (e.g. 100 Continue) are not the final
+	// status code.  Delegate them to the underlying ResponseWriter but do
+	// not record them as the response status or call observeWriteHeader,
+	// mirroring the behaviour of net/http's own responseWriter.
+	if code >= 100 && code < 200 {
+		r.ResponseWriter.WriteHeader(code)
+		return
+	}
 	if r.observeWriteHeader != nil && !r.wroteHeader {
 		// Only call observeWriteHeader for the 1st time. It's a bug if
 		// WriteHeader is called more than once, but we want to protect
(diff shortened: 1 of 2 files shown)


nghiack7:
Addressed the two CI failures:

1. **Lint / gofmt**: Fixed alignment in the `trackingResponseWriter` stub in the test file — gofmt prefers different column widths than I had.
2. **DCO**: Added `Signed-off-by: Nghĩa Nguyễn Ngọc <nghiann@firegroup.io>` to the commit per the project's contribution requirement.

Both fixes are in the latest push. All `promhttp` tests still pass.

nghiack7:
Friendly ping — all CI checks are green ✅ (15/15 passed, DCO signed). Happy to revise anything.

PR #2005, opened 2026-05-13
promhttp: don't panic when instrumenting with non-exemplar observers

### LINKED ISSUE #1258: Can't use an ObserverVec with promhttp.InstrumentHandlerDuration()
I'm trying to instrument my `http.Handler` with `promhttp.InstrumentHandlerDuration`, like so:

```go
duration :=  prometheus.NewSummaryVec(
	prometheus.SummaryOpts{
		Name:       "request_duration_seconds",
		Help:       "Request duration",
		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
	},
	[]string{"code", "method"},
)

// ...
http.Handle("/", promhttp.InstrumentHandlerDuration(duration, handler))
```

However, this causes a runtime panic because `prometheus.summary` doesn't conform to the `ExemplarObserver` interface.

```
2023/04/22 14:47:25 http: panic serving 127.0.0.1:56256: interface conversion: *prometheus.summary is not prometheus.ExemplarObserver: missing method ObserveWithExemplar
```

It's not clear to me how to resolve this issue - unfortunately the documentation for `InstrumentHandlerDuration`  doesn't discuss the matter, nor does the example provided (I'd be curious if the example as written even works!)

spor3006:
## What

`InstrumentHandlerDuration` and `InstrumentHandlerCounter` panic at request time when their `ObserverVec`/`CounterVec` is backed by a type that does not implement `ExemplarObserver`/`ExemplarAdder`. The most common trigger is a `SummaryVec`, which is a valid `prometheus.ObserverVec` but whose underlying `*summary` does not implement `ObserveWithExemplar` — summaries cannot carry exemplars in the Prometheus exposition format.

Reproduction from #1258:

```go
duration := prometheus.NewSummaryVec(
    prometheus.SummaryOpts{
        Name:       "request_duration_seconds",
        Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
    },
    []string{"code", "method"},
)
http.Handle("/", promhttp.InstrumentHandlerDuration(duration, handler))
```

panics on the first request with:

```
interface conversion: *prometheus.summary is not prometheus.ExemplarObserver: missing method ObserveWithExemplar
```

## Why

The unsafe type assertion was inconsistent with the rest of the codebase. `Timer.ObserveDurationWithExemplar` (`prometheus/timer.go:72`) has handled the same situation correctly for years using a safe-cast + fallback:

```go
eo, ok := t.observer.(ExemplarObserver)
if ok && exemplar != nil {
    eo.ObserveWithExemplar(d.Seconds(), exemplar)
    return d
}
if t.observer != nil {
    t.observer.Observe(d.Seconds())
}
```

The two helpers in `promhttp/instrument_server.go` (`observeWithExemplar`, `addWithExemplar`) did `obs.(prometheus.ExemplarObserver)` without the `ok` check, turning a documented exemplar-not-supported case into a runtime panic in the request path.

## Changes

1. `observeWithExemplar` / `addWithExemplar` use the safe-cast + fallback pattern. If the observer/counter implements the exemplar interface, the exemplar is attached; otherwise the exemplar is silently dropped and the value is recorded with the plain `Observe`/`Add` path. **No public API change.**
2. Doc comments on both helpers describe the fallback explicitly and cross-reference `Timer.ObserveDurationWithExemplar` for consistency.
3. CHANGELOG `[BUGFIX]` entry under Unreleased.

## Tests

Three regression tests in `instrument_server_test.go`:

- `TestMiddlewareAPI_SummaryWithExemplars` — exact issue reproduction: `InstrumentHandlerDuration(summaryVec, …, WithExemplarFromContext(…))` + one HTTP request. Fails on `main` with the panic; passes here.
- `TestObserveWithExemplar_NonExemplarObserverFallsBack` — unit-level contract for the helper.
- `TestAddWithExemplar_NonExemplarAdderFallsBack` — same for the counter helper.

All three are stable under `go test -count=10 -race`.

## Verification

- `go test ./prometheus/promhttp/...` — passes (existing + new).
- `go test -race ./prometheus/promhttp/...` — passes.
- `go vet ./prometheus/promhttp/...` — clean.
- `gofmt` — clean.

Note: `go test ./prometheus/collectors/...` fails on this branch *and* on `main` because Go 1.25 added runtime metrics (`htmlmetacontenturlescape`, `httpcookiemaxnum`, `urlmaxqueryparams`) that aren't in the `expectedMetrics` fixture. Pre-existing, unrelated to this fix.

## Trade-offs considered

- **Silently dropping the exemplar.** The helpers have no error return and `promhttp` has no logger. Silent fallback matches `Timer.ObserveDurationWithExemplar`. The new behavior is documented in the godoc.
- **Making `*summary` implement `ExemplarObserver`.** Rejected: the Prometheus exposition format does not support exemplars on summaries, so accepting them would either silently drop at write time (worse — bug moves to a less visible site) or fabricate a different code path. Keeping the interface as the contract is the right design.

Closes #1258.

/cc @bwplotka @kakkoyun @vesari

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * [FEATURE] HTTP handlers created by `promhttp` package now support metrics filtering by providing one or more `name[]` query parameters. The default behavior when none are provided remains the same, returning all metrics. #1925
+* [BUGFIX] promhttp: `InstrumentHandlerDuration` and `InstrumentHandlerCounter` no longer panic when given an observer/counter that does not implement `ExemplarObserver`/`ExemplarAdder` (e.g. a `SummaryVec`). The exemplar is dropped and the value is recorded via the plain `Observe`/`Add` path, matching the safe-cast already used by `Timer.ObserveDurationWithExemplar`. #2005
 
 ## Unreleased `exp` module
 
--- a/prometheus/promhttp/instrument_server.go
+++ b/prometheus/promhttp/instrument_server.go
@@ -28,24 +28,36 @@ import (
 // magicString is used for the hacky label test in checkLabels. Remove once fixed.
 const magicString = "zZgWfBxLqvG8kc8IMv3POi2Bb0tZI3vAnBx+gBaFi9FyPzB/CzKUer1yufDa"
 
-// observeWithExemplar is a wrapper for [prometheus.ExemplarAdder.ExemplarObserver],
-// which falls back to [prometheus.Observer.Observe] if no labels are provided.
+// observeWithExemplar records val on obs. If labels is non-nil and obs
+// implements [prometheus.ExemplarObserver], the exemplar is attached via
+// ObserveWithExemplar; otherwise the exemplar is dropped and the value is
+// recorded with a plain [prometheus.Observer.Observe]. This mirrors the
+// safe-cast pattern in [prometheus.Timer.ObserveDurationWithExemplar] and
+// ensures we never panic when callers pass an ObserverVec backed by a
+// summary, which cannot carry exemplars in the Prometheus exposition format.
 func observeWithExemplar(obs prometheus.Observer, val float64, labels map[string]string) {
-	if labels == nil {
-		obs.Observe(val)
-		return
+	if labels != nil {
+		if eo, ok := obs.(prometheus.ExemplarObserver); ok {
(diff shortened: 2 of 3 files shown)


bwplotka (approved):
Thanks! Solid fix.

PR #2006, opened 2026-05-16
prometheus: add Registry.DisableHelpTextValidation opt-in

### LINKED ISSUE #1820: Metric help text validation causes issues with multiple versions
There was a [change](https://github.com/prometheus/client_golang/pull/1583/commits/8514bd76953accf6aa84f1e2dd875306964e8787) back in August 2024 that added a (seemingly innocent) period to some help text.

Unfortunately with the [strict help text match checking](https://github.com/prometheus/client_golang/blob/main/prometheus/registry.go#L640) that happens in the client that makes interoperability between different golang applications utilizing different versions of the client problematic (as the validation fails).

I'm not close enough to know why validation of the help text is necessary, but are you open to a PR to either remove it or have an option to disable strict checking?

spor3006:
## What

`Registry.Gather` currently returns an error when two collectors emit metrics with the same fully-qualified name but slightly different help text:

```
collected metric foo {…} has help "X." but should have "X"
```

The most common trigger is two libraries — or two versions of the same library — registering compatible metrics into the same registry. A trivial difference like a help string gaining or losing a trailing period across a `client_golang` upgrade turns a metadata mismatch into a hard `/metrics` outage for the whole binary (see #1820).

The Prometheus server itself is lenient about help-text drift on scrape, so this strict check makes inter-library interop harder than it needs to be.

## Design

This PR is **strictly additive** to the public API and leaves the default behavior unchanged. It adds one new method:

```go
reg.DisableHelpTextValidation()
```

When the toggle is enabled, `processMetric` aggregates later metrics into the family the first metric created and keeps the first help text seen. Other validation (type checks, suffix collisions, dim-hash consistency at `Register` time, pedantic per-`Desc` checks) is unchanged.

### Why opt-in rather than relaxing the default

The strict default has value: it surfaces accidental help-text changes during development and testing. Silently overwriting a refactored help string under the user is a debuggability footgun. Forcing callers to consciously opt in to the lenient semantics keeps that safety net for everyone who isn't fighting the interop problem.

### Pedantic registries

Calling `DisableHelpTextValidation` on a Registry returned by `NewPedanticRegistry` has **no effect**. The pedantic caller has opted into the strictest set of checks; mixing that with lenient help validation would be internally contradictory. The override is computed once per `Gather` under the existing `RLock`:

```go
helpTextValidationDisabled := r.helpTextValidationDisabled && !r.pedanticChecksEnabled
```

### What this PR does NOT change

- `Gatherers.Gather` (cross-Gatherer merge) and `checkDescConsistency` (pedantic per-`Desc` check) are intentionally left strict. They cover different concerns from `processMetric`'s per-fqName check, and changing them is out of scope.
- `Register()` still rejects mismatched dim-hashes (which includes help text) at registration time for checked collectors. The toggle only affects what `Gather` does with metrics that nevertheless reach `processMetric` with mismatched help — in practice, the unchecked-collector path.

## Tests

Four new tests in `prometheus/registry_test.go`, all stable under `go test -count=10 -race`:

| Test | What it pins |
|---|---|
| `TestGatherRejectsHelpTextMismatchByDefault` | Backwards compat: default Registry still errors on help mismatch. |
| `TestGatherAcceptsHelpTextMismatchWhenDisabled` | Regression for #1820: with the toggle on, Gather succeeds and aggregates. |
| `TestGatherWithDisabledHelpTextValidationKeepsFirstHelp` | Pins the documented "first help text wins" guarantee using a single collector for deterministic emit order. |
| `TestPedanticRegistryOverridesHelpTextValidationToggle` | Pedantic mode ignores the toggle. |

Two unexported test helpers are added:

- `helpTextCollector` — unchecked collector emitting one metric with configurable fqName/help/label, for the two-collectors-into-one-registry scenarios.
- `orderedHelpTextCollector` — single unchecked collector emitting multiple metrics with deterministic source order, for the "first wins" assertion (two separate collectors would race because the registry fans collection out across goroutines).

## Verification

- `go test ./prometheus/...` — passes (existing + new).
- `go test -race ./prometheus/...` — passes.
- `go vet ./prometheus/...` — clean.
- `gofmt` — clean.

Note: `prometheus/collectors` and `TestGoCollector*` continue to fail on this branch *and* on `main` because Go 1.25 added/removed runtime metrics (`htmlmetacontenturlescape`, `httpcookiemaxnum`, `urlmaxqueryparams`) that aren't in the `expectedMetrics` fixture. Pre-existing, unrelated to this fix.

## Trade-offs considered

- **Relax the default instead.** Rejected on the first iteration after a self-review: a silent default-behavior change to a registry-level contract risks breaking callers who rely on the existing error as a configuration-bug detector, and removes a useful safety net during local development.
- **Add a registry-package `Opts` struct / new constructor (`NewLenientRegistry`).** Rejected as heavier API surface for a single boolean. A method on `*Registry` matches the existing `NewPedanticRegistry`-style pattern: one entry point for the strict mode, one method to opt into the lenient mode.
- **Log a warning instead of silent acceptance.** Rejected because the `prometheus` package does not have a logger in scope here, and threading one through purely for this case would be a much larger change. Documented as the explicit trade-off in the godoc.
- **Also relax `Gatherers.Gather` / `checkDescConsistency`.** Out of scope. The motivating use case is single-registry interop; cross-Gatherer merging is a separate user contract. Happy to file a follow-up if maintainers think the same reasoning applies there.

Closes #1820.

/cc @bwplotka @kakkoyun @vesari

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * [FEATURE] HTTP handlers created by `promhttp` package now support metrics filtering by providing one or more `name[]` query parameters. The default behavior when none are provided remains the same, returning all metrics. #1925
+* [FEATURE] prometheus: `Registry` exposes a new `DisableHelpTextValidation` method to opt out of the strict per-fqName help-text consistency check during `Gather`. Useful when multiple libraries (or different versions of the same library) register compatible metrics into the same registry. Default behavior is unchanged, and `NewPedanticRegistry` always overrides the toggle. #2006
 
 ## Unreleased `exp` module
 
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -271,12 +271,38 @@ func (errs MultiError) MaybeUnwrap() error {
 // Registry implements Collector to allow it to be used for creating groups of
 // metrics. See the Grouping example for how this can be done.
 type Registry struct {
-	mtx                   sync.RWMutex
-	collectorsByID        map[uint64]Collector // ID is a hash of the descIDs.
-	descIDs               map[uint64]struct{}
-	dimHashesByName       map[string]uint64
-	uncheckedCollectors   []Collector
-	pedanticChecksEnabled bool
+	mtx                        sync.RWMutex
+	collectorsByID             map[uint64]Collector // ID is a hash of the descIDs.
+	descIDs                    map[uint64]struct{}
+	dimHashesByName            map[string]uint64
+	uncheckedCollectors        []Collector
+	pedanticChecksEnabled      bool
+	helpTextValidationDisabled bool
+}
+
+// DisableHelpTextValidation opts this Registry out of the strict check that
+// every metric sharing a fully-qualified name also share the same help text.
+// When disabled, Gather keeps the help text from the first metric it sees for
+// a given fqName and aggregates later metrics into the same family even if
+// their help differs.
+//
+// This is useful when multiple libraries — or different versions of the same
+// library — register compatible metrics into a shared registry. A trivial
+// difference such as a trailing period gained across a client_golang upgrade
+// can otherwise turn a metadata mismatch into a hard /metrics outage. The
+// Prometheus server itself applies the same lenient semantics on scrape (see
+// https://github.com/prometheus/client_golang/issues/1820).
+//
+// The strict default is recommended for development and testing, where it
+// surfaces accidental help-text changes early. Calling this method on a
+// Registry returned by NewPedanticRegistry has no effect: pedantic mode
+// always enforces help-text consistency. The change applies to subsequent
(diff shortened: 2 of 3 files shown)


bwplotka:
Thanks for this! 

Not sure if I agree with the implementation --  but I do know this would help Otel SDK Go (cc @dashpole) as now they workaround this before hand https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/prometheus/exporter.go#L294-L302

Perhaps one way of doing this would be change `DisableHelpTextValidation` with `WithOnHelpTextMismatch(onMismatchFunc func (old, new string) (string, error))`

Also I wonder, should it be method or constructor variadic argument in `NewRegistry()` 🤔

PR #2009, opened 2026-05-28
docs: fix godoc indentation and typos in timer.go and wrap.go

immanuwell:
Three small doc fixes:

- timer.go: the TimeMeWithExemplar godoc example has extra tabs, renders with wrong indentation (extra level vs the TimeMe example above it)
- timer.go: "casted" -> "cast", cast is an irregular verb, no -ed needed
- wrap.go: "errer" typo -> "error"

Reproduce the indentation issue:

    go doc github.com/prometheus/client_golang/prometheus.NewTimer

second example block shows up with extra tab indent before this fix ngl

cc @vesari @bwplotka @kakkoyun

### DIFF
--- a/prometheus/timer.go
+++ b/prometheus/timer.go
@@ -37,10 +37,10 @@ type Timer struct {
 // or
 //
 //	func TimeMeWithExemplar() {
-//		    timer := NewTimer(myHistogram)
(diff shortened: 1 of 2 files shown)


bwplotka (approved):
Thanks!

PR #2012, opened 2026-05-30
fix: use keyed fields in SamplePair struct literals in api_test.go

immanuwell:
`go vet ./...` fails on 15 unkeyed `model.SamplePair` struct literals in `TestSamplesJSONSerialization`. The bench test in the same package already uses keyed syntax, so this just brings `api_test.go` in line with it.

Reproduce:

```
go vet ./api/prometheus/v1/...
# api/prometheus/v1 [api/prometheus/v1.test]
api/prometheus/v1/api_test.go:1622:14: github.com/prometheus/common/model.SamplePair struct literal uses unkeyed fields
...
```

Fix: add `Timestamp:` and `Value:` field names to all 15 literals. No behavior change, tests still pass.

@bwplotka @beorn7

### DIFF
--- a/api/prometheus/v1/api_test.go
+++ b/api/prometheus/v1/api_test.go
@@ -1619,63 +1619,63 @@ func TestSamplesJSONSerialization(t *testing.T) {
 		expected string
 	}{
 		{
-			point:    model.SamplePair{0, 0},
+			point:    model.SamplePair{Timestamp: 0, Value: 0},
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):

kakkoyun:
LGTM. However, mac os tests are failing. We need to check.

kakkoyun (approved):

PR #2018, opened 2026-06-04
Add registry option to skip metric sorting

### LINKED ISSUE #483: Consider optional disabling of metric sorting
With extremely large exposition (example: kube-state-metrics on a large cluster will expose hundreds of MiB of metrics data), sorting of metrics becomes a quite significant computational burden. As metrics sorting is not required for valid expositions, a registry could provide an option to not sort after gathering metrics.

ibobgunardi:
Fixes #483.

Adds `RegistryOpts.DisableMetricSorting` with option-aware constructors for vanilla and pedantic registries. The existing constructors keep the current sorted output; the new option skips the final MetricFamily/Metric sort for callers that can accept unspecified ordering.

Tests:
- `go test ./prometheus -run TestRegistryDisableMetricSorting -count=1`
- `go test ./prometheus -run 'TestRegistry|TestRegisterUnregisterCollector|TestCollectorOnMetricPanic|TestCheckMetricConsistency|TestGatherDoesNotLeakGoroutines' -count=1`
- `go test ./prometheus/internal -count=1`

Note: `go test ./prometheus -count=1` fails locally on Go 1.25.1 because the expected Go runtime metric set has 180 entries while this toolchain exposes 177.

### DIFF
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -65,10 +65,25 @@ func init() {
 // NewRegistry creates a new vanilla Registry without any Collectors
 // pre-registered.
 func NewRegistry() *Registry {
+	return NewRegistryWithOptions(RegistryOpts{})
+}
+
+// RegistryOpts configures a Registry.
+type RegistryOpts struct {
+	// DisableMetricSorting skips sorting MetricFamilies and their Metrics in
+	// Gather. The gathered MetricFamilies remain valid for exposition, but their
+	// order is unspecified.
+	DisableMetricSorting bool
+}
+
+// NewRegistryWithOptions creates a new vanilla Registry without any Collectors
+// pre-registered.
+func NewRegistryWithOptions(opts RegistryOpts) *Registry {
 	return &Registry{
-		collectorsByID:  map[uint64]Collector{},
-		descIDs:         map[uint64]struct{}{},
-		dimHashesByName: map[string]uint64{},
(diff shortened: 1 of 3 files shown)


prometheus/registry.go
```diff
@@ -448,6 +471,7 @@ func (r *Registry) Gather() ([]*dto.MetricFamily, error) {
 		wg                  sync.WaitGroup
 		safeErrs            = &SafeMultiError{} // To collect errors in a threadsafe way
 		registeredDescIDs   map[uint64]struct{} // Only used for pedantic checks
+		disableSorting      = r.disableMetricSorting
```
kakkoyun: nit: we don't need this additional variable.

kakkoyun (approved):
LGTM. 

Really solid PR.

kakkoyun:
I will wait for another maintainers' opinion before merging it @vesari @bwplotka

bwplotka (changes requested):
Thanks for this, solid PR, but we add bunch of new APIs and increase amount of options. 

Do we know, for what exactly benefit? Can we add a quick micro benchmark to see how much we would really save in terms of time/memory?

The issue is https://github.com/prometheus/client_golang/issues/483 old - notably KSM moved to a hand coded exposition, so we don't need to support it? Curious how much benefit we have for 100 metric families vs 10k metric families.

PR #2019, opened 2026-06-07
fix metric vec label input aliasing

### LINKED ISSUE #1951: bug: metrics with no label constraint are exposed to potential race conditions
Latest version and `main` are impacted and any releases since Jun 27, 2023 (bug introduced in #1296).

Both `func constrainLabels` and `func constrainLabels` return the reference to the user's label list or map (when there's no label constraints), instead of creating a new copy (this is the root cause of the potential race conditions):
https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L665-L669

https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L684-L688

Creating a handful of potential race conditions in functions utilizing these two functions like `func (m *MetricVec) GetMetricWith`:
https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L238-L248

At best, when a race conditions is encountered, metrics will be unintentionally written with incorrect labels. At worst:
1. potentially crashing the program if the race condition is hit just right, triggering a `fatal error: concurrent map iteration and map write`
2. metric labels values could change to a previous written value set, triggering `collected before with the same name and label values` errors

For example, this is a reproducible race condition in  `func (m *MetricVec) GetMetricWith`:
1. initial call [`myCounterVec.With(myLabels)`](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/counter.go#L293C1-L293C41) with `myLabels := map[string]string{"foo": "bar"}` and assuming there no label constraints
   1 which will call  [`c, err := v.GetMetricWith(labels)`](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/counter.go#L294) and will make these internal calls
      1. [`labels, closer := constrainLabels(m.desc, labels)`](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L239)
         1. `labels` still points to the same map as `myLabels` due to there being [no label constraints being used](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L666-L669)
      2. [`h, err := m.hashLabels(labels)`](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L242)
         2. `h` being the hash of the labels values of `map[string]string{"foo": "bar"}`, i.e.: hash of `"bar"`
      3. [`return m.getOrCreateMetricWithLabels(h, labels, m.curry), nil`](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L247C2-L247C63), will make the following calls
         1. assuming a metric didn't exist for the hash value `hash`, then [`metric, ok := m.getMetricWithHashAndLabels(hash, labels, curry)`](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L526) will results in `ok := False`
         2. leading to the creation of a new metric [`lvs := extractLabelValues(m.desc, labels, curry)`](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L528)
            1. since `labels` still points to the same map as `myLabels`, then a value change from `"foo": "bar"` to `"foo": "cake"` in `myLabels` will be reflected in `labels`
         3. [`m.metrics[hash] = append(m.metrics[hash], metricWithLabelValues{values: lvs, metric: metric})`](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L530)
            1. assuming that label value change occurred in `myLabels` right before [`lvs := extractLabelValues(m.desc, labels, curry)`](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/vec.go#L528) was executed, then you'll have a miss match between `hash` and the new metric `metric = m.newMetric(lvs...)`

The miss match between `hash` and the new metric, will either:
1. assuming the new metric previously didn't exist
   1. potentially crash the program due to a `fatal error: concurrent map iteration and map write` if the race condition is hit just right
   2. the new metric will unintentionally be written with the wrong labels
2. assuming it did exist
   1. again, potentially triggering a `fatal error: concurrent map iteration and map write` crash
   2. [`mfs, done, err := reg.Gather()`](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/promhttp/http.go#L183) returns a `collected before with the same name and label values` error due to its `checkMetricConsistency` call and will either, depending the [prom http setting](https://github.com/prometheus/client_golang/blob/bf37be4fecc0a3d89f980252e206b67806990e56/prometheus/promhttp/http.go#L190-L202)
      1. PanicOnError - panic due to the error
      2. HTTPErrorOnError -  return 500s with the body as the error when metrics are scraped 
      3. ContinueOnError - ignores the error (besides logging it which is all three cases do) or report the error if no metrics have been gathered

Saithej2k:
## Summary

- clone caller-provided `Labels` maps when `MetricVec` has no label constraints
- clone caller-provided label-value slices in the no-constraint `GetMetricWithLabelValues` path
- add regression tests that mutate caller-owned inputs between hashing and metric creation

## Why

When no label constraints are configured, `constrainLabels` and `constrainLabelValues` returned the caller-owned map or slice directly. `MetricVec` hashes those values first and can then read the same caller-owned input again while creating or matching a metric. If the caller mutates that input concurrently, the internal reads can race and can associate the hash with mutated label values.

Fixes #1951.

## Validation

- `go test -race ./prometheus -run 'TestMetricVecGetMetricWith.*Copies' -count=1`
- `go test -race ./prometheus -run 'TestMetricVec' -count=1`

I also tried `go test -race ./prometheus -count=1`, but my local Go runtime metrics differed from the checked-in Go collector fixtures, causing unrelated `go_collector_latest_test.go` failures.

### DIFF
--- a/prometheus/vec.go
+++ b/prometheus/vec.go
@@ -15,6 +15,8 @@ package prometheus
 
 import (
 	"fmt"
+	"maps"
+	"slices"
 	"sync"
 
 	"github.com/prometheus/common/model"
@@ -667,7 +669,7 @@ var labelsPool = &sync.Pool{
 func constrainLabels(desc *Desc, labels Labels) (Labels, func()) {
 	if len(desc.variableLabels.labelConstraints) == 0 {
 		// Fast path when there's no constraints
-		return labels, func() {}
+		return maps.Clone(labels), func() {}
 	}
 
 	constrainedLabels := labelsPool.Get().(Labels)
@@ -686,7 +688,7 @@ func constrainLabels(desc *Desc, labels Labels) (Labels, func()) {
 func constrainLabelValues(desc *Desc, lvs []string, curry []curriedLabelValue) []string {
 	if len(desc.variableLabels.labelConstraints) == 0 {
 		// Fast path when there's no constraints
-		return lvs
+		return slices.Clone(lvs)
 	}
 
 	constrainedValues := make([]string, len(lvs))
--- a/prometheus/vec_test.go
+++ b/prometheus/vec_test.go
@@ -446,6 +446,79 @@ func TestCounterVecEndToEndWithCollision(t *testing.T) {
 	}
 }
 
+func TestMetricVecGetMetricWithCopiesLabels(t *testing.T) {
+	vec := NewCounterVec(
+		CounterOpts{
+			Name: "test",
+			Help: "helpless",
+		},
+		[]string{"l"},
+	)
+
+	labels := Labels{"l": "bar"}
+	mutated := false
+	vec.hashAdd = func(h uint64, s string) uint64 {
+		if s == "bar" && !mutated {
+			labels["l"] = "cake"
+			mutated = true
+		}
+		return hashAdd(h, s)
+	}
+
+	metric, err := vec.GetMetricWith(labels)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !mutated {
+		t.Fatal("test hook was not reached")
+	}
+	assertMetricLabelValue(t, metric, "bar")
+}
+
+func TestMetricVecGetMetricWithLabelValuesCopiesLabelValues(t *testing.T) {
+	vec := NewCounterVec(
+		CounterOpts{
+			Name: "test",
+			Help: "helpless",
+		},
+		[]string{"l"},
+	)
+
+	labelValues := []string{"bar"}
+	mutated := false
+	vec.hashAdd = func(h uint64, s string) uint64 {
+		if s == "bar" && !mutated {
+			labelValues[0] = "cake"
+			mutated = true
+		}
+		return hashAdd(h, s)
+	}
+
+	metric, err := vec.GetMetricWithLabelValues(labelValues...)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


bwplotka:
Thanks but we prefer performance vs making it ultra safe for all cases. However some commentary would be useful. See https://github.com/prometheus/client_golang/issues/1951#issuecomment-4763695341

PR #2022, opened 2026-06-17
Relax suffix collision check for native-only histograms

npazosmendez:


krajorama (commented):
It seems we cannot do this as in text formats native only histogram do show up with _count, _sum and _bucket (with the +Inf le only).

PR #2029, opened 2026-06-26
Recover panics from wrapped collectors (match Gather safeCollect)

mahirhir:
`wrappingCollector.Collect`/`.Describe` (`prometheus/wrap.go`) run the inner collector in a child goroutine, so the panic recovery #1961 added to `Registry.Gather` (`safeCollect`) can't catch it — recover is per-goroutine. A collector registered via `WrapRegistererWith` (the standard const-label/prefix API) that panics still crashes the whole process.

#1961 hardened the unwrapped path; the wrapped path was the symmetric leftover. The fix applies the same recovery to the wrapped collect/describe. Verified both ways: old panics crash the test binary (exit 1); fixed surfaces them as a Gather error. Full `./prometheus/` suite passes, 0 regression.

### DIFF
--- a/prometheus/wrap.go
+++ b/prometheus/wrap.go
@@ -15,6 +15,7 @@ package prometheus
 
 import (
 	"fmt"
+	"runtime"
 	"sort"
 
 	"github.com/prometheus/client_golang/prometheus/internal"
@@ -160,8 +161,23 @@ type wrappingCollector struct {
 func (c *wrappingCollector) Collect(ch chan<- Metric) {
 	wrappedCh := make(chan Metric)
 	go func() {
+		defer close(wrappedCh)
+		defer func() {
+			// The wrapped Collect runs in its own goroutine, so a panic
+			// here cannot be recovered by safeCollect in the Registry (a
+			// recover only catches panics from the same goroutine). Recover
+			// it locally and surface it as an invalid metric, mirroring how
+			// safeCollect handles panics from unwrapped collectors. Without
+			// this, a single misbehaving collector wrapped via
+			// WrapRegistererWith crashes the whole process during Gather.
+			if r := recover(); r != nil {
+				buf := make([]byte, 64<<10) // 64 KB
+				n := runtime.Stack(buf, false)
+				err := fmt.Errorf("prometheus collector panic recovered: type=%T: error=%v\nstack trace=%s", c.wrappedCollector, r, buf[:n])
+				wrappedCh <- NewInvalidMetric(NewInvalidDesc(err), err)
+			}
+		}()
 		c.wrappedCollector.Collect(wrappedCh)
-		close(wrappedCh)
 	}()
 	for m := range wrappedCh {
 		ch <- &wrappingMetric{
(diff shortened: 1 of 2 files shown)


mahirhir:
This has had no review since June, so here is the case in one place. The whole argument is inside this package, so nothing here needs to be taken on my word.

`Gather` already protects against a collector that panics:

```go
// prometheus/registry.go:604
func safeCollect(c Collector, ch chan<- Metric) (err error) {
	defer func() {
		if r := recover(); r != nil {
			buf := make([]byte, 64<<10) //	64 KB
			n := runtime.Stack(buf, false)
			err = fmt.Errorf("prometheus collector panic recovered: type=%T: error=%v\nstack trace=%s", c, r, buf[:n])
			ch <- NewInvalidMetric(NewInvalidDesc(err), err)
		}
	}()
	c.Collect(ch)
```

called from `registry.go:479` and `:481`. There is a test for it, `TestCollectorOnMetricPanic`, so this is settled behaviour rather than an accident.

A collector reached through `WrapRegistererWith` does not get that protection:

```console
$ grep -c 'recover()' prometheus/wrap.go
0
```

```go
// prometheus/wrap.go:162
go func() {
	c.wrappedCollector.Collect(wrappedCh)
	close(wrappedCh)
}()

// prometheus/wrap.go:177
go func() {
	c.wrappedCollector.Describe(wrappedCh)
	close(wrappedCh)
}()
```

The reason is not that anyone forgot. `safeCollect` runs on the registry's goroutine and the wrapped collector runs on a new one, and a `recover` only catches a panic raised in the goroutine that deferred it. So the existing guard is structurally unable to reach the wrapped call, no matter how it is written.

The effect is that the same misbehaving collector is a reported error when registered directly and a process crash when registered through a wrapper. That difference is invisible at the call site, since wrapping is usually added for labels or a prefix rather than as a change in failure semantics.

This PR puts the `safeCollect` body inside those two goroutines. It is deliberately a copy of the existing recovery rather than a new idea, so the two paths report a panic identically: same 64 KB stack buffer, same message format, same `NewInvalidMetric(NewInvalidDesc(err), err)`. `Describe` surfaces its panic as an invalid `Desc` carrying the error, which is how invalid descriptors already travel.

CI as this repository reports it:

```
Linux (stable) / Linux (oldstable) / Linux arm64  = pass
Windows (stable) / Windows (oldstable)            = pass
Analyze (go) / CodeQL / Validate / DCO            = pass
mergeable                                         = MERGEABLE
```

Ten of ten green, including the two probe tests in this PR. Both register a collector that panics through `WrapRegistererWith`: one asserts `Gather` returns an error containing `panic recovered` rather than taking the process down, the other asserts the same for `Register`, since registering a checked collector calls `Describe`.

One thing I am not claiming: I have no Go toolchain on the machine I am on right now, so I have not re-run the suite against current `main` myself. The run above is yours, on this branch.

Happy to rebase onto current `main`, or to close this if you would rather solve it by having the wrapper collect on the caller's goroutine instead. No attribution needed.

PR #2030, opened 2026-06-26
fix(api): fall back to GET on forbidden POSTs

### LINKED ISSUE #1619: DoGetFallback: fallback to GET when POST is forbidden (e.g. when behind a RBAC proxy)
Version used for observed issue: v1.19.0 (looking at the `main` code, it's still an issue: https://github.com/prometheus/client_golang/blob/main/api/prometheus/v1/api.go#L1453

Currently, some of the API functions to query metrics use a POST method, and fallback to GET when a status "Not allowed" or "Not implemented" is returned; cf `DoGetFallback` function.

When Prometheus is used behind a RBAC proxy, users might have permissions for GET (associated with reading) but not for POST (associated with writing). So a POST query might return a 403 error while the GET would succeed.

I'd suggest adding a check for 403 response code as a valid trigger for the GET fallback. What do you think? I'm happy to open a PR if that sounds good to you.

immanuwell:
cc @bwplotka

Closes #1619.

Tiny fix, but real. Some setups put Prometheus behind an RBAC proxy, so `GET` is allowed but `POST` gets `403`. Right now `DoGetFallback` only retries on `405` and `501`, so read-only calls can just fail there, kinda rough.

This adds `403` to the existing POST to GET fallback path. No new API, just extends the compat behavior already in place.

Repro:
1. Put Prometheus behind a proxy that allows `GET` and denies `POST`.
2. Call a `DoGetFallback` backed read API like `Query`, `QueryRange`, `LabelNames`, `Series`, `QueryExemplars`, or `FormatQuery`.
3. Before this patch it returns `403`, no retry.
4. After this patch it retries with `GET` and works.

Tests:
- `go test ./api/prometheus/v1`
- `go test ./...`

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -1545,8 +1545,8 @@ func (h *apiClientImpl) Do(ctx context.Context, req *http.Request) (*http.Respon
 	return resp, []byte(result.Data), result.Warnings, err
 }
 
-// DoGetFallback will attempt to do the request as-is, and on a 405 or 501 it
-// will fallback to a GET request.
+// DoGetFallback will attempt to do the request as-is, and on a 403, 405, or
+// 501 it will fallback to a GET request.
 func (h *apiClientImpl) DoGetFallback(ctx context.Context, u *url.URL, args url.Values) (*http.Response, []byte, Warnings, error) {
 	encodedArgs := args.Encode()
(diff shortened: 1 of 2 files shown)


bwplotka (approved):
That makes sense to me 👍🏽 

Thanks!

Issue #1810, opened 2025-05-08
Memory Leak with DeletePartialMatch()


rsylim804:
Hello I am using the library's DeletePartialMatch() function to clean up metrics which I consider expired after a certain time. However, I am experiencing a memory leak somewhere on the API server where I host these metrics. Is there something I'm missing about the usage in this deletion function? I'm lost on where this memory leak could be occurring. Are there any ideas on what could be potentially causing this. I will note there is high cardinality on the metrics being stored (but according to my understanding this should not matter because I am periodically deleting "expired" metrics with DeletePartialMatch())?

kakkoyun:
@rsylim804 Thanks for the report. Would it be possible to provide a profile from the binary that you are experiencing this?

PR #2039, opened 2026-07-01
prometheus: DeletePartialMatch no longer deletes colliding non-matching metrics

arpitjain099:
When two series hash into the same bucket and you call DeletePartialMatch with labels that only match one of them, the other one gets deleted too. deleteByLabels drops the whole bucket on any partial match instead of removing just the matching entry, so a colliding series that should have stayed is silently wiped, and the returned count is off when a single bucket holds more than one match. This is likely part of what the reporter saw in #1810, so I framed it as the concrete data-loss bug rather than claiming it is the entire cause of that memory growth.

The fix mirrors what deleteByHashWithLabels already does: splice out only the matching element when the bucket has more than one metric, delete the bucket only when it holds a single element, and count every metric actually removed.

I added TestDeletePartialMatchWithCollisions, which forces a collision and checks the non-matching series survives with its value and that multiple matches in one bucket are all counted. It fails before the change and passes after; the full prometheus package still passes with -race.

@vesari would you mind taking a look? Thanks.

### DIFF
--- a/prometheus/vec.go
+++ b/prometheus/vec.go
@@ -416,13 +416,26 @@ func (m *metricMap) deleteByLabels(labels Labels, curry []curriedLabelValue) int
 	var numDeleted int
 
 	for h, metrics := range m.metrics {
-		i := findMetricWithPartialLabels(m.desc, metrics, labels, curry)
-		if i >= len(metrics) {
-			// Didn't find matching labels in this metric slice.
-			continue
+		// A single hash bucket can hold more than one distinct metric on a
+		// hash collision, so remove only the elements that actually match
+		// instead of dropping the whole bucket.
+		for {
+			i := findMetricWithPartialLabels(m.desc, metrics, labels, curry)
+			if i >= len(metrics) {
+				// No (more) matching labels in this metric slice.
+				break
+			}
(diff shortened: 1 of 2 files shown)


bwplotka:
Thanks for attempting to address this.

I am skeptical here. Can you reproduce hash collision on normal set of metrics? It's quite unlikely, and the damage radius is minimal 🤔 

Also how https://github.com/prometheus/client_golang/issues/1810 is connected? If we clear more than one metric, why would we leak?

arpitjain099:
You're right on both counts, thanks for pushing back.

On #1810: my connection there doesn't hold up. Over-deleting frees entries, so it wouldn't cause the growth in that report. I shouldn't have tied the two together, sorry for the misleading framing.

On reachability: agreed. This only triggers on an actual hash collision landing two series in one bucket, which I can't reproduce on a normal metric set (the test forces the collision synthetically). It's a genuine correctness edge, the whole bucket gets dropped when only one colliding series matches, but the real-world damage radius is small as you say, and it's not worth carrying on that basis.

I'll close this. Happy to revisit if a real collision report ever surfaces. Thanks for taking a look.

PR #2040, opened 2026-07-01
api/prometheus: clamp out-of-range formatted timestamps

### LINKED ISSUE #951: API passes time.Earliest incorrectly
If you pass `model.Earliest.Time()` to prometheus/client_golang it sends -9223372036854776, which is 192 milliseconds less than the original.  This value will cause an overflow when read by `prometheus/web/api/v1`, turning it into a very large _positive_ time.

I am using github.com/prometheus/client_golang v1.11.0
and github.com/prometheus/client_model v0.2.0.

To illustrate:
https://go.dev/play/p/b_75bSN2Gyr

```go
package main

import (
	"fmt"
	"strconv"
	"time"

)

// As found in github.com/prometheus/common/model
func formatTime(t time.Time) string {
	return strconv.FormatFloat(float64(t.Unix())+float64(t.Nanosecond())/1e9, 'f', -1, 64)
}

// As found in github.com/prometheus/prometheus/model/timestamp
func FromTime(t time.Time) int64 {
	return t.Unix()*1000 + int64(t.Nanosecond())/int64(time.Millisecond)
}

func main() {
	t1 := model.Earliest.Time()
	fmt.Println(formatTime(t1))
	t2 := time.Unix(-9223372036854776, 0)  // this is the number output by the previous line
	fmt.Println(t1, FromTime(t1))
	fmt.Println(t2, FromTime(t2))
}
```

I believe the problem arises because `strconv.FormatFloat()` is asked to round the number as if it came from a `float64`, and `model.Earliest` is the largest negative `int64` so it doesn't fit in a `float64`, and the milliseconds get dropped.

This arose while trying to work round #621

happysnaker:
## Summary

Clamp extremely low/high `time.Time` values to Prometheus's documented parseable API boundary timestamps before formatting query parameters.

## What this fixes

Issue #951 reports that passing `model.Earliest.Time()` through the client ends up formatting to a rounded float value like `-9223372036854776`, which overflows back into a large positive timestamp on the server side.

This patch keeps normal timestamp formatting in the existing numeric API style, but special-cases values at or beyond the Prometheus API boundary sentinels so they serialize to the same RFC3339 strings that the server already recognizes explicitly.

## Details

- keep regular timestamps in the current numeric form
- avoid relying on `time.Time.UnixNano()` for out-of-range values
- clamp values at/beyond the effective API boundaries to the server-compatible `minTimeFormatted` / `maxTimeFormatted` strings
- add focused regression tests for:
  - normal timestamps
  - negative fractional timestamps
  - `model.Earliest.Time()`
  - `model.Latest.Time()`

Fixes #951

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -1575,6 +1575,50 @@ func (h *apiClientImpl) DoGetFallback(ctx context.Context, u *url.URL, args url.
 	return resp, body, warnings, err
 }
 
+var (
+	// Prometheus's HTTP API accepts these boundary timestamps explicitly even
+	// though Go's RFC3339 parser cannot round-trip years with more than four
+	// digits. Keep these values aligned with the server-side defaults so very
+	// low/high sentinels like model.Earliest and model.Latest map to something
+	// the API can parse without overflowing.
+	minTime = time.Unix(math.MinInt64/1000+62135596801, 0).UTC()
+	maxTime = time.Unix(math.MaxInt64/1000-62135596801, 999999999).UTC()
+
+	minTimeFormatted = minTime.Format(time.RFC3339Nano)
+	maxTimeFormatted = maxTime.Format(time.RFC3339Nano)
+)
+
 func formatTime(t time.Time) string {
-	return strconv.FormatFloat(float64(t.Unix())+float64(t.Nanosecond())/1e9, 'f', -1, 64)
+	switch {
+	case !t.After(minTime):
+		return minTimeFormatted
+	case !t.Before(maxTime):
+		return maxTimeFormatted
+	}
+
+	// Avoid t.UnixNano here: it is undefined for times that cannot be
+	// represented as an int64 nanosecond offset even though time.Unix and
+	// Nanosecond still expose the second/nanosecond components we need.
+	return formatTimeString(t.Unix(), t.Nanosecond())
+}
+
+func formatTimeString(sec int64, nsec int) string {
+	if sec >= 0 {
+		return formatTimeParts("", uint64(sec), nsec)
+	}
+	if nsec == 0 {
+		return strconv.FormatInt(sec, 10)
+	}
+
+	return formatTimeParts("-", uint64(-(sec + 1)), int(time.Second)-nsec)
+}
+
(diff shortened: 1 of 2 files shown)


happysnaker:
Maintainer context for review:

- issue #951 points out that float-based formatting can round `model.Earliest` / `model.Latest` into values that overflow back on the server side
- this patch keeps the existing numeric timestamp style for normal values
- it only clamps values at or beyond the effective Prometheus API boundaries to the same parseable boundary strings the server already special-cases

So the behavior change is intentionally narrow: fix the extreme out-of-range edge case without changing normal query parameter formatting.

api/prometheus/v1/api.go
```diff
@@ -1575,6 +1575,50 @@ func (h *apiClientImpl) DoGetFallback(ctx context.Context, u *url.URL, args url.
 	return resp, body, warnings, err
 }
 
+var (
+	// Prometheus's HTTP API accepts these boundary timestamps explicitly even
+	// though Go's RFC3339 parser cannot round-trip years with more than four
+	// digits. Keep these values aligned with the server-side defaults so very
+	// low/high sentinels like model.Earliest and model.Latest map to something
+	// the API can parse without overflowing.
```
bwplotka, line 1583: Let's mention the relevant server side code

```suggestion
	// Prometheus's HTTP API accepts these boundary timestamps explicitly even
	// though Go's RFC3339 parser cannot round-trip years with more than four
	// digits. Keep these values aligned with the server-side defaults so very
	// low/high sentinels like model.Earliest and model.Latest map to something
	// the API can parse without overflowing.
	// Reference: https://github.com/prometheus/prometheus/blob/10cbff99a9d5998c4a5ad998b6cf41ea269bde08/web/api/v1/api.go#L834
```

bwplotka (approved):
This is interesting one! It's not great if HTTP client changes the time ranges you ask for underhood.

But for general practicality that would make sense. People use lowest and highest for "any" range purpose. Let's fix this this way 👍🏽 

One nit only, thanks!

kakkoyun:
@happysnaker Would you have a look? We will cut a release soon and this could be included.

PR #2044, opened 2026-07-02
chores: remove example Dockerfile and container_description.yaml

### LINKED ISSUE #1938: container_description.yml workflow failing

bwplotka:
It's not hooked into our CI, it does not work, causing container_description.yaml to be [synced](https://github.com/prometheus/prometheus/blob/19273691bb2a1f7584c12134dd23cb6a31de8d52/scripts/sync_repo_files.sh#L209)
and [unnecessary GH action to fail](https://github.com/prometheus/client_golang/actions/workflows/container_description.yml).

It's gives low value (it's trivial to recreate), so let's remove to have a bit less things to maintain.

Fixes https://github.com/prometheus/client_golang/issues/1938


vesari (approved):
Sounds like a good idea to me, thanks!

PR #2050, opened 2026-07-08
test: fix two flaky tests (darwin start_time regex, memstats HeapReleased drift)

kakkoyun:
Two pre-existing flaky tests, each fixed independently.

Darwin `process_start_time_seconds` (`process_collector_darwin_test.go`): the check used `[0-9.]{10,}`, but expfmt renders large gauges in scientific notation (e.g. `1.783414e+09`), whose leading digit run is only eight characters, so it failed intermittently on macOS. Changed to `[0-9]`, matching the sibling checks in the same test; that is a presence check that does not depend on float formatting.

`TestMemStatsEquivalence` (`go_collector_latest_test.go`): the test read `runtime.ReadMemStats` and `runtime/metrics` separately and compared them within 5%, and the code itself noted it hoped no GC happened in between. `HeapReleased` is driven by the background scavenger and drifted between the two reads, failing on Linux arm64. It now forces a full scavenge with `debug.FreeOSMemory` and disables GC for the measurement window, so both reads observe the same runtime state.

Test plan: `go test -race ./prometheus/...` passes; `TestMemStatsEquivalence` run 200 times with `-race` is stable.

### DIFF
--- a/prometheus/go_collector_latest_test.go
+++ b/prometheus/go_collector_latest_test.go
@@ -21,6 +21,7 @@ import (
 	"reflect"
 	"regexp"
 	"runtime"
+	"runtime/debug"
 	"runtime/metrics"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


kakkoyun:
Blocked by #2051

bwplotka (approved):
Thanks!

PR #2051, opened 2026-07-08
Replace deprecated model.NameValidationScheme with explicit UTF8Validation

kakkoyun:
`model.NameValidationScheme` is deprecated in prometheus/common. staticcheck flags each use, and under some golangci-lint cache states the suppressing `//nolint` directives are themselves reported as unused, which makes lint flaky.

This replaces the three uses (`desc.go`, `labels.go`, `push/push.go`) with the explicit `model.UTF8Validation` scheme and removes the `//nolint` directives.

`NameValidationScheme` already defaults to `UTF8Validation`, so default behavior is unchanged. Users who set it to `LegacyValidation` no longer get that enforcement at metric/label/grouping construction; UTF-8 validation is more permissive, so no name that validates today starts failing.

Test plan: `go test -race ./prometheus/... ./prometheus/push/...` passes and `golangci-lint run` is clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+* [CHANGE] prometheus: Name validation now always uses the UTF-8 scheme instead of the deprecated `model.NameValidationScheme` global. Default behavior is unchanged; code that set `NameValidationScheme = LegacyValidation` no longer gets legacy enforcement at metric, label, and push-grouping construction. #2051
 * [FEATURE] HTTP handlers created by `promhttp` package now support metrics filtering by providing one or more `name[]` query parameters. The default behavior when none are provided remains the same, returning all metrics. #1925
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
 
+* [CHANGE] prometheus: Name validation now always uses the UTF-8 scheme instead of the deprecated `model.NameValidationScheme` global. Default behavior is unchanged; code that set `NameValidationScheme = LegacyValidation` no longer gets legacy enforcement at metric, label, and push-grouping construction. #2051
```
bwplotka, line 3: Hm, so does that change anything? 🤔  I think not, right?

bwplotka: I am going to kill it later in the release step 👍🏽

PR #2052, opened 2026-07-10
back Counter with sharded per-P counters from argc.dev/goexp/sync

cagedmantis:
Summary

  This replaces the atomic uint64 fields backing the plain prometheus.Counter with sharded per-P counters from argc.dev/goexp/sync. Under concurrent writers, a single atomic counter
  becomes a cache-coherency hotspot: every Inc()/Add() contends on the same cache line. The sharded counter stripes writes across per-P shards so increments stay contention-free, summing
  the shards only at read (collection) time.

  Counter is the ideal — and essentially only — good fit for this in the repository:

  - Inc/Add are on the hottest write paths in instrumented code, while reads happen only at scrape time (seconds apart). The counter is monotonic and never needs a consistent snapshot or
  a reset — a textbook match for a write-optimized, approximate-read sharded counter.
  - Histograms and summaries are intentionally not changed. They depend on exact atomic hot/cold index swaps and consistent-snapshot draining of cold buckets; the sharded counter's
  approximate Value() and lack of atomic-swap semantics would break that logic.
  - Gauges are intentionally not changed. They decrease as well as increase and need Set plus exact reads, which a sharded counter does not provide.

  Changes

  - counter.valBits/counter.valInt change from uint64 atomics to *xsync.Counter[float64] / *xsync.Counter[int64], initialized in both the NewCounter and NewCounterVec factories.
  - Add, Inc, and get delegate to the sharded counter's Add() / Value(). The existing integer/float split is preserved, so the exact-integer fast path and float precision are retained.
  (int64 replaces uint64 because the package is generic over int64 | float64; this only differs in the astronomical 9.2e18–1.8e19 range, far beyond any realistic counter.)
  - Removed the now-obsolete struct-alignment comment and the unused math import.
  - Updated the white-box assertions in counter_test.go to read via .Value().
  - go mod tidy adds a single require line; module-graph pruning keeps goexp's unrelated dependencies out of go.mod since only the sync subpackage is imported.

  Behavioral note for reviewers

  Value() is approximate under concurrent writers — it sums shards without a global lock, so a scrape taken mid-write may read shards from slightly different instants. For a monotonic
  counter scraped every ~15s this is harmless and is the intended trade for contention-free writes, but it is a genuine semantic change from the old strictly-atomic read and is called
  out here deliberately.

  Validation

  - go build ./... — clean
  - go vet ./prometheus/ — clean
  - go test -race ./prometheus/ — PASS (127s)

Validation                                                                                                                                                                       [0/843]

  - go build ./... — clean
  - go vet ./prometheus/ — clean
  - go test -race ./prometheus/ — PASS (127s)

  Benchmarks

  go test -bench=... -benchmem -cpu=8 -count=3 (Apple Silicon, darwin):

   | Benchmark | Before (atomic) | After (sharded) | Change |
  |---|---|---|---|
  | `BenchmarkParallelCounter` (contended `Inc`) | ~48.1 ns/op | ~0.81 ns/op | ~59× faster |
  | `BenchmarkCounterNoLabels` (uncontended) | ~6.95 ns/op | ~6.00 ns/op | ~14% faster |

  Zero allocations before and after in both benchmarks.

  <details>
  <summary>Raw benchmark output</summary>

  === OLD (atomic) ===
  BenchmarkCounterNoLabels-8    170848102                6.965 ns/op           0 B/op          0 allocs/op
  BenchmarkCounterNoLabels-8    173221360                6.946 ns/op           0 B/op          0 allocs/op
  BenchmarkCounterNoLabels-8    172775521                6.945 ns/op           0 B/op          0 allocs/op
  BenchmarkParallelCounter-8    26206380                48.18 ns/op            0 B/op          0 allocs/op
  BenchmarkParallelCounter-8    26129320                47.30 ns/op            0 B/op          0 allocs/op
  BenchmarkParallelCounter-8    29907597                48.78 ns/op            0 B/op          0 allocs/op

  === NEW (sharded) ===
  BenchmarkCounterNoLabels-8    196301019                6.003 ns/op           0 B/op          0 allocs/op
  BenchmarkCounterNoLabels-8    200686598                6.007 ns/op           0 B/op          0 allocs/op
  BenchmarkCounterNoLabels-8    200671132                5.996 ns/op           0 B/op          0 allocs/op
  BenchmarkParallelCounter-8    1000000000               0.8075 ns/op          0 B/op          0 allocs/op
  BenchmarkParallelCounter-8    1000000000               0.8266 ns/op          0 B/op          0 allocs/op
  BenchmarkParallelCounter-8    1000000000               0.8082 ns/op          0 B/op          0 allocs/op
  </details>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/prometheus/client_golang
 go 1.25.0
 
 require (
+	argc.dev/goexp v0.0.0-20260710001111-8c1f88208172
 	github.com/beorn7/perks v1.0.1
 	github.com/cespare/xxhash/v2 v2.3.0
 	github.com/google/go-cmp v0.7.0
--- a/go.sum
+++ b/go.sum
@@ -1,3 +1,5 @@
+argc.dev/goexp v0.0.0-20260710001111-8c1f88208172 h1:PET7L5h+6EjqOnYjG2Z5sxuMHx2NFmn4OjA/ls/ywEI=
+argc.dev/goexp v0.0.0-20260710001111-8c1f88208172/go.mod h1:5KW5TTv68+fSFL3H+qywujHM1tkwM9JWcc2+ZtT2Vrc=
 github.com/beorn7/perks v1.0.1 h1:VlbKKnNfV8bJzeqoa4cOKqO6bYr3WgKZxO8Z16+hsOM=
 github.com/beorn7/perks v1.0.1/go.mod h1:G2ZrVWU2WbWT9wwq4/hrbKbnv/1ERSJQ0ibhJ6rlkpw=
 github.com/cespare/xxhash/v2 v2.3.0 h1:UL815xU9SqsFlibzuggzjXhog7bL6oX9BbNZnL2UFvs=
--- a/prometheus/counter.go
+++ b/prometheus/counter.go
@@ -15,10 +15,10 @@ package prometheus
 
 import (
 	"errors"
-	"math"
 	"sync/atomic"
 	"time"
 
+	xsync "argc.dev/goexp/sync"
 	dto "github.com/prometheus/client_model/go"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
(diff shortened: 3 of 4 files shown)


cagedmantis:
This PR was opened in error as part of an experiment.

Issue #763, opened 2020-06-03
Check status code before marshalling and returning response


bwplotka:
I think this is a bug that we don't check status code before we start to do things in our client APIs code against Prometheus. Related discussion: https://github.com/prometheus/client_golang/pull/755#discussion_r427176238

The problem is that users can access the API through various proxies, so we can easily get 404, 403, and other 4xx or 5xx code without proper return type. This will cause very confusing unmarshalling error without even mentioning the response code. I think this is quite a serious problem and something that potential blocks Thanos to switch to those APIs.

AC: 
* Code like this https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go#L568 should never ignore status code
* Make sure Body is always closed and exhausted (I am not sure if unmarshall gives us that)

Help wanted, I can guide someone through.

cc @beorn7 @lilic

simonpasquier:
It looks like the underlying implementation of  [`Do()`](https://github.com/prometheus/client_golang/blob/84c6b9db90c01eae0373263aa46088a0850c7724/api/prometheus/v1/api.go#L924-L931) already checks the HTTP response code.
From what I can tell no caller of the `Do()` and `DoGetFallback()` functions (from the [`apiClient`](https://github.com/prometheus/client_golang/blob/84c6b9db90c01eae0373263aa46088a0850c7724/api/prometheus/v1/api.go#L879-L883) interface) use the returned `*http.Response` so maybe it shouldn't be returned at all and the functions should take care of closing the body?

bwplotka:
Ah good point, @simonpasquier thanks for this.

Yea, let's check how we can fix second AC. There are two options:

1. Caller should close and exhaust as you proposed
2. We read body and close, and pass just buffered bytes. That makes sense only if we are sure that client does not stream. e.g for remote read series, that is NOT true (we recommend streaming)

max-neverov:
@bwplotka it seems the client already does [the second option](https://github.com/prometheus/client_golang/blob/84c6b9db90c01eae0373263aa46088a0850c7724/api/client.go#L114).

What interesting is that the client closes the body twice. The [response can be ignored](https://github.com/golang/go/blob/master/src/net/http/client.go#L566) if the go http client returns an error.

The body is parsed in a different goroutine, but the result is [awaited for anyway even if the context was cancelled](https://github.com/prometheus/client_golang/blob/84c6b9db90c01eae0373263aa46088a0850c7724/api/client.go#L120).

[...]

PR #2062, opened 2026-07-21
api: return HTTP status on non-JSON API error bodies

dpacgdm:
## Summary
* When HTTP 400/422 responses carry a non-JSON body (common via proxies), `apiClientImpl.Do` now returns a status-based error with the body as `Detail` instead of a confusing JSON unmarshal failure.
* Existing 404/5xx paths already skipped unmarshalling; this closes the remaining gap for Prometheus API error statuses that previously tried to decode a JSON API envelope first.

Addresses prometheus/client_golang#763 (status-before-unmarshal for proxy/non-JSON error bodies). Body close/exhaust is already handled by the lower HTTP client and is out of scope here.

## Test plan
- [x] `go test ./api/prometheus/v1/ -run TestAPIClientDo`
- [x] `go test ./api/prometheus/v1/ -run TestAPIs`

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -1521,6 +1521,16 @@ func (h *apiClientImpl) Do(ctx context.Context, req *http.Request) (*http.Respon
 
 	if http.StatusNoContent != code {
 		if jsonErr := json.Unmarshal(body, &result); jsonErr != nil {
+			// Non-2xx API error codes (400/422) with a non-JSON body (e.g. HTML from a
+			// proxy) should surface the HTTP status instead of a confusing unmarshal error.
+			if code/100 != 2 {
+				errorType, errorMsg := errorTypeAndMsgFor(resp)
+				return resp, body, nil, &Error{
+					Type:   errorType,
+					Msg:    errorMsg,
+					Detail: string(body),
+				}
+			}
 			return resp, body, nil, &Error{
 				Type: ErrBadResponse,
 				Msg:  jsonErr.Error(),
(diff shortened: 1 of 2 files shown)


api/prometheus/v1/api.go
```diff
@@ -1521,6 +1521,16 @@ func (h *apiClientImpl) Do(ctx context.Context, req *http.Request) (*http.Respon
 
 	if http.StatusNoContent != code {
 		if jsonErr := json.Unmarshal(body, &result); jsonErr != nil {
+			// Non-2xx API error codes (400/422) with a non-JSON body (e.g. HTML from a
+			// proxy) should surface the HTTP status instead of a confusing unmarshal error.
+			if code/100 != 2 {
```
gnanirahulnutakki, line 1526: This still allows a non-2xx response to return a nil error when the body is valid JSON but not a Prometheus API envelope. For HTTP 400 with `{"message":"proxy error"}`, `json.Unmarshal` succeeds, leaves `result.Status` empty, and neither status check below sets `err`. I reproduced this with a `TestAPIClientDo` case; it failed with `expected error, but got none`.

Could the 400/422 path also fall back to `errorTypeAndMsgFor` when the decoded body is not a valid API error response, with this valid-but-non-envelope JSON case added as a regression test? That would preserve issue #763's requirement that the status code is never ignored.

gnanirahulnutakki (commented):
I reproduced one remaining status-before-decode gap: syntactically valid non-envelope JSON on HTTP 400/422 can still return a nil error. Details and a regression case are inline. The existing `api/prometheus/v1` package test suite passes on the PR branch.

api/prometheus/v1/api.go
liggitt: I incorporated this change into the error path improvements in https://github.com/prometheus/client_golang/pull/2120, which ensure an error is always returned on non-200 responses

PR #1027, opened 2022-04-11
go.mod: Exclude prometheus/client_golang v1.12.1

mrueg:
See also https://github.com/prometheus/client_golang/issues/1012
As suggested in https://github.com/prometheus/client_golang/issues/1012#issuecomment-1090482644

Issue #1012, opened 2022-03-23
Cyclic dependency in packages is leading to use of vulnerable packages


rhim:
To address the following vulnerability that exists in associated package
```
CVE-2021-3121 github.com/gogo/protobuf v1.1
```

we need to address client_golang and common's cyclic dependency, which is as follows:

```
github.com/prometheus/common@v0.4.1 github.com/gogo/protobuf@v1.1.1
github.com/prometheus/client_golang@v1.0.0 github.com/prometheus/common@v0.4.1
github.com/prometheus/common@v0.10.0 github.com/prometheus/client_golang@v1.0.0 
github.com/prometheus/client_golang@v1.7.1 github.com/prometheus/common@v0.10.0 
github.com/prometheus/common@v0.26.0 github.com/prometheus/client_golang@v1.7.1 
github.com/prometheus/client_golang@v1.11.0 github.com/prometheus/common@v0.26.0 
github.com/prometheus/common@v0.32.1 github.com/prometheus/client_golang@v1.11.0 
github.com/prometheus/client_golang@v1.12.1 github.com/prometheus/common@v0.32.1 
```

Can someone suggest a way to remove this cycle?

bwplotka:
Yea I think there is no other way than trying to copy a code around to not import one on another. I don't know what common is importing from client_golang, but it might make more sense to get rid of client_golang usage of common - meaning to copy data model from it.

beorn7:
The vulnerable package is not actually used. It's just part of the graph due to the way the Go Modules dependency solver works. The cyclic dependency between _modules_ is allowed (as long as there are no cyclic dependencies between _packages_). The bug is in the security scanner, not in prometheus/client_golang.

I don't think we should invest efforts into solving it here (if it is possible at all). The broken security scanners will run into the same problem with other Go Modules in the same situation.

rhim:
How to prove definitively that package isn’t used?

The vulnerability scanner is indeed going off go mod graph output. They would need to be told how to fix the issue.

***@***.***
about.me/rajhimanshu
________________________________

rcousineau-xandr:
I was running into this as well... there are several closed issues on this repo + some closed issues on the prometheus/common repo too.

Linked to one of them was this pr: https://github.com/wavesplatform/gowaves/pull/678/files

I followed what they did in their project and added `exclude github.com/gogo/protobuf v1.1.1` to my go.mod and that fixed it.

kakkoyun:
[...]

PR #2077, opened 2026-07-26
go.mod: remove obsolete exclude rule

thaJeztah:
relates to;

- https://github.com/prometheus/client_golang/pull/1027
- https://github.com/prometheus/client_golang/issues/1012#issuecomment-1090482644

The exclude was originally added in 02fec464ad5c015332dcbd4db0cf53b96967e3f2 to prevent traversing client_golang v1.12.1. No current dependency requires that version, making the exclude obsolete.

To verify;

    go list -m all

which produced no changes.

Also verified that:

    go mod tidy
    go test ./...

produce no changes and all tests pass.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,5 +30,3 @@ require (
 	golang.org/x/oauth2 v0.36.0 // indirect
 	golang.org/x/text v0.40.0 // indirect
 )
-
-exclude github.com/prometheus/client_golang v1.12.1


bwplotka:
Thanks. What's the motivation for this change? Just cleanup?

> No current dependency requires that version, making the exclude obsolete.

It's more about users who should skip this version, when updating client_golang in the past as it was broken.  Not sure why `go list -m all` would help figuring out if anyone else still imports e..g 1.11 and won't upgrade to a broken version.

Still, it's unlikely given time, but is there harm leaving?

thaJeztah:
Hi! Thanks for looking!


> What's the motivation for this change? Just cleanup?

Yes, cleanup because it's no longer needed, but also because it partially breaks the module (I'll post an example).

> It's more about users who should skip this version, when updating client_golang in the past as it was broken.

The exclude won't help with that; exclude helps preventing the module's own "requires" to skip the version (so that `go.mod` doesn't use it), and helps during version resolution, but for users of the module, it has no effect; https://go.dev/doc/modules/gomod-ref#exclude

> Use the exclude and replace directives to control build-time dependency resolution when building the current module. **These directives are ignored in modules that depend on the current module.**

But .. I found some time ago that, despite the documentation mentioning the directives are "ignored", they _do_ have an effect, and make the module partially broken, because `go install` will refuse instaling if it finds an `exclude` rule;

```bash
go install github.com/prometheus/client_golang/examples/simple@latest
go: github.com/prometheus/client_golang/examples/simple@latest (in github.com/prometheus/client_golang@v1.24.1):
	The go.mod file for the module providing named packages contains one or
	more exclude directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

> Not sure why go list -m all would help figuring out if anyone else still imports e..g 1.11 and won't upgrade to a broken version.

The `go list` was used to check that the module itself no longer had the broken version in its graph (so to check if the `exclude` was still relevant); it does indeed not show if any _user_ of the module would have it, but (as mentioned above), the `exclude` won't help with that.

thaJeztah:
> It's more about users who should skip this version, when updating client_golang in the past as it was broken

Actually; in case I misunderstood the intent; was `client_golang` v1.12.1 itself broken? Because if that's the case, then probably the correct thing is to _retract_ the version (i.e., mark that version as "faulty release; do not use")
https://go.dev/doc/modules/gomod-ref#retract

e.g.;

```go.mod
retract v1.12.1 // Broken release, do not use.
```

I don't think that prevents users from installing it, but at least will provide the correct metadata, pkg.go.dev will show it as retracted;

Screenshot 2026-08-18 at 13 35 04

bwplotka (approved):
No, you are right I mixed exclude with retract -- TIL that breaks install (but not go run).

Let's do this, thanks!

thaJeztah:
> TIL that breaks install (but not go run).

Same for me (well not _Today_ but some time ago); I thought `exclude` was "local" so no side-effect for consumers.. Then ran into that error.

Thanks!

Issue #2004, opened 2026-05-13
Question: What are the best practices for testing exposed metrics compatibility?


iypetrov:
What’s the recommended way to write compatibility tests for all metrics an app exposes? I want to detect breaking changes like renamed metrics, removed metrics, changed labels/types, etc. This should includes metrics from:
- Built-in collectors (`go_*`, `process_*`)
- Custom app metrics
- 3rd-party metrics (gRPC, OpenTelemetry, etc.)

P.S. I was thinking about using the [Describe](https://pkg.go.dev/github.com/prometheus/client_golang@v1.23.2/prometheus#Collector.Describe) method from the Collector interface to discover all metrics a collector can expose. But I couldn’t find a way to ask a Registry for the list of registered collectors. Would it make sense to add something like this to support this kind of use case, or is there a recommended way to achieve it already?

iypetrov:
@bwplotka wdyt?

bwplotka:
Good question! 

It depends on what lifecycle you want to do it in. I assume you're thinking about unit tests?

One way of doing this is through /metrics. See https://prometheus.io/docs/prometheus/latest/command-line/promtool/#promtool-check-metrics

> P.S. I was thinking about using the [Describe](https://pkg.go.dev/github.com/prometheus/client_golang@v1.23.2/prometheus#Collector.Describe) method from the Collector interface to discover all metrics a collector can expose. But I couldn’t find a way to ask a Registry for the list of registered collectors. Would it make sense to add something like this to support this kind of use case, or is there a recommended way to achieve it already?

There is https://pkg.go.dev/github.com/prometheus/client_golang@v1.23.2/prometheus#Registry.Describe but all fields are private. You could use `Gather` and it's similar to scrape - I think that's a reasonable approach.

Long term we thought about the direction of Otel schema:
* https://kccnceu2026.sched.com/event/2CVzR/schema-inference-and-automation-a-new-era-for-telemetry-management-nicolas-takashi-coralogix-arthur-silva-sens-grafana-labs
* https://github.com/prometheus/prometheus/issues/18349

When all metrics are defined in a known schema common Otel tools like `weaver` could be used. For intermitted state we could try to infer (see talk about) schema from metrics by scraping and then Otel tools would work too.

cc @ArthurSens @nicolastakashi

MistaTwista:
i think that function proposal https://github.com/prometheus/client_golang/issues/2079 can be useful for your case too

iypetrov:
[...]

Issue #516, opened 2018-12-17
feature: get descriptor attributes functions 


greenpau:
To ease the documentation of exporters, I read descriptors in the following way:

```
        fmt.Fprintf(os.Stdout, "| Metric | Description | Labels |\n")
        fmt.Fprintf(os.Stdout, "| ------ | ------- | ------ |\n")
        e := &exporter.MyMetric{}
        metrics := make(chan *prometheus.Desc, 100000)
        e.Describe(metrics)
        close(metrics)
        for metric := range metrics {
            m := metric.String()
            fmt.Fprintf(os.Stdout, "%s\n", m)
        }
        os.Exit(0)
```

That yields:

```
| Metric | Description | Labels |
| ------ | ------- | ------ |
Desc{fqName: "net_node_up", help: "Is node up and responding to queries (1) or is it down (0).", constLabels: {}, variableLabels: [node]}
```

I could parse out relevant parts to get `fqName`, `help`, etc. However, I would love to have helper functions.

Here, at https://github.com/prometheus/client_golang/blob/master/prometheus/desc.go#L169, there is `String()` function.

I propose adding the following "public" functions:

* GetName()
* GetHelp()
* GetLabels()

I can submit a PR for it. Worth it?

beorn7:
Thanks for your thoughts.

Descs are up for a major rehaul in the upcoming v0.10, see #222. I would prefer to not try to make the current Descs more sophisticated at this point, just to break everybody anyway in a few months. Depending on the exact new way Descs will be handled, accessors as you suggest might be something we could add, or they might become not needed (if, for example, a Desc is created by an Opts struct whose fields are exported anyway). Cannot really say at the moment, as the design is still evolving.

beorn7:
I'll close this now, but it will be taken into account when working on #222.

greenpau:
@beorn7 , actually i made it prettier and. More useful here. 
https://github.com/ovnworks/gobgp_exporter/blob/master/pkg/gobgp_exporter/metric_table.go

This way it could be integrated as help.

beorn7:
Many thanks. I'll keep it in mind.

PR #2094, opened 2026-08-25
prometheus: expose descriptor metadata via Desc.Info

nicolastakashi:
## What

Adds a way to ask a `Collector` what it declares, without parsing `Desc.String()`.

- `Desc.Info() DescInfo` returns name, help, unit, variable label names, const labels, and the metric type.
- `Desc` now records the metric type. Typed constructors set it; a `Desc` from `NewDesc` reports `UNTYPED`.
- `Registry.DescribeAll()` returns the descriptors of every registered checked Collector.
- `testutil.CollectAndDescribe()` does the same for a single Collector.

Purely additive. No exported symbol changes signature or behaviour.

## Why

**There is no structured way to read a descriptor.** Anything that wants a metric's declared name, help, unit, or labels has to parse `Desc.String()`. That format is not an API and it has changed: one release renders `variableLabels: [a]` with no unit, another renders `unit: "..."` and `variableLabels: {a}`. A test built on it breaks on a `client_golang` bump.

**`Gather` cannot see a metric that has not produced a sample.** A `CounterVec` with no children reports nothing, so you cannot check its type, its labels, or its help. One Prometheus package declares 8 metrics, and 2 of them show up in a scrape of an idle instance. Anything checking metric compatibility against gathered output covers a quarter of that package.

`Describe` already reaches all 8, but nothing could inspect a descriptor, and a descriptor could not say what type the metric was anyway, because the type lived only on the concrete metric. That is why `Desc` now carries one.

#2004 asks for this directly: compatibility tests for renamed metrics, removed metrics, and changed labels or types. The blocker in that thread is not being able to ask a `Registry` what is registered. The breaking change it cites, FluentBit's `fluentbit_hot_reloaded_times` changing from gauge to counter, is a type change.

There is a working consumer at https://github.com/prometheus/prometheus/pull/19523. It holds a package's metrics to a declared schema and catches renames, removals, added and removed labels, and type changes, including on metrics that never produce a sample.

### DIFF
--- a/prometheus/desc.go
+++ b/prometheus/desc.go
@@ -49,6 +49,11 @@ type Desc struct {
 	help string
 	// unit provides the unit of this metric.
 	unit string
+	// metricType is the type of the metric this Desc belongs to. It is
+	// MetricType_UNTYPED unless a typed constructor of this package set it,
+	// as a Desc created via NewDesc does not carry a type: the type of a
+	// const metric is provided per sample, e.g. via MustNewConstMetric.
+	metricType dto.MetricType
 	// constLabelPairs contains precalculated DTO label pairs based on
 	// the constant labels.
 	constLabelPairs []*dto.LabelPair
@@ -75,6 +80,15 @@ type DescOpt func(*Desc)
 func WithUnit(unit string) DescOpt {
 	return func(d *Desc) {
 		d.unit = unit
+	}
+}
+
+// withType sets the metric type for a Desc. It is unexported on purpose: only
+// the typed constructors of this package know the type at Desc construction
+// time.
+func withType(t dto.MetricType) DescOpt {
+	return func(d *Desc) {
+		d.metricType = t
 	}
 }
 
@@ -106,6 +120,7 @@ func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, const
 		fqName:         fqName,
 		help:           help,
 		variableLabels: variableLabels.compile(),
+		metricType:     dto.MetricType_UNTYPED,
 	}
 
 	for _, opt := range opts {
@@ -195,8 +210,59 @@ func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, const
 // a Collector to signal inability to describe itself.
 func NewInvalidDesc(err error) *Desc {
 	return &Desc{
-		err: err,
+		err:        err,
+		metricType: dto.MetricType_UNTYPED,
 	}
+}
+
+// DescInfo is a read-only view of the meta-data a Desc declares. It is meant
+// for introspection, e.g. to check that a Collector describes the metrics a
+// schema says it should. Info allocates on every call and is not meant for
+// hot paths.
+type DescInfo struct {
+	// FQName is the fully-qualified name of the metric.
+	FQName string
+	// Help is the help string of the metric.
+	Help string
+	// Unit is the unit of the metric, empty if none was set.
+	Unit string
+	// Type is the type of the metric. It is MetricType_UNTYPED for a Desc
+	// created via NewDesc, which does not carry a type.
+	Type dto.MetricType
+	// VariableLabels are the names of the labels whose values vary per
+	// metric, in the order they were provided at construction. It is nil if
+	// the metric has no variable labels.
+	VariableLabels []string
+	// ConstLabels are the labels fixed at construction time. It is nil if
+	// the metric has no const labels.
+	ConstLabels Labels
+	// Err is the error that occurred during construction, if any. The
+	// remaining fields may be zero if Err is non-nil.
+	Err error
+}
+
+// Info returns a read-only view of the meta-data of the Desc. In contrast to
+// String, the returned value is structured and therefore suitable for
+// programmatic inspection.
+func (d *Desc) Info() DescInfo {
+	info := DescInfo{
+		FQName: d.fqName,
+		Help:   d.help,
+		Unit:   d.unit,
+		Type:   d.metricType,
+		Err:    d.err,
+	}
+	if d.variableLabels != nil && len(d.variableLabels.names) > 0 {
+		info.VariableLabels = make([]string, len(d.variableLabels.names))
+		copy(info.VariableLabels, d.variableLabels.names)
+	}
+	if len(d.constLabelPairs) > 0 {
+		info.ConstLabels = make(Labels, len(d.constLabelPairs))
+		for _, lp := range d.constLabelPairs {
+			info.ConstLabels[lp.GetName()] = lp.GetValue()
+		}
+	}
+	return info
 }
 
 // Err returns an error that occurred during construction, if any.
--- a/prometheus/registry.go
+++ b/prometheus/registry.go
@@ -595,6 +595,26 @@ func (r *Registry) Describe(ch chan<- *Desc) {
 	}
 }
 
+// DescribeAll returns the Desc of every checked Collector registered with r.
+// It is a convenience wrapper around Describe for callers that want a slice
+// rather than draining a channel, e.g. tests that assert on the metrics an
+// application declares. Unchecked Collectors are not included, as they do not
+// report any Desc.
+func (r *Registry) DescribeAll() []*Desc {
+	ch := make(chan *Desc)
+	go func() {
+		r.Describe(ch)
+		close(ch)
+	}()
+
+	var descs []*Desc
+	for d := range ch {
+		descs = append(descs, d)
+	}
+
+	return descs
+}
+
 // Helper wrapper around Collector.Collect.
 // It tries to collect from the channel, recovers on panic and
 // if it has recovered from a panic, then it sends an InvalidMetric into
--- a/prometheus/testutil/testutil.go
+++ b/prometheus/testutil/testutil.go
@@ -45,6 +45,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"slices"
 
 	"github.com/kylelemons/godebug/diff"
 	dto "github.com/prometheus/client_model/go"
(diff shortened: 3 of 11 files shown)


prometheus/desc.go
```diff
@@ -195,8 +210,59 @@ func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, const
...
+		err:        err,
+		metricType: dto.MetricType_UNTYPED,
+	}
+}
+
+// DescInfo is a read-only view of the meta-data a Desc declares. It is meant
+// for introspection, e.g. to check that a Collector describes the metrics a
+// schema says it should. Info allocates on every call and is not meant for
+// hot paths.
+type DescInfo struct {
```
bwplotka, line 222: This needs more thoughts. We never exposed those info publicly. Apparently the reason was the need for changes in this structure. I wonder if this is still true 🤔 

Some fact digged by AI:

### Key Reasons Why `Desc` Fields Were Kept Private

#### 1. `Desc` was viewed as an internal implementation detail, not a public user-facing abstraction
In [PR #1309](https://github.com/prometheus/client_golang/pull/1309#issuecomment-1631405903) (when a user asked to expose variable label names on `Desc`), `@beorn7` summarized the history and design philosophy directly:

> *"The longer history here is that I was never really happy with `Desc`. It felt like an implementation detail that the user should never see. Therefore, I've tried to keep it as much out of the way of the users as possible, and thought a lot about how to change the use cases where you still have to touch it in the future v2 version of this library.*
>
> *Adding exported methods to `Desc` would go into the exact opposite direction.*
>
> *Another fundamental concern is that this library already has a lot of knobs for the users to turn. Adding more knobs requires strong justification, like simplifying a fairly common use case.*
>
> *In this case, it feels we would help a relatively small number of users to simplify a relatively niche use case that could also be solved with a relatively simple work-around. All of that at the cost of increasing the cognitive load for the many many users of this library that do not need it."*

#### 2. Avoiding user dependencies on internal details that were slated for fundamental rework
Whenever users asked to export `Desc` fields or add getters, the maintainers held off because `Desc` was slated for a major redesign in [Issue #222 ("Rework Desc and ConstMetrics")](https://github.com/prometheus/client_golang/issues/222) intended for a future v2:

* In [Issue #322 ("Expose Desc fields")](https://github.com/prometheus/client_golang/issues/322), when `@AlekSi` requested read access to fields like `fqName`:
  > *"The way Desc work will change fundamentally with #222. Tests in code using client_golang (in contrast to tests that test client_golang itself) must not depend on implementation details of client_golang. Simply making internal fields exported in Desc would only allow users to depend on all those internal implementation details (that are about to change anyway).*
  >
  > *Right now, exposing Desc fields would be a step in the wrong direction."*

* In [PR #326 ("Add getter methods for Desc() struct")](https://github.com/prometheus/client_golang/pull/326), when getters were proposed:
  > *"I'm reworking this for #222. I don't want to expose an internal structure in the main repo that is doomed to get changed in a couple of months."*

* In [Issue #516 ("feature: get descriptor attributes functions")](https://github.com/prometheus/client_golang/issues/516), when `GetName()`, `GetHelp()`, and `GetLabels()` were proposed:
  > *"Descs are up for a major rehaul in the upcoming v0.10, see #222. I would prefer to not try to make the current Descs more sophisticated at this point, just to break everybody anyway in a few months. Depending on the exact new way Descs will be handled, accessors as you suggest might be something we could add, or they might become not needed (if, for example, a Desc is created by an Opts struct whose fields are exported anyway)."*

#### 3. Preserving strict immutability and thread safety
By contract, `Desc` represents the immutable metadata of a metric shared across collectors and goroutines during scrapes.
* Exporting fields directly would allow callers to mutate them (e.g., altering `variableLabels` slices or label values), violating the concurrency guarantees required during `Collect()` and `Gather()` operations.
* In [Issue #269](https://github.com/prometheus/client_golang/issues/269#issuecomment-272097033), `@beorn7` noted that modifying existing descriptors would *"violate the contract about Descs being immutable and Write being concurrency-safe."*

#### 4. Enforcing registry invariants and preventing anti-patterns
`Desc` was introduced in 2014 ([commit `5d40912f`](https://github.com/prometheus/client_golang/commit/5d40912fd2772fbfbfb8a36675573608ec2609c2)) to calculate precomputed hashes (`id` and `dimHash`) so the `Registry` could detect collisions and validate consistency (e.g. matching label dimensions and help text for identical metric names) at registration time rather than scrape time.
* Exposing `Desc` internals to encourage dynamic label routing or building ad-hoc registries was viewed as encouraging anti-patterns (e.g., PR #326 and PR #1309).

prometheus/desc.go
```diff
@@ -78,6 +83,15 @@ func WithUnit(unit string) DescOpt {
 	}
 }
 
+// withType sets the metric type for a Desc. It is unexported on purpose: only
+// the typed constructors of this package know the type at Desc construction
+// time.
+func withType(t dto.MetricType) DescOpt {
```
bwplotka, line 89: Because `withType` is unexported and `NewDesc` defaults to `UNTYPED`, any collector built with `NewDesc` and `NewConstMetric` (such as `GoCollector`, `ProcessCollector`, or custom exporters) will always report `UNTYPED` here. This means `Desc.Info().Type` cannot catch type changes for const metrics, which was one of the motivating examples from issue #2004.

prometheus/desc.go
```diff
@@ -195,8 +210,59 @@ func (v2) NewDesc(fqName, help string, variableLabels ConstrainableLabels, const
...
+	ConstLabels Labels
+	// Err is the error that occurred during construction, if any. The
+	// remaining fields may be zero if Err is non-nil.
+	Err error
+}
+
+// Info returns a read-only view of the meta-data of the Desc. In contrast to
+// String, the returned value is structured and therefore suitable for
+// programmatic inspection.
+func (d *Desc) Info() DescInfo {
```
bwplotka: If `d` is nil, this will panic on `d.fqName`. Should we guard against nil and return a zero `DescInfo` with `Type: dto.MetricType_UNTYPED` and an error?

prometheus/registry.go
```diff
@@ -595,6 +595,26 @@ func (r *Registry) Describe(ch chan<- *Desc) {
...
 }
 
+// DescribeAll returns the Desc of every checked Collector registered with r.
+// It is a convenience wrapper around Describe for callers that want a slice
+// rather than draining a channel, e.g. tests that assert on the metrics an
+// application declares. Unchecked Collectors are not included, as they do not
+// report any Desc.
+func (r *Registry) DescribeAll() []*Desc {
+	ch := make(chan *Desc)
+	go func() {
```
bwplotka, line 605: Consider deferring `close(ch)` so the channel is guaranteed to close even if `r.Describe` panics:
```suggestion
	go func() {
		defer close(ch)
		r.Describe(ch)
	}()
```

prometheus/testutil/testutil.go
```diff
@@ -120,6 +121,37 @@ func ToFloat64(c prometheus.Collector) float64 {
...
+// That makes it suitable for asserting that a Collector still declares the
+// metrics it is expected to, e.g. to catch a renamed metric or a changed label
+// set before it reaches users.
+//
+// There is no GatherAndDescribe counterpart because the prometheus.Gatherer
+// interface does not expose descriptors. Use prometheus.Registry.DescribeAll
+// to inspect a whole Registry.
+func CollectAndDescribe(c prometheus.Collector, metricNames ...string) []prometheus.DescInfo {
+	ch := make(chan *prometheus.Desc)
+	go func() {
```
bwplotka, line 138: Consider deferring `close(ch)` so `ch` is closed even if `c.Describe` panics:
```suggestion
	go func() {
		defer close(ch)
		c.Describe(ch)
	}()
```

prometheus/testutil/testutil.go
```diff
@@ -120,6 +121,37 @@ func ToFloat64(c prometheus.Collector) float64 {
...
+		c.Describe(ch)
+		close(ch)
+	}()
+
+	var infos []prometheus.DescInfo
+	for desc := range ch {
```
bwplotka, line 144: If a collector emits a nil `*Desc`, calling `desc.Info()` will panic and leave the sending goroutine blocked on unbuffered `ch`. Maybe check if `desc == nil` before inspecting it?

prometheus/testutil/testutil.go
```diff
@@ -120,6 +121,37 @@ func ToFloat64(c prometheus.Collector) float64 {
...
+//
+// In contrast to the GatherAnd… functions, it also reports metrics that have
+// not produced a sample yet and would therefore be absent from an exposition.
+// That makes it suitable for asserting that a Collector still declares the
+// metrics it is expected to, e.g. to catch a renamed metric or a changed label
+// set before it reaches users.
+//
+// There is no GatherAndDescribe counterpart because the prometheus.Gatherer
+// interface does not expose descriptors. Use prometheus.Registry.DescribeAll
+// to inspect a whole Registry.
```
bwplotka: Note that `Registry.DescribeAll` returns `[]*Desc` while `CollectAndDescribe` returns `[]prometheus.DescInfo`. Callers following the doc comment to inspect a registry have to do their own conversion loops.

prometheus/desc.go
bwplotka: One way to solve this is to add withType to NewConstMetric based on value type. It would break immutability though, but would work for majority of users 🤔

prometheus/desc.go
```diff
@@ -49,6 +49,11 @@ type Desc struct {
 	help string
 	// unit provides the unit of this metric.
 	unit string
+	// metricType is the type of the metric this Desc belongs to. It is
+	// MetricType_UNTYPED unless a typed constructor of this package set it,
+	// as a Desc created via NewDesc does not carry a type: the type of a
+	// const metric is provided per sample, e.g. via MustNewConstMetric.
+	metricType dto.MetricType
```
bwplotka, line 56: Interestingly this was always [the plan in 2.0](https://github.com/prometheus/client_golang/issues/222#issue-171455482) but adding it without other changes is a bit of a hack.

It's also has a bit of performance penalty, but probably manageable if there is a good motivation

bwplotka (commented):
Some more thoughts

prometheus/desc.go
vesari: Another thing to consider on top of the gap you describe: `MetricType_UNTYPED` is also a legitimate declared type in the exposition format. So `Info().Type` means two things, "declared untyped" and "no type recorded", and a schema check cannot separate them. That makes the gap a bit worse than missing coverage: for the collectors you mention (which #2004 explicitly lists in its scope), a type change comes back as "no change" rather than as "cannot check".
Maybe a solution could be implementing `DescInfo` in this round without the `Type` field. Name, help, unit, labels, and Err already cover the rename, removal, and label-change cases from #2004. Then a follow-up PR could add `Type`, provided that the const-metric mechanism is settled in the meantime, including the immutability trade-off you mention (for example as `*dto.MetricType` with nil for "unknown"?). If instead we absolutely want to add `Type` already, then most likely we need a new test case with `NewDesc` plus `MustNewConstMetric` that surfaces the `UNTYPED` behaviour, so the limitation is visible in code.

prometheus/testutil/testutil.go
```diff
@@ -120,6 +121,40 @@ func ToFloat64(c prometheus.Collector) float64 {
...
+func CollectAndDescribe(c prometheus.Collector, metricNames ...string) []prometheus.DescInfo {
```
vesari, line 136: Unlike the other `CollectAnd*` functions, this one does not register `c` and I assume that's a deliberate choice, but maybe this diverging behaviour would be worth briefly documenting somewhere?

prometheus/registry.go
```diff
@@ -595,6 +595,26 @@ func (r *Registry) Describe(ch chan<- *Desc) {
 	}
 }
 
+// DescribeAll returns the Desc of every checked Collector registered with r.
+// It is a convenience wrapper around Describe for callers that want a slice
+// rather than draining a channel, e.g. tests that assert on the metrics an
+// application declares. Unchecked Collectors are not included, as they do not
```
vesari, line 601: The comment says that unchecked Collectors are not included. Can we add a test that shows this? Unless I'm overlooking something, no current test exercises `DescribeAll` with an unchecked Collector. The already existing `uncheckedCollector` helper could be used for this.

Issue #2105, opened 2026-09-05
Move from unmaintained json-iterator custom encoders to json/v2 encoding


liggitt:
This library adds a dependency on the unmaintained json-iterator library in the `api/prometheus/v1` package

It looks like it does this for two reasons:
1. To add more efficient marshalers/unmarshalers for specific types from `github.com/prometheus/common/model`
2. To do more efficient internal marshaling/unmarshaling for types declared in `api/prometheus/v1`

In addition to being unmaintained, the json-iterator library also achieves some of its efficiency at the cost of correctness and proper use of Go APIs (there is extensive use of unsafe, as well as linking to internal go functions that have caused breakages when bumping Go minor versions). This library is the [last one in the Kubernetes tree](https://github.com/kubernetes/kubernetes/blob/b2ec8b6fefac451a2dedafc4dd71f2f16c7a6abe/hack/unwanted-dependencies.json#L159-L161) that pulls in json-iterator as a dependency, and we're trying to eliminate it from our dependency tree if we can.

With Go 1.27, json/v2 provides the ability to implement custom streaming json encode/decode methods that are very close to the custom json-iterator implementations in performance. In my experiments, they were ~equivalent to the json-iterator allocs and slightly better in bytes used (in addition to being more correct and matching the default json serializations for `github.com/prometheus/common/model`). POC branches:
* https://github.com/liggitt/client_golang/commits/json-iter/
* https://github.com/liggitt/prometheus-common/tree/json-v2

I propose we:
1. Benchmark json and json/v2 in `api/prometheus/v1/api_bench_test.go`
2. Implement MarshalJSONTo/UnmarshalJSONFrom methods on `github.com/prometheus/common/model` types in go1.27-build-tagged files with equivalence tests to ensure identical behavior to the MarshalJSON/UnmarshalJSON methods
3. Once benchmarks demonstrate that we have good enough performance using stdlib with streaming json implementations, drop the json-iterator implementations and the json-iterator dependency

PRs:
- https://github.com/prometheus/client_golang/pull/2107
- https://github.com/prometheus/client_golang/pull/2109
- https://github.com/prometheus/client_golang/pull/2110
- https://github.com/prometheus/client_golang/pull/2111

---

Quick benchmark showing the before / after of stdlib json performance on Go 1.27 with streaming json/v2 implementations:
```
                                                                │   main.txt   │             jsonv2.txt              │
                                                                │    sec/op    │    sec/op     vs base               │
[...]

PR #2109, opened 2026-09-08
RuleGroup: use stdlib for json decoding, optimize decoding

liggitt:
cc @bwplotka 

part of https://github.com/prometheus/client_golang/issues/2105

1. Adds a benchmark for RuleGroup decoding
2. Switches to stdlib
3. Avoids decoding rule bytes multiple times
4. Avoids struct copies

Allocs and bytes consumed are significantly better than json-iter on both Go 1.26 and 1.27

CPU is worse on Go 1.26, better on Go 1.27.

```
goos: darwin
goarch: arm64
pkg: github.com/prometheus/client_golang/api/prometheus/v1
cpu: Apple M4 Pro
                       │    main     │              go1.26.6               │              go1.27.1              │
                       │   sec/op    │    sec/op     vs base               │   sec/op     vs base               │
RuleGroup/streaming-14   944.5µ ± 1%   1551.7µ ± 0%  +64.28% (p=0.002 n=6)   787.3µ ± 3%  -16.64% (p=0.002 n=6)
RuleGroup/unmarshal-14   885.9µ ± 1%   1521.7µ ± 1%  +71.76% (p=0.002 n=6)   722.0µ ± 2%  -18.50% (p=0.002 n=6)
geomean                  914.7µ         1.537m       +67.98%                 754.0µ       -17.58%

                       │     main      │              go1.26.6               │              go1.27.1               │
                       │     B/op      │     B/op      vs base               │     B/op      vs base               │
RuleGroup/streaming-14   1293.9Ki ± 0%   832.7Ki ± 0%  -35.64% (p=0.002 n=6)   620.5Ki ± 0%  -52.04% (p=0.002 n=6)
RuleGroup/unmarshal-14    998.0Ki ± 0%   578.3Ki ± 0%  -42.05% (p=0.002 n=6)   363.5Ki ± 0%  -63.57% (p=0.002 n=6)
geomean                   1.110Mi        693.9Ki       -38.93%                 475.0Ki       -58.20%

                       │     main     │              go1.26.6              │              go1.27.1              │
                       │  allocs/op   │  allocs/op   vs base               │  allocs/op   vs base               │
RuleGroup/streaming-14   27.396k ± 0%   9.614k ± 0%  -64.91% (p=0.002 n=6)   3.868k ± 0%  -85.88% (p=0.002 n=6)
RuleGroup/unmarshal-14   27.314k ± 0%   9.601k ± 0%  -64.85% (p=0.002 n=6)   3.850k ± 0%  -85.90% (p=0.002 n=6)
geomean                   27.35k        9.607k       -64.88%                 3.859k       -85.89%
```

### DIFF
--- a/api/prometheus/v1/api.go
+++ b/api/prometheus/v1/api.go
@@ -761,36 +761,57 @@ func (rg *RuleGroup) UnmarshalJSON(b []byte) error {
 	rg.Interval = v.Interval
 
 	for _, rule := range v.Rules {
-		alertingRule := AlertingRule{}
-		if err := gojson.Unmarshal(rule, &alertingRule); err == nil {
+		ruleType, err := unmarshalRuleType(rule)
+		if err != nil {
+			return err
+		}
+		switch ruleType {
+		case RuleTypeAlerting:
+			alertingRule := AlertingRule{}
+			if err := alertingRule.unmarshalTypeCheckedJSON(rule); err != nil {
+				return err
+			}
 			rg.Rules = append(rg.Rules, alertingRule)
-			continue
-		}
-		recordingRule := RecordingRule{}
-		if err := gojson.Unmarshal(rule, &recordingRule); err == nil {
+		case RuleTypeRecording:
+			recordingRule := RecordingRule{}
+			if err := recordingRule.unmarshalTypeCheckedJSON(rule); err != nil {
+				return err
+			}
 			rg.Rules = append(rg.Rules, recordingRule)
-			continue
+		default:
+			return errors.New("failed to decode JSON into an alerting or recording rule")
 		}
-		return errors.New("failed to decode JSON into an alerting or recording rule")
 	}
 
 	return nil
 }
 
-func (r *AlertingRule) UnmarshalJSON(b []byte) error {
+func unmarshalRuleType(b []byte) (RuleType, error) {
 	v := struct {
 		Type string `json:"type"`
 	}{}
 	if err := gojson.Unmarshal(b, &v); err != nil {
-		return err
+		return RuleType(""), err
 	}
 	if v.Type == "" {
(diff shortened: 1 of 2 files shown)


api/prometheus/v1/api.go
```diff
@@ -745,13 +746,13 @@ type Stat struct {
...
 func (rg *RuleGroup) UnmarshalJSON(b []byte) error {
 	v := struct {
-		Name     string            `json:"name"`
-		File     string            `json:"file"`
-		Interval float64           `json:"interval"`
-		Rules    []json.RawMessage `json:"rules"`
+		Name     string              `json:"name"`
+		File     string              `json:"file"`
+		Interval float64             `json:"interval"`
+		Rules    []gojson.RawMessage `json:"rules"`
```
bwplotka, line 752: How realistic it would be to stick to jsoniter for <1.27 Go versions, given significant CPU overhead in those versions without v2 version? 

I assume it's possible, just is it worth an intrusive change?

liggitt: I'm not sure a difference on the order of microseconds matters on a client operation dwarfed by network transport, but I can do it that way if you prefer... I'd do something like the following:

* build-tag a typedef and function just to do this internal first-level decode to make go1.26 use json-iterator and go1.27 use stdlib
* plan to drop the go 1.26 / json-iterator variant when go 1.26 goes out of support in ~February 2027

liggitt: Just let me know if you think it's necessary to do that for this client operation

bwplotka: ack, I wanted to capture our discussion in https://github.com/prometheus/client_golang/issues/2105

bwplotka (approved):
Thanks!

PR #2104, opened 2026-09-03
Synchronize common files from prometheus/prometheus

prombot:
Propagating changes from prometheus/prometheus default branch.

*Source can be found [here](https://github.com/prometheus/prometheus/blob/main/scripts/sync_repo_files.sh).*

To check out this branch locally and push changes back:
```
git remote add prombot https://github.com/prombot/prometheus_client_golang.git
git fetch prombot repo_sync
git checkout -b repo_sync prombot/repo_sync
```

PR #2115, opened 2026-09-13
Improve linting

SuperQ:
* Update golangci-lint to match upstream Prometheus.
* Enable `govet` linter.
* Enable `modernize` linter.
* Remove `revive` exclusions.
* Make `govet` settings match Prometheus.
* Apply linting fixes.
* Remove outdated Go collector tests.


SuperQ:
This will help with https://github.com/prometheus/client_golang/pull/2104.

vesari (approved):
Thanks for narrowing down the scope of the extra rules. I don't know whether it's worth noting in a comment the rationale behind it, so in the future we don't widen it again thinking it's ok to align it with how it is in prometheus/prometheus.

kakkoyun (approved):

PR #2121, opened 2026-09-17
Delete unused custom json-iterator marshaling

liggitt:
These methods are not used by client_golang (it only unmarshals), and did not match stdlib encoding of common/model types.

Dropping these and and the json-iterator marshal benchmarks as part of https://github.com/prometheus/client_golang/issues/2105

If json/v2 streaming marshaling improvements are added in a follow-up, that should be done in the common/model repo/package, with benchmarks in that repo, and should be byte-identical to the stdlib json encoders.

The benchmark change is best reviewed ignoring whitespace: https://github.com/prometheus/client_golang/pull/2121/changes?w=1

cc @bwplotka


bwplotka (approved):
I think this is OK. We did some investigation and those are not used in Prometheus, so I am not sure why we optimized marshalling for those. 

The bucket omit empty logic and 0.xxx decimals is the only difference for those who depended on those. It shouldn't be functional.

Perhaps @krajorama has more input on this, but he is PTO. The risk is low and we can bring it back if needed in some form, so LGTM

PR #2125, opened 2026-09-18
testutil: finalize OpenMetrics output

ajavanma:
`CollectAndFormat` and `GatherAndFormat` omit the required `# EOF` line when producing OpenMetrics output. An empty selection returns no bytes, which is also an incomplete OpenMetrics document.

Close the encoder in the shared `GatherAndFormat` helper after encoding all metric families and before returning the buffer, propagating any finalization error. This follows the existing `promhttp` pattern. Table-driven regressions cover both helpers, unmatched selections, an empty registry, multiple families with one final terminator, and unchanged plain-text output.

Validation on Linux ARM64 with Go 1.27.1 and the declared module dependencies:

- Six OpenMetrics cases fail before the fix and pass afterward; all five plain-text controls pass in both versions.
- `make test GOOPTS=-count=1` passes for the root and `exp/` modules.
- `make lint` passes with golangci-lint 2.13.2.
- `go test -race ./prometheus/testutil -run 'Test(Collect|Gather)AndFormat$' -count=1` passes.
- A separate before/after compatibility check passes all 28 cases for both helpers across plain text and all three protobuf formats.

cc @bwplotka


bwplotka (approved):
This semantically makes sense, but I have a feeling this is a breaking change. Not a critical one, but annoying one potentially (someone depending in their tests on lack of newline/EOF).

Scheduled some AI agent work to test it out on kubernetes/kubernetes.

prometheus/testutil/testutil.go
```diff
@@ -266,6 +266,11 @@ func GatherAndFormat(g prometheus.Gatherer, format expfmt.FormatType, metricName
 			return nil, fmt.Errorf("encoding gathered metrics failed: %w", err)
 		}
 	}
+	if closer, ok := enc.(expfmt.Closer); ok {
```
bwplotka: Tested this commit (`1a59775c31d1`) against `kubernetes/kubernetes` (`master` @ `ca581156293`):
- All 115+ packages importing `prometheus/client_golang` and `k8s.io/component-base/metrics/testutil` pass (`k8s.io/component-base/metrics/...`, `hack/tools/instrumentation/...`, `pkg/kubelet/...`, `pkg/scheduler/...`, `k8s.io/apiserver/...`, etc.).
- Note: `kubernetes/kubernetes` uses `testutil.CollectAndCompare`, `testutil.GatherAndCompare`, `testutil.ScrapeAndCompare`, and `testutil.ToFloat64` (which encode using `expfmt.TypeTextPlain`), and does not directly call `CollectAndFormat` or `GatherAndFormat` with `expfmt.TypeOpenMetrics`.

bwplotka (approved):
TL;DR: newline on text should not break anything.

--- 
Tested this commit (`1a59775c31d1`) against `kubernetes/kubernetes` (`master` @ `ca581156293`):
- All 115+ packages importing `prometheus/client_golang` and `k8s.io/component-base/metrics/testutil` pass (`k8s.io/component-base/metrics/...`, `hack/tools/instrumentation/...`, `pkg/kubelet/...`, `pkg/scheduler/...`, `k8s.io/apiserver/...`, etc.).
- Note: `kubernetes/kubernetes` uses `testutil.CollectAndCompare`, `testutil.GatherAndCompare`, `testutil.ScrapeAndCompare`, and `testutil.ToFloat64` (which encode using `expfmt.TypeTextPlain`), and does not directly call `CollectAndFormat` or `GatherAndFormat` with `expfmt.TypeOpenMetrics`.

PR #2127, opened 2026-09-18
push: preserve dot segments in job and grouping values

jakezwang:
A job name or grouping value equal to `.` or `..` becomes a literal URL path segment. Routers that clean paths can then redirect the push to a different path; for example, job `..` turns `/metrics/job/..` into `/metrics/`.

Use the existing Pushgateway base64 encoding for these two values. Embedded dots retain their current encoding. The regression test sends pushes through `http.ServeMux` and checks the received path and method.

Validation:
- The four dot-segment cases fail before the fix; the ordinary-dot control passes.
- Go 1.27.1: `go test -race ./...` passes in both the root and `exp` modules.
- `GOOS=linux make lint` passes with golangci-lint 2.13.1. The Darwin lint run reports two existing findings in `process_collector_darwin.go`.
- Go 1.26.4: the push package passes with `-race`; the full suite encounters an unrelated runtime-metric expectation mismatch for `go_godebug_non_default_behavior_fips140ems_events_total`.

@kakkoyun


prometheus/push/dot_components_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus/push"
+)
+
+func TestPushDotSegments(t *testing.T) {
```
bwplotka, line 24: Can we add this to the existing file, it's not a major functionality to test, it's just an edge case.

bwplotka: or even better - add to existing tests for / (or create one for encodeComponent)

bwplotka (approved):
I don't think pragmatically we should expect anyone to have `..` label value, but the fix is cheap and it's definitely broken now.

LGTM, just testing nits.

PR #2132, opened 2026-09-22
promhttp: update prometheus/common and support OpenMetrics 2.0

bwplotka:
Update `github.com/prometheus/common` to include OpenMetrics 2.0 support across all metric types.

> NOTE: This is preview, it would be better if we release common first.

- Switch `promhttp.HandlerFor` from deprecated `expfmt.Negotiate` / `expfmt.NegotiateIncludingOpenMetrics` to `expfmt.NegotiateAccept`.
- Support OpenMetrics 2.0 content negotiation when `EnableOpenMetrics: true`.
- Pre-instantiate accepted format slices to avoid per-request allocations.
- Add table-driven tests for OpenMetrics 2.0 negotiation and exposition.


prometheus/promhttp/http.go
```diff
@@ -316,9 +352,9 @@ func HandlerForTransactional(reg prometheus.TransactionalGatherer, opts HandlerO
 
 		var contentType expfmt.Format
 		if opts.EnableOpenMetrics {
-			contentType = expfmt.NegotiateIncludingOpenMetrics(req.Header)
+			contentType = expfmt.NegotiateAccept(req.Header, openMetricsAcceptedFormats...)
```
bwplotka, line 355: Open question:  Can we introduce OM2 by default (experimental stage)? Given our feature flag on collection side, we probably can and should. cc @dashpole @krajorama

prometheus/promhttp/http.go
```diff
@@ -50,6 +50,42 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp/internal"
 )
 
+var (
```
bwplotka, line 53: This `expfmt.NewFormat(` abstraction is odd. I think I will try to propose a different flow.

